PR #2, opened 2021-11-05
Bundle workflow release

mmirecki:


fromanirh:
In this case not much to say, but in general let's always try to have at least one-sentence commit message.

PR #3, opened 2021-11-05
add olm-deploy make target

mmirecki:
The patch adds a olm-deploy make target.
The target will prepare an index image containing numaresources-operator,
and deploy the operator via a Subscription

To  be used as part of "make test-on-ci" to run tests on ocp. This will be used by the ci in this [release repo pr](https://github.com/mmirecki/release/pull/new/numa_tests)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,3 +224,9 @@ catalog-push: ## Push a catalog image.
 .PHONY: deps-update
 deps-update:
 	go mod tidy && go mod vendor
+
+.PHONY: olm-deploy
+olm-deploy:
+	hack/olm-deploy.sh
+
+test-on-ci: olm-deploy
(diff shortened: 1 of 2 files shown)


fromanirh:
OCP CI supports OLM bundles installation (documented in https://docs.ci.openshift.org/docs/how-tos/onboarding-a-new-component/) and we should use this functionality to simplify our flows.

I'd still like to keep the work here as fallback/documentation, but our default approach should do leverage the above builtin functionalty, like we do for the performance-addon-operator.

fromanirh:
@mmirecki do we still need this PR?

mmirecki:
> @mmirecki do we still need this PR?

No, closed

PR #4, opened 2021-11-08
manifests: add bundle

mmirecki:
appropriate parameters provided

To generate the files in this commit, the following parameters
were used with the command:
VERSION=4.10.0 IMG=quay.io/fromani/numaresource-operator:latest make bundle

Note that when running the command for the first time or when the
bundle files are removed, the command will ask additional questions,
asking for:

operator name
operator description
provider name
provider url
maintainers
operator keywords

NOTE: no controller binary args are set for now, this will be added in another PR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.1
+VERSION ?= 4.10-snapshot
 
 # CHANNELS define the bundle channels used in the bundle.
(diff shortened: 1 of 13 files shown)


fromanirh:
Thanks, almost there. Let's fix the commit message and we should be good. It should read:
```
manifests: add bundle

The bundle files can be generated using "make bundle" with appropriate parameters provided.

To generate the files in this commit, the following parameters
were used with the command:
VERSION=4.10.0 IMG=quay.io/fromani/numaresource-operator:latest make bundle
```

mmirecki:
> Thanks, almost there. Let's fix the commit message and we should be good. It should read:
> 
> ```
> manifests: add bundle
> 
> The bundle files can be generated using "make bundle" with appropriate parameters provided.
> 
> To generate the files in this commit, the following parameters
> were used with the command:
> VERSION=4.10.0 IMG=quay.io/fromani/numaresource-operator:latest make bundle
> ```

fixed

fromanirh:
thanks @mmirecki , looks fine now!

PR #6, opened 2021-11-11
Manifests.Update() on fresh Manifests copy

Tal-or:
The operator is performing multiple updates on the same manifests instance - IOW the Update() function processed a dirty copy of Manifests object if more than a single Update() call being executed.
This led to multiple bugs which I initially thought their source is in the deployer:
1. https://github.com/k8stopologyawareschedwg/deployer/pull/55
2. https://github.com/k8stopologyawareschedwg/deployer/pull/56

In order to solve this issue we change the Update() logic to always work on a fresh copy of Manifests object, hence we won't carry former changes during the Update() call.

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -53,14 +53,15 @@ const (
 // NUMAResourcesOperatorReconciler reconciles a NUMAResourcesOperator object
 type NUMAResourcesOperatorReconciler struct {
 	client.Client
-	Log          logr.Logger
-	Scheme       *runtime.Scheme
-	Platform     platform.Platform
-	APIManifests apimanifests.Manifests
-	RTEManifests rtemanifests.Manifests
-	Helper       *deployer.Helper
-	Namespace    string
-	ImageSpec    string
+	Log                 logr.Logger
+	Scheme              *runtime.Scheme
+	Platform            platform.Platform
+	APIManifests        apimanifests.Manifests
+	RTEManifests        rtemanifests.Manifests
+	InitialRTEManifests rtemanifests.Manifests
+	Helper              *deployer.Helper
+	Namespace           string
+	ImageSpec           string
 }
 
 // TODO: narrow down
@@ -133,7 +134,7 @@ func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctr
 func (r *NUMAResourcesOperatorReconciler) RenderManifests(namespace string) rtemanifests.Manifests {
 	logger := r.Log.WithValues("rte", namespace)
 	logger.Info("Updating manifests")
-	mf := r.RTEManifests.Update(rtemanifests.UpdateOptions{
+	mf := r.InitialRTEManifests.Update(rtemanifests.UpdateOptions{
 		Namespace: namespace,
 	})
 	rtestate.UpdateDaemonSetImage(mf.DaemonSet, r.ImageSpec)
--- a/main.go
+++ b/main.go
@@ -124,11 +124,12 @@ func main() {
 
 	if renderManifestsFor != "" {
 		reconciler := &controllers.NUMAResourcesOperatorReconciler{
-			Log:          ctrl.Log.WithName("controllers").WithName("RTE"),
-			APIManifests: apiManifests,
-			RTEManifests: rteManifests,
-			Platform:     clusterPlatform,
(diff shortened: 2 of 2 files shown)


controllers/numaresourcesoperator_controller.go
```diff
@@ -53,14 +53,15 @@ const (
...
-	RTEManifests rtemanifests.Manifests
-	Helper       *deployer.Helper
-	Namespace    string
-	ImageSpec    string
+	Log                 logr.Logger
+	Scheme              *runtime.Scheme
+	Platform            platform.Platform
+	APIManifests        apimanifests.Manifests
+	RTEManifests        rtemanifests.Manifests
+	InitialRTEManifests rtemanifests.Manifests
```
fromanirh, line 61: we will probably need manifests for each host group we define in the API (cc @cynepco3hahue )

main.go
```diff
@@ -124,11 +124,12 @@ func main() {
...
 		reconciler := &controllers.NUMAResourcesOperatorReconciler{
-			Log:          ctrl.Log.WithName("controllers").WithName("RTE"),
-			APIManifests: apiManifests,
-			RTEManifests: rteManifests,
-			Platform:     clusterPlatform,
-			ImageSpec:    images.ResourceTopologyExporterDefaultImageSHA,
+			Log:                 ctrl.Log.WithName("controllers").WithName("RTE"),
+			APIManifests:        apiManifests,
+			RTEManifests:        rteManifests,
+			InitialRTEManifests: rteManifests.Clone(),
```
fromanirh, line 130: but `RTEManifests` is going to be overwritten each time, isnt' it? So what about just
```
Log:                 ctrl.Log.WithName("controllers").WithName("RTE"),
APIManifests:        apiManifests,
InitialRTEManifests: rteManifests,
```

main.go
```diff
@@ -160,14 +161,15 @@ func main() {
...
-		RTEManifests: rteManifests,
-		Platform:     clusterPlatform,
-		Helper:       deployer.NewHelperWithClient(mgr.GetClient(), "", tlog.NewNullLogAdapter()),
-		ImageSpec:    imageSpec,
+		Client:              mgr.GetClient(),
+		Scheme:              mgr.GetScheme(),
+		Log:                 ctrl.Log.WithName("controllers").WithName("RTE"),
+		APIManifests:        apiManifests,
+		RTEManifests:        rteManifests,
+		InitialRTEManifests: rteManifests.Clone(),
```
fromanirh, line 169: same

fromanirh (commented):
when we add support to multiple host groups (cc @cynepco3hahue ) we will probably need to revisit the reconcile loop architecture to support multiple copies of the RTE manifests.

controllers/numaresourcesoperator_controller.go
cynepco3hahue: yep, I am currently thinking about how to do it.
Like in general under expected state we should have a number of resources as only a single one should exist under the cluster and some can have multiple instances. Something like
### Single
- service account
- role
- role binding
- scc

### Multiple
- daemon set
- machine config
- config map

fromanirh: the `deployer`'s `manifest` package should offer enough flexibility for us to compose the functionality the way we need, should not that be the case we can totally send PRs to `deployer` to enable this degree of flexibility.

main.go
Tal-or: We need to keep the mutated ```RTEManifests``` somewhere.
This is why we must have two copies:
one which we always mutate - ```RTEManifests```
two which we never mutate, the original manifest - ```InitialRTEManifests```

fromanirh: sure, I meant only about the initialization. The member attribute should be kept of course, for the very reason you mentioned.

Tal-or: So in the beginning ```RTEManifests``` is going to be empty and then we'll assign the mutated object to it.
But we are about to change the way of how ```Update()``` works anyway - it won't return the mutated object so we'll need to clone it anyway don't we?

Tal-or:
A lot is going on in this area that might fix the problem:
https://github.com/k8stopologyawareschedwg/deployer/pull/60 
closing the PR for now.

Tal-or:
A lot is going on in this area that might fix the problem:
https://github.com/k8stopologyawareschedwg/deployer/pull/60 
closing the PR for now.

PR #9, opened 2021-11-12
gh actions: build also the manifest bundle

fromanirh:
For the operator to be deployable, we need to also build and publish the manifest bundle corresponding to the last image we built.

We probably need to do some manifest processing to make sure to reference the image we just built.


mmirecki:
/lgtm

although we already have this in: #2

fromanirh:
> /lgtm
> 
> although we already have this in: #2

ok, so not needed, closing

PR #10, opened 2021-11-12
Rename operator bundle image to use openshift-kni quay repo

mmirecki:
The image is now stored in quay.io/openshift-kni


fromanirh (commented):
/approve
/lgtm

PR #11, opened 2021-11-14
Consume rte e2e

Tal-or:
Implement the necessary changes in order for the E2E tests to be able to consume and run the RTE-specific E2E tests.


Tal-or:
```run golangci-lint
  Running [/home/runner/golangci-lint-1.41.1-linux-amd64/golangci-lint run --out-format=github-actions --timeout=15m0s --verbose] in [] ...
  Error: SA4023: this comparison is always true (staticcheck)
  Error: SA4023(related information): the lhs of the comparison is the 1st return value of this function call (staticcheck)```

Getting this error in the ```CI GO / lint``` lane, but it's not clear from the log where the issue is.

**Comment by Tal-or** [comment]
> ```
>   Running [/home/runner/golangci-lint-1.41.1-linux-amd64/golangci-lint run --out-format=github-actions --timeout=15m0s --verbose] in [] ...
>   Error: SA4023: this comparison is always true (staticcheck)
>   Error: SA4023(related information): the lhs of the comparison is the 1st return value of this function call (staticcheck)
> ```
> 
> Getting this error in the `CI GO / lint` lane, but it's not clear from the log where the issue is.

@fromanirh Any idea why/where it's happening? the error doesn't provide the line of the faulty code

**Comment by fromanirh** [comment]
> > Getting this error in the `CI GO / lint` lane, but it's not clear from the log where the issue is.
> 
> @fromanirh Any idea why/where it's happening? the error doesn't provide the line of the faulty code

Indeed, we should improve the lane to have better output. Meantime it's just simpler to run the tool in your local checkout.

**Comment by fromanirh** [comment]
Ah, nice, the lint errors are displayed inline: https://github.com/openshift-kni/numaresources-operator/pull/11/commits/a3fe88171680f5a0de9cbc57426960d0df61f4e0

**Comment by Tal-or** [comment]
> Ah, nice, the lint errors are displayed inline: [a3fe881](https://github.com/openshift-kni/numaresources-operator/commit/a3fe88171680f5a0de9cbc57426960d0df61f4e0)

What do I see here? it's just the last commit I added

**Comment by fromanirh** [comment]
> > Ah, nice, the lint errors are displayed inline: [a3fe881](https://github.com/openshift-kni/numaresources-operator/commit/a3fe88171680f5a0de9cbc57426960d0df61f4e0)
> 
> What do I see here? it's just the last commit I added

image

No sure I see this because I have admin rights on the repo, though.

File: test/e2e/basic_install/install.go [hunk 1]
```diff
@@ -18,132 +18,61 @@ package basic_install
...
 
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
-	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
-	machineconfigclientset "github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned/typed/machineconfiguration.openshift.io/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/kubernetes/test/e2e/framework"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
fromanirh, line 28: previously, we were using the k8s framework to enable the import of the RTE testsuite, which is based on k8s framework (in turn to maximize the interoparability with the NFD testsuite)

I still like consuming the rte e2e suite as separate binary (!) so let's keep this direction

test/e2e/basic_install/install.go
```diff
@@ -18,132 +18,61 @@ package basic_install
 
 import (
 	"context"
+	"log"
```
fromanirh, line 21: we should use klog even in the tests, but we can improve later

test/e2e/utils/runtime/runtime.go
```diff
@@ -0,0 +1,32 @@
...
+	"runtime"
+)
+
+func GetBinariesPath() (string, error) {
+	_, file, _, ok := runtime.Caller(0)
+	if !ok {
+		return "", fmt.Errorf("cannot retrieve tests directory")
+	}
+	basedir := filepath.Dir(file)
+	return filepath.Abs(filepath.Join(basedir, "..", "..", "..", "..", "bin"))
```
fromanirh, line 31: this looks fragile, we will need a more clever approach. But we can improve later.

fromanirh (approved):
/lgtm
/approve
good work here, and really big effort.
I still like it runnint the rte e2e suite as separate binary, so let's go this direction.
Minor comments inside which we can improve later, let's have this merged.

PR #13, opened 2021-11-16
Provide a new NodeGroups API

cynepco3hahue:
The API allows specifying groups of nodes that are chosen by the machine config pool selector. For each node group, the controller will deploy a separate daemon set, machine config, and config map. The PR also introduces changes under the API status section to reflect the fact that we can have a number of daemon sets.

New API example:

```yaml
...
spec:
  nodeGroups:
  - machineConfigPoolSelector:
      matchLabels:
        test: test
...
```


config/rbac/role.yaml
```diff
@@ -11,67 +11,43 @@ rules:
...
   - configmaps
   verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
+  - '*'
```
fromanirh, line 14: ok, I see the point here, but let's make sure to have a good run later to actually REDUCE these permissions to the bare minimum, if possible. Now, probably for `configmaps` we cannot remove any permission, but in general our RBAC permissions will need some serious review later on.

fromanirh (approved):
awesome work! there are minor comments inside, but considering the size of the PR I'd rather merge it once the e2e tests pass and improve the (minor) comments later.

fromanirh:
```ERRO[2021-11-16T15:40:13Z] Timed out waiting for image pull secrets in the test namespace. ```
it seems like a glitch?

**Comment by fromanirh** [comment]
ok, so prow is still misbehaving:
```
error: build error: failed to pull image: After retrying 2 times, Pull image still failed due to error: Error reading signatures: Error downloading signatures for sha256:0795d0168de07e5d16f4d0fefd43a0615618a8b8bee259eae008063872d2a852 in image-registry.openshift-image-registry.svc:5000/ci-op-r7wxd3wi/pipeline: error parsing HTTP 404 response body: unexpected end of JSON input: ""
```
however the e2e test failure seems real.

**Comment by cynepco3hahue** [comment]
> so, the e2e tests failures are real and the lane found some bugs! :)
> 
> 1. let's NOT set the references for the CRD: CRDs are cluster-scoped anyway:

The question what should we do with the CRD that the operator installed, if someone updates it or when we delete the operator we also should delete the CRD that it created. And it is much easier to do with the owner's reference.

> 
> ```
> diff --git a/controllers/numaresourcesoperator_controller.go b/controllers/numaresourcesoperator_controller.go
> index bf0422a..3844d9e 100644
> --- a/controllers/numaresourcesoperator_controller.go
> +++ b/controllers/numaresourcesoperator_controller.go
> @@ -19,9 +19,10 @@ package controllers
>  import (
>         "context"
>         "fmt"
> +       "time"
> +
>         metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
>         "k8s.io/apimachinery/pkg/labels"
> -       "time"
>  
>         "github.com/k8stopologyawareschedwg/deployer/pkg/deployer"
>         "github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
> @@ -191,10 +192,6 @@ func (r *NUMAResourcesOperatorReconciler) syncNodeResourceTopologyAPI(instance *
>         existing := apistate.FromClient(context.TODO(), r.Client, r.Platform, r.APIManifests)
>  
>         for _, objState := range existing.State(r.APIManifests) {
> -               if err := controllerutil.SetControllerReference(instance, objState.Desired, r.Scheme); err != nil {
> -                       return errors.Wrapf(err, "failed to set controller reference to %s %s", objState.Desired.GetNamespace(), objState.Desired.GetName())
> -               }
> -
>                 if _, err := apply.ApplyObject(context.TODO(), r.Client, objState); err != nil {
>                         return errors.Wrapf(err, "could not create %s", objState.Desired.GetObjectKind().GroupVersionKind().String())
> ```
> 
> 1. we need to deploy a couple more CRDs:

no need, it my bad that I always watch OpenShift only resources.

> 
> ```
> kubectl create -f https://raw.githubusercontent.com/openshift/api/master/security/v1/0000_03_security-openshift_01_scc.crd.yaml
> kubectl create -f https://raw.githubusercontent.com/openshift/machine-config-operator/master/install/0000_80_machine-config-operator_01_machineconfig.crd.yaml
> ```
> 
> let's add makefile targets (and helper scripts?) to deploy/undeploy all of these. We need separate entrypoints, which the top-level `deploy` target will depend on 3. we need to add permissions for MCPs:
> 
> ```
> E1116 17:03:49.735063       1 reflector.go:138] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:245: Failed to watch *v1.MachineConfigPool: failed to list *v1.MachineConfigPool: machineconfigpools.machineconfiguration.openshift.io is forbidden: User "system:serviceaccount:numaresources:numaresources-controller-manager" cannot list resource "machineconfigpools" in API group "machineconfiguration.openshift.io" at the cluster scope
> ```
> 
> fix:
> 
> ```
> -//+kubebuilder:rbac:groups=machineconfiguration.openshift.io,resources=machineconfigs,verbs=*
> +//+kubebuilder:rbac:groups=machineconfiguration.openshift.io,resources=machineconfigs,verbs=get;list;watch
> +//+kubebuilder:rbac:groups=machineconfiguration.openshift.io,resources=machineconfigpools,verbs=get;list;watch
> ```
> 
> EDIT: these changes are sufficient to pass the existing tests, which is in turn enough (provided the other lane pass) to merge this PR. We will need to add (many) more tests anyway.

Goot catch regarding to RBAC.

**Comment by fromanirh** [comment]
> > so, the e2e tests failures are real and the lane found some bugs! :)
> > 
> > 1. let's NOT set the references for the CRD: CRDs are cluster-scoped anyway:
> 
> The question what should we do with the CRD that the operator installed, if someone updates it or when we delete the operator we also should delete the CRD that it created. And it is much easier to do with the owner's reference.

I'm not sure kube allows. If it is possible, let's take care in a followup PR. Let's have the (good) foundation in and then we can improve, the re are quite some minor improvements all around we can implement as followup.

**Comment by cynepco3hahue** [comment]
> I'm not sure kube allows. If it is possible, let's take care in a followup PR. Let's have the (good) foundation in and then we can improve, the re are quite some minor improvements all around we can implement as followup.

In general it is possible, but not in our case because
```
FailedAPISync: failed to set controller reference to  noderesourcetopologies.topology.node.k8s.io: cluster-scoped resource must not have a namespace-scoped owner, owner's namespace rte-6715
```
and it brings another topic, do we want to change operator CR to have cluster scope instead of namespace scope?

config/rbac/role.yaml
cynepco3hahue: In general, all resource that the operator creates we can not reduce any permissions, because we need to create, update, get, watch, list and delete resources

fromanirh:
/lgtm
let's improve incrementally the pending (mostly minor) items

fromanirh:
> > I'm not sure kube allows. If it is possible, let's take care in a followup PR. Let's have the (good) foundation in and then we can improve, the re are quite some minor improvements all around we can implement as followup.
> 
> In general it is possible, but not in our case because
> 
> ```
> FailedAPISync: failed to set controller reference to  noderesourcetopologies.topology.node.k8s.io: cluster-scoped resource must not have a namespace-scoped owner, owner's namespace rte-6715
> ```
> 
> and it brings another topic, do we want to change operator CR to have cluster scope instead of namespace scope?

good point. Let's discuss in another forum

PR #14, opened 2021-11-17
Nodegroups minor cleanups

fromanirh:
minor cleanups emerged during the review of #13

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -219,10 +219,8 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 			return nil, errors.Wrapf(err, "could not apply (%s) %s/%s", objState.Desired.GetObjectKind().GroupVersionKind(), objState.Desired.GetNamespace(), objState.Desired.GetName())
 		}
 
-		if obj.GetObjectKind().GroupVersionKind().Kind == "DaemonSet" {
-			if nname, ok := rte.NamespacedNameFromObject(obj); ok {
-				daemonSetsNName = append(daemonSetsNName, nname)
-			}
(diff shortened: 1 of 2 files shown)


config/rbac/role.yaml
```diff
@@ -49,7 +49,9 @@ rules:
   resources:
   - machineconfigs
   verbs:
-  - '*'
+  - get
```
cynepco3hahue, line 52: it will not work, because our operator should created machine configs

controllers/numaresourcesoperator_controller.go
```diff
@@ -77,7 +78,7 @@ type NUMAResourcesOperatorReconciler struct {
 // Cluster Scoped
 //+kubebuilder:rbac:groups=topology.node.k8s.io,resources=noderesourcetopologies,verbs=get;list;create;update
 //+kubebuilder:rbac:groups=config.openshift.io,resources=clusterversionss,verbs=list
-//+kubebuilder:rbac:groups=machineconfiguration.openshift.io,resources=machineconfigs,verbs=*
+//+kubebuilder:rbac:groups=machineconfiguration.openshift.io,resources=machineconfigs,verbs=get;list;watch
```
cynepco3hahue, line 81: see my comment above

config/rbac/role.yaml
fromanirh: ah, right. Good point. Do we also need delete permissions, though?

controllers/numaresourcesoperator_controller.go
fromanirh: sure

config/rbac/role.yaml
cynepco3hahue: I think we need all permissions, like create, update, patch, watch, delete, list and get

PR #15, opened 2021-11-18
Move the CRD to cluster scoped

fromanirh:
Since "enable numa-aware scheduling" is a cluster property, we
move the scope to cluster. Previously it was Namespaced for no
reall well thought reason.


fromanirh:
too simplistic. Need to think more about the namespace management.

PR #16, opened 2021-11-22
API: allow to override rte image

fromanirh:
Add a field in our API to optionally allow to override the RTE image to deploy, mirroring wath the secondary-scheduler-operator does.
This gives us flexibiltiy at almost no cost, allowing easier test, enabling troubleshooting and hotfix delivery.

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -192,6 +192,8 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 		return nil, errors.Wrapf(err, "failed to get machine config pools related to instance node groups")
 	}
 
+	rtestate.UpdateDaemonSetImage(r.RTEManifests.DaemonSet, r.getExporterImage(instance.Spec.ExporterImage))
+
 	var daemonSetsNName []nropv1alpha1.NamespacedName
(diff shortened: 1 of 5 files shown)


yanirq:
Aren't we missing generated code here when it comes to API changes ?

fromanirh:
> Aren't we missing generated code here when it comes to API changes ?

AFAIK no, because the client imports the API defs, so the changes are consumed that way.
xref: https://github.com/openshift-kni/numaresources-operator/blob/main/pkg/k8sclientset/generated/clientset/versioned/typed/numaresourcesoperator/v1alpha1/numaresourcesoperator.go#L24

controllers/numaresourcesoperator_controller.go
```diff
@@ -210,6 +212,15 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 	return daemonSetsNName, nil
 }
 
+func (r *NUMAResourcesOperatorReconciler) getExporterImage(imgSpec string) string {
+	imageSpec := imgSpec
+	if imageSpec == "" {
+		imageSpec = r.ImageSpec
+	}
+	klog.InfoS("Using Exporter", "image", imageSpec)
```
Tal-or, line 220: The log message isn't completely clear.
I would let the user know if we override the default image using the API filed or not

fromanirh: ok, I'll revisit the log line.

fromanirh:
/hold cancel
re-uploading remove the lgtm anyway

PR #17, opened 2021-11-22
Change the NRO CRD scope from namespace to cluster scope

cynepco3hahue:
The PR also introduces the functionality to fetch the namespace from pod ENV variables and use fetched namespace for deploying RTE components.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -33,6 +33,11 @@ spec:
         name: manager
         securityContext:
           allowPrivilegeEscalation: false
+        env:
+        - name: NAMESPACE
+          valueFrom:
+            fieldRef:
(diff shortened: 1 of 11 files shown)


config/manager/manager.yaml
```diff
@@ -33,6 +33,11 @@ spec:
         name: manager
         securityContext:
           allowPrivilegeEscalation: false
+        env:
+        - name: NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
```
fromanirh, line 40: nice and simple solution

fromanirh (approved):
/lgtm
/hold
great, turned out simpler than I thought! Holding to let other reviewers chime in

PR #18, opened 2021-11-23
Add ci dockerfiles

fromanirh:
Add dockefiles which will be consumed by prow CI.


fromanirh:
will revisit if actually needed

PR #20, opened 2021-11-23
makefile: add e2e tests target

mmirecki:


Makefile
```diff
@@ -96,6 +96,9 @@ test: manifests generate fmt vet envtest ## Run tests.
 test-unit: envtest
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./controllers/... ./pkg/...
 
+test-e2e:
+	go test ./test/e2e
```
fromanirh, line 100: I'd do like we do in the github actions kind flow:

```
test-e2e: binary-e2e binary-e2e-rte
	./bin/e2e.test -ginkgo.v
```

While I'm fine adding this make target (and re-using it in the aforementioned github actions flow) I'm not sure it's necessary for the prow config

PR #24, opened 2021-11-23
Detect topology manager policy 

Tal-or:
We'll intersect between the MCP label and kubeletconfig MCP selector in order to find the correct kubeletconfig.
Then we'll retrieve the tm policy from the kubeletconfig object and update it in the correct DaemonSets.


fromanirh:
absorbed in https://github.com/openshift-kni/numaresources-operator/pull/40

PR #33, opened 2021-11-25
sysinfo: allow configure from the command line

fromanirh:
So we have options: we can either reconfigure the tool
using only command-line arguments or (re)generate the configmap,
whatever it's easier in the short term until OCP rebases
on top of kube 1.23.


fromanirh:
replaced by https://github.com/openshift-kni/numaresources-operator/pull/38

PR #34, opened 2021-11-25
rte: move kni-specific packages to klog

fromanirh:
it's just ugly to mix `log` and `klog`, and
we have no real reason to be stuck with `log`.

### DIFF
--- a/rte/pkg/config/config.go
+++ b/rte/pkg/config/config.go
@@ -17,9 +17,9 @@ package config
 import (
 	"errors"
 	"io/ioutil"
-	"log"
 	"os"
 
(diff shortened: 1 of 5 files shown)


cynepco3hahue:
oo yee
/lgtm

Tal-or:
We also have some here https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/e2e_suite_test.go#L22

fromanirh:
> We also have some here https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/e2e_suite_test.go#L22

thanks, updated

PR #35, opened 2021-11-25
Improve logic around MachineConfigPools

cynepco3hahue:
The current controller logic will be:
- create machine configs
- wait for MCPs updates
- create all other components

TODO:
- [x] unittests
- [x] verify the logic, I can see some issues to pass MCP conditions


pkg/validation/validation.go
```diff
@@ -0,0 +1,109 @@
...
+	for _, nodeGroup := range nodeGroups {
+		if nodeGroup.MachineConfigPoolSelector == nil {
+			return fmt.Errorf("one of the node groups does not have machineConfigPoolSelector")
+		}
+	}
+
+	return nil
+}
+
+// TODO: move it under the validation webhook once we will have one
```
fromanirh, line 82: I'm not sure we want to invest into a validation webhook considering the purpose of this operator, but i'm open to such opportunity if there is need and we have capacity

cynepco3hahue: in general, it is not too hard and provides us some guarantees regarding to API values
Anyway we can discuss it in the future

fromanirh: I'm open to that in general, let's just prioritize properly and not overinvest if we can help it.

api/numaresourcesoperator/v1alpha1/numaresourcesoperator_types.go
```diff
@@ -57,12 +59,22 @@ type NodeGroup struct {
 
 // NUMAResourcesOperatorStatus defines the observed state of NUMAResourcesOperator
 type NUMAResourcesOperatorStatus struct {
-	DaemonSets []NamespacedName `json:"daemonsets,omitempty"`
-
+	DaemonSets         []NamespacedName    `json:"daemonsets,omitempty"`
+	MachineConfigPools []MachineConfigPool `json:"machineconfigpools,omitempty"`
```
fromanirh, line 63: next time please elaborate in the commit message *why* we do like this, which problem we want to solve/mitigate and in general everything that will allow us our future selves in 6 months for now to reconstruct our thinking now.

fromanirh (commented):
/lgtm
/approve

PR #36, opened 2021-11-25
rte: use operator image as rte image

fromanirh:
Revamp the code handling user-supplied RTE image, to streamline the flow and make room for future enhancement needed to fully fix the operator-as-operand flow.

Expose and adjust the environment variables we use to introspect the image.

Important fix: We need to passthrough the ImagePullPolicy in the operator-as-operand flow; this is correct in general, the image needs to be handled consistently, and unbreaks our CI flows.

Change UID/GID when running RTE reusing the same operator image:
The RTE need to access entries in the filesystem which are owned by root - surely the podresources socket, on k8s also the kubelet config.

However the operator should run as non-root, and its configuration reflects so.

When we use the operator image as operand (RTE), this settings "leaks" (positively for a change!) into RTE.

The access to the files (podresources socket) needed by RTE is guarded by both SELinux and unix permission control.
We already fixed the former when needed, but to pass the latter, we need to run RTE with uid and gid 0 (as root).

Note we do NOT want full root privileges: we rely on general kube settings to drop all the other privileges; should unneeded
privilege not have been dropped, it will be fixed on separate PRs.

### DIFF
--- a/Dockerfile.openshift.rte
+++ b/Dockerfile.openshift.rte
@@ -1,0 +1,6 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal
+COPY exporter /bin/resource-topology-exporter
+RUN mkdir /etc/resource-topology-exporter/ && \
+    touch /etc/resource-topology-exporter/config.yaml
+RUN microdnf install pciutils
+ENTRYPOINT ["/bin/resource-topology-exporter"]
--- a/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
@@ -55,6 +55,14 @@ spec:
         - apiGroups:
           - ""
           resources:
+          - pods
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - ""
+          resources:
           - serviceaccounts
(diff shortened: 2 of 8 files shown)


fromanirh:
/hold
broken:
```
2021-11-25T17:15:37.047Z	INFO	setup	unable to find current image, using hardcoded	{"error": "the cache is not started, can not read objects"}
2021-11-25T17:15:37.047Z	INFO	setup	using RTE image	{"spec": ""}
```

fromanirh:
/hold
the kind failure seems real

cynepco3hahue:
Looks good to me, only one question, IIUC correctly you want to have an option to change the RTE image on the fly, without need to restart the operator, right?

fromanirh:
> Looks good to me, only one question, IIUC correctly you want to have an option to change the RTE image on the fly, without need to restart the operator, right?

The flow we want to enable is:
_When we create the cluster-scoped CR_, allow the user to override the default RTE image. This is to enable hotfixes or allow to try out things. We expect this feature to be used rarely, but must be there (and should be covered by e2e tests...).
The vast majority of time (99%) the operator should use the defaults, so no `ImageSpec` in the CR, so the RTE image should be the same as the operator (operator-as-operand)

PR #38, opened 2021-11-29
WIP Reconcile kubelet config

fromanirh:
- [X] Cover letter (this message)
- [X] Add configmap mount into daemonset
- [ ] Test coverage
- [ ] Handle config change in RTE

revamp how we handle the configuration of the RTE.
We need to supply to the RTE a configuration file to tell it:
1. details about the TM configuration, which it cannot figure out otherwise (policy, scope...)
2. parameters needed by the `sysinfo` package, which it cannot determine from the system because they depend on the kubelet configuration.
    
The second set is going to be phased out when OCP exposed the latest changes in the podresources API, but the first set will always be needed. So we will need some form of configuration management anyway.
    
So the operator will translate the information from the Kubeletconfig object, if available, to the RTE config.yaml and make it available as configmap.
    
Most important design decisions:
a. always translate as much configuration as possible, since it's cheap once we have access to the relevant kubeconfig. This for example include always reporting `reservedCPUs`, even when the RTE will no longer need this information (podresources' GetAllocatable available). RTE will just ignore the fields it doesn't need anymore
b. build the configuration only on OpenShift; on Kubernetes, we can just mount the kubelet config on the RTE pod and let it figure out.
c. on the other hand, always mount the configMap, marking it `optional`, so we will not need platform-dependent changes on the DaemonSet.
    
Summary: always try to mount the configMap from the RTE daemonset, ignore errors if not available, create it only on OCP translating as much as possible (greedily).
    
In order to process the kubeletconfig, we introduced a new controller.
The new controller will watch the kubeletconfigs and reconcile the relevant configmaps.
We add a new controller to not overload the existing controller, and because, per above design decisions, we can nicely split the responsabilities and make the two reconciliation loops (almost) completely independent: the numaresources main controller will deploy the bigger part of the stack; the kubeletconfig controller will provide configuration, if needed and possible.
(xref: "One controller per custom resource definition" from https://sdk.operatorframework.io/docs/best-practices/best-practices/)
 
Open issues:
- We need to trigger the kubeletconfig processing somehow when we create the manifests the first time.
A possible approach is to is to make sure to run our own kubeletconfig reconciliation loop runs before to create the daemonsets, because this way the daemonset has a very high chance to start up picking the correct configuration right from the start.
We can either process the kubeletconfig explicitely or trigger a reconciliation.
- RTEs do not pick up  ConfigMaps updates automatically yet.

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -104,6 +104,8 @@ type NUMAResourcesOperatorReconciler struct {
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.2/pkg/reconcile
 func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	_ = context.Background()
+	klog.V(3).InfoS("Starting NUMAResourcesOperator reconcile loop", "object", req.NamespacedName)
+	defer klog.V(3).InfoS("Finish NUMAResourcesOperator reconcile loop", "object", req.NamespacedName)
 
 	instance := &nropv1alpha1.NUMAResourcesOperator{}
 	err := r.Get(context.TODO(), req.NamespacedName, instance)
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -60,6 +60,7 @@ import (
 
 const (
 	defaultNUMAResourcesOperatorCrName = "numaresourcesoperator"
+	RetryPeriod                        = 30 * time.Second
 )
 
 // NUMAResourcesOperatorReconciler reconciles a NUMAResourcesOperator object
@@ -92,7 +93,6 @@ type NUMAResourcesOperatorReconciler struct {
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles,verbs=*
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings,verbs=*
 //+kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=*
-//+kubebuilder:rbac:groups="",resources=configmaps,verbs=*
 //+kubebuilder:rbac:groups=nodetopology.openshift.io,resources=numaresourcesoperators,verbs=*
 //+kubebuilder:rbac:groups=nodetopology.openshift.io,resources=numaresourcesoperators/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=nodetopology.openshift.io,resources=numaresourcesoperators/finalizers,verbs=update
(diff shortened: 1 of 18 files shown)


controllers/reconcile_configmap.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+
+		if selector.Matches(labels.Set(mcp.Labels)) {
+			return mcp, nil
+		}
+	}
+	return nil, fmt.Errorf("no match")
+}
+
+func syncConfigMapFromNROPInstance(ctx context.Context, cli client.Client, instance *nropv1alpha1.NUMAResourcesOperator) error {
```
cynepco3hahue, line 93: NROP?

pkg/objectstate/rte/rte.go
```diff
@@ -180,21 +179,27 @@ func (em *ExistingManifests) State(mf rtemanifests.Manifests, instance *nropv1al
...
-			desiredConfigMap := mf.ConfigMap.DeepCopy()
-			desiredConfigMap.Name = generatedName
+		if mf.MachineConfig != nil {
+			if mcp.Spec.MachineConfigSelector == nil {
+				klog.Warningf("the machine config pool %q does not have machine config selector", mcp.Name)
+				continue
+			}
+			desiredMachineConfig := mf.MachineConfig.DeepCopy()
+			// prefix machine config name to guarantee that we will have an option to override it
+			desiredMachineConfig.Name = fmt.Sprintf("51-%s", generatedName)
```
cynepco3hahue, line 189: Can you use https://github.com/openshift-kni/numaresources-operator/blob/1ff7e5148afe0b3b8ca0ffc33e227818ad17f125/pkg/objectstate/rte/rte.go#L334?

cynepco3hahue:
Gave some initial review, I will pass again once the CI will pass. Can you please clarify what the expected flow should be?
IIUC the controller will monitor kubelet configs and update accordingly the config map that the RTE daemon set will use as a source of truth to get the topology manager policy, reserved CPUs, and probably reserved memory?

controllers/reconcile_configmap.go
fromanirh: Numa Resources OPerator

controllers/numaresourcesoperator_controller.go
```diff
@@ -160,22 +162,16 @@ func (r *NUMAResourcesOperatorReconciler) updateStatus(ctx context.Context, inst
...
 	var err error
 	err = r.syncNodeResourceTopologyAPI()
 	if err != nil {
-		return ctrl.Result{}, status.ConditionDegraded, errors.Wrapf(err, "FailedAPISync")
+		return ctrl.Result{RequeueAfter: RetryPeriod}, status.ConditionDegraded, errors.Wrapf(err, "FailedAPISync")
+	}
+
+	err = syncConfigMapFromNROPInstance(ctx, r.Client, instance)
+	if err != nil {
+		return ctrl.Result{RequeueAfter: RetryPeriod}, status.ConditionDegraded, errors.Wrapf(err, "FailedCFGSync")
```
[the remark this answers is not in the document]
fromanirh: thanks, removed

pkg/objectstate/rte/rte.go
fromanirh: This was actually a rebase glitch, removed

fromanirh:
continuing on https://github.com/openshift-kni/numaresources-operator/pull/40
(disclosure: `git push -d` instead of `git push -f` :\ )

PR #40, opened 2021-12-01
Reconcile kubelet config

fromanirh:
- [X] Cover letter (this message)
- [X] Add configmap mount into daemonset
- [X] ([very] basic)Test coverage

revamp how we handle the configuration of the RTE.
We need to supply to the RTE a configuration file to tell it:
1. details about the TM configuration, which it cannot figure out otherwise (policy, scope...)
2. parameters needed by the `sysinfo` package, which it cannot determine from the system because they depend on the kubelet configuration.
    
The second set is going to be phased out when OCP exposed the latest changes in the podresources API, but the first set will always be needed. So we will need some form of configuration management anyway.
    
So the operator will translate the information from the Kubeletconfig object, if available, to the RTE config.yaml and make it available as configmap.
    
Most important design decisions:
a. always translate as much configuration as possible, since it's cheap once we have access to the relevant kubeconfig. This for example include always reporting `reservedCPUs`, even when the RTE will no longer need this information (podresources' GetAllocatable available). RTE will just ignore the fields it doesn't need anymore
b. build the configuration only on OpenShift; on Kubernetes, we can just mount the kubelet config on the RTE pod and let it figure out.
    
In order to process the kubeletconfig, we introduced a new controller.
The new controller will watch the kubeletconfigs and reconcile the relevant configmaps.
We add a new controller to not overload the existing controller, and because, per above design decisions, we can nicely split the responsabilities and make the two reconciliation loops (almost) completely independent: the numaresources main controller will deploy the bigger part of the stack; the kubeletconfig controller will provide configuration, if needed and possible.
(xref: "One controller per custom resource definition" from https://sdk.operatorframework.io/docs/best-practices/best-practices/)

NOTE: we will handle the RTE side (noticing the update configuration) on a separate PR

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -15,6 +15,13 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - events
+  verbs:
+  - create
+  - patch
+- apiGroups:
+  - ""
+  resources:
   - serviceaccounts
   verbs:
   - '*'
@@ -36,6 +43,14 @@ rules:
   - clusterversions
   verbs:
   - list
+- apiGroups:
+  - machineconfiguration.openshift.io
+  resources:
+  - kubeletconfigs
+  verbs:
+  - get
+  - list
+  - watch
 - apiGroups:
   - machineconfiguration.openshift.io
   resources:
--- a/controllers/kubeletconfig_controller.go
+++ b/controllers/kubeletconfig_controller.go
@@ -1,0 +1,200 @@
+/*
+
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/klog/v2"
+	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	"github.com/ghodss/yaml"
+	"github.com/pkg/errors"
+
+	rtemanifests "github.com/k8stopologyawareschedwg/deployer/pkg/manifests/rte"
+	nropv1alpha1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1alpha1"
+	"github.com/openshift-kni/numaresources-operator/pkg/apply"
+	cfgstate "github.com/openshift-kni/numaresources-operator/pkg/objectstate/cfg"
+	rtestate "github.com/openshift-kni/numaresources-operator/pkg/objectstate/rte"
+	rteconfig "github.com/openshift-kni/numaresources-operator/rte/pkg/config"
+	"github.com/openshift-kni/numaresources-operator/rte/pkg/sysinfo"
+	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+)
+
+const (
+	kubeletConfigRetryPeriod = 30 * time.Second
+)
+
+// KubeletConfigReconciler reconciles a KubeletConfig object
+type KubeletConfigReconciler struct {
+	client.Client
+	Scheme    *runtime.Scheme
+	Recorder  record.EventRecorder
+	Namespace string
+}
+
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=*
+//+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+//+kubebuilder:rbac:groups=machineconfiguration.openshift.io,resources=kubeletconfigs,verbs=get;list;watch
+
+func (r *KubeletConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	klog.V(3).InfoS("Starting KubeletConfig reconcile loop", "object", req.NamespacedName)
+	defer klog.V(3).InfoS("Finish KubeletConfig reconcile loop", "object", req.NamespacedName)
+
+	nname := types.NamespacedName{
+		Name: defaultNUMAResourcesOperatorCrName,
+	}
+	instance := &nropv1alpha1.NUMAResourcesOperator{}
+	err := r.Get(ctx, nname, instance)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// so we need to have a NUMAResourceOperator instance to be able to set owner refs
+			// correctly. And we just wait forever until it comes, which is expected to be "soon".
+			return ctrl.Result{RequeueAfter: kubeletConfigRetryPeriod}, nil
+		}
+		// Error reading the object - requeue the request.
+		return ctrl.Result{}, err
+	}
+
+	// KubeletConfig changes are expected to be sporadic, yet are important enough
+	// to be made visible at kubernetes level. So we generate events to handle them
+
+	cm, err := r.reconcileConfigMap(ctx, instance, req.NamespacedName)
+	if err != nil {
+		klog.ErrorS(err, "failed to reconcile configmap", "controller", "kubeletconfig")
+
+		msg := fmt.Sprintf("Failed to update RTE config from kubelet config %s/%s", req.NamespacedName.Namespace, req.NamespacedName.Name)
+		r.Recorder.Event(instance, "Warning", "ProcessFailed", msg)
+		return ctrl.Result{}, err
+	}
+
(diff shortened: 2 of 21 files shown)


fromanirh:
reason of failure:
```
Events:
  Type     Reason       Age               From               Message
  ----     ------       ----              ----               -------
  Normal   Scheduled    72s               default-scheduler  Successfully assigned numaresources/numaresourcesoperator-mcp-test-xxkcs to kind-worker
  Warning  FailedMount  9s (x8 over 72s)  kubelet            MountVolume.SetUp failed for volume "rte-config" : configmap "numaresourcesoperator-mcp-test" not found
```

This is actually good because it fails fast and reliably. It is indeed caused by the pending design decisions :)

fromanirh:
note this is fixed (and the fix is debatable) on kube, but it's silently broken on OCP.

controllers/kubeletconfig_controller.go
```diff
@@ -0,0 +1,97 @@
...
+		return ctrl.Result{RequeueAfter: RetryPeriod}, err
+	}
+
+	msg := fmt.Sprintf("Updated RTE config %s/%s from kubelet config %s/%s", cm.Namespace, cm.Name, req.NamespacedName.Namespace, req.NamespacedName.Name)
+	r.Recorder.Event(instance, "Normal", "ProcessOK", msg)
+	return ctrl.Result{}, nil
+}
+
+func (r *KubeletConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	b := ctrl.NewControllerManagedBy(mgr).
```
yanirq, line 93: Can you explain the return syntax here ?    what is the diff between using :
```
ctrl.NewControllerManagedBy(mgr).
		For(&mcov1.KubeletConfig{}).
                Owns(&corev1.ConfigMap{}).
                Complete(r)
```

controllers/kubeletconfig_controller.go
```diff
@@ -0,0 +1,97 @@
...
+		For(&mcov1.KubeletConfig{})
+	return b.Owns(&corev1.ConfigMap{}).
```
yanirq, line 95: any predicats needed for config map here ?  (update/delete/...)

controllers/kubeletconfig_controller.go
fromanirh: we can switch back to this syntax actually. The current code is due to a past attempt never fully cleaned up.

controllers/kubeletconfig_controller.go
fromanirh: defaults should be fine, because any change on kubelet config should be reflected on the configmap(s) this controller owns

controllers/reconcile_configmap.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+type configMapHelper struct {
+	cli       client.Client
+	scm       *runtime.Scheme
+	instance  *nropv1alpha1.NUMAResourcesOperator
+	namespace string
+}
+
+func (cmh *configMapHelper) reconcileConfigMap(ctx context.Context, kcKey client.ObjectKey) (*corev1.ConfigMap, error) {
```
yanirq, line 60: So basically this is a call from the kubelet reconcile loop right ?  not a second actual reconcile loop ?  (just trying to understand by the PR description and the code de-facto)

main.go
```diff
@@ -187,6 +187,16 @@ func main() {
...
 		os.Exit(1)
 	}
+	if err = (&controllers.KubeletConfigReconciler{
+		Client:    mgr.GetClient(),
+		Scheme:    mgr.GetScheme(),
+		Recorder:  mgr.GetEventRecorderFor("kubeletconfig-controller"),
+		Namespace: namespace,
+	}).SetupWithManager(mgr); err != nil {
+		klog.ErrorS(err, "unable to create controller", "controller", "KubeletConfig")
+		os.Exit(1)
```
yanirq, line 197: maybe we can swap all  error logs + os.Exit(1)  with klog.Exitf  , but that's for a different PR. not urgent.

controllers/reconcile_configmap.go
```diff
@@ -0,0 +1,144 @@
...
+		selector, err := metav1.LabelSelectorAsSelector(mcoKc.Spec.MachineConfigPoolSelector)
+		if err != nil {
+			return nil, err
+		}
+
+		if selector.Matches(labels.Set(mcp.Labels)) {
+			return mcp, nil
+		}
+	}
+	return nil, fmt.Errorf("no match")
```
yanirq, line 122: we can use a more elaborate error message here

controllers/reconcile_configmap.go
fromanirh: We have a completely separate controller which watches kubeletconfigs and maintains configMaps which are then mounted by daemonsets, handled by the existing controller.
At this point in time, there is no sync whatsoever between the two controllers, but I think we will need some form of code sharing in the future, and anticipating this the code is arranged to allow that already

main.go
fromanirh: note we are using structured logging here (`ErrorS` vs `Exitf`)

controllers/reconcile_configmap.go
fromanirh: yeah, I'll be handling on a separate PR

controllers/kubeletconfig_controller.go
```diff
@@ -0,0 +1,97 @@
...
+		namespace: r.Namespace,
+		instance:  instance,
+	}
+	cm, err := helper.reconcileConfigMap(ctx, req.NamespacedName)
+	if err != nil {
+		klog.ErrorS(err, "failed to reconcile configmap", "controller", "kubeletconfig")
+
+		msg := fmt.Sprintf("Failed to update RTE config from kubelet config %s/%s", req.NamespacedName.Namespace, req.NamespacedName.Name)
+		r.Recorder.Event(instance, "Warning", "ProcessFailed", msg)
+		return ctrl.Result{RequeueAfter: RetryPeriod}, err
```
yanirq, line 84: RetryPeriod will be taken from the numaresourcesoperator_controller,  if they might have different values then consider adding the constant also here.

controllers/reconcile_configmap.go
```diff
@@ -0,0 +1,144 @@
...
+			ReservedCPUs: klConfig.ReservedSystemCPUs,
+		},
+		TopologyManagerPolicy: klConfig.TopologyManagerPolicy,
+		TopologyManagerScope:  klConfig.TopologyManagerScope,
+	}
+	data, err := yaml.Marshal(conf)
+	if err != nil {
+		return nil, err
+	}
+	return rtemanifests.CreateConfigMap(namespace, name, string(data)), nil
```
yanirq, line 143: Do we need to handle also deletion of config maps ?

bundle/manifests/nodetopology.openshift.io_numaresourcesoperators.yaml
```diff
@@ -179,6 +179,55 @@ spec:
                       type: string
                   type: object
                 type: array
+              machineconfigpools:
```
cynepco3hahue, line 182: hm looks like rebasing issue

config/rbac/role.yaml
```diff
@@ -36,6 +43,14 @@ rules:
   - clusterversions
   verbs:
   - list
+- apiGroups:
```
cynepco3hahue, line 46: hm does it work, I was sure that for cluster scope resources you need cluster role

controllers/kubeletconfig_controller.go
cynepco3hahue: we will ignore the RetryPeriod here, see - https://github.com/kubernetes-sigs/controller-runtime/blob/8f43f26d53ffb08a96ffce1298df0cd761f8be0d/pkg/internal/controller/controller.go#L311

controllers/kubeletconfig_controller.go
```diff
@@ -0,0 +1,93 @@
...
+			return ctrl.Result{RequeueAfter: RetryPeriod}, nil
+		}
+		// Error reading the object - requeue the request.
+		return ctrl.Result{}, err
+	}
+
+	// KubeletConfig changes are expected to be sporadic, yet are important enough
+	// to be made visible at kubernetes level. So we generate events to handle them
+
+	helper := configMapHelper{
```
cynepco3hahue, line 68: Any reason to introduce a helper here? Can we  just do something like
```
func (r *KubeletConfigReconciler) syncConfigMap(ctx, instance) ...
```

fromanirh: yes, fixing

config/rbac/role.yaml
fromanirh: let me check the annotations

bundle/manifests/nodetopology.openshift.io_numaresourcesoperators.yaml
fromanirh: probably this should be handled by a separate PR, on it

fromanirh: https://github.com/openshift-kni/numaresources-operator/pull/41

config/rbac/role.yaml
fromanirh: note `config/rbac/role.yaml` is deceptive: we actually have a `ClusterRole` in here

controllers/reconcile_configmap.go
fromanirh: handled by the owner reference (or it's a bug)

controllers/kubeletconfig_controller.go
fromanirh: yeah, old code never updated. Removed. Thanks!

cynepco3hahue:
/hold
/lgtm
feel free to unhold when you feel comfortable ;)

fromanirh:
> /hold /lgtm feel free to unhold when you feel comfortable ;)

sure, want to add some tests before

controllers/kubeletconfig_controller.go
```diff
@@ -0,0 +1,200 @@
...
+
+func (r *KubeletConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&mcov1.KubeletConfig{}).
+		Owns(&corev1.ConfigMap{}).
+		Complete(r)
+}
+
+func (r *KubeletConfigReconciler) reconcileConfigMap(ctx context.Context, instance *nropv1alpha1.NUMAResourcesOperator, kcKey client.ObjectKey) (*corev1.ConfigMap, error) {
+	mcoKc := &mcov1.KubeletConfig{}
```
Tal-or, line 112: Do we know for certain that there's only one?

fromanirh: zero or one, yes. `KubeletConfig` is cluster scoped, so it must be unique by name.

PR #44, opened 2021-12-03
e2e: wait for MCP resources

fromanirh:
we need to wait longer and explicitely for MCPs to be updated on OCP; otherwise the test is misleading - and prone to false
negatives.

The behaviour of the wait is controlled by env variables, to allow tuning without code changes.

This PR includes some refactoring - code movement - to allow to reuse the same code we have in the controllers.

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -131,7 +131,7 @@ func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctr
 		return r.updateStatus(ctx, instance, status.ConditionDegraded, validation.NodeGroupsError, err.Error())
 	}
 
-	mcps, err := getNodeGroupsMCPs(ctx, r.Client, instance.Spec.NodeGroups)
+	mcps, err := GetNodeGroupsMCPs(ctx, r.Client, instance.Spec.NodeGroups)
 	if err != nil {
 		return r.updateStatus(ctx, instance, status.ConditionDegraded, validation.NodeGroupsError, err.Error())
 	}
@@ -260,7 +260,7 @@ func (r *NUMAResourcesOperatorReconciler) syncMachineConfigPoolsStatuses(instanc
 			Conditions: mcp.Status.Conditions,
 		})
 
-		if !isMachineConfigPoolUpdated(instance.Name, mcp) {
+		if !IsMachineConfigPoolUpdated(instance.Name, mcp) {
 			allMCPsUpdated = false
 			break
 		}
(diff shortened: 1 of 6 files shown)


fromanirh:
once this PR lands, we will need to make sure the rte e2e suite which we import works correctly on OCP; since it is passing on https://github.com/openshift-kni/resource-topology-exporter , it should boil down to make sure we run the suite the same way.
Once that is done, we can claim the e2e suite works on OCP, so we can the enable prow lane at last!
cc @Tal-or @mmirecki @cynepco3hahue

controllers/numaresourcesoperator_controller.go
```diff
@@ -248,6 +250,24 @@ func (r *NUMAResourcesOperatorReconciler) syncMachineConfigs(ctx context.Context
 	return nil
 }
 
+func (r *NUMAResourcesOperatorReconciler) syncMachineConfigsStatuses(instance *nropv1alpha1.NUMAResourcesOperator, mcps []*machineconfigv1.MachineConfigPool) bool {
```
cynepco3hahue, line 253: `syncMachineConfigPoolsStatuses`

test/e2e/e2e_suite_test.go
```diff
@@ -19,21 +19,41 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"os"
 	"testing"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
cynepco3hahue, line 28: nit: sort imports

test/e2e/e2e_suite_test.go
```diff
@@ -43,29 +63,116 @@ var _ = BeforeSuite(func() {
...
+
+	if clusterPlatform == platform.Kubernetes {
+		mcpObj := objects.TestMCP()
+		By(fmt.Sprintf("creating the machine config pool object: %s", mcpObj.Name))
+		err := e2eclient.Client.Create(context.TODO(), mcpObj)
+		Expect(err).NotTo(HaveOccurred())
+		matchLabels = map[string]string{"test": "test"}
+	}
+	if clusterPlatform == platform.OpenShift {
+		// TODO: should this be configurable?
```
cynepco3hahue, line 79: probably yes, but we should leave it for the additional PR

controllers/numaresourcesoperator_controller.go
fromanirh: done

test/e2e/e2e_suite_test.go
fromanirh: done

PR #45, opened 2021-12-03
add numaresourcesscheduler API

fromanirh:
Sketch out the numaresourcesscheduler API and bootstrap empty controller

We want to offer a simple alternative to the [secondary scheduler operator](https://github.com/openshift/secondary-scheduler-operator). This controller will have the strictest subset of features needed to deploy the [noderesourcetopologies scheduler plugin](https://github.com/openshift-kni/scheduler-plugins).

This controller will offer basic functionaliy: installation and complete cleanup. There is no seamless upgrade path to the secondary scheduler operator: the cluster admin must have to deinstall the scheduler plugin using this operator and then reinstall uisng the secondary scheduler plugin.

The scheduler plugin is stateless and quick to setup, so this operation is expected to cause minimal to none downtime.


pkg/numaresourcesscheduler/manifests/manifests.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	if namespace != "" {
+		sa.Namespace = namespace
+	}
+	return sa, nil
+}
+
+func ClusterRole() (*rbacv1.ClusterRole, error) {
+	obj, err := loadObject(filepath.Join("yaml", "clusterrole.yaml"))
```
Tal-or, line 52: ```clusterrole.nrt.yaml```

fromanirh: fixed

PR #46, opened 2021-12-07
bundle: download the operator sdk

cynepco3hahue:


Makefile
```diff
@@ -203,6 +213,17 @@ bundle-build: ## Build the bundle image.
 bundle-push: ## Push the bundle image.
 	$(MAKE) container-push IMG=$(BUNDLE_IMG)
 
+.PHONY: operator-sdk
+operator-sdk:
```
fromanirh, line 217: nice catch, but please let's be consistent to what the Makefile already does for `opm`. So either let's cache all the tools or none of the tools.

bundle/manifests/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -256,13 +256,13 @@ spec:
         serviceAccountName: numaresources-controller-manager
     strategy: deployment
   installModes:
-  - supported: false
+  - supported: true
```
fromanirh, line 259: please elaborate in the commit message about why this change is needed

cynepco3hahue: done

Makefile
cynepco3hahue: done

fromanirh (commented):
I don't like the caching too much, it backfired more than once. But we can fix later.

bundle/manifests/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -256,13 +256,13 @@ spec:
...
     type: OwnNamespace
-  - supported: false
+  - supported: true
     type: SingleNamespace
   - supported: false
     type: MultiNamespace
-  - supported: true
+  - supported: false
```
fromanirh, line 265: yes, this makes sense with the direction of the operator as it we see at the moment.

fromanirh (commented):
/lgtm

PR #47, opened 2021-12-07
E2E bootstrap scripts

Tal-or:
Add scrips for e2e tests bootstrap
Couple of scripts that will bootstrap the environment so we'll be able to run e2e in it:
1. `run-e2e-kni.sh` - This is the main script that fires all the others. Any environment variable that is relevant for the e2e tests should be added here.
2. `kube-update.sh` - Creates a kubeletconfig resource with necessary configuration for RTE/NFD deployment.
3. `label-custom-kubelet.sh` - Labels MCP with a unique label that kubeletconfig resource (from 2) will look for in the machineConfigPoolSelector field.
4. `wait-for-mcp` - Waits for MCP to apply the new MC changes.
5. `setup-e2e.sh` - Setups a namespace where all tests components should be deployed in.

### DIFF
--- a/hack/label-custom-kubelet.sh
+++ b/hack/label-custom-kubelet.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+# expect oc to be in PATH by default
+OC_TOOL="${OC_TOOL:-oc}"
+
+# Label worker MCP
+echo "[INFO]: Labeling worker MCP with custom-kubelet"
+mcp=$(${OC_TOOL} get mcp --selector='pools.operator.machineconfiguration.openshift.io/worker=' -o name)
+
+${OC_TOOL} label --overwrite "$mcp" custom-kubelet=enabled
--- a/hack/setup-e2e.sh
+++ b/hack/setup-e2e.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+set -xue
+
+# expect oc to be in PATH by default
+OC_TOOL="${OC_TOOL:-oc}"
+
+E2E_NAMESPACE_NAME="${E2E_NAMESPACE_NAME:-numaresources-operator-e2e}"
+
+function create_ns() {
+echo "Create $E2E_NAMESPACE_NAME namespace"
+  cat << EOF | $OC_TOOL apply -f -
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: "$E2E_NAMESPACE_NAME"
+EOF
+}
+
+create_ns
--- a/hack/wait-for-mcp.sh
+++ b/hack/wait-for-mcp.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+set -e
+
+# expect oc to be in PATH by default
+OC_TOOL="${OC_TOOL:-oc}"
+
+success=0
+iterations=0
+sleep_time=10
+max_iterations=180 # results in 30 minute timeout
+
+# Let's give the operator some time to do its work before we unpause the MCP (see below)
+echo "[INFO] Waiting a bit for letting the operator do its work"
+sleep 30
+
+until [[ $success -eq 1 ]] || [[ $iterations -eq $max_iterations ]]
+do
+
+  echo "[INFO] Checking if MCP picked up the custom Kubelet configuration"
+  mc=$("$OC_TOOL" get mcp worker -o jsonpath='{.spec.configuration.source[?(@.name=="99-worker-generated-kubelet")]}')
+  # No output means that the new Kubelet config wasn't picked by MCO yet
+  if [ -z "$mc" ]
+  then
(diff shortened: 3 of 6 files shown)


hack/wait-for-mcp.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Let's give the operator some time to do its work before we unpause the MCP (see below)
+echo "[INFO] Waiting a bit for letting the operator do its work"
+sleep 30
+
+until [[ $success -eq 1 ]] || [[ $iterations -eq $max_iterations ]]
+do
+
+  echo "[INFO] Checking if MCP picked up the custom Kubelet configuration"
+  mc=$("$OC_TOOL" get mcp worker -o jsonpath='{.spec.configuration.source[?(@.name=="99-worker-generated-kubelet")]}')
```
cynepco3hahue, line 21: I think it better to waiting first for updating state and after for updated state

hack/setup-e2e.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
cynepco3hahue, line 1: not need this one, the CI will create the namespace for us

hack/wait-for-mcp.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
fromanirh, line 1: we need to check overlaps with https://github.com/openshift-kni/numaresources-operator/pull/44/commits/cb2d540cf0a87a63def2fcfcf6373ef5dbfe9258

hack/wait-for-mcp.sh
```diff
@@ -0,0 +1,52 @@
...
+  then
+    iterations=$((iterations + 1))
+    iterations_left=$((max_iterations - iterations))
+    echo "[INFO] Custom Kubelet configuration not picked up yet. $iterations_left retries left."
+    sleep $sleep_time
+    continue
+  fi
+
+  echo "[INFO] Checking if MCP is updated"
+  if ! ${OC_TOOL} wait mcp/worker --for condition=Updated --timeout 1s &> /dev/null
```
cynepco3hahue, line 33: I remember it was some issues to use wait, but I do not remember what was the issue :(
Like we can just do
`${OC_TOOL} wait mcp/worker --for condition=Updated --timeout 15m`

hack/label-custom-kubelet.sh
```diff
@@ -0,0 +1,12 @@
...
+set -e
+
+# expect oc to be in PATH by default
+OC_TOOL="${OC_TOOL:-oc}"
+
+# Label worker MCP
+echo "[INFO]: Labeling worker MCP with custom-kubelet"
+mcp=$(${OC_TOOL} get mcp --selector='pools.operator.machineconfiguration.openshift.io/worker=' -o name)
+
+${OC_TOOL} label --overwrite "$mcp" custom-kubelet=enabled
```
fromanirh, line 12: we should minimize node reboots, because we should try hard to make the e2e lane as fast as possible. Fast lane -> we ran it often -> good signal, less bitrot.

cynepco3hahue: hh, do you have proposals? Like probably we can do this kind of configuration from test:
1. Pause MCP
2. Create NRO CR
3. Create custom kubeletconfig and label the MCP
4. Unpause the MCP

Tal-or: Sorry but why we would have node reboot here? it's just labeling the MCP

hack/setup-e2e.sh
Tal-or: But how do we passing the ns created by the CI to the test itself?

hack/label-custom-kubelet.sh
fromanirh: @cynepco3hahue  yes, pausing, then creating NRO, then create custom kubelet config+ label (or the previous two in any order) then unpause is propably the best we can do. I'd not complicate the flow any further until we try this out.
@Tal-or the labeling per se won't of course cause a reboot, and neither will the kubeletconfig alone, but once MCO notices that both the `kubeletconfig` matches a MCP through the label, it will trigger a reboot.

fromanirh:
probably some of these scripts should be ported to golang code inside the testsuite - like we added the wait for MCP

hack/label-custom-kubelet.sh
Tal-or: I see. so we can save one reboot by applying this changes and the one from https://github.com/openshift-kni/numaresources-operator/commit/cb2d540cf0a87a63def2fcfcf6373ef5dbfe9258 together.

Tal-or:
> probably some of these scripts should be ported to golang code inside the testsuite - like we added the wait for MCP

Which part do you think is important to get ported?

Tal-or:
Deprecated by #48 and #49

PR #48, opened 2021-12-08
e2e: make sure that we always call install NRO test suite first

cynepco3hahue:
Refactor e2e tests by adding additional test suits for install and uninstall.
Provide the script to run function tests, under the script we call separately
install test suite first, after all other e2e tests and the script always calls
to uninstall if script exited or terminated.


test/e2e/rte/rte_e2e_suite_test.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package rte_e2e
+package rte
```
fromanirh, line 17: why?

test/install/install.go
```diff
@@ -0,0 +1,109 @@
+/*
```
fromanirh, line 1: why we are removing the `e2e` component of the path?

test/utils/consts/consts.go
```diff
@@ -0,0 +1,69 @@
...
+	defaultMCPUpdateInterval = 30 * time.Second
+)
+
+var (
+	Platform                        platform.Platform
+	MachineConfigPoolUpdateTimeout  time.Duration
+	MachineConfigPoolUpdateInterval time.Duration
+)
+
+func init() {
```
fromanirh, line 44: it feels awkward to have this code running in the `consts` subpackage

fromanirh (commented):
I think the overall idea is good, but there are some implementation detail that I'd like to discuss

fromanirh:
what's the relationship between this PR and https://github.com/openshift-kni/numaresources-operator/pull/47 ?

hack/run-functests.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+source hack/common.sh
+
+NO_COLOR=""
```
Tal-or, line 5: Is this part really necessary?
To me, it looks like it's just some additional burden on the eyes when reading the code.

Tal-or:
> what's the relationship between this PR and #47 ?

#47 makes sure to add some additional configuration to the cluster in general, like kubeletconfig with the needed parameters(reserverd CPU/Memory, etc.)
But both PRs can't coexist without significant changes.
I think I'll remove #47 and create the kubeletconfig as part of the install test_suite, WDYT?

hack/run-functests.sh
```diff
@@ -0,0 +1,31 @@
...
+fi
+
+
+echo "Running Functional Tests: ${GINKGO_SUITS}"
+# -v: print out the text and location for each spec before running it and flush output to stdout in realtime
+# -r: run suites recursively
+# --failFast: ginkgo will stop the suite right after the first spec failure
+# --flakeAttempts: rerun the test if it fails
+# -requireSuite: fail if tests are not executed because of missing suite
+${BIN_DIR}/e2e-rte.test ${NO_COLOR} --ginkgo.v --ginkgo.failFast --ginkgo.flakeAttempts=2 --ginkgo.reportFile=/tmp/artifacts/e2e
```
Tal-or, line 31: nit: new line

fromanirh:
> > what's the relationship between this PR and #47 ?
> 
> #47 makes sure to add some additional configuration to the cluster in general, like kubeletconfig with the needed parameters(reserverd CPU/Memory, etc.) But both PRs can't coexist without significant changes. I think I'll remove #47 and create the kubeletconfig as part of the install test_suite, WDYT?

yes, let's go this direction please.

hack/run-functests.sh
fromanirh: it can be reintroduced later, but I see the value here. If the tests are trying to color the logs while running on an environment which doesn't support so (CI), the final result is almost unreadable.

Makefile
```diff
@@ -100,8 +100,8 @@ test: manifests generate fmt vet envtest ## Run tests.
 test-unit: envtest
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./controllers/... ./pkg/...
 
-test-e2e: binary-e2e binary-rte-e2e
-	./bin/e2e.test -ginkgo.v
+test-e2e: binary-e2e-install binary-e2e-rte binary-e2e-uninstall
+	hack/run-functests.sh
```
fromanirh, line 104: let's rename this to be like `run-e2e.sh` or `run-e2e-test.sh` or something like this.

test/install/install.go
cynepco3hahue: will move it back

test/utils/consts/consts.go
cynepco3hahue: Somehow took it from PAO :) https://github.com/openshift-kni/performance-addon-operators/blob/master/functests/utils/consts.go

because we import const package anyway, do you want to rename consts to something else?

test/e2e/rte/rte_e2e_suite_test.go
fromanirh: sure, but let's have either `_e2e` in the package name or move it back under `test/e2e` like you already want to do (https://github.com/openshift-kni/numaresources-operator/pull/48#discussion_r765575929) . I prefer the latter.

test/utils/consts/consts.go
fromanirh: PAO does a lot of things right and nice, but not everything (yet!) :)
In this case, I prefer to have a `consts` package which doesn't run any code, if possible. Just renaming/moving the code around is fine by me.

Makefile
cynepco3hahue: done

hack/run-functests.sh
cynepco3hahue: done

test/utils/consts/consts.go
cynepco3hahue: renamed to `configuration`, please let me know if it good enough

test/install/install.go
cynepco3hahue: done

test/e2e/rte/rte_e2e_suite_test.go
cynepco3hahue: you can see that RTE tests are already placed under `test/e2e/rte` does it good enough or do you want some additional changes?

fromanirh: nope, that's good enough thanks!

test/utils/configuration/configuration.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
fromanirh, line 2: tha new name is cool!

test/e2e/install/install.go
```diff
@@ -0,0 +1,109 @@
...
+
+		}
+		initialized = true
+	})
+
+	Context("with a running cluster with all the components", func() {
+		It("should perform overall deployment and verify the condition is reported as available", func() {
+			var matchLabels map[string]string
+
+			if configuration.Platform == platform.Kubernetes {
```
Tal-or, line 54: This is fine that we're creating MCP resources in k8s platform, but what about the CRD itself?

test/e2e/install/install.go
```diff
@@ -0,0 +1,109 @@
...
+				Expect(err).NotTo(HaveOccurred())
+				matchLabels = map[string]string{"test": "test"}
+			}
+
+			if configuration.Platform == platform.OpenShift {
+				// TODO: should this be configurable?
+				matchLabels = map[string]string{"pools.operator.machineconfiguration.openshift.io/worker": ""}
+			}
+
+			nroObj := objects.TestNRO(matchLabels)
```
Tal-or, line 67: Same here, who make sure the CRD is already deployed?

test/e2e/install/install.go
Tal-or: BTW this can be addressed on a separate PR, since this code was just copied.

test/e2e/install/install.go
Tal-or: BTW this can be addressed on a separate PR, since this code was just copied.

test/e2e/install/install.go
cynepco3hahue: Its part of the deployment procedure, the same true regarding NRO CRD

Tal-or: What deployment procedure? 
we're using kustomize, but I don't see MCP CRD here: https://github.com/openshift-kni/numaresources-operator/tree/main/config/crd

cynepco3hahue: https://github.com/openshift-kni/numaresources-operator/blob/939709fee7a1c4b58463cb66e353f723139c2c90/Makefile#L155

PR #49, opened 2021-12-09
e2e: create kubeletconfig resource

Tal-or:
During the e2e test initialization we should create a kubeletconfig object that will apply all the necessary configuration for a proper functionality of RTE

### DIFF
--- a/test/e2e/install/install.go
+++ b/test/e2e/install/install.go
@@ -21,13 +21,14 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
 	nropv1alpha1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1alpha1"
 	"github.com/openshift-kni/numaresources-operator/pkg/status"
 	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
@@ -65,9 +66,21 @@ var _ = Describe("[Install]", func() {
 			}
 
 			nroObj := objects.TestNRO(matchLabels)
+			kcObj, err := objects.TestKC(matchLabels)
+			Expect(err).To(Not(HaveOccurred()))
 
-			By(fmt.Sprintf("Creating the NRO object: %s", nroObj.Name))
-			err := e2eclient.Client.Create(context.TODO(), nroObj)
+			unpause, err := machineconfigpools.PauseMCPs(nroObj.Spec.NodeGroups)
+			Expect(err).NotTo(HaveOccurred())
+
+			By(fmt.Sprintf("creating the KC object: %s", kcObj.Name))
+			err = e2eclient.Client.Create(context.TODO(), kcObj)
+			Expect(err).NotTo(HaveOccurred())
+
+			By(fmt.Sprintf("creating the NRO object: %s", nroObj.Name))
+			err = e2eclient.Client.Create(context.TODO(), nroObj)
+			Expect(err).NotTo(HaveOccurred())
+
+			err = unpause()
 			Expect(err).NotTo(HaveOccurred())
 
 			err = e2eclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(nroObj), nroObj)
--- a/test/e2e/uninstall/uninstall.go
+++ b/test/e2e/uninstall/uninstall.go
@@ -19,19 +19,21 @@ package uninstall
 import (
 	"context"
 
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
+	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
 	"github.com/openshift-kni/numaresources-operator/controllers"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectstate/rte"
 	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
 	"github.com/openshift-kni/numaresources-operator/test/utils/configuration"
+	"github.com/openshift-kni/numaresources-operator/test/utils/machineconfigpools"
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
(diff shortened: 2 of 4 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -79,11 +79,19 @@ var _ = BeforeSuite(func() {
...
+
+	By(fmt.Sprintf("creating the KC object: %s", kcObj.Name))
+	err = e2eclient.Client.Create(context.TODO(), kcObj)
+	Expect(err).NotTo(HaveOccurred())
 
 	By(fmt.Sprintf("creating the NRO object: %s", nroObj.Name))
 	err = e2eclient.Client.Create(context.TODO(), nroObj)
 	Expect(err).NotTo(HaveOccurred())
 
+	pauseMCPs(nroObj.Spec.NodeGroups, false)
```
fromanirh, line 93: next time consider creating two functions:
`func pauseMCPs(nodeGroups []nropv1alpha1.NodeGroup)`
and
`func unpauseMCPs(nodeGroups []nropv1alpha1.NodeGroup)`
The reason is that the bool argument feels clunky. This approach may be more expressive and we can share quite a lot of common code anyway

test/e2e/utils/objects/objects.go
```diff
@@ -66,3 +73,49 @@ func TestMCP() *machineconfigv1.MachineConfigPool {
...
+				Raw: getEncodedKCConfig(),
+			},
+			MachineConfigPoolSelector: &metav1.LabelSelector{
+				MatchLabels: matchLabels,
+			},
+		},
+	}
+}
+
+func getEncodedKCConfig() []byte {
```
fromanirh, line 98: next time please consider creating a function which just returns a fully initialized `kubeletconfigv1beta1.KubeletConfiguration` object and encode it in place in `TestKC`. This will make the code more composable

test/e2e/e2e_suite_test.go
```diff
@@ -79,11 +79,19 @@ var _ = BeforeSuite(func() {
 	}
 
 	nroObj := objects.TestNRO(matchLabels)
+	kcObj := objects.TestKC(matchLabels)
+	pauseMCPs(nroObj.Spec.NodeGroups, true)
```
fromanirh, line 83: I think we should pause BEFORE creating NRO and KC objects

fromanirh (commented):
just one comment about the right moment on which we should pause, everything else is good enough IMO

test/e2e/e2e_suite_test.go
Tal-or: Initially, I went with this approach but then I saw the functions are 100% the same but the ```pause``` field.
So I thought it's a huge waste of repetitive code.

test/e2e/e2e_suite_test.go
Tal-or: This is not the creation, we're just initializing the structs.
The real creation happens later when an API create call is executed by the controller-runtime client

fromanirh: ...which we are btw. We do pause before to submit the objects to the cluster.

test/e2e/e2e_suite_test.go
```diff
@@ -174,3 +187,21 @@ func getMCPUpdateValueFromEnv(envVar string, fallback time.Duration) time.Durati
 	Expect(err).NotTo(HaveOccurred())
 	return timeout
 }
+
+func pauseMCPs(nodeGroups []nropv1alpha1.NodeGroup, pause bool) {
+	mcps, err := controllers.GetNodeGroupsMCPs(context.TODO(), e2eclient.Client, nodeGroups)
+	if err != nil {
+		klog.Warningf("failed to get the MCPs resources: %v", err)
+		return
```
fromanirh, line 195: on second thought we should `ExpectNoError` here. Failing to get this information is worth aborting the test. Can be fixed later.

test/e2e/e2e_suite_test.go
```diff
@@ -174,3 +187,21 @@ func getMCPUpdateValueFromEnv(envVar string, fallback time.Duration) time.Durati
...
+	}
+
+	if len(mcps) == 0 {
+		klog.Warningf("no MCPs resources found", err)
```
fromanirh, line 199: again this is more likely a bug than something else. We should `Expect` 1+ MCPs to be found.

test/e2e/e2e_suite_test.go
fromanirh: What I'd probably do is something like
```
func pauseMCPs(nodeGroups []nropv1alpha1.NodeGroup) func() {
	mcps, err := controllers.GetNodeGroupsMCPs(context.TODO(), e2eclient.Client, nodeGroups)
	Expect(len(mcps).To(BeGreatherThan(0).ICantRememberThe(Syntax).Please(Check))
	for i := range mcps {
		mcps[i].Spec.Paused = true
		err = e2eclient.Client.Update(context.TODO(), mcps[i])
	}
	unpause := func() {
		for i := range mcps {
			mcps[i].Spec.Paused = false
			err = e2eclient.Client.Update(context.TODO(), mcps[i])
		}
	}
	return unpause
}
```
Use like

unpause := pauseMCPs(nroObj.Spec.NodeGroups)

// do stuff

unpause()
```

the advantage is we make sure to unpause the same set of MCPs we pause. It's unlikely some MCPs get created in between, but from theoretical perspective this is pretty clean and not too clunky. Note this is a pretty common golang/kube pattern (returning an anonymous function which does cleanup), not inventing anything here.

Another option is to save the MCPs we pause. Not strong feelings against that, it's ok as well.

Comment (fromanirh) [reply test/e2e/e2e_suite_test.go:93]: if we go the fancy way with the anonymous func, let's make sure we capture the MCPs properly in the closure, this is a common golang bug.

**Review by fromanirh — COMMENTED** [review]
suggested a possible improvement I will NOT push for.

File: test/e2e/e2e_suite_test.go (hunk 3 above)
Comment (fromanirh) [reply test/e2e/e2e_suite_test.go:83]: > This is not the creation, we're just initializing the structs here. The real creation happens later when an API create call is executed by the controller-runtime client

aye, I read too hastily. What you're doing here is fine.

File: test/e2e/e2e_suite_test.go (hunk 1 above)
Comment (Tal-or) [reply test/e2e/e2e_suite_test.go:93]: This is such a neat trick! I'll totally adopt it.

> let's make sure we capture the MCPs properly in the closure

What do you mean by that, from your code example it looks like we're working on the same group of MCPs so it looks ok to me

File: test/e2e/e2e_suite_test.go [hunk 6]
```diff
@@ -174,3 +187,21 @@ func getMCPUpdateValueFromEnv(envVar string, fallback time.Duration) time.Durati
...
+	}
+
+	for i := range mcps {
+		mcps[i].Spec.Paused = pause
+		err = e2eclient.Client.Update(context.TODO(), mcps[i])
```
Tal-or, line 204: @fromanirh is there a way to update all the MCPs together, using ```machineConfigList```?

fromanirh: I actually don't know

test/e2e/e2e_suite_test.go
fromanirh: We need to make sure we don't introduce hard-to see bugs caused to shadowing. `for` loops and closures are the usual suspects for this kind of bugs. Code looks correct, yet it misbehaves. It's one of the most common golang gotchas. See for example: https://blog.charmes.net/post/scope-shadowing-go/

fromanirh: that said, it is possible my code is actually correct, but I didn't spent much thought on it, so it must be taken with a big grain of salt.

fromanirh (approved):
/lgtm
/approve

PR #53, opened 2021-12-14
RTE: Consume config changes

fromanirh:
Add facilities to react to config file changes in turn generated by changes of the `kubeletconfig` objects, which are already being handled by the kubeletconfig controller.


fromanirh:
/hold
I'd like a e2e test here

fromanirh:
/hold cancel
added tests in between (and uncovered bugs in the process leading to separate fixes)

PR #55, opened 2021-12-15
logs: init klog flags

Tal-or:

### DIFF
--- a/main.go
+++ b/main.go
@@ -94,6 +94,7 @@ func main() {
 		Development: true,
 	}
 	opts.BindFlags(flag.CommandLine)
+	klog.InitFlags(nil)
 	flag.Parse()
 
 	// if it is unknown, it's fine


main.go
```diff
@@ -94,6 +94,7 @@ func main() {
 		Development: true,
 	}
 	opts.BindFlags(flag.CommandLine)
+	klog.InitFlags(nil)
```
fromanirh, line 97: I'm, not sure the argument is correct

Tal-or: Same as in PAO https://github.com/openshift-kni/performance-addon-operators/blob/master/main.go#L89

fromanirh: OK, we do what I'd expect inside the function itself: https://github.com/Tal-or/numaresources-operator/blob/387e57987d0485850698e61239c75d1e3adc7f9e/vendor/k8s.io/klog/v2/klog.go#L425

PR #60, opened 2021-12-20
nro controller: generate the machine config label if it impossible to get labels from the machine config pool

cynepco3hahue:
Provide a backup plan when the targeted machine config pool does not have a machine config
selector with matchLabels field. In the above case, we will generate the label and drop a
warning message under the log.

It should be important for machine config pools that target more than one machine config
label. For example, our worker-cnf machine config pool definition looks, like this:
```
...
  machineConfigSelector:
    matchExpressions:
      - key: machineconfiguration.openshift.io/role
        operator: In
        values: [worker,worker-cnf]
...
```

It is important to say that the code also validates if the generated labels matched by
MachineConfigPool machine config selector, if it not matches the NRO will have
degraded state until the user will update the machine config pool or NRO object with
correct data.

### DIFF
--- a/pkg/objectstate/rte/rte.go
+++ b/pkg/objectstate/rte/rte.go
@@ -39,6 +39,9 @@ import (
 	"github.com/openshift-kni/numaresources-operator/pkg/objectstate/merge"
 )
 
+// MachineConfigLabelKey contains the key of generated label for machine config
+const MachineConfigLabelKey = "machineconfiguration.openshift.io/role"
+
 type daemonSetManifest struct {
 	daemonSet      *appsv1.DaemonSet
 	daemonSetError error
@@ -73,7 +76,7 @@ func (em *ExistingManifests) MachineConfigsState(mf rtemanifests.Manifests, inst
 			desiredMachineConfig := mf.MachineConfig.DeepCopy()
 			// prefix machine config name to guarantee that we will have an option to override it
 			desiredMachineConfig.Name = mcName
-			desiredMachineConfig.Labels = mcp.Spec.MachineConfigSelector.MatchLabels
+			desiredMachineConfig.Labels = GetMachineConfigLabel(mcp)
 
 			existingMachineConfig, ok := em.machineConfigs[mcName]
 			if !ok {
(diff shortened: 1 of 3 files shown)


cynepco3hahue:
We can improve it by adding events under the NRO object.

fromanirh:
> We can improve it by adding events under the NRO object.

yes please. Events AND logs seems the best solution.

pkg/objectstate/rte/rte.go
```diff
@@ -96,6 +96,19 @@ func (em *ExistingManifests) MachineConfigsState(mf rtemanifests.Manifests, inst
 	return ret
 }
 
+func getMachineConfigLabel(mcp *machineconfigv1.MachineConfigPool) map[string]string {
```
fromanirh, line 99: actually reading the code, I think we should just fail and put the conditions of the object in Degraded state.
This may require some internal refactoring, but better to avoid guesswork if we can help it.

fromanirh:
> > We can improve it by adding events under the NRO object.
> 
> yes please. Events AND logs seems the best solution.

...From reporting perspective in general. For this specific issue, we should probably just stop and put the NRO conditions to Degraded (see inline comments)

pkg/objectstate/rte/rte.go
cynepco3hahue: The problem is that we totally can have machine config pools as one that I specified under the description and I am assuming that we will want to use PAO together with the NRO.
Do you have any ideas on how to handle the above machine config pool?

fromanirh: Ah, right, that's another story entirely. We need to think about this

cynepco3hahue:
@fromanirh I refactored a little bit of the code to drop the NRO resource into the degraded state if the generated label does not satisfy the MCP machine config selector.

fromanirh:
> @fromanirh I refactored a little bit of the code to drop the NRO resource into the degraded state if the generated label does not satisfy the MCP machine config selector.

thanks! this seems the best approach

fromanirh (commented):
/approve
/lgtm
I think this is the best approach we can pursue atm.
We will revisit later if this proves insufficient or we need further changes.

PR #59, opened 2021-12-20
WIP: test e2e

fromanirh:
WIP TBD

PR #62, opened 2021-12-20
NUMAResourcesscheduler controller 

Tal-or:
This controller is supposed to replace the role of [SSOP](https://github.com/openshift/secondary-scheduler-operator) in our TAS solution until SSOP will catch up with the necessary changes for our needs.

The purpose of this controller is to be able to deploy a secondary scheduler (with a given image) with all the required components.

The code was rebased on top of https://github.com/openshift-kni/numaresources-operator/pull/45


fromanirh:
re: e2e test: feel free to cherry-pick https://github.com/openshift-kni/numaresources-operator/pull/59/commits/2a0639e4b2b5e2d81ebc56e5caf9c588d7d88386

pkg/objectstate/sched/sched.go
```diff
@@ -0,0 +1,50 @@
+package sched
```
fromanirh, line 1: please add the license boilerplate

pkg/objectstate/sched/sched.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	nrsv1alpha1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1alpha1"
+)
+
+const DefaultSchedulerPluginImage = "quay.io/openshift-kni/scheduler-plugins:4.10-snapshot"
```
fromanirh, line 12: there is not a safe fallback to be hardcoded; better to have this in the example CR rather than hardcoded; additionally, the `SchedulerImage` field should be mandatory.

pkg/objectstate/sched/sched.go
```diff
@@ -0,0 +1,50 @@
...
+	// There is only a single container
+	cnt := &dp.Spec.Template.Spec.Containers[0]
+	if userImageSpec != "" {
+		// we don't really know what's out there, so we minimize the changes.
+		cnt.Image = userImageSpec
+		klog.V(3).InfoS("Exporter image", "reason", "user-provided", "pullSpec", userImageSpec)
+		return
+	}
+
+	if builtinImageSpec == "" {
```
fromanirh, line 24: please remove

pkg/objectstate/sched/sched.go
```diff
@@ -0,0 +1,50 @@
...
+		cnt.Image = DefaultSchedulerPluginImage
+		klog.V(3).InfoS("Exporter image", "reason", "default", "pullSpec", builtinImageSpec)
+		return
+	}
+
+	cnt.Image = builtinImageSpec
```
fromanirh, line 30: for the RTE, we need this when we use the operator image as RTE image. For the scheduler we don't need to change the pullPolicy, it's sufficient to set it as we want in the `deployment.yaml`. Please remove.

pkg/objectstate/sched/sched.go
```diff
@@ -0,0 +1,50 @@
...
+	cnt.ImagePullPolicy = builtinPullPolicy
+	klog.V(3).InfoS("Exporter image", "reason", "builtin", "pullSpec", builtinImageSpec, "pullPolicy", builtinPullPolicy)
+}
+
+func UpdateDeploymentConfigMapSettings(dp *appsv1.Deployment, cmName string) {
+	for i := range dp.Spec.Template.Spec.Volumes {
+		if dp.Spec.Template.Spec.Volumes[i].Name == "etckubernetes" {
```
fromanirh, line 37: let's make this a constant, or let's just replace the `Volume` definition entirely:
```
func UpdateDeploymentConfigMap(dp *appsv1.Deployment, configMapName string) {
	spec := &dp.Spec.Template.Spec // shortcut
     // ...

	spec.Volumes[0] = newSchedConfigVolume(schedVolumeConfigName, configMapName)
}

func newSchedConfigVolume(schedVolumeConfigName, configMapName string) corev1.Volume {
	return corev1.Volume{
		Name: schedVolumeConfigName,
		VolumeSource: corev1.VolumeSource{
			ConfigMap: &corev1.ConfigMapVolumeSource{
				LocalObjectReference: corev1.LocalObjectReference{
					Name: configMapName,
				},
			},
		},
	}
}

File: controllers/numaresourcesscheduler_controller.go [hunk 6]
```diff
@@ -18,19 +18,46 @@ package controllers
...
+	"github.com/openshift-kni/numaresources-operator/pkg/objectstate/sched"
+	"github.com/openshift-kni/numaresources-operator/pkg/status"
 )
 
 // NUMAResourcesSchedulerReconciler reconciles a NUMAResourcesScheduler object
 type NUMAResourcesSchedulerReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme             *runtime.Scheme
+	Platform           platform.Platform
```
fromanirh, line 55: we don't need the platform here

controllers/numaresourcesscheduler_controller.go
```diff
@@ -18,19 +18,46 @@ package controllers
...
+	SchedulerManifests schedmanifests.Manifests
+	Helper             *deployer.Helper
+	Namespace          string
+	ImageSpec          string
+	ImagePullPolicy    corev1.PullPolicy
```
fromanirh, line 60: please remove

controllers/numaresourcesscheduler_controller.go
```diff
@@ -47,16 +74,110 @@ type NUMAResourcesSchedulerReconciler struct {
...
+
+func (r *NUMAResourcesSchedulerReconciler) reconcileResource(ctx context.Context, instance *nrsv1alpha1.NUMAResourcesScheduler) (reconcile.Result, string, error) {
+	klog.Info("SchedulerSync start")
+
+	deplomentInfo, err := r.syncNUMASchedulerResources(ctx, instance)
+	if err != nil {
+		return ctrl.Result{}, status.ConditionDegraded, errors.Wrapf(err, "FailedSchedulerSync")
+	}
+
+	instance.Status.Deployment = []nrsv1alpha1.NamespacedName{}
```
fromanirh, line 146: we will have exactly one deployment, so let's simplify this code.
I'm not sure we need this feature, but since it's already there, let's keep it.
Let's NOT add more features to this API and controller - we need to support and test them, and this controller has a fixed and limited timespan. Hardly worth the extra effort.

main.go
```diff
@@ -210,9 +211,17 @@ func main() {
 	}
```
fromanirh, line 212: what about the `render` support?

controllers/numaresourcesscheduler_controller.go
```diff
@@ -60,6 +60,13 @@ type NUMAResourcesSchedulerReconciler struct {
 	ImagePullPolicy    corev1.PullPolicy
 }
 
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles,verbs=*
```
fromanirh, line 63: unneeded, this controller does not need to access roles

controllers/numaresourcesscheduler_controller.go
```diff
@@ -60,6 +60,13 @@ type NUMAResourcesSchedulerReconciler struct {
...
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=rolebindings,verbs=*
```
fromanirh, line 64: unneeded, this controller does not need to access rolebindings

controllers/numaresourcesscheduler_controller.go
```diff
@@ -60,6 +60,13 @@ type NUMAResourcesSchedulerReconciler struct {
...
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles,verbs=*
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings,verbs=*
+//+kubebuilder:rbac:groups=apps,resources=deployments,verbs=*
+//+kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=*
+//+kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch
```
fromanirh, line 69: unneeded, this controller does not need to access pods

controllers/numaresourcesscheduler_controller.go
fromanirh, line 63: we miss the configmaps permissions:
```
//+kubebuilder:rbac:groups="",resources=configmaps,verbs=*
```

controllers/numaresourcesscheduler_controller.go
```diff
@@ -112,8 +112,26 @@ func (r *NUMAResourcesSchedulerReconciler) Reconcile(ctx context.Context, req ct
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *NUMAResourcesSchedulerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// we want to initiate reconcile loop only on change under labels or spec of the object
```
fromanirh, line 115: so, did you figure out why the `Own` did not work before? what was the root cause?

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/openshift-kni/numaresources-operator
 go 1.17
 
 require (
+	github.com/fsnotify/fsnotify v1.4.9
```
fromanirh, line 6: why do we need this change?

fromanirh (commented):
nice!

controllers/numaresourcesscheduler_controller.go
fromanirh: Additionally, did we extend the types definition to add this field to the `Status`?

main.go
fromanirh: actually let's postpone this to a later PR

controllers/numaresourcesscheduler_controller.go
Tal-or: We have it in the kubeletconfig_controller and the permissions for all the controllers are rendered under the same ```role.yaml``` file

controllers/numaresourcesscheduler_controller.go
Tal-or: Yes we do! I used the ```SetOwnerReference``` function instead of the ```SetControllerReference``` which doesn't trigger the controller's reconcile loop.

Thanks to @cynepco3hahue  for that.

go.mod
Tal-or: I'm not sure, need to check

controllers/numaresourcesscheduler_controller.go
fromanirh: sure. And each controller should add the annotations for each permission they need. Let `controller-gen` and the other automation coalesce them as needed.

controllers/numaresourcesscheduler_controller.go
fromanirh: Great, TIL.

go.mod
fromanirh: If not obviously needed, please remove to keep the PR minimal.

controllers/numaresourcesscheduler_controller.go
Tal-or: Yes we do https://github.com/Tal-or/numaresources-operator/blob/numares-sched-controller/api/numaresourcesoperator/v1alpha1/numaresourcesscheduler_types.go#L32

pkg/objectstate/sched/sched.go
Tal-or: Yes I noticed that. didn't get the chance to fix yet.

pkg/objectstate/sched/sched.go
Tal-or: This is not the ```pullPolicy``` but the image. I'll remove the ```pullPolicy```  field

controllers/numaresourcesscheduler_controller.go
Tal-or: In second thought I'll add it in order to be more explicit

go.mod
Tal-or: It was already required, please see some lines below.
Why it becomes from an indirect to a direct one, maybe someone forgot to run ```go mod``` but it wouldn't affect the compilation since only the ```//indirect``` comment was removed.

fromanirh: sure, and this change doesn't belong to this PR. So it should be moved on a separate PR and/or expunged by this one.

pkg/objectstate/sched/sched.go
fromanirh: please remove the `builtinImageSpec` from this function. We really need to require the user to set a `imageSpec`

pkg/objectstate/sched/sched.go
fromanirh: this still seems relevant

controllers/numaresourcesscheduler_controller.go
```diff
@@ -18,19 +18,36 @@ package controllers
...
 )
 
 // NUMAResourcesSchedulerReconciler reconciles a NUMAResourcesScheduler object
 type NUMAResourcesSchedulerReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme             *runtime.Scheme
+	SchedulerManifests schedmanifests.Manifests
+	Namespace          string
+	ImageSpec          string
```
fromanirh, line 54: did we fix the JSON tags in the type definition? (removing `omitempty`)

controllers/numaresourcesscheduler_controller.go
fromanirh: please change the `Status.Deployment` to be only a single object - not a slice

go.mod
fromanirh: please remove this commit from the series

fromanirh (commented):
thanks for the updates. There are some unanswered comments, bumped them.

Tal-or:
> thanks for the updates. There are some unanswered comments, bumped them.

I only resolved the comments that were already addressed.
I'll let you know when the PR will be ready for another review cycle

controllers/numaresourcesscheduler_controller.go
Tal-or: Is this change should be manually or I should use the operator-sdk framework for that?

pkg/objectstate/sched/sched.go
Tal-or: The builtinImageSpec does exactly that:
https://github.com/Tal-or/numaresources-operator/blob/numares-sched-controller/controllers/numaresourcesscheduler_controller.go#L168

I can change the name to be ```imageSpec``` if that's what you mean

controllers/numaresourcesscheduler_controller.go
fromanirh: The way it works best for you. I usually edit the struct and then regenerate the artifacts out of it (`make generate` , `make manifests`, `make bundle`...)

pkg/objectstate/sched/sched.go
fromanirh: we don't want a default here. We need the user to explicitely set it or fail.

controllers/numaresourcesscheduler_controller.go
Tal-or: yes we do

pkg/objectstate/sched/sched.go
```diff
@@ -0,0 +1,66 @@
...
+	cnt := &dp.Spec.Template.Spec.Containers[0]
+	cnt.Image = userImageSpec
+	klog.V(3).InfoS("Exporter image", "reason", "user-provided", "pullSpec", userImageSpec)
+}
+
+func UpdateDeploymentConfigMapSettings(dp *appsv1.Deployment, cmName string) {
+	spec := &dp.Spec.Template.Spec // shortcut
+	for i := range spec.Volumes {
+		if spec.Volumes[i].Name == SchedulerConfigMapVolumeName {
+			spec.Volumes[i] = newSchedConfigVolume(SchedulerConfigMapVolumeName, cmName)
```
fromanirh, line 41: add `return` here.
Or, since we control the deployment, just always unconditionally override the first volume

controllers/numaresourcesscheduler_controller.go
```diff
@@ -22,13 +22,17 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
```
fromanirh, line 25: let's remove this entire commit from the series please. I want to have a deeper dive in to the owner references. We can easily add it later.

fromanirh (commented):
looks good now, thanks for updating so fast! mostly minor comments pending, then we will need to rebase after #45 merges. We are very close.

fromanirh:
not sure this will merge cleanly without a rebase, let's see

PR #64, opened 2021-12-21
numaresourcesscheduler: post-merge minor cleanups

fromanirh:
minor post-merge cleanups


main.go
```diff
@@ -152,12 +153,23 @@ func main() {
...
 
 	if renderMode {
-		if err := renderObjects(
-			renderManifests(
-				rteManifests,
-				renderNamespace,
-				renderImage,
-			).ToObjects()); err != nil {
+		var objs []client.Object
+		if enableScheduler {
```
Tal-or, line 157: We'll only render the scheduler manifests here without the RTE one.
is that what we want?

fromanirh: Actually no because this is a different semantic than without the `-render` flag enabled. Let me fix it.

PR #69, opened 2022-01-02
Improve E2E coverage

Tal-or:
We need to control the order of the NUMAResourcesScheduler tests:
1. Install
2. Other tests
3. Uninstall

For that, we need each one of the above to be placed in a separate suite and call the suites in the desired order.
This PR intended to address exactly that.

In addition, the NUMAResourcesScheduler e2e tests are set as an opt-in for the e2e script.
Current behavior stays the same, i.e. NUMAResourcesScheduler e2e tests are not running.

### DIFF
--- a/test/e2e/sched/install/install.go
+++ b/test/e2e/sched/install/install.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package install
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	nropv1alpha1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1alpha1"
+	"github.com/openshift-kni/numaresources-operator/pkg/status"
+	schedutils "github.com/openshift-kni/numaresources-operator/test/e2e/sched/utils"
+	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
+	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
+)
+
+var _ = Describe("[Scheduler] install", func() {
+	Context("with a running cluster with all the components", func() {
+		It("should perform the scheduler deployment and verify the condition is reported as available", func() {
+			var err error
+			nroSchedObj := objects.TestNROScheduler()
+
+			By(fmt.Sprintf("creating the NRO Scheduler object: %s", nroSchedObj.Name))
+			err = e2eclient.Client.Create(context.TODO(), nroSchedObj)
+			Expect(err).NotTo(HaveOccurred())
+
+			err = e2eclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(nroSchedObj), nroSchedObj)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("checking that the condition Available=true")
+			Eventually(func() bool {
+				updatedNROObj := &nropv1alpha1.NUMAResourcesScheduler{}
+				err := e2eclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(nroSchedObj), updatedNROObj)
+				if err != nil {
+					klog.Warningf("failed to get the RTE resource: %v", err)
+					return false
+				}
+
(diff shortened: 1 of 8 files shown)


Tal-or:
The last commit in this PR is a hack for triggering the new tests. 
Once we'll see the tests are passing it shall be removed.

Tal-or:
/retest
looks like infra issue

test/e2e/sched/install/install.go
```diff
@@ -0,0 +1,55 @@
...
+
+				cond := status.FindCondition(updatedNROObj.Status.Conditions, status.ConditionAvailable)
+				if cond == nil {
+					klog.Warningf("missing conditions in %v", updatedNROObj)
+					return false
+				}
+
+				klog.Infof("condition: %v", cond)
+
+				return cond.Status == metav1.ConditionTrue
```
fromanirh, line 51: we should also add some extra testing (here or elsewhere) to test that if the condition available is true, well, the relevant pods are also running; having a condition set but no objects actually running is not that good :)

Could be done on a separate PR (it's probably the best approach actually)

test/e2e/sched/install/install.go
```diff
@@ -0,0 +1,55 @@
+package install
```
fromanirh, line 1: missing copyright boilerplate

test/e2e/sched/uninstall/uninstall.go
```diff
@@ -0,0 +1,35 @@
+package uninstall
```
fromanirh, line 1: missing copyright boilerplate

test/e2e/sched/uninstall/uninstall.go
```diff
@@ -0,0 +1,35 @@
...
+			// failed to get the NROScheduler object, nothing else we can do
+			if err := e2eclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(nroSchedObj), nroSchedObj); err != nil {
+				if !errors.IsNotFound(err) {
+					klog.Warningf("failed to get the NUMA resource scheduler %q: %w", nroSchedObj.Name, err)
+				}
+				return
+			}
+
+			err := e2eclient.Client.Delete(context.TODO(), nroSchedObj)
+			Expect(err).ToNot(HaveOccurred())
```
fromanirh, line 32: ok, and now we should check no dependent objects are left in the namespace, shouldn't we?

test/e2e/sched/sched.go
```diff
@@ -21,61 +21,74 @@ import (
...
+	. "github.com/onsi/gomega"
+
 	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
 )
 
-var _ = Describe("[Scheduler] install", func() {
-	var initialized bool
+// TODO change to an image from openshift-kni registry
+const newTestImage = "quay.io/titzhak/kube-scheduler:test"
```
fromanirh, line 37: yes, we need to have this fixed. We probably need a test image (created from the NROP sources by CI)

fromanirh (commented):
overall looks good, but there are some missing/pending parts inside.

test/e2e/sched/sched.go
Tal-or: If the test image is temporary, i.e. created during the CI run and deleted once the run is done, we'll have a problem with running the tests locally.

Maybe we can create an image and store it in a local registry?

fromanirh: I think we just need to have a separate image with RTE only (not the default one with NROP+RTE) that we can use freely for this kind of test. This image should be built by CI, uploaded on quay and kept here.

Tal-or: >This image should be built by CI, uploaded on quay and kept here.

AFAIK we don't have a procedure for that

fromanirh: yep, we should set it up as prerequisite for this PR.
Alternatively, we need to get rid of the tests requiring the separate image from this PR, and postpone until we have CI set.

Tal-or: I prefer the former because I need that for additional tests I'm writing.

Do we have an example for something similar in other projects/repo?

fromanirh: we "just" need to build a image using a Dockerfile we already have (may require some updates/fixes): https://github.com/openshift-kni/numaresources-operator/blob/main/Dockerfile.openshift.rte
The PAO flow builds quite a lot of images in CI, could be a good starting point

Tal-or: BTW I just noticed, we don't need RTE image here, but scheduler-plugins one:
https://github.com/openshift/release/pull/24949

fromanirh: right, I misread the imagespec

Tal-or: Now that we have the new tag, we should trigger it somehow

test/e2e/sched/sched.go
```diff
@@ -18,64 +18,57 @@ package sched
...
 			err = e2eclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(nroSchedObj), nroSchedObj)
-			Expect(err).NotTo(HaveOccurred())
+			Expect(err).ToNot(HaveOccurred())
+			Expect(nroSchedObj.GetUID()).To(BeEquivalentTo(uid))
 
-			By("checking that the condition Available=true")
 			Eventually(func() bool {
-				updatedNROObj := &nropv1alpha1.NUMAResourcesScheduler{}
-				err := e2eclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(nroSchedObj), updatedNROObj)
+				// find deployment by the ownerReference
```
fromanirh, line 63: this is pretty obvious from the code itself, but it's OK

fromanirh (commented):
/approve
/lgtm

PR #70, opened 2022-01-03
read tm scope from conf file

Tal-or:
In case the topology manager scope is configured in the conf file we should read the value from it instead of the default.


fromanirh (commented):
/lgtm
/approve

PR #72, opened 2022-01-03
Add nodeAffinity to run only on master

marioferh:
Add nodeAffinity to run only on master

### DIFF
--- a/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
@@ -229,6 +229,13 @@ spec:
               labels:
                 control-plane: controller-manager
             spec:
+              affinity:
+                nodeAffinity:
+                  requiredDuringSchedulingIgnoredDuringExecution:
+                    nodeSelectorTerms:
+                    - matchExpressions:
+                      - key: node-role.kubernetes.io/master
+                        operator: Exists
               containers:
               - args:
                 - --leader-elect
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -29,6 +29,13 @@ spec:
       tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: node-role.kubernetes.io/master
+                operator: Exists
       containers:
       - command:
         - /bin/numaresources-operator


fromanirh:
at glance the change is correct, but misplaced. We need to make sure this snipped is preserved once we do `make generate` and/or `make manifests`.

cynepco3hahue:
I think you need to set change here https://github.com/openshift-kni/numaresources-operator/blob/main/config/manager/manager.yaml and re-generate manifests.

fromanirh:
> I think you need to set change here https://github.com/openshift-kni/numaresources-operator/blob/main/config/manager/manager.yaml and re-generate manifests.

yes, this seems the right place. Thanks for checking quickly, @cynepco3hahue

marioferh:
> I think you need to set change here https://github.com/openshift-kni/numaresources-operator/blob/main/config/manager/manager.yaml and re-generate manifests.

Done, thanks

bundle/metadata/annotations.yaml
```diff
@@ -5,7 +5,6 @@ annotations:
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: numaresources-operator
   operators.operatorframework.io.bundle.channels.v1: alpha
-  operators.operatorframework.io.bundle.channel.default.v1: alpha
```
fromanirh, line 8: why was this removed?

cynepco3hahue, line 8: @fromanirh I remember you added it, probably we should add some parameter to avoid override by `make bundle`

marioferh: override by make bundle

fromanirh: good point, need to check why is being deleted

fromanirh: handled here: https://github.com/openshift-kni/numaresources-operator/pull/75 
not blocking for this PR

bundle/manifests/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -257,9 +264,6 @@ spec:
                   initialDelaySeconds: 5
                   periodSeconds: 10
                 resources:
-                  limits:
-                    cpu: 200m
-                    memory: 100Mi
```
fromanirh, line 262: I like this, the operator pod should really not have limits until/unless we have a good rationale - and proof the limits will not backfire. And since AFAIK we don't have this data, better not to guess. The only problem is, this should happen more explicitely :\

marioferh: This come from here: https://github.com/openshift-kni/numaresources-operator/pull/71
Was generated after make bundle

fromanirh: yup. For the reasons above, it's good we remove the limits. But we should remove the limitis explicitely, with a separate commit/PR and proper explanation, and not causally by incident on an apparently unrelated PR which happened to need to run `make bundle`.

marioferh: Ok, done: https://github.com/openshift-kni/numaresources-operator/pull/77

marioferh: Rebase and fix it

bundle/metadata/annotations.yaml
marioferh: Fixed it

fromanirh:
besides the conflict, the test failures look real

marioferh:
> besides the conflict, the test failures look real

It is, investigating

PR #76, opened 2022-01-03
Delete DaemonSets for deleted NodeGroups

jlojosnegros:
We should remove daemon sets  for the node group  that was removed from the NumaResourceOperator spec.

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -38,6 +38,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -281,6 +282,11 @@ func (r *NUMAResourcesOperatorReconciler) syncMachineConfigPoolsStatuses(instanc
 func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx context.Context, instance *nropv1alpha1.NUMAResourcesOperator, mcps []*machineconfigv1.MachineConfigPool) ([]nropv1alpha1.NamespacedName, error) {
 	klog.Info("RTESync start")
 
+	errorList := r.deleteUnusedDaemonSets(ctx, instance, mcps)
+	if len(errorList) > 0 {
+		klog.ErrorS(fmt.Errorf("failed to delete unused daemonsets"), "errors", errorList)
+	}
+
 	var daemonSetsNName []nropv1alpha1.NamespacedName
 
 	err := rtestate.UpdateDaemonSetUserImageSettings(r.RTEManifests.DaemonSet, instance.Spec.ExporterImage, r.ImageSpec, r.ImagePullPolicy)
@@ -303,6 +309,46 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 		}
 	}
 	return daemonSetsNName, nil
+}
+
+func (r *NUMAResourcesOperatorReconciler) deleteUnusedDaemonSets(ctx context.Context, instance *nropv1alpha1.NUMAResourcesOperator, mcps []*machineconfigv1.MachineConfigPool) []error {
+	klog.V(3).Info("Delete DS start")
+	var errors []error
+	var daemonSetList appsv1.DaemonSetList
+	if err := r.List(ctx, &daemonSetList, &client.ListOptions{Namespace: instance.Namespace}); err != nil {
+		klog.ErrorS(err, "error while getting DS list")
+		return append(errors, err)
+	}
+
+	// Generate the names of the DaemonSets that should be running,
+	// one for each MachineConfigPool
+	expectedDaemonSetNames := sets.NewString()
+	for _, mcp := range mcps {
+		expectedDaemonSetNames = expectedDaemonSetNames.Insert(rtestate.GetComponentName(instance.Name, mcp.Name))
+	}
+
+	for _, ds := range daemonSetList.Items {
+		if !expectedDaemonSetNames.Has(ds.Name) {
+			if isOwnedBy(ds.GetObjectMeta(), instance) {
+
+				if err := r.Client.Delete(ctx, &ds); err != nil {
+					klog.ErrorS(err, "error while deleting daemonset", "DaemonSet", ds.Name)
+					errors = append(errors, err)
+				}
+				klog.V(3).Infof("DS[%s] deleted", ds.Name)
+			}
+		}
+	}
+	return errors
+}
+
+func isOwnedBy(element metav1.Object, owner metav1.Object) bool {
+	for _, ref := range element.GetOwnerReferences() {
+		if ref.UID == owner.GetUID() {
+			return true
+		}
+	}
+	return false
 }
 
 // SetupWithManager sets up the controller with the Manager.
--- a/controllers/numaresourcesoperator_controller_test.go
+++ b/controllers/numaresourcesoperator_controller_test.go
@@ -126,15 +126,170 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
 		})
 	})
 
+	Context("with correct NRO and more than one NodeGroup", func() {
+		var nro *nrov1alpha1.NUMAResourcesOperator
+		var mcp1 *machineconfigv1.MachineConfigPool
+		var mcp2 *machineconfigv1.MachineConfigPool
+
+		var reconciler *NUMAResourcesOperatorReconciler
+		var label1, label2 map[string]string
+
+		BeforeEach(func() {
+			label1 = map[string]string{
+				"test1": "test1",
+			}
+			label2 = map[string]string{
+				"test2": "test2",
+			}
+
+			nro = testutils.NewNUMAResourcesOperator(defaultNUMAResourcesOperatorCrName, []*metav1.LabelSelector{
+				{MatchLabels: label1},
+				{MatchLabels: label2},
+			})
+
+			mcp1 = testutils.NewMachineConfigPool("test1", label1, &metav1.LabelSelector{MatchLabels: label1}, &metav1.LabelSelector{MatchLabels: label1})
+			mcp2 = testutils.NewMachineConfigPool("test2", label2, &metav1.LabelSelector{MatchLabels: label2}, &metav1.LabelSelector{MatchLabels: label2})
+
+			var err error
+			reconciler, err = NewFakeNUMAResourcesOperatorReconciler(platform.OpenShift, nro, mcp1, mcp2)
+			Expect(err).ToNot(HaveOccurred())
+
+			key := client.ObjectKeyFromObject(nro)
+			firstLoopResult, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: key})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(firstLoopResult).To(Equal(reconcile.Result{RequeueAfter: time.Minute}))
+
+			// Ensure mcp1 is ready
+			Expect(reconciler.Client.Get(context.TODO(), client.ObjectKeyFromObject(mcp1), mcp1)).ToNot(HaveOccurred())
+			mcp1.Status.Configuration.Source = []corev1.ObjectReference{
+				{
+					Name: rte.GetMachineConfigName(nro.Name, mcp1.Name),
+				},
+			}
+			mcp1.Status.Conditions = []machineconfigv1.MachineConfigPoolCondition{
+				{
+					Type:   machineconfigv1.MachineConfigPoolUpdated,
+					Status: corev1.ConditionTrue,
+				},
+			}
+			Expect(reconciler.Client.Status().Update(context.TODO(), mcp1))
+
+			// ensure mcp2 is ready
+			Expect(reconciler.Client.Get(context.TODO(), client.ObjectKeyFromObject(mcp2), mcp2)).ToNot(HaveOccurred())
+			mcp2.Status.Configuration.Source = []corev1.ObjectReference{
+				{
+					Name: rte.GetMachineConfigName(nro.Name, mcp2.Name),
+				},
+			}
+			mcp2.Status.Conditions = []machineconfigv1.MachineConfigPoolCondition{
+				{
+					Type:   machineconfigv1.MachineConfigPoolUpdated,
+					Status: corev1.ConditionTrue,
+				},
+			}
+			Expect(reconciler.Client.Status().Update(context.TODO(), mcp2))
+
+			secondLoopResult, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: key})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(secondLoopResult).To(Equal(reconcile.Result{RequeueAfter: 5 * time.Second}))
+
+			By("Check DaemonSets are created")
+			mcp1DSKey := client.ObjectKey{
+				Name:      rte.GetComponentName(nro.Name, mcp1.Name),
+				Namespace: testNamespace,
+			}
+			ds := &appsv1.DaemonSet{}
+			Expect(reconciler.Client.Get(context.TODO(), mcp1DSKey, ds)).ToNot(HaveOccurred())
+
+			mcp2DSKey := client.ObjectKey{
+				Name:      rte.GetComponentName(nro.Name, mcp2.Name),
+				Namespace: testNamespace,
+			}
+			Expect(reconciler.Client.Get(context.TODO(), mcp2DSKey, ds)).ToNot(HaveOccurred())
+		})
+		When("a NodeGroup is deleted", func() {
+			BeforeEach(func() {
+				// check we have at least two NodeGroups
+				Expect(len(nro.Spec.NodeGroups)).To(BeNumerically(">", 1))
+
+				By("Update NRO to have just one NodeGroup")
+				key := client.ObjectKeyFromObject(nro)
+				nro := &nrov1alpha1.NUMAResourcesOperator{}
+				Expect(reconciler.Client.Get(context.TODO(), key, nro)).NotTo(HaveOccurred())
+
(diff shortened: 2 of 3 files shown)


controllers/numaresourcesoperator_controller.go
```diff
@@ -294,6 +299,35 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
...
 
+func (r *NUMAResourcesOperatorReconciler) deleteUnusedDaemonSets(ctx context.Context, instance *nropv1alpha1.NUMAResourcesOperator, mcps []*machineconfigv1.MachineConfigPool) error {
+
+	var daemonSetList appsv1.DaemonSetList
+	err := r.List(ctx, &daemonSetList)
+	if err != nil {
+		return err
+	}
+
+	for _, mcp := range mcps {
```
cynepco3hahue, line 310: I think it should be other way, something like:
```
expectedDaemonSetNames := map[string]bool{}
for _, mcp := range mcps {
        generatedDSName := rtestate.GetComponentName(instance.Name, mcp.Name)
        expectedDaemonSetNames[generatedDSName] = true
}
      
for _, ds := range daemonSetList.Items {
    if _, ok := expectedDaemonSetNames[ds.Name]; !ok {
        // delete ds and handle properly errors
    }
}        
```

controllers/numaresourcesoperator_controller.go
```diff
@@ -294,6 +299,35 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 	return daemonSetsNName, nil
 }
 
+func (r *NUMAResourcesOperatorReconciler) deleteUnusedDaemonSets(ctx context.Context, instance *nropv1alpha1.NUMAResourcesOperator, mcps []*machineconfigv1.MachineConfigPool) error {
+
+	var daemonSetList appsv1.DaemonSetList
+	err := r.List(ctx, &daemonSetList)
```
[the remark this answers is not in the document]
jlojosnegros: yep sorry you are right

controllers/numaresourcesoperator_controller.go
```diff
@@ -299,6 +304,34 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 	return daemonSetsNName, nil
 }
 
+func (r *NUMAResourcesOperatorReconciler) deleteUnusedDaemonSets(ctx context.Context, instance *nropv1alpha1.NUMAResourcesOperator, mcps []*machineconfigv1.MachineConfigPool) error {
+
+	var daemonSetList appsv1.DaemonSetList
+	err := r.List(ctx, &daemonSetList, &client.ListOptions{Namespace: instance.Namespace})
+	if err != nil {
```
cynepco3hahue, line 311: nit: you can use one liner
```
if err := r.List(ctx, &daemonSetList, &client.ListOptions{Namespace: instance.Namespace}); err != nil {
    // do something
}

File: controllers/numaresourcesoperator_controller.go [hunk 4]
```diff
@@ -299,6 +304,34 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
...
+	for _, mcp := range mcps {
+		generatedDSName := rtestate.GetComponentName(instance.Name, mcp.Name)
+		expectedDaemonSetNames[generatedDSName] = struct{}{}
+	}
+
+	for _, ds := range daemonSetList.Items {
+		if _, ok := expectedDaemonSetNames[ds.Name]; !ok {
+
+			err := r.Client.Delete(ctx, &ds)
+			if err != nil {
```
cynepco3hahue, line 327: nit: you can use one liner

controllers/numaresourcesoperator_controller.go
```diff
@@ -299,6 +304,34 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
...
+
+	// Generate the names of the DaemonSets that should be running,
+	// one for each MachineConfigPool
+	expectedDaemonSetNames := map[string]struct{}{}
+	for _, mcp := range mcps {
+		generatedDSName := rtestate.GetComponentName(instance.Name, mcp.Name)
+		expectedDaemonSetNames[generatedDSName] = struct{}{}
+	}
+
+	for _, ds := range daemonSetList.Items {
```
cynepco3hahue, line 323: please filter by owner reference, to be sure that we are only deleting daemon set related to our instance

controllers/numaresourcesoperator_controller.go
jlojosnegros: Yeah I know but It seems a little bit harder to read to me than the two line expression ...

controllers/numaresourcesoperator_controller.go
jlojosnegros: done

controllers/numaresourcesoperator_controller.go
```diff
@@ -305,6 +311,53 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 	return daemonSetsNName, nil
 }
 
+func (r *NUMAResourcesOperatorReconciler) deleteUnusedDaemonSets(ctx context.Context, instance *nropv1alpha1.NUMAResourcesOperator, mcps []*machineconfigv1.MachineConfigPool) error {
+	klog.Info("Delete DS start")
```
fromanirh, line 315: let's log this only if verbosity is > 3

controllers/numaresourcesoperator_controller.go
```diff
@@ -305,6 +311,53 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
...
+	var daemonSetList appsv1.DaemonSetList
+	if err := r.List(ctx, &daemonSetList, &client.ListOptions{Namespace: instance.Namespace}); err != nil {
+		klog.Infof("error while getting DS list: %v", err)
```
fromanirh, line 318: `ErrorS` seems better here

controllers/numaresourcesoperator_controller.go
```diff
@@ -305,6 +311,53 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
...
+		return err
+	}
+
+	// Generate the names of the DaemonSets that should be running,
+	// one for each MachineConfigPool
+	expectedDaemonSetNames := map[string]struct{}{}
```
fromanirh, line 324: you may want to use `sets.String` from the k8s apimachinery package (we should have examples around)

pkg/ownerreferences/ownerreferences.go
```diff
@@ -0,0 +1,74 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+// This came from
+// https://github.com/kubernetes-sigs/controller-runtime/blob/3c54acbad091d40ef80d467943c5d7126c4b8291/pkg/controller/controllerutil/controllerutil.go#L158
```
fromanirh, line 14: and why we are copypasting it here? this deserves a comment

pkg/ownerreferences/ownerreferences.go
```diff
@@ -0,0 +1,74 @@
+package ownerreferences
```
fromanirh, line 1: missing copyright boilerplate (should be the same of the controller-runtime package)

controllers/numaresourcesoperator_controller.go
```diff
@@ -281,9 +282,14 @@ func (r *NUMAResourcesOperatorReconciler) syncMachineConfigPoolsStatuses(instanc
 func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx context.Context, instance *nropv1alpha1.NUMAResourcesOperator, mcps []*machineconfigv1.MachineConfigPool) ([]nropv1alpha1.NamespacedName, error) {
 	klog.Info("RTESync start")
 
+	err := r.deleteUnusedDaemonSets(ctx, instance, mcps)
+	if err != nil {
+		klog.Warningf("failed to delete unused DaemonSets %v", err)
```
fromanirh, line 287: This should probably be `ErrorS`

controllers/numaresourcesoperator_controller.go
```diff
@@ -305,6 +311,53 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
...
+				}
+				klog.Infof("DS[%s] deleted", ds.Name)
+			}
+		}
+	}
+	return err
+}
+
+func (r *NUMAResourcesOperatorReconciler) isOwnedBy(element metav1.Object, owner metav1.Object) (bool, error) {
+	ownerReference, err := ownerreferences.CreateOwnerReference(owner, r.Scheme)
```
Tal-or, line 352: I think you can settle for ``` return element.GetOwnerReferences()  == owner.GetUID()

controllers/numaresourcesoperator_controller.go
```diff
@@ -305,6 +311,53 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
...
+					continue
+				}
+				klog.Infof("DS[%s] deleted", ds.Name)
+			}
+		}
+	}
+	return err
+}
+
+func (r *NUMAResourcesOperatorReconciler) isOwnedBy(element metav1.Object, owner metav1.Object) (bool, error) {
```
Tal-or, line 351: I think you can settle for 
``` return element.GetOwnerReferences()  == owner.GetUID()```

Comment (fromanirh) [reply controllers/numaresourcesoperator_controller.go:351]: If this works, I like it much much more - and would probably make the `ownerreferences` borrowed code no longer necessary.

File: controllers/numaresourcesoperator_controller.go (hunk 8 above)
Comment (Tal-or) [reply controllers/numaresourcesoperator_controller.go:324]: An example https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/2a8d8ac49582d5b32bcfedb87341b3f4d6315c03/pkg/resourcemonitor/resourcemonitor.go#L61

File: controllers/numaresourcesoperator_controller.go (hunk 13 above)
Comment (cynepco3hahue) [reply controllers/numaresourcesoperator_controller.go:351]: I think more correct is `element.GetOwnerReferences().GetUID() == owner.GetUID()`

File: controllers/numaresourcesoperator_controller_test.go [hunk 14]
```diff
@@ -296,6 +298,83 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
...
 					})
+
+					When("a NodeGroup is deleted", func() {
+						var thirdLoopResult reconcile.Result
+						BeforeEach(func() {
+							// check reconcile second loop result
+							Expect(secondLoopResult).To(Equal(reconcile.Result{RequeueAfter: 5 * time.Second}))
+
+							// check we have at least two NodeGroups
+							Expect(len(nro.Spec.NodeGroups)).To(BeNumerically(">", 1))
```
cynepco3hahue, line 309: Can you explicitly create a new NRO resource with two node groups that you will add under the reconcile, just to be sure that deleting the node group will not affect other tests?

jlojosnegros: Deleting NodeGroups will only afffect those test cases under this "When" clause, so I think there would be no collision with other testcases

controllers/numaresourcesoperator_controller.go
```diff
@@ -305,6 +311,46 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
...
+		expectedDaemonSetNames = expectedDaemonSetNames.Insert(rtestate.GetComponentName(instance.Name, mcp.Name))
+	}
+
+	for _, ds := range daemonSetList.Items {
+		if !expectedDaemonSetNames.Has(ds.Name) {
+			if r.isOwnedBy(ds.GetObjectMeta(), instance) {
+
+				if err := r.Client.Delete(ctx, &ds); err != nil {
+					klog.Infof("error while deleting [%s]: %v", ds.Name, err)
+					errors = append(errors, err)
```
fromanirh, line 336: I'm not sure we want to accumulate errors here, need to think about it.
If we are not sure this case is likely and we really want to keep iterating and delete unused nodegroups, I'd start very simple and just bail out on the first error.

controllers/numaresourcesoperator_controller.go
```diff
@@ -305,6 +311,46 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
...
+				}
+				klog.V(3).Infof("DS[%s] deleted", ds.Name)
+			}
+		}
+	}
+	return errors
+}
+
+func (r *NUMAResourcesOperatorReconciler) isOwnedBy(element metav1.Object, owner metav1.Object) bool {
```
fromanirh, line 345: let's make this a free function

controllers/numaresourcesoperator_controller_test.go
fromanirh: While the above is true, I'm also in favor of an explicit test case.

controllers/numaresourcesoperator_controller.go
jlojosnegros: Well I just thought of accumulate them because an error while deleting a DS could stop the process preventing the deletion of the others that may work properly. Also, even if all of them fail, it could be interesting to have all the information at once, unless the deletion operation would take too much resources does not seem a bad exchange.
Also I start from the base that an error while deleting DS resources will be just a "warning" at operator level as this is just to improve resource utilization but no interferes with the operator functionality.

controllers/numaresourcesoperator_controller.go
jlojosnegros: sure

controllers/numaresourcesoperator_controller_test.go
jlojosnegros: Ok, If both of you agree on that I will change it.
I laid out tests this way to avoid code repetition... I mean, I agree that some repetition is ok in test cases and I usually prefer explicit to implicit but being the sut a kind of state machine I though using hierarchical testing features from Ginkgo to go through the different states and assert proper status on each one.

controllers/numaresourcesoperator_controller.go
fromanirh: I need to think more about this

fromanirh:
it seems we have still unanswered comments? perhaps a new version wasn't pushed?

controllers/numaresourcesoperator_controller.go
jlojosnegros: done

controllers/numaresourcesoperator_controller.go
jlojosnegros: done

controllers/numaresourcesoperator_controller.go
jlojosnegros: > it seems we have still unanswered comments? perhaps a new version wasn't pushed?

Sorry forgot to change the "ErrorS" comments. 
I think all comments, but the one regarding the convenience to use a list of errors or not, have been addressed now.

jlojosnegros:
> it seems we have still unanswered comments? perhaps a new version wasn't pushed?

Sorry forgot to change the "ErrorS" comments. 
I think all comments, but the one regarding the convenience to use a list of errors or not, have been addressed now.

controllers/numaresourcesoperator_controller.go
```diff
@@ -281,6 +282,11 @@ func (r *NUMAResourcesOperatorReconciler) syncMachineConfigPoolsStatuses(instanc
 func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx context.Context, instance *nropv1alpha1.NUMAResourcesOperator, mcps []*machineconfigv1.MachineConfigPool) ([]nropv1alpha1.NamespacedName, error) {
 	klog.Info("RTESync start")
 
+	errorList := r.deleteUnusedDaemonSets(ctx, instance, mcps)
+	if len(errorList) > 0 {
+		klog.ErrorS(fmt.Errorf("some errors found while deleting unused daemonsets"), "errors", errorList)
```
fromanirh, line 287: `fmt.Errorf` looks redundant, and I'd drop the "some errors" part: `"failed to delete unused daemonsets", "errors", errorList`
but not worth reuploading for this

controllers/numaresourcesoperator_controller.go
fromanirh: fixed?

controllers/numaresourcesoperator_controller.go
fromanirh: I'm still not sold, but let's keep the accumulation for now.

controllers/numaresourcesoperator_controller_test.go
```diff
@@ -126,15 +126,168 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
...
+				Namespace: testNamespace,
+			}
+			Expect(reconciler.Client.Get(context.TODO(), mcp2DSKey, ds)).ToNot(HaveOccurred())
+		})
+		When("a NodeGroup is deleted", func() {
+			BeforeEach(func() {
+				// check we have at least two NodeGroups
+				Expect(len(nro.Spec.NodeGroups)).To(BeNumerically(">", 1))
+
+				By("Update NRO to have just one NodeGroup")
```
fromanirh, line 215: I'd expect an actual update here, but I guess it's good enough.

fromanirh (commented):
it seems the commit message describing the change got lost in the rebase/squash, please restore it and we can merge.

controllers/numaresourcesoperator_controller.go
jlojosnegros: Ok with the "some errors" part but ErrorS expect an error as first parameter so that's why I have use fmt.Errorf ... maybe I could use klog.Errorf instead and avoid the use of fmt.Errorf...

fromanirh: nope, I misread, keep it this way.

controllers/numaresourcesoperator_controller_test.go
jlojosnegros: using Update now.

fromanirh:
thanks, looks ready now. Please restore the commit message text, which apparently got lost in the rebase/squash, and we can merge.

fromanirh (commented):
/lgtm
/approve

PR #77, opened 2022-01-04
Apply make bundle to remove limits

marioferh:
The operator pod should really not have limits until/unless we have a good rationale and proof the limits will not backfire. 

Apply make bundle to remove limits after https://github.com/openshift-kni/numaresources-operator/pull/71


fromanirh:
/lgtm
/approve
ideally: this and #71 should have been on the same PR, with a more descriptive commit message (e.g. WHY we want to remove the limits? what change since we introduced them? why we believe now we will do better without them).

PR #78, opened 2022-01-04
doc: fix kubelet config example to include memory manager policy

cynepco3hahue:

### DIFF
--- a/doc/examples/kubeletconfig.yaml
+++ b/doc/examples/kubeletconfig.yaml
@@ -10,6 +10,7 @@ spec:
     cpuManagerPolicy: "static"
     cpuManagerReconcilePeriod: "5s"
     reservedSystemCPUs: "0,1"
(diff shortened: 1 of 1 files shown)


doc/examples/kubeletconfig.yaml
```diff
@@ -10,6 +10,7 @@ spec:
     cpuManagerPolicy: "static"
     cpuManagerReconcilePeriod: "5s"
     reservedSystemCPUs: "0,1"
+    memoryManagerPolicy: "Static"
```
fromanirh, line 13: please add to the commit message something like "because otherwise the memory resources are not reported correctly by the kubelet, thus by RTE"

cynepco3hahue: done

fromanirh: thanks!

PR #80, opened 2022-01-04
doc: fix reserved memory value to be equal to NodeAllocatable

cynepco3hahue:
The kubelet start fails if the reserved memory is not equal to
NodeAllocatable(system-reserved + kube-reserved + hard eviction threshold).


fromanirh (commented):
/lgtm
/approve
yeah, I was uncareful when I wrote the example :\

PR #82, opened 2022-01-05
Re-organize Numaresources operator tests

jlojosnegros:
Organize NRO tests to avoid some code repetition and help in adding
new tests that could need the same sut setup as the ones already in the
suite

### DIFF
--- a/controllers/numaresourcesoperator_controller_test.go
+++ b/controllers/numaresourcesoperator_controller_test.go
@@ -154,51 +154,51 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 			Context("on the first iteration", func() {
-				var result reconcile.Result
+				var firstLoopResult reconcile.Result
 				BeforeEach(func() {
 					var err error
 
 					key := client.ObjectKeyFromObject(nro)
-					result, err = reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: key})
+					firstLoopResult, err = reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: key})
 					Expect(err).ToNot(HaveOccurred())
 				})
 				It("should create CRD, machine configs and wait for MCPs updates", func() {
 					// check reconcile loop result
-					Expect(result).To(Equal(reconcile.Result{RequeueAfter: time.Minute}))
+					Expect(firstLoopResult).To(Equal(reconcile.Result{RequeueAfter: time.Minute}))
 
 					// check CRD is created
 					crd := &apiextensionsv1.CustomResourceDefinition{}
-					key := client.ObjectKey{
+					crdKey := client.ObjectKey{
 						Name: "noderesourcetopologies.topology.node.k8s.io",
 					}
-					Expect(reconciler.Client.Get(context.TODO(), key, crd)).ToNot(HaveOccurred())
+					Expect(reconciler.Client.Get(context.TODO(), crdKey, crd)).ToNot(HaveOccurred())
 
 					// check MachineConfigs are created
 					mc := &machineconfigv1.MachineConfig{}
-					key = client.ObjectKey{
+					mc1Key := client.ObjectKey{
 						Name: rte.GetMachineConfigName(nro.Name, mcp1.Name),
 					}
-					Expect(reconciler.Client.Get(context.TODO(), key, mc)).ToNot(HaveOccurred())
+					Expect(reconciler.Client.Get(context.TODO(), mc1Key, mc)).ToNot(HaveOccurred())
 
-					key = client.ObjectKey{
+					mc2Key := client.ObjectKey{
 						Name: rte.GetMachineConfigName(nro.Name, mcp2.Name),
 					}
-					Expect(reconciler.Client.Get(context.TODO(), key, mc)).ToNot(HaveOccurred())
+					Expect(reconciler.Client.Get(context.TODO(), mc2Key, mc)).ToNot(HaveOccurred())
 				})
 			})
 			Context("on the second iteration", func() {
-				var result reconcile.Result
+				var secondLoopResult reconcile.Result
 				When("machine config pools still are not ready", func() {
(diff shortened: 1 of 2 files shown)


controllers/numaresourcesoperator_controller_test.go
```diff
@@ -138,224 +140,237 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
...
+					})
+					It("should continue with creation of additional components", func() {
+						// check reconcile second loop result
+						Expect(secondLoopResult).To(Equal(reconcile.Result{RequeueAfter: 5 * time.Second}))
+
+						// Check All the additional components are created
+						key := client.ObjectKey{
+							Name:      "rte",
+							Namespace: testNamespace,
+						}
```
fromanirh, line 254: I'd NOT reuse objects, it's bad practice, even if all the names are the same. In this case I prefer to create temporary object keys for each object (there is a helper already for this, see ObjectKeyFromObject)

jlojosnegros: I also prefer not to reuse them ... but was just trying not to modify to much of the existing code in the reorganize process...

fromanirh: it's fine, as long as you do refactoring on it's own commit. Better that than reusing objects :)

jlojosnegros: I have already changed some names in the previous commit so in the end I have changed them all ...

cynepco3hahue:
/approve
Probably we need to organize commits and drop renaming under the separate commit.

controllers/numaresourcesoperator_controller_test.go
```diff
@@ -138,224 +141,242 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
...
+				reconciler, err = NewFakeNUMAResourcesOperatorReconciler(platform.OpenShift, nro, mcp1, mcp2)
+				Expect(err).ToNot(HaveOccurred())
+			})
+			Context("on the first iteration", func() {
+				var firstLoopResult reconcile.Result
+				BeforeEach(func() {
+					var err error
 
-				mc := &machineconfigv1.MachineConfig{}
+					// call first reconcile loop
```
fromanirh, line 163: this is already pretty evident from the code, though :)

controllers/numaresourcesoperator_controller_test.go
```diff
@@ -138,224 +141,242 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
 				{MatchLabels: label1},
 				{MatchLabels: label2},
 			})
+			nroKey = client.ObjectKeyFromObject(nro)
```
fromanirh, line 144: this is a cheap operation, can be done inline. On top of that, we don't care SO MUCH about performance in test code, so in general let's be a bit biased towards inlining (https://about.sourcegraph.com/go/advanced-testing-in-go/#Repeat-yourself)

fromanirh (commented):
Is this a general cleanup or are you making room for upcoming tests?

This looks like a good cleanup, but there are two things:
1. (most important) this feels unnecessarily hard to read, because we seem to do many things in one go. Any chance to split this commit in a series of focused commit (code reorganization, test changes...)?
2. Even though in this case the benefits seem to far outweight the drawbacks, I'd be wary about factoring the test code too much. But again we seem OK here (bar the above point).

jlojosnegros:
> Is this a general cleanup or are you making room for upcoming tests?
> 
> This looks like a good cleanup, but there are two things:
> 
> 1. (most important) this feels unnecessarily hard to read, because we seem to do many things in one go. Any chance to split this commit in a series of focused commit (code reorganization, test changes...)?
> 2. Even though in this case the benefits seem to far outweight the drawbacks, I'd be wary about factoring the test code too much. But again we seem OK here (bar the above point).

It is a general cleanup triggered when I was looking for the proper way to introduce a new tests. 
I need to test the deletion of DaemonSets and while I was reading the tests I found many sut setup was done in the "it" steps, and so many code was repeated. To create a new test I would have to repeat code again so I though i would be a nice idea to split the sut setups in phases and use the capacities of Ginkgo for hierarchical setup with "BeforeEach"

Reagarding the commit split ... I can identify the test code reorganization and the renaming of certain local variables and maybe these two changes could go in different commits or even PRs if you prefer.

fromanirh:
> It is a general cleanup triggered when I was looking for the proper way to introduce a new tests. I need to test the deletion of DaemonSets and while I was reading the tests I found many sut setup was done in the "it" steps, and so many code was repeated. To create a new test I would have to repeat code again so I though i would be a nice idea to split the sut setups in phases and use the capacities of Ginkgo for hierarchical setup with "BeforeEach"

OK, in general this specific change looks like a good change, but the tradeoffs in test code are different than from the production code, so let's not push too much in this direction (this applies for future work ofc).

> Reagarding the commit split ... I can identify the test code reorganization and the renaming of certain local variables and maybe these two changes could go in different commits or even PRs if you prefer.

If you can do this easily, this will help a lot the review, yes.

controllers/numaresourcesoperator_controller_test.go
jlojosnegros: yep, I will delete that comment.

controllers/numaresourcesoperator_controller_test.go
jlojosnegros: I was not concern about performance here. I just thought that as "nroKey" was a derived from "nro" only it could be a good idea to do as with "nro", create it once and use it, that said I have no strong opinion against create the object multiple times in the test where it is used.

fromanirh: OK, then it is ok to create it once, and the explanation you just added should be added as a comment (comments which explain "why" are almost always good), something like (random example)
```
the key is bound to nro, it doesn't change and has the same lifetime, so it makes sense to create alongside it
```

jlojosnegros:
I have made variable renaming in another commit. 
Hope that helps

fromanirh (commented):
/lgtm
/approve
thanks!

PR #84, opened 2022-01-06
Observability  enhancement

Tal-or:
In order to improve state observability, let's add events for the user to understand why (and if) certain actions/operations failed and to notify in case of success.

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -38,6 +38,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -73,6 +74,7 @@ type NUMAResourcesOperatorReconciler struct {
 	Namespace       string
 	ImageSpec       string
 	ImagePullPolicy corev1.PullPolicy
+	Recorder        record.EventRecorder
 }
 
 // TODO: narrow down
@@ -143,9 +145,7 @@ func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctr
 	if condition != "" {
 		// TODO: use proper reason
 		reason, message := condition, messageFromError(err)
-		if err := status.Update(ctx, r.Client, instance, condition, reason, message); err != nil {
-			klog.InfoS("Failed to update numaresourcesoperator status", "Desired condition", condition, "error", err)
-		}
+		_, _ = r.updateStatus(ctx, instance, condition, reason, message)
 	}
 	return result, err
 }
@@ -178,15 +178,19 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
 	var err error
 	err = r.syncNodeResourceTopologyAPI()
 	if err != nil {
+		r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedCRDInstall", "Failed to install Node Resource Topology CRD: %v", err)
 		return ctrl.Result{}, status.ConditionDegraded, errors.Wrapf(err, "FailedAPISync")
 	}
+	r.Recorder.Eventf(instance, corev1.EventTypeNormal, "SuccessfulCRDInstall", "Node Resource Topology CRD installed")
 
 	if r.Platform == platform.OpenShift {
 		// we need to create machine configs first and wait for the MachineConfigPool updates
 		// before creating additional components
 		if err := r.syncMachineConfigs(ctx, instance, mcps); err != nil {
+			r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedMCSync", "Failed to set up machine configuration for worker nodes: %v", err)
 			return ctrl.Result{}, status.ConditionDegraded, errors.Wrapf(err, "failed to sync machine configs")
 		}
+		r.Recorder.Eventf(instance, corev1.EventTypeNormal, "SuccessfulMCSync", "Enabled machine configuration for worker nodes")
 
 		// MCO need to update SELinux context and other stuff, and need to trigger a reboot.
 		// It can take a while.
@@ -198,8 +202,10 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
 
 	daemonSetsInfo, err := r.syncNUMAResourcesOperatorResources(ctx, instance, mcps)
 	if err != nil {
+		r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedRTECreate", "Failed to create Resource-Topology-Exporter DaemonSets: %v", err)
 		return ctrl.Result{}, status.ConditionDegraded, errors.Wrapf(err, "FailedRTESync")
 	}
+	r.Recorder.Eventf(instance, corev1.EventTypeNormal, "SuccessfulRTECreate", "Created Resource-Topology-Exporter DaemonSets")
 
 	instance.Status.DaemonSets = []nropv1alpha1.NamespacedName{}
 	for _, nname := range daemonSetsInfo {
--- a/controllers/numaresourcesoperator_controller_test.go
+++ b/controllers/numaresourcesoperator_controller_test.go
@@ -35,11 +35,13 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -68,6 +70,8 @@ func NewFakeNUMAResourcesOperatorReconciler(plat platform.Platform, initObjects 
 		return nil, err
 	}
 
+	recorder := record.NewFakeRecorder(bufferSize)
+
 	return &NUMAResourcesOperatorReconciler{
 		Client:       fakeClient,
 		Scheme:       scheme.Scheme,
@@ -77,6 +81,7 @@ func NewFakeNUMAResourcesOperatorReconciler(plat platform.Platform, initObjects 
 		Helper:       helper,
 		Namespace:    testNamespace,
 		ImageSpec:    testImageSpec,
+		Recorder:     recorder,
 	}, nil
 }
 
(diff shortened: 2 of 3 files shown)


config/rbac/role.yaml
```diff
@@ -17,6 +17,7 @@ rules:
   resources:
   - events
   verbs:
+  - '*'
```
cynepco3hahue, line 20: remove create and patch if you set `*`

cynepco3hahue (commented):
Check unittests and extend them to handle events.

controllers/numaresourcesoperator_controller.go
```diff
@@ -143,9 +145,7 @@ func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctr
 	if condition != "" {
 		// TODO: use proper reason
 		reason, message := condition, messageFromError(err)
-		if err := status.Update(ctx, r.Client, instance, condition, reason, message); err != nil {
-			klog.InfoS("Failed to update numaresourcesoperator status", "Desired condition", condition, "error", err)
-		}
+		_, _ = r.updateStatus(ctx, instance, condition, reason, message)
```
fromanirh, line 149: this emits the event AND also logs, right?

controllers/numaresourcesoperator_controller.go
```diff
@@ -155,6 +155,7 @@ func (r *NUMAResourcesOperatorReconciler) updateStatus(ctx context.Context, inst
 
 	if err := status.Update(ctx, r.Client, instance, condition, reason, message); err != nil {
 		klog.InfoS("Failed to update numaresourcesoperator status", "Desired condition", status.ConditionDegraded, "error", err)
+		r.Recorder.Eventf(instance, corev1.EventTypeWarning, "UpdateStatusFailed", "failed to update numaresourcesoperator status with desired condition: %s", condition)
```
fromanirh, line 159: this doesn't deserve a event I believe, users can barely do anything about that.

controllers/numaresourcesoperator_controller.go
```diff
@@ -178,15 +179,19 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
 	var err error
 	err = r.syncNodeResourceTopologyAPI()
 	if err != nil {
+		r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedAPISync", "failed to sync noderesourcetopology api: %v", err)
 		return ctrl.Result{}, status.ConditionDegraded, errors.Wrapf(err, "FailedAPISync")
 	}
+	r.Recorder.Eventf(instance, corev1.EventTypeNormal, "SuccessfulAPISync", "noderesourcetopology synchronization completed")
```
fromanirh, line 186: this and the above probably deserve a event, but we should revisit the message to make it less low-level.
Something like "Node Resource Topology CRD installed" or so. Need to think a bit about it.

controllers/numaresourcesoperator_controller.go
```diff
@@ -178,15 +179,19 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
...
 
 	if r.Platform == platform.OpenShift {
 		// we need to create machine configs first and wait for the MachineConfigPool updates
 		// before creating additional components
 		if err := r.syncMachineConfigs(ctx, instance, mcps); err != nil {
+			r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedMCSync", "failed to sync machine-config: %v", err)
 			return ctrl.Result{}, status.ConditionDegraded, errors.Wrapf(err, "failed to sync machine configs")
 		}
+		r.Recorder.Eventf(instance, corev1.EventTypeNormal, "SuccessfulMCSync", "machine-config synchronization completed")
```
fromanirh, line 195: not sure about these two, inclined to yes though. The problem is again drawing a line between meaninful events and not spamming the users.

controllers/numaresourcesoperator_controller.go
```diff
@@ -198,8 +203,10 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
 
 	daemonSetsInfo, err := r.syncNUMAResourcesOperatorResources(ctx, instance, mcps)
 	if err != nil {
+		r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedRTESync", "failed to sync resource-topology-exporter: %v", err)
 		return ctrl.Result{}, status.ConditionDegraded, errors.Wrapf(err, "FailedRTESync")
 	}
+	r.Recorder.Eventf(instance, corev1.EventTypeNormal, "SuccessfulRTESync", "resource-topology-exporter synchronization completed")
```
fromanirh, line 210: same

config/rbac/role.yaml
fromanirh: yep, this makes sense, but why do we need `*` in the first place? IOW, why `create` and `patch` are not sufficient?

controllers/numaresourcesoperator_controller_test.go
```diff
@@ -77,6 +81,7 @@ func NewFakeNUMAResourcesOperatorReconciler(plat platform.Platform, initObjects
 		Helper:       helper,
 		Namespace:    testNamespace,
 		ImageSpec:    testImageSpec,
+		Recorder:     recorder,
```
fromanirh, line 84: this is correct, but is something missing here?

fromanirh (commented):
This is a good idea and a good change. The only thing that needs a bit more polishing is which events we want to emit. Events are user-observable behaviour, so we need to guarantee and support them, kinda like APIs. So as rule of thumb let's start small, and add as needed, because adding is easier than removing.

controllers/numaresourcesoperator_controller.go
Tal-or: Yes.
But following the next comment of yours, I'll probably remove the event and keep the log

controllers/numaresourcesoperator_controller_test.go
Tal-or: I followed the convention from here:https://github.com/openshift-kni/numaresources-operator/blob/main/controllers/kubeletconfig_controller_test.go#L50

fromanirh: ok, in both cases we are not actually testing the events. Good enough for now.

Tal-or: I was trying to test the actual events, but it seems like this fake recorder, doesn't actually create the event (```client.List(ctx, eventList...)``` returns empty), but at least it allows you to run the tests without crashing when calling the recorder.

config/rbac/role.yaml
Tal-or: we can revert, no biggy

controllers/numaresourcesoperator_controller_test.go
cynepco3hahue: Here you can see how you can check event
https://github.com/openshift-kni/performance-addon-operators/blob/e5294c90a8e103d19318d3cf035c132b3052d1bb/controllers/performanceprofile_controller_test.go#L121

controllers/numaresourcesoperator_controller.go
```diff
@@ -178,15 +178,19 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
 	var err error
 	err = r.syncNodeResourceTopologyAPI()
 	if err != nil {
+		r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedCRDInstall", "Failed to install Node Resource Topology CRD: %v", err)
 		return ctrl.Result{}, status.ConditionDegraded, errors.Wrapf(err, "FailedAPISync")
 	}
+	r.Recorder.Eventf(instance, corev1.EventTypeNormal, "SuccessfulCRDInstall", "Node Resource Topology CRD installed")
```
fromanirh, line 184: ok, this is still a bit low level, but good enough for now

controllers/numaresourcesoperator_controller.go
```diff
@@ -178,15 +178,19 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
...
 
 	if r.Platform == platform.OpenShift {
 		// we need to create machine configs first and wait for the MachineConfigPool updates
 		// before creating additional components
 		if err := r.syncMachineConfigs(ctx, instance, mcps); err != nil {
+			r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedMCSync", "Failed to sync Machine-Config: %v", err)
```
fromanirh, line 190: "failed to set up machine configuration for worker nodes"?

controllers/numaresourcesoperator_controller.go
```diff
@@ -178,15 +178,19 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
...
 			return ctrl.Result{}, status.ConditionDegraded, errors.Wrapf(err, "failed to sync machine configs")
 		}
+		r.Recorder.Eventf(instance, corev1.EventTypeNormal, "SuccessfulMCSync", "Machine-Config synchronization completed")
```
fromanirh, line 193: "enabled machine configuration for worker nodes"?

controllers/numaresourcesoperator_controller.go
```diff
@@ -198,8 +202,10 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
 
 	daemonSetsInfo, err := r.syncNUMAResourcesOperatorResources(ctx, instance, mcps)
 	if err != nil {
+		r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedRTECreate", "Failed to create Resource-Topology-Exporter DaemonSet: %v", err)
 		return ctrl.Result{}, status.ConditionDegraded, errors.Wrapf(err, "FailedRTESync")
 	}
+	r.Recorder.Eventf(instance, corev1.EventTypeNormal, "SuccessfulRTECreate", "Created Resource-Topology-Exporter DaemonSet: %s", daemonSetsInfo[0].String())
```
fromanirh, line 208: if we have more than an option (we do: daemonSetsInfo is a slice) and no option is clearly more important than others (it is not) better to log none of them. Hence
"Created Resource-Topology-Exporter DaemonSets"

controllers/numaresourcesoperator_controller_test.go
Tal-or: Super cool! on it!

fromanirh: we can (and probably should) postponing the event tests to a separate PR (or PRs, considering also the kubeletconfig controller needs work)

fromanirh (commented):
thanks for the improvements. Let's refine further the event texts, because they are going to be part of the API contract.

controllers/numaresourcesoperator_controller_test.go
Tal-or: OK i'll make sure to issue a separate PR for both controllers

controllers/numaresourcesoperator_controller.go
Tal-or: This raises another question (which is unnecessarily related to this PR directly) why we should have a slice of DS when we only have one?
IIRC how DS is working, we have a single DS which makes sure to deploy an RTE pod on every single worker node.

controllers/numaresourcesoperator_controller_test.go
fromanirh: I can handle the kubeletconfig controller part, no worries

controllers/numaresourcesoperator_controller.go
fromanirh: we have a 1:1 between node groups and daemonsets. In the test suite/CI we create only one node group, because it was the simplest test. This is actually a test gap.

Tal-or: Right! I forgot about the node groups. Thanks!

Tal-or: > This is actually a test gap
IMHO it is extensive enough to be tested by QE

controllers/numaresourcesoperator_controller_test.go
fromanirh: here: https://github.com/openshift-kni/numaresources-operator/pull/86

controllers/numaresourcesoperator_controller.go
fromanirh: well, it surely doesn't belong in this PR :)

fromanirh (commented):
/lgtm
/approve
thanks for the changes

PR #85, opened 2022-01-10
Makefile: use the most updated tag as default

Tal-or:
When a user is running ```make deploy``` without specifiying any specific option we should deploy the controller with the most updated version.
The most updated version that is getting build (automatically by the CI) is with the 4.10-snapshot tag, hence that should be the default.

Not running with the most updated version out of the box, might lead to gaps between what specified in the manifests (flags, args, etc.) and what the binary in the image actually contains

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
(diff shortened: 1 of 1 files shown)


fromanirh:
the idea is good, but it clashes with our (and OCP's?) tagging policy

Tal-or:
> the idea is good, but it clashes with our (and OCP's?) tagging policy

it seems so, but I don't understand why in the CI it's using the ```VERSION``` from the Makefile.
I thought we should inject and use the latest build from prow

Tal-or:
Deprecated by https://github.com/openshift/release/pull/25141

PR #86, opened 2022-01-10
tests: kubeletconfig: add tests for events

fromanirh:
Add basic coverage for the events emitted by the kubeletconfig
controller. Being part of the user-visible behaviour, we need
to have at least _some_ test coverage.


fromanirh:
/hold
this is not e2e

fromanirh:
/hold cancel
fixed everything but the branch name, which would require a new PR. I guess we can deal with that

PR #89, opened 2022-01-12
Control component log verbosity

Tal-or:
This PR adds flags in both the CRs to let the cluster-admin control the verbosity of the RTE and of the scheduler plugin containers.

The operator will translate the flag into the proper command-line setting.

The API won't allow maximum granularity for the log level rather it provides **much more** readable, sufficient, and self-explanatory options.


fromanirh:
so far, flawless. Please make sure this code is exercised by at least one e2e test (existing or new), and please make sure to address the CI failure.

Tal-or:
/hold
depend on https://github.com/openshift-kni/numaresources-operator/pull/91

fromanirh:
needs rebase

pkg/objectstate/loglevel/loglevel.go
```diff
@@ -0,0 +1,61 @@
+/*
```
fromanirh, line 1: this is looks like a fine package, but I don't think it belongs to `objectstate`. Let's move up one level:
`pkg/loglevel/...`

fromanirh (commented):
nice work. We need a minor mechanical change and we are good to go.

test/utils/objects/objects.go
```diff
@@ -113,6 +115,17 @@ func TestKC(matchLabels map[string]string) (*machineconfigv1.KubeletConfig, erro
 	}, nil
 }
 
+func IsOwnedBy(obj metav1.ObjectMeta, owner metav1.ObjectMeta) bool {
```
fromanirh, line 118: this works and I'm fine with it. I'd have used the daemonset names reported in the Status, though. No need to change, I don't have strong opinions and this approach has no obvious issues.

fromanirh:
/lgtm
let's move forward and do the code movement in a later PR

PR #90, opened 2022-01-12
Fix uninstall e2e test

marioferh:
Uninstall test fail due delete MCP object is done before PauseMCP for k8s
In fact PauseMCP doesn't find the object.
Fix implement right logic. PauseMCP - Delete nro and kc object - DeleteMCP

CI test were unstable, sometimes test passed. No root cause found. 


jlojosnegros:
/lgtm

There is something I still do not understand. If the problem is that MachineConfigPool was deleted before NRO and so not found, how could this test pass sometimes?

PR #94, opened 2022-01-13
Delete MachineConfigs for deleted NodeGroups

jlojosnegros:
Delete the corresponding MachineConfig for any already deleted NodeGroup
in NRO

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -287,6 +287,11 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 		klog.ErrorS(fmt.Errorf("failed to delete unused daemonsets"), "errors", errorList)
 	}
 
+	errorList = r.deleteUnusedMachineConfigs(ctx, instance, mcps)
+	if len(errorList) > 0 {
+		klog.ErrorS(fmt.Errorf("failed to delete unused machineconfigs"), "errors", errorList)
+	}
+
 	var daemonSetsNName []nropv1alpha1.NamespacedName
 
 	err := rtestate.UpdateDaemonSetUserImageSettings(r.RTEManifests.DaemonSet, instance.Spec.ExporterImage, r.ImageSpec, r.ImagePullPolicy)
@@ -312,11 +317,11 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 }
 
 func (r *NUMAResourcesOperatorReconciler) deleteUnusedDaemonSets(ctx context.Context, instance *nropv1alpha1.NUMAResourcesOperator, mcps []*machineconfigv1.MachineConfigPool) []error {
-	klog.V(3).Info("Delete DS start")
+	klog.V(3).Info("Delete Daemonsets start")
 	var errors []error
 	var daemonSetList appsv1.DaemonSetList
 	if err := r.List(ctx, &daemonSetList, &client.ListOptions{Namespace: instance.Namespace}); err != nil {
-		klog.ErrorS(err, "error while getting DS list")
+		klog.ErrorS(err, "error while getting Daemonset list")
 		return append(errors, err)
 	}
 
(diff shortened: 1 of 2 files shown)


controllers/numaresourcesoperator_controller.go
```diff
@@ -334,8 +339,40 @@ func (r *NUMAResourcesOperatorReconciler) deleteUnusedDaemonSets(ctx context.Con
...
+		}
+	}
+	return errors
+}
+
+func (r *NUMAResourcesOperatorReconciler) deleteUnusedMachineConfigs(ctx context.Context, instance *nropv1alpha1.NUMAResourcesOperator, mcps []*machineconfigv1.MachineConfigPool) []error {
+	klog.V(3).Info("Delete MC start")
+	var errors []error
+	var machineConfigList machineconfigv1.MachineConfigList
+	if err := r.List(ctx, &machineConfigList, &client.ListOptions{Namespace: instance.Namespace}); err != nil {
```
Tal-or, line 355: MachineConfigs are cluster scoped. 
You don't need to mention a Namespace

controllers/numaresourcesoperator_controller.go
```diff
@@ -334,8 +339,40 @@ func (r *NUMAResourcesOperatorReconciler) deleteUnusedDaemonSets(ctx context.Con
...
+				}
+			}
+		}
+	}
+	return errors
+}
+
+func (r *NUMAResourcesOperatorReconciler) deleteUnusedMachineConfigs(ctx context.Context, instance *nropv1alpha1.NUMAResourcesOperator, mcps []*machineconfigv1.MachineConfigPool) []error {
+	klog.V(3).Info("Delete MC start")
+	var errors []error
```
Tal-or, line 353: Is there a special reason why errors are a slice?
Usually, if we're hitting an error we should return immediately

jlojosnegros: I thought this and PR #76  where intended as an improvement to delete unused resources but not as a primary task for the operator, that's why I thought in do it in "best-effort" in case of error and try to delete as much resources as possible not stopping at the first error. 
Also an error on this function or the one deleting DaemonSets do not make the sync function to return as an error.

@fromanirh was also not too sold on this either...

controllers/numaresourcesoperator_controller.go
jlojosnegros: yeah, I will delete it.

controllers/numaresourcesoperator_controller.go
```diff
@@ -334,8 +339,40 @@ func (r *NUMAResourcesOperatorReconciler) deleteUnusedDaemonSets(ctx context.Con
 				if err := r.Client.Delete(ctx, &ds); err != nil {
 					klog.ErrorS(err, "error while deleting daemonset", "DaemonSet", ds.Name)
 					errors = append(errors, err)
+				} else {
+					klog.V(3).Infof("DS[%s] deleted", ds.Name)
```
fromanirh, line 343: let's use the full "Daemonset" (or "DaemonSet") here

controllers/numaresourcesoperator_controller.go
fromanirh: yes, I'm still not sold, but I don't fight against it either. I'm open to small experiments. However, If we decide to change, let's change everything in one go.

The reason I'm sold is that doing this way breaks the expected flow, so one stops and pauses and in the back of their head the "why is this different?" thought starts. It breaks the flow.
Obviously some cases _are_ special enough to break the rules. This case doesn't strike me as obviously special, yet.

jlojosnegros: I mean I just did it that way just because I understood the initiative as a best-effort to delete unused resources, but maybe I misunderstood the intention and then it is better to change it to keep things homogeneous.

controllers/numaresourcesoperator_controller.go
jlojosnegros: sure.

controllers/numaresourcesoperator_controller.go
fromanirh: no worries, it's ok. Let's try this approach and we will have a more informed opinion for the future.

controllers/numaresourcesoperator_controller.go
```diff
@@ -334,8 +339,40 @@ func (r *NUMAResourcesOperatorReconciler) deleteUnusedDaemonSets(ctx context.Con
...
+
+	// Generate the names of the MachineConfigs that should be running,
+	// one for each MachineConfigPool
+	expectedMachineConfigNames := sets.NewString()
+	for _, mcp := range mcps {
+		expectedMachineConfigNames = expectedMachineConfigNames.Insert(rtestate.GetMachineConfigName(instance.Name, mcp.Name))
+	}
+
+	for _, mc := range machineConfigList.Items {
+		if !expectedMachineConfigNames.Has(mc.Name) {
```
fromanirh, line 368: this works. In general we prefer to keep the expected path on the left:
```
if expectedMachineConfigNames.Has(mc.Name) {
	continue
}
if !isOwnedBy(mc.GetObjectMeta(), instance) {
	continue
}
err := r.Client.Delete(ctx, &mc)
if err != nil {
	klog.ErrorS(err, "error while deleting machineconfig", "MachineConfig", mc.Name)
	errors = append(errors, err)
	continue
}
klog.V(3).Infof("Machineconfig [%s] deleted", mc.Name)
```

fromanirh (commented):
/lgtm
/approve

fromanirh (commented):
/lgtm

PR #95, opened 2022-01-13
e2e: start the serial suite

fromanirh:
The "serial" lane will have e2e tests which check the whole scheduler + RTE solution, thus aiming to validate
the real e2e, cluster level solution and not just the individual components.

For this reason, each test need to change the cluster state, even though is responsability of each test to clean up everything.

Hence, each test is potentially disruptive and must run in complete isolation with respect any other cluster activity, not just any other test.

Mimicing(ish) what k8s upstream does, we collect these tests in a "sefial" suite, to be called in a strictly serial and exclusive
(as in mutual exclusion) way. 
The latter has to be addressed using proper cluster and test runner configuration, hence it is out of scope for this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,6 +149,9 @@ binary-e2e-uninstall:
 binary-e2e-sched:
 	go test -c -v -o bin/e2e-sched.test ./test/e2e/sched
 
+binary-e2e-serial:
+	go test -c -v -o bin/e2e-serial.test ./test/e2e/serial
+
 binary-all: binary binary-rte
 
 build: generate fmt vet binary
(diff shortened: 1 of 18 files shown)


fromanirh (commented):
/hold
actually holding again because of the FIXMEs in turn caused by lack of kube 1.23 clusters

fromanirh:
I'll move all the infra work on https://github.com/openshift-kni/numaresources-operator/pull/101, so we can move forward with other PRs once we can cleanly remove the FIXMEs and un-hold this one

fromanirh:
missing parts split in
https://github.com/openshift-kni/numaresources-operator/pull/111
https://github.com/openshift-kni/numaresources-operator/pull/110
https://github.com/openshift-kni/numaresources-operator/pull/109

PR #96, opened 2022-01-14
Change typo affecting NUMAResourcesScheduler tests

jlojosnegros:


fromanirh:
it seems we are starting to flakes, we should start tackling them

test/utils/objects/objects.go
```diff
@@ -37,7 +37,7 @@ func TestNROScheduler() *nropv1alpha1.NUMAResourcesScheduler {
 			APIVersion: nropv1alpha1.GroupVersion.String(),
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "numaresourcesscheudler",
+			Name: "numaresourcesscheduler",
```
fromanirh, line 40: is the CR name checked by the numaresources scheduler controller? if so, how come the CI is not constantly failing because of the typo?
/cc @jlojosnegros @Tal-or

jlojosnegros: I think it is not failing because the NumaresourceScheduler test are not executed unless a environment variable y set to true 

```bash
ENABLE_SCHED_TESTS="${ENABLE_SCHED_TESTS:-false}"

if [ "$ENABLE_SCHED_TESTS" = true ]; then
  test_sched
fi
```

Tal-or: I made the numaresourcesscheduler tests as an opt-in (and mentioned that in the PR when I added them) since they are used as a temporary solution until SSOP kicks in.

Tal-or: Thank you @jlojosnegros for catching this

fromanirh: That was a good idea at time, but it seems our in-house solution will stay with us longer than anticipated. Let's create a new CI lane for testing the scheduler controller, enabling these e2e tests.

fromanirh: Next time let's add this information (why the typo isn't breaking CI, because CI isn't running) in the PR description

Tal-or: @fromanirh I'm not in favor of creating an additional lane since logically it should be part of the E2E test controller flow.
In addition to that, it won't increase the E2E flow significantly since the ```numaresourcesscheduler-controller``` is not dealing with machine-config changes which is the part that takes most of the test's time.

fromanirh: these are valid concerns. Let's address them offline.

PR #98, opened 2022-01-17
Kind E2E test: flakiness resolution attempt   

Tal-or:
This PR intends to address a couple of things:
1. The environment that is provided by GH action CI is extremely poor from a resources perspective.
When NROP deployed as a Burstable pod, it failed to get deployed from time to time due to a lack of resources.
We introduce a separate kustomize overlay for deployment over kind, that won't ask for any resources.

2. In case of test failure, create and export the logs and ```describe``` output from every pod running under the  ```$E2E_NAMESPACE_NAME``` into it's corresponding log file under ```LOG_DIR```


fromanirh (commented):
/lgtm
/approve

fromanirh (commented):
/lgtm
/approve

fromanirh:
the extended timeout doesn't seem to help too much, but the extra logs should. So I'd like to keep this PR.

Tal-or:
> the extended timeout doesn't seem to help too much, but the extra logs should. So I'd like to keep this PR.

The extra logs are not being exported properly and the timeout doesn't help.
I need to revert the timeout change (no point to extend test times for nothing)  and revisit the commit that export the logs

.github/workflows/e2e.yml
```diff
@@ -59,12 +60,24 @@ jobs:
         export KUBECONFIG=${HOME}/.kube/config
         make test-e2e
 
+    - name: Export E2E Tests logs
+      if: ${{ failure() }}
+      run: |
+        mkdir -p ${LOG_DIR}
+        # tail --lines=+2 - do not print first line
+        # awk  '{ print $1 }' - print only first columns
+        for pod in $(kubectl get pods -n $E2E_NAMESPACE_NAME | tail --lines=+2 | awk  '{ print $1 }')
```
fromanirh, line 69: this works, but for the future you can get the very same result using the kubectl format options, simplifying the command line. E.g.: `--no-headers=true -o custom-columns=...`

config/kind-ci/noReservedResources.yaml
```diff
@@ -0,0 +1,17 @@
...
+  selector:
+    matchLabels:
+      control-plane: controller-manager
+  template:
+    spec:
+      containers:
+        - name: manager
+          resources:
+            requests:
+              cpu: 0m
```
fromanirh, line 16: does this disable the requests (OK) or sets them to 0? I'm worried about side effects/unwanted behaviour when we DO set requests with zero value, which I'm not sure how common it is

Tal-or: It is the simplest way for patching.
I need to check how to delete the ```resources``` stanza with patching.

.github/workflows/e2e.yml
Tal-or: I'll look into it

config/kind-ci/noReservedResources.yaml
fromanirh: I don't mind using this way if the end result is the same as removing the requests. If we can demonstrate this referring to the docs/APIs then it's fine and no need to change.

Tal-or: I'm always in favor of doing things the right way. 
And I learned a new thing on the way. check the commit, it's kinda cool.

config/kind-ci/noReservedResources.yaml
```diff
@@ -0,0 +1,18 @@
...
+      control-plane: controller-manager
+  template:
+    spec:
+      containers:
+        - name: manager
+          resources:
+            requests:
+              cpu: 100m
+              memory: 20Mi
+            $patch: delete
```
fromanirh, line 18: nice!

Makefile
```diff
@@ -191,10 +193,10 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 	kubectl apply -f $(CRD_MACHINE_CONFIG_POOL_URL)
 	kubectl apply -f $(CRD_KUBELET_CONFIG_URL)
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | kubectl apply -f -
+	$(KUSTOMIZE) build $(KUSTOMIZE_DEPLOY_DIR)
```
fromanirh, line 196: I can't recall if/how it worked before, without this change. Probably it didn't.

Tal-or: it didn't :)

fromanirh:
> @fromanirh It's up to you if you want to revert this change: #104

I will, but let us wait like a week or so before

Makefile
fromanirh, line 196: ah, the uncareful reviewer missed this bug!

Tal-or: the uncareful author forgot the E2E GH action lane was disabled!

PR #99, opened 2022-01-17
sched: report the scheduler name in the status

fromanirh:
Expose the scheduler name in the NUMAResourcesScheduler Status.
This makes easier for any workload, including e2e tests, to
consume the scheduler.

Workload wishing to be scheduled using the Topology/NUMA aware
scheduler need to be using the same name.

### DIFF
--- a/pkg/numaresourcesscheduler/objectstate/sched/sched.go
+++ b/pkg/numaresourcesscheduler/objectstate/sched/sched.go
@@ -25,11 +25,17 @@ import (
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/k8stopologyawareschedwg/deployer/pkg/manifests"
+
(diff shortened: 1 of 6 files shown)


pkg/numaresourcesscheduler/objectstate/sched/sched.go
```diff
@@ -148,6 +154,32 @@ func DeploymentNamespacedNameFromObject(obj client.Object) (nrsv1alpha1.Namespac
 	return res, ok
 }
 
+func SchedulerNameFromObject(obj client.Object) (string, bool) {
```
Tal-or, line 157: some unit-tests for this function can be useful

Tal-or:
The flaky test fails again I suggest reviewing  https://github.com/openshift-kni/numaresources-operator/pull/98

pkg/numaresourcesscheduler/objectstate/sched/sched.go
fromanirh: Indeed, added! thanks for raising.

PR #100, opened 2022-01-17
Verify TAS deployed properly with default values

jlojosnegros:
Improve verification of deployed resources when NRO use default values.
We need to check if all the resources has been properly deployed. 
There are multiple resources to check but we are going to start checking Deployments, DemonSets And CRDs

### DIFF
--- a/test/e2e/install/install_test.go
+++ b/test/e2e/install/install_test.go
@@ -25,7 +25,6 @@ import (
 	. "github.com/onsi/gomega"
 
 	v1 "k8s.io/api/apps/v1"
-	apiextensionv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -38,12 +37,12 @@ import (
 	"github.com/openshift-kni/numaresources-operator/pkg/status"
 	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
 	"github.com/openshift-kni/numaresources-operator/test/utils/configuration"
+	"github.com/openshift-kni/numaresources-operator/test/utils/crds"
 	"github.com/openshift-kni/numaresources-operator/test/utils/machineconfigpools"
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
 )
 
 const (
-	crdName      = "noderesourcetopologies.topology.node.k8s.io"
 	newTestImage = "quay.io/openshift-kni/numaresources-operator:v0.1.0"
 )
 
@@ -58,9 +57,8 @@ var _ = Describe("[Install] continuousIntegration", func() {
 	})
 
 	Context("with a running cluster with all the components", func() {
-		It("should perform overall deployment and verify the condition is reported as available", func() {
+		It("[test_id: 47574] should perform overall deployment and verify the condition is reported as available", func() {
 			deployedObj := overallDeployment()
-
 			var nname client.ObjectKey
 			for _, obj := range deployedObj {
 				if nroObj, ok := obj.(*nropv1alpha1.NUMAResourcesOperator); ok {
@@ -70,8 +68,9 @@ var _ = Describe("[Install] continuousIntegration", func() {
 			Expect(nname.Name).ToNot(BeEmpty())
 
 			By("checking that the condition Available=true")
+			var updatedNROObj *nropv1alpha1.NUMAResourcesOperator
 			Eventually(func() bool {
-				updatedNROObj := &nropv1alpha1.NUMAResourcesOperator{}
+				updatedNROObj = &nropv1alpha1.NUMAResourcesOperator{}
 				err := e2eclient.Client.Get(context.TODO(), nname, updatedNROObj)
 				if err != nil {
 					klog.Warningf("failed to get the RTE resource: %v", err)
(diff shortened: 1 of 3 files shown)


test/e2e/install/install_test.go
```diff
@@ -58,8 +60,21 @@ var _ = Describe("[Install] continuousIntegration", func() {
 	})
 
 	Context("with a running cluster with all the components", func() {
-		It("should perform overall deployment and verify the condition is reported as available", func() {
-			deployedObj := overallDeployment()
+		var deployedObj []client.Object
+		BeforeEach(func() {
```
fromanirh, line 64: why was this moved to `beforeeach`?
I realize it could be confusing (and actually wrong) to leverage the side effect the suite is leaving TAS installed on the system; this should be at least documented.

test/e2e/install/install_test.go
```diff
@@ -300,3 +329,30 @@ func getDaemonSetByOwnerReference(uid types.UID) (*v1.DaemonSet, error) {
...
+func getWorkerNodes() ([]*corev1.Node, error) {
+	nodeList := &corev1.NodeList{}
+	var ret []*corev1.Node
+
+	if err := e2eclient.Client.List(context.TODO(), nodeList); err != nil {
+		return nil, fmt.Errorf("failed to get nodes: %w", err)
+	}
+
+	for _, node := range nodeList.Items {
+		_, ok := node.Labels["node-role.kubernetes.io/worker"]
```
fromanirh, line 351: you can filter by labels using `ListOptions`.
This code is wrong because is too generic: we need to fetch the nodes referred by the `NodeGroups` we configured.
There is a fair bit of refactoring/code movement that can be useful for tests like this

test/e2e/sched/install/install_test.go
```diff
@@ -71,9 +74,18 @@ var _ = Describe("[Scheduler] install", func() {
...
 			Expect(err).NotTo(HaveOccurred())
 			Expect(deploy.Status.ReadyReplicas).To(BeIdenticalTo(int32(1)))
+
+			By("checking the NumaResourcesScheduler CRD is deployed")
+			crd := &apiextensionv1.CustomResourceDefinition{}
+			key := client.ObjectKey{
+				Name: crdName,
+			}
+			err = e2eclient.Client.Get(context.TODO(), key, crd)
+			Expect(err).NotTo(HaveOccurred())
```
fromanirh, line 88: I'm inclined to think we may need a new utility function. My own threshold is "used in two sites", and we already hit the threshold. More will likely come in the next batch of e2e tests.

test/e2e/sched/install/install_test.go
```diff
@@ -71,9 +74,18 @@ var _ = Describe("[Scheduler] install", func() {
 			err = e2eclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(nroSchedObj), nroSchedObj)
 			Expect(err).NotTo(HaveOccurred())
 
+			By("checking the NumaResourcesScheduler Deployment is correctly deployed")
 			deploy, err := schedutils.GetDeploymentByOwnerReference(nroSchedObj.UID)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(deploy.Status.ReadyReplicas).To(BeIdenticalTo(int32(1)))
+
+			By("checking the NumaResourcesScheduler CRD is deployed")
```
fromanirh, line 82: we also need to check the scheduler deployment

fromanirh (commented):
we need more context both in the commit message and in the PR cover letter (the first message)

test/e2e/install/install_test.go
fromanirh: documented -> not in this PR (separate one)

jlojosnegros: Well I just saw newly tests introduced in the package were following this pattern and I though it would be a good idea to make them all work in the same way and also clean-up the system after the test. 
I did not add any comment because I though it would not be needed as I was just using an already used pattern.
I will add some comments.

test/e2e/sched/install/install_test.go
jlojosnegros: Sorry but scheduler deployment is not the one checked in L78?

test/e2e/sched/install/install_test.go
jlojosnegros: ok

test/e2e/sched/install/install_test.go
fromanirh: right, overlooked this part

test/e2e/install/install_test.go
fromanirh: ok, no problem. In this case we need a separate PR to document we intentionally leave the cluster state mutated after the install suite runs, and to explain why we do like we do.

jlojosnegros: Ok I though you want to comment why I was changing this to BeforeEach ... Now I see the point.

test/e2e/install/install_test.go
```diff
@@ -38,12 +38,14 @@ import (
...
 	"github.com/openshift-kni/numaresources-operator/test/utils/configuration"
+	"github.com/openshift-kni/numaresources-operator/test/utils/crds"
 	"github.com/openshift-kni/numaresources-operator/test/utils/machineconfigpools"
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
 )
 
 const (
-	crdName      = "noderesourcetopologies.topology.node.k8s.io"
+	crdNRTName   = "noderesourcetopologies.topology.node.k8s.io"
+	crdNROName   = "numaresourcesoperators.nodetopology.openshift.io"
```
fromanirh, line 48: not sure these belongs here or in the new `crds` package. But it's minor.

test/e2e/install/install_test.go
```diff
@@ -90,12 +92,25 @@ var _ = Describe("[Install] continuousIntegration", func() {
...
 			Expect(err).NotTo(HaveOccurred())
+
+			By("checking Daemonset is up&running")
+			workerNodes, err := getWorkerNodes()
+			Expect(err).NotTo(HaveOccurred())
+
+			ds, err := getDaemonSetByOwnerReference(updatedNROObj.UID)
+			Expect(err).NotTo(HaveOccurred())
+
+			workerNodesNumber := int32(len(workerNodes))
```
fromanirh, line 109: rather than using this value, we should use the `desiredNumberScheduled` as target. We need to assume the value is correctly computed (and maybe test it on a separate test). IOW, we don't need to count the worker nodes (which we are doing incorrectly here).

test/e2e/install/install_test.go
```diff
@@ -90,12 +92,25 @@ var _ = Describe("[Install] continuousIntegration", func() {
...
+			Expect(ds.Status.DesiredNumberScheduled).To(Equal(workerNodesNumber))
+			Expect(ds.Status.CurrentNumberScheduled).To(Equal(workerNodesNumber))
+			Expect(ds.Status.NumberReady).To(Equal(workerNodesNumber))
```
fromanirh, line 112: this works in the happy path: lines 106-112 should be wrapped in a `Eventually` block.

test/e2e/sched/install/install_test.go
```diff
@@ -71,9 +74,14 @@ var _ = Describe("[Scheduler] install", func() {
 			err = e2eclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(nroSchedObj), nroSchedObj)
 			Expect(err).NotTo(HaveOccurred())
 
+			By("checking the NumaResourcesScheduler Deployment is correctly deployed")
 			deploy, err := schedutils.GetDeploymentByOwnerReference(nroSchedObj.UID)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(deploy.Status.ReadyReplicas).To(BeIdenticalTo(int32(1)))
```
fromanirh, line 80: just noticed that this check has the same issue of the daemonset check. Can be fixed in a later PR.

test/e2e/install/install_test.go
jlojosnegros: yeah you are totally right ... will change it

test/e2e/install/install_test.go
jlojosnegros: Ok, understood. I will change it.

For the sake of curiosity... will this be a proper way to found the nodes where daemonset should be running ?

```go
func getWorkerNodes(labelSelector map[string]string) ([]corev1.Node, error) {
	nodeList := &corev1.NodeList{}

	if err := e2eclient.Client.List(context.TODO(), nodeList, &client.ListOptions{LabelSelector: labels.SelectorFromSet(labelSelector)}); err != nil {
		return nil, fmt.Errorf("failed to get nodes: %w", err)
	}
	return nodeList.Items, nil
}
```
using  the MachineConfigPool.Spec.NodeSelector.MatchLabels as `labelSelector`

test/e2e/install/install_test.go
jlojosnegros: done

test/e2e/install/install_test.go
```diff
@@ -90,12 +89,45 @@ var _ = Describe("[Install] continuousIntegration", func() {
...
+					klog.Infof(" CurrentNumberScheduled %d (should be %d)", ds.Status.CurrentNumberScheduled, ds.Status.DesiredNumberScheduled)
+					return false
+				}
+
+				if ds.Status.NumberReady != ds.Status.DesiredNumberScheduled {
+					klog.Infof(" NumberReady %d (should be %d)", ds.Status.CurrentNumberScheduled, ds.Status.DesiredNumberScheduled)
+					return false
+				}
+				return true
+			}, DSCheckTimeout, DSCheckPollingPeriod).Should(BeTrue(), "DaemonSet Status was not correct")
```
fromanirh, line 130: for the next time: "DaemonSet status is not correct" is too generic. The test should clearly signal at glance what the expectation is and what the actual state is. Yes, we (correctly) do logging, but in some test runners we may want not to enable the logs by default, so better to reiterate the failed condition as precisely as possible.

fromanirh (commented):
/lgtm
/approve
we will need a followup to move the deployment verification in Eventually

fromanirh (commented):
/lgtm

test/e2e/install/install_test.go
fromanirh: there is code in the `controllers` package (soon to be moved on its own package) to resolve nodegroups which we can use here

PR #102, opened 2022-01-18
e2e: introduce NO_TEARDOWN option

Tal-or:
In order to explain the motivation of this PR let's define the following terms:
Clean state: The cluster state, **before** any MC changes and/or TAS components were applied.
Dirty state: The cluster state, **after** any/all MC changes and/or TAS components were applied.

In the ususal test cases, we would always want the cluster to be at clean state, after the test run.
But there are special cases, like local tests and GH action tests, that we would like to keep the cluster in dirty state so we could perform various investigation, exporting logs, events, etc.

For that, we'll introduce the NO_TEARDOWN option, that will allow us to keep the cluster in dirty state after test run.
This is totaly optional and keep the current behavior intact.

### DIFF
--- a/hack/run-test-e2e.sh
+++ b/hack/run-test-e2e.sh
@@ -3,10 +3,11 @@
 source hack/common.sh
 
 ENABLE_SCHED_TESTS="${ENABLE_SCHED_TESTS:-false}"
+NO_TEARDOWN="${NO_TEARDOWN:-false}"
 
 function test_sched() {
   # Make sure that we always properly clean the environment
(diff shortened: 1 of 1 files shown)


fromanirh (commented):
/lgtm
the composed effect of double negative logic (`if NO_SOMETHING == false`) and nested logic is a bit hard to read. The idea and the logic lgtm though.

Tal-or:
@fromanirh Please retest CI E2E / e2e test when you can.
I don't have permissions

PR #103, opened 2022-01-18
numaresourcesscheduler: configurable scheduler name

Tal-or:
This PR contains the necessary changes in the ```NUMAResourceScheduler``` CR and controller in order for the user to be able to configure the ```SchedulerName``` field.

PR contains both implementation and unit tests.

E2E tests will be added on a separate PR later on since there is a lot of activity in this area atm and there is no point with stepping each other toes.

### DIFF
--- a/controllers/numaresourcesscheduler_controller.go
+++ b/controllers/numaresourcesscheduler_controller.go
@@ -152,11 +152,18 @@ func isDeploymentRunning(ctx context.Context, c client.Client, key nrsv1alpha1.N
 }
 
 func (r *NUMAResourcesSchedulerReconciler) syncNUMASchedulerResources(ctx context.Context, instance *nrsv1alpha1.NUMAResourcesScheduler) (nrsv1alpha1.NamespacedName, string, error) {
+	var deploymentNName nrsv1alpha1.NamespacedName
+	schedulerName := instance.Spec.SchedulerName
+
 	schedstate.UpdateDeploymentImageSettings(r.SchedulerManifests.Deployment, instance.Spec.SchedulerImage)
 	schedstate.UpdateDeploymentConfigMapSettings(r.SchedulerManifests.Deployment, r.SchedulerManifests.ConfigMap.Name)
+	if schedulerName != "" {
+		err := schedstate.UpdateSchedulerName(r.SchedulerManifests.ConfigMap, instance.Spec.SchedulerName)
+		if err != nil {
+			return nrsv1alpha1.NamespacedName{}, schedulerName, err
+		}
+	}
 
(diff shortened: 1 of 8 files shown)


pkg/numaresourcesscheduler/objectstate/sched/sched.go
```diff
@@ -180,6 +181,45 @@ func SchedulerNameFromObject(obj client.Object) (string, bool) {
 	return "", false
 }
 
+func UpdateSchedulerName(cm *corev1.ConfigMap, name string) error {
+	if name == "" {
+		return nil
```
fromanirh, line 186: let's error out here. Empty scheduler name is almost always a mistake unless we have a crystal clear use case for it (nothing in sight)

controllers/numaresourcesscheduler_controller.go
```diff
@@ -152,11 +152,16 @@ func isDeploymentRunning(ctx context.Context, c client.Client, key nrsv1alpha1.N
 }
 
 func (r *NUMAResourcesSchedulerReconciler) syncNUMASchedulerResources(ctx context.Context, instance *nrsv1alpha1.NUMAResourcesScheduler) (nrsv1alpha1.NamespacedName, string, error) {
+	var deploymentNName nrsv1alpha1.NamespacedName
+	schedulerName := instance.Spec.SchedulerName
+
 	schedstate.UpdateDeploymentImageSettings(r.SchedulerManifests.Deployment, instance.Spec.SchedulerImage)
 	schedstate.UpdateDeploymentConfigMapSettings(r.SchedulerManifests.Deployment, r.SchedulerManifests.ConfigMap.Name)
+	err := schedstate.UpdateSchedulerName(r.SchedulerManifests.ConfigMap, instance.Spec.SchedulerName)
```
fromanirh, line 160: let's make this more explicit: if not name was set excplicitely, don't update and stick with whatever default we have in the YAML asset. So let's move the check `if schedulerName != '"` here in the calling site...

pkg/numaresourcesscheduler/objectstate/sched/sched.go
```diff
@@ -180,6 +181,45 @@ func SchedulerNameFromObject(obj client.Object) (string, bool) {
 	return "", false
 }
 
+func UpdateSchedulerName(cm *corev1.ConfigMap, name string) error {
+	if name == "" {
```
fromanirh, line 185: ...and let's error out if we receive a empty name

fromanirh (commented):
looks almost good!

fromanirh (commented):
/lgtm
thanks for the updates!

PR #104, opened 2022-01-19
gh actions: opt in the kind e2e flow

fromanirh:
We always considered the kind e2e flow a extra sanity check
and a tier-2 platform because this operator has -and will have-
a number of unavoidable OCP dependencies.

Since the kind e2e lane is flaking recently and no good
resolution is in sight, we make the kind e2e-lane opt in.

We totally want to keep deploying on kind, and we still
recommend to test also on kind.

gh docs references:
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push

inspired by:
https://github.com/veggiemonk/skip-commit/issues/5

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -54,6 +54,7 @@ jobs:
         IMG=${built_image} KUSTOMIZE_DEPLOY_DIR="config/kind-ci/" make deploy
 
     - name: E2E Tests
+      if: "contains(github.event.head_commit.message, '[ci-run-e2e-kind]')"
       run: |
         export KUBECONFIG=${HOME}/.kube/config
         make test-e2e


Tal-or:
It doesn't seems like it's working.
IIUC the lane should run only if there's a ```ci-run-e2e-kind``` in the head of the commit message

fromanirh:
> It doesn't seems like it's working. IIUC the lane should run only if there's a `ci-run-e2e-kind` in the head of the commit message

there is not (yet) a easy way to exit early the lane, so the only option is to skip steps. I'm starting by skipping only the e2e run step, and intentionally running all the other steps (build, deploy...) to at least continuosly get _some_ signal from the vanilla k8s environment.
image

fromanirh:
/override ci/prow/ci-e2e
1. this change cannot affect anything but the gh actions flow (already verified)
2. the overridden lane passed just before merging a unrelated change which triggered the rerun

so let's (exceptionally) save time and unblock the pending PRs.

PR #105, opened 2022-01-19
Make sched deployment checking more time robust

jlojosnegros:
Follow-up for https://github.com/openshift-kni/numaresources-operator/pull/100#discussion_r786541738 where it was found this check could fail due to timeout.
Wrapping scheduler deployment check in an Eventually block make it more resilience to a slow CI

### DIFF
--- a/test/e2e/sched/install/install_test.go
+++ b/test/e2e/sched/install/install_test.go
@@ -75,9 +75,21 @@ var _ = Describe("[Scheduler] install", func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			By("checking the NumaResourcesScheduler Deployment is correctly deployed")
-			deploy, err := schedutils.GetDeploymentByOwnerReference(nroSchedObj.UID)
-			Expect(err).NotTo(HaveOccurred())
-			Expect(deploy.Status.ReadyReplicas).To(BeIdenticalTo(int32(1)))
+			const deploymentCheckTimeout = 5 * time.Minute
+			const deploymentCheckPollPeriod = 10 * time.Second
+			Eventually(func() bool {
+				deploy, err := schedutils.GetDeploymentByOwnerReference(nroSchedObj.UID)
+				if err != nil {
+					klog.Warningf("unable to get deployment by owner reference: %v", err)
+					return false
+				}
+
+				if deploy.Status.ReadyReplicas != *deploy.Spec.Replicas {
+					klog.Warningf("Invalid number of ready replicas: desired: %d, actual: %d", *deploy.Spec.Replicas, deploy.Status.ReadyReplicas)
+					return false
+				}
+				return true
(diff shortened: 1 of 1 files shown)


test/e2e/sched/install/install_test.go
```diff
@@ -75,9 +75,21 @@ var _ = Describe("[Scheduler] install", func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			By("checking the NumaResourcesScheduler Deployment is correctly deployed")
-			deploy, err := schedutils.GetDeploymentByOwnerReference(nroSchedObj.UID)
-			Expect(err).NotTo(HaveOccurred())
-			Expect(deploy.Status.ReadyReplicas).To(BeIdenticalTo(int32(1)))
+			const DeploymentCheckTimeout = 5 * time.Minute
+			const DeploymentCheckPollPeriod = 10 * time.Second
```
fromanirh, line 79: C++-ism :) I'd rather avoid it but not a big deal

test/e2e/sched/install/install_test.go
```diff
@@ -75,9 +75,21 @@ var _ = Describe("[Scheduler] install", func() {
...
+			Eventually(func() bool {
+				deploy, err := schedutils.GetDeploymentByOwnerReference(nroSchedObj.UID)
+				if err != nil {
+					klog.Warningf("unable to get deployment by owner reference: %v", err)
+					return false
+				}
+
+				if deploy.Status.ReadyReplicas != 1 {
```
fromanirh, line 87: Let's not hardcodec constants: let's check the ready replicas are equal to the expected replicas (can't remember the field name atm)

test/e2e/sched/install/install_test.go
fromanirh: on second thought we can also keep them
not a big deal

fromanirh (commented):
it's almost ready, let's just not hardcoded the expected replica count. Everything else is actually ok as it is.

test/e2e/sched/install/install_test.go
jlojosnegros: maybe I should change the names to start with a lowercase letter to be more go-ish style

fromanirh: I think yes, but it's minor.

test/e2e/sched/install/install_test.go
jlojosnegros: ok

fromanirh:
/lgtm
/approve
we have some utilities which do the same check, and which may want to reuse, but better consolidate with a later PR

jlojosnegros:
> /lgtm /approve we have some utilities which do the same check, and which may want to reuse, but better consolidate with a later PR

Sure, once #95 is merged we could reconsider all these tests, and maybe consolidate all the different "utilities" folders under test.

PR #109, opened 2022-01-21
e2e: serial: initial tests for workload resilience

fromanirh:
Add test to make sure that if the topology-aware
scheduler is undeployed (or is somehow unavailable) while
the workload is running, the aforementioned workload is
unaffected; also new workloads requiring that scheduler
must be kept pending.

### DIFF
--- a/test/e2e/serial/sched_removal_test.go
+++ b/test/e2e/serial/sched_removal_test.go
@@ -1,0 +1,214 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


fromanirh:
partially replaces #95

test/e2e/serial/sched_removal_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+				updatedDp := &appsv1.Deployment{}
+				err = fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(dp), updatedDp)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(e2ewait.IsDeploymentComplete(dp, &updatedDp.Status)).To(BeTrue(), "deployment %q become unready", dp.Name)
+			}
+		})
+
+		It("should keep new scheduled workloads pending", func() {
```
shajmakh, line 105: can we recreate the scheduler after this step then verify the pending pods are scheduled?
I am aware that this step wasn't initially in the TC, but I thought it would be useful to check it along with the scenario. another option would be having a separate TC covering this check. wdyt?

fromanirh: yes, it's a reasonnable thing to test. Will add a new test case.

shajmakh: Thank you. I have created the TC & the id is 48069.

fromanirh: thanks, done.

PR #111, opened 2022-01-21
pkg: controllers: move MCP code in its own package

fromanirh:
Extract MCP utilities from the controllers to its own package,
so the test code doesn't depend on the controllers anymore.


fromanirh:
replaced by https://github.com/openshift-kni/numaresources-operator/pull/113

PR #116, opened 2022-01-25
e2e: padder package

Tal-or:
The padder package allows controlling the number of available resources under the nodes.
It's doing that by padding the nodes with idle pods, until the node reaches the desired amount of available resources.

This package comes very handy on cases when we need specific prerequisites to be apply on nodes,
wrt the amount of available resources they have prior to the test run.

Usage example:
```
pad := New(client, test-ns)
pad.Nodes(2).UntilAvailableReach(corev1.ResourceCPU, 3).UntilAvailableReach(corev1.ResourceMemory, 100Mi).Pad()
```
or
```
rl := corev1.ResourceList {
corev1.ResourceCPU: 10,
corev1.ResourceMemory: 100Mi
}
pad.Nodes(2).UntilAvailableReachWithResourceList(rl).Pad()
```

In an ideal situation, this package would deserve some unit tests, but unit tests are complex to implement here, and anyway we can trust that if there is a bug in the package, the test that is using it won't pass.

### DIFF
--- a/test/utils/noderesourcetopologies/noderesourcetopologies.go
+++ b/test/utils/noderesourcetopologies/noderesourcetopologies.go
@@ -208,7 +208,7 @@ func FilterAnyZoneMatchingResources(nrts []nrtv1alpha1.NodeResourceTopology, req
 	for _, nrt := range nrts {
 		matches := 0
 		for _, zone := range nrt.Zones {
-			if !zoneResourcesMatchesRequest(zone.Resources, requests) {
+			if !ZoneResourcesMatchesRequest(zone.Resources, requests) {
 				continue
 			}
 			klog.Infof(" ----> node %q zone %q provides at least %#v", nrt.Name, zone.Name, ResourceListToString(requests))
@@ -233,16 +233,17 @@ func FindFromList(nrts []nrtv1alpha1.NodeResourceTopology, name string) (*nrtv1a
 	return nil, fmt.Errorf("failed to find NRT for %q", name)
 }
 
-func findZoneByName(nrt nrtv1alpha1.NodeResourceTopology, zoneName string) (*nrtv1alpha1.Zone, error) {
-	for idx := 0; idx < len(nrt.Zones); idx++ {
-		if nrt.Zones[idx].Name == zoneName {
-			return &nrt.Zones[idx], nil
-		}
+// AvailableFromZone returns a ResourceList of all available resources under the zone
+func AvailableFromZone(z nrtv1alpha1.Zone) corev1.ResourceList {
+	rl := corev1.ResourceList{}
+
+	for _, ri := range z.Resources {
+		rl[corev1.ResourceName(ri.Name)] = ri.Available
 	}
-	return nil, fmt.Errorf("cannot find zone %q", zoneName)
+	return rl
 }
 
-func zoneResourcesMatchesRequest(resources []nrtv1alpha1.ResourceInfo, requests corev1.ResourceList) bool {
+func ZoneResourcesMatchesRequest(resources []nrtv1alpha1.ResourceInfo, requests corev1.ResourceList) bool {
 	for resName, resQty := range requests {
 		zoneQty, ok := findResourceAvailableByName(resources, string(resName))
 		if !ok {
@@ -253,6 +254,15 @@ func zoneResourcesMatchesRequest(resources []nrtv1alpha1.ResourceInfo, requests 
 		}
 	}
 	return true
+}
+
+func findZoneByName(nrt nrtv1alpha1.NodeResourceTopology, zoneName string) (*nrtv1alpha1.Zone, error) {
+	for idx := 0; idx < len(nrt.Zones); idx++ {
+		if nrt.Zones[idx].Name == zoneName {
+			return &nrt.Zones[idx], nil
+		}
+	}
+	return nil, fmt.Errorf("cannot find zone %q", zoneName)
 }
 
 func findResourceAvailableByName(resources []nrtv1alpha1.ResourceInfo, name string) (resource.Quantity, bool) {
--- a/test/utils/padder/padder.go
+++ b/test/utils/padder/padder.go
@@ -1,0 +1,212 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package padder
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	nrtv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
+	"github.com/openshift-kni/numaresources-operator/test/utils/fixture"
+	nrtutil "github.com/openshift-kni/numaresources-operator/test/utils/noderesourcetopologies"
+	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
+	"github.com/openshift-kni/numaresources-operator/test/utils/objects/wait"
+)
+
+// This package allows to control the amount of available allocationTarget under the nodes.
+// it's doing that by "padding" the nodes with guaranteed pods until node reaches the desired amount of available allocationTarget.
+//
+// Usage example:
+// pad := New(client, "test-ns")
+// pad.Nodes(2).UntilAvailableIsResource(corev1.ResourceCPU, "3").UntilAvailableIsResource(corev1.ResourceMemory, "100Mi").Pad()
+// or
+// rl := corev1.ResourceList {
+//		corev1.ResourceCPU: "10",
+//		corev1.ResourceMemory: "100Mi"
+//	}
+// pad.Nodes(2).UntilAvailableIsResourceList(rl).Pad()
+
+const PadderLabel = "nrop-test-pad-pod"
+
+type Padder struct {
+	// Client defines the API client to run CRUD operations, that will be used for testing
+	Client    client.Client
+	namespace string
+	*padRequest
+}
+
+type padRequest struct {
+	nNodes           int
+	allocationTarget corev1.ResourceList
+}
+
+// New return new padder object
+// ns is where padding pods will get deployed
+func New(cli client.Client, ns string) (*Padder, error) {
+	return &Padder{
+		Client:     cli,
+		namespace:  ns,
+		padRequest: &padRequest{allocationTarget: make(corev1.ResourceList)},
+	}, nil
+}
+
+// Nodes number of nodes to pad
+func (p *Padder) Nodes(n int) *Padder {
+	p.padRequest.nNodes = n
+	return p
+}
+
+// UntilAvailableIsResource will pad pods into nodes until reach the desired available allocationTarget
+func (p *Padder) UntilAvailableIsResource(resName corev1.ResourceName, quantity string) *Padder {
+	quan := resource.MustParse(quantity)
+	p.allocationTarget[resName] = quan
+	return p
+}
+
+// UntilAvailableIsResourceList is like UntilAvailableIsResource but with a complete ResourceList as a parameter
+func (p *Padder) UntilAvailableIsResourceList(resources corev1.ResourceList) *Padder {
+	p.allocationTarget = resources
+	return p
+}
+
+// Pad will create pad pods in order to align the nodes
+// with the requested amount of available allocationTarget
+func (p *Padder) Pad() error {
+	if p.nNodes == 0 {
+		klog.Warningf("no nodes for padding were found. please specify at least one node")
+		return nil
+	}
+
+	nrtList, err := nrtutil.GetUpdated(p.Client, nrtv1alpha1.NodeResourceTopologyList{}, time.Second*10)
+	if err != nil {
+		return err
+	}
+
+	singleNumaNrt := filterSingleNumaNodePolicyNrts(nrtList.Items)
+
+	for i := 0; i < len(singleNumaNrt) && p.nNodes > 0; i++ {
+		nrt := singleNumaNrt[i]
+		nodePadded := false
+		for _, zone := range nrt.Zones {
+			// check that zone has at least the amount of allocationTarget that needed
+			if nrtutil.ZoneResourcesMatchesRequest(zone.Resources, p.allocationTarget) {
+				availResList := nrtutil.AvailableFromZone(zone)
+				diffList, err := diffAvailableToExpected(availResList, p.allocationTarget)
+				if err != nil {
+					return err
+				}
+
+				// create a pod that asks for exact amount of allocationTarget as the diff
+				// in order to reach to desired amount of available allocationTarget in the node
+				padPod := objects.NewTestPodPause(p.namespace, fixture.RandomName("padder"))
+
+				// label the pod with a pad label, so it will be easier to identify later
+				labelPod(padPod, map[string]string{PadderLabel: ""})
+
+				padPod.Spec.Containers[0].Resources.Limits = diffList
+				padPod.Spec.Containers[0].Resources.Requests = diffList
+
+				// place the pod on the selected node
+				//padPod.Spec.NodeSelector = map[string]string{nodes.LabelHostname: nrt.Name}
+				padPod.Spec.NodeSelector = map[string]string{"kubernetes.io/hostname": nrt.Name}
+				if err := p.Client.Create(context.TODO(), padPod); err != nil {
+					return err
+				}
+				if _, err = wait.ForPodPhase(p.Client, p.namespace, padPod.Name, corev1.PodRunning, time.Minute); err != nil {
+					return err
+				}
+				klog.InfoS("created pod", "pod", fmt.Sprintf("%s/%s", padPod.Namespace, padPod.Name), "node", nrt.Name)
+				nodePadded = true
+			} else {
+				klog.Warningf("node: %q zone: %q, doesn't have enough available allocationTarget", nrt.Name, zone.Name)
+			}
+		}
+		if nodePadded {
+			p.nNodes--
+		}
+	}
+	return nil
+}
+
+// Clean can be called after test finished
+// in order to clean all the padding pod in an easier way
+func (p *Padder) Clean() error {
+	pod := &corev1.Pod{}
+	opts := []client.DeleteAllOfOption{
+		client.InNamespace(p.namespace),
+		client.MatchingLabels{PadderLabel: ""},
+		client.GracePeriodSeconds(5),
+	}
+	err := p.Client.DeleteAllOf(context.TODO(), pod, opts...)
+	if errors.IsNotFound(err) {
+		err = nil
+	}
+	return err
+}
+
+// return a resourceList of differences between the available and expected amount of resources
+// it will return an error if requested allocationTarget are more than available
+func diffAvailableToExpected(availableResLst corev1.ResourceList, expectedResLst corev1.ResourceList) (corev1.ResourceList, error) {
+	// this list will be requested by the pad pod
+	// in order to align the nodes resources with the expectedResLst
+	resourceList := corev1.ResourceList{}
+
+	for n, q1 := range availableResLst {
+		if q2, ok := expectedResLst[n]; ok {
+			// resource exist in expected list
+			// calc the diff
+			q1.Sub(q2)
+			if q1.Sign() == -1 {
+				return resourceList, fmt.Errorf("expected resource: %s=%v is higher than available: %s=%v", n, q2.String(), n, q1.String())
+			}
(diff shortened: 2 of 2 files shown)


test/utils/padder/padder.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// New return new padder object
+// ns is where padding pods will get deployed
+func New(ns string) (*Padder, error) {
+	padder := &Padder{
+		Namespace: ns,
+	}
+
+	if !e2eclient.ClientsEnabled {
```
fromanirh, line 68: i'd pass to this function a `client.Client`, leaving to the calling code how to create it (`e2eclient` or anything else)

test/utils/padder/padder.go
```diff
@@ -0,0 +1,168 @@
...
+// pad.Nodes(2).UntilAvailableReachWithResourceList(rl).Pad()
+
+type Padder struct {
+	// Client defines the API client to run CRUD operations, that will be used for testing
+	Client    client.Client
+	Namespace string
+	*PadRequest
+}
+
+type PadRequest struct {
```
fromanirh, line 56: do we need this type to be public?

test/utils/padder/padder.go
```diff
@@ -0,0 +1,168 @@
...
+
+// Nodes number of nodes to pad
+func (p *Padder) Nodes(n int) *Padder {
+	p.PadRequest.nNodes = n
+	return p
+}
+
+// UntilAvailableReach will pad pods into nodes until reach the desired available resources
+func (p *Padder) UntilAvailableReach(resName corev1.ResourceName, quantity string) *Padder {
+	quan := resource.MustParse(quantity)
```
fromanirh, line 85: panicing here is not a great UX, though.

test/utils/padder/padder.go
```diff
@@ -0,0 +1,168 @@
...
+	nrtutil "github.com/openshift-kni/numaresources-operator/test/utils/noderesourcetopologies"
+	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
+)
+
+// This package allows to control the amount of available resources under the nodes.
+// it's doing that by "padding" the nodes with guaranteed pods until node reaches the desired amount of available resources.
+//
+// Usage example:
+// pad := New("test-ns")
+// pad.Nodes(2).UntilAvailableReach(corev1.ResourceCPU, "3").UntilAvailableReach(corev1.ResourceMemory, "100Mi").Pad()
```
fromanirh, line 41: Not a huge fan of the fluent API, but not against them either. We can try this and see how it turns out.

fromanirh (commented):
very initial pass, need to have a deep look still.

test/utils/padder/padder.go
Tal-or: Probably not, let me fix that.

test/utils/padder/padder.go
Tal-or: Agree. 
But:
1. It's a testing utility that is used for testing purposes only, so we basically know the string is valid.
2. If we need to return an error here, it kinda breaks the fluent  API design.

fromanirh: Valid points. Let's keep like it is (circling a bit back to "not a huge fan of the fluent api design", but still)

Tal-or:
@fromanirh in general, do you think that a utility package like this one required a unit-tests of its own.
It's kinda testing the tests

fromanirh:
> @fromanirh in general, do you think that a utility package like this one required a unit-tests of its own. It's kinda testing the tests

not unless the logic is supercomplex. In that case it should probably moved on its own package (with unit tests :) ). Testing the test should be exceptional for the cases we care about.

test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -186,16 +186,17 @@ func FindFromList(nrts []nrtv1alpha1.NodeResourceTopology, name string) (*nrtv1a
 	return nil, fmt.Errorf("failed to find NRT for %q", name)
 }
 
-func findZoneByName(nrt nrtv1alpha1.NodeResourceTopology, zoneName string) (*nrtv1alpha1.Zone, error) {
-	for idx := 0; idx < len(nrt.Zones); idx++ {
-		if nrt.Zones[idx].Name == zoneName {
-			return &nrt.Zones[idx], nil
-		}
+// AvailableFromZone returns a ResourceList of all available resources under the zone
+func AvailableFromZone(z nrtv1alpha1.Zone) corev1.ResourceList {
```
fromanirh, line 190: this is fine, but we should (briefly) expain "why/how" in the commit message rather than "what", because "what" must be evident from the PR itself, and if the PR is too big to make it evident, we will need to split it.

test/utils/padder/padder.go
```diff
@@ -0,0 +1,202 @@
...
+//
+// Usage example:
+// pad := New(client, "test-ns")
+// pad.Nodes(2).UntilAvailableReach(corev1.ResourceCPU, "3").UntilAvailableReach(corev1.ResourceMemory, "100Mi").Pad()
+// or
+// rl := corev1.ResourceList {
+//		corev1.ResourceCPU: "10",
+//		corev1.ResourceMemory: "100Mi"
+//	}
+// pad.Nodes(2).UntilAvailableReachWithResourceList(rl).Pad()
```
fromanirh, line 47: better: `UntilAvailableIsResourceList`

test/utils/padder/padder.go
```diff
@@ -0,0 +1,202 @@
...
+	nrtutil "github.com/openshift-kni/numaresources-operator/test/utils/noderesourcetopologies"
+	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
+)
+
+// This package allows to control the amount of available untilAvailableReach under the nodes.
+// it's doing that by "padding" the nodes with guaranteed pods until node reaches the desired amount of available untilAvailableReach.
+//
+// Usage example:
+// pad := New(client, "test-ns")
+// pad.Nodes(2).UntilAvailableReach(corev1.ResourceCPU, "3").UntilAvailableReach(corev1.ResourceMemory, "100Mi").Pad()
```
fromanirh, line 41: Better: `UntilAvailableIsResource`

test/utils/padder/padder.go
```diff
@@ -0,0 +1,202 @@
...
+//		corev1.ResourceMemory: "100Mi"
+//	}
+// pad.Nodes(2).UntilAvailableReachWithResourceList(rl).Pad()
+
+const PadderLabel = "nrop-test-pad-pod"
+
+type Padder struct {
+	// Client defines the API client to run CRUD operations, that will be used for testing
+	Client    client.Client
+	Namespace string
```
fromanirh, line 54: do these need to be public?

test/utils/padder/padder.go
```diff
@@ -0,0 +1,202 @@
...
+// New return new padder object
+// ns is where padding pods will get deployed
+func New(cli client.Client, ns string) (*Padder, error) {
+	padder := &Padder{
+		Namespace: ns,
+	}
+
+	padder.Client = cli
+	padder.padRequest = &padRequest{untilAvailableReach: make(corev1.ResourceList)}
+	return padder, nil
```
fromanirh, line 72: why not just
```
return &Padder{
  Namespace: ns,
  Client: cli,
  padRequest: ...
}, nil
```
?

test/utils/padder/padder.go
```diff
@@ -0,0 +1,202 @@
...
+type Padder struct {
+	// Client defines the API client to run CRUD operations, that will be used for testing
+	Client    client.Client
+	Namespace string
+	*padRequest
+}
+
+type padRequest struct {
+	nNodes              int
+	untilAvailableReach corev1.ResourceList
```
fromanirh, line 60: I'd use something like `allocationTarget`

test/utils/padder/padder.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	nrtList, err := nrtutil.GetUpdated(p.Client, nrtv1alpha1.NodeResourceTopologyList{}, time.Second*10)
+	if err != nil {
+		return err
+	}
+
+	// create a list of only nodes that have single-numa-node policy
+	singleNumaNrt := nrtutil.FilterTopologyManagerPolicy(nrtList.Items, nrtv1alpha1.SingleNUMANodeContainerLevel)
+	singleNumaNrt = append(singleNumaNrt, nrtutil.FilterTopologyManagerPolicy(nrtList.Items, nrtv1alpha1.SingleNUMANodePodLevel)...)
```
fromanirh, line 109: is this correct? it seems convoluted

test/utils/padder/padder.go
```diff
@@ -0,0 +1,202 @@
...
+				padPod.Spec.Containers[0].Resources.Requests = diffList
+
+				// place the pod on the selected node
+				//padPod.Spec.NodeSelector = map[string]string{nodes.LabelHostname: nrt.Name}
+				padPod.Spec.NodeSelector = map[string]string{"kubernetes.io/hostname": nrt.Name}
+				if err := p.Client.Create(context.TODO(), padPod); err != nil {
+					return err
+				}
+				klog.InfoS("created pod", "pod", fmt.Sprintf("%s/%s", padPod.Namespace, padPod.Name), "node", nrt.Name)
+				nodePadded = true
```
fromanirh, line 140: unfortunately not so easy. We need to wait for the pod to go running, to make sure the Kubelet (and the system in general) accpeted it and allocated the resources

test/utils/padder/padder.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+		if nodePadded {
+			p.nNodes--
+		}
+	}
+	return nil
+}
+
+// Clean can be called after test finished
+// in order to clean all the padding pod in an easier way
```
fromanirh, line 153: if we create the padding pod in yet another namespace managed by the `Padder`, we can just delete the parent namespace (wait for the pod to be gone so the resources are released) and call it a day

fromanirh (commented):
the idea is nice, and the API is nicer than I thought initially, but unfortunately there are many pitfalls in the implementation, let's try to avoid as much as possible of them

Tal-or:
> > @fromanirh in general, do you think that a utility package like this one required a unit-tests of its own. It's kinda testing the tests
> 
> no, unless the logic is supercomplex. In that case it should probably moved on its own package (with unit tests :) ). Testing the test should be exceptional for the cases we care about.

you mean we should have it under ```pkg``` and not ```test``` ?

fromanirh:
> > > @fromanirh in general, do you think that a utility package like this one required a unit-tests of its own. It's kinda testing the tests
> > 
> > 
> > no, unless the logic is supercomplex. In that case it should probably moved on its own package (with unit tests :) ). Testing the test should be exceptional for the cases we care about.
> 
> you mean we should have it under `pkg` and not `test` ?

should we get to this stage, yes.

test/utils/padder/padder.go
Tal-or: The namespace is not managed by the padder, it was given to it by the user.

Think of a case when there are other components inside the namespace.
We don't want to delete them behind the back of the user.
We only want to delete the pods we care about- i.e. the padding pod.

Tal-or:
> > 
> > you mean we should have it under `pkg` and not `test` ?
> 
> should we get to this stage, yes.

But then we would import other utils from ```test``` into ```pkg``` (because padder is using other packages under ```test/utils```) is that ok?

test/utils/padder/padder.go
Tal-or: We want all nodes with a ```single-numa-node``` policy, we don't care about the scope policy.

test/utils/padder/padder.go
fromanirh: correct, but what's the use case for letting the client manage the namespace?
padding pods are logically part of the test fixture. What's the benefit in NOT having the padder package manage the namespace?
If needed (can't see the usecase yet, though) the padder package can return the relevant namespace and the label(s) needed to get the padding pods in the padding namespace

test/utils/padder/padder.go
fromanirh: ok, let's try to simplify and/or make the code more explicit though.

fromanirh:
> > > you mean we should have it under `pkg` and not `test` ?
> > 
> > 
> > should we get to this stage, yes.
> 
> But then we would import other utils from `test` into `pkg` (because padder is using other packages under `test/utils`) is that ok?

in general it is. In practice, we will have to evaluate on a case-by-case basis.

test/utils/padder/padder.go
fromanirh: I still tend to see the padding pods as a fixture implementation detail, hence I see little merit but some drawbacks in expecting a namespace instead of self-managing, but is not wrong either. Let's start this way and let's see how it works out.

test/utils/padder/padder.go
Tal-or: > We need to wait for the pod to go running
Easy task.
> make sure the Kubelet (and the system in general) accepted it and allocated the resources
Hard. We can't tell in which zone the pod land, hence we can't track the changes under the NRT object.

fromanirh: > > We need to wait for the pod to go running
> 
> Easy task.

Let's do in parallel to avoid consuming too much time. And we should likely fail (as in `Expect` failed) if 1+ pods fail to go running. We can reuse some code/ideas from https://github.com/openshift-kni/numaresources-operator/pull/126/commits/41bb6c430cf8be7fa2d65bd8d523419af0d46af2


> > make sure the Kubelet (and the system in general) accepted it and allocated the resources
> 
> Hard. We can't tell in which zone the pod land, hence we can't track the changes under the NRT object.

Aye, hence https://github.com/openshift-kni/numaresources-operator/pull/126

test/utils/padder/padder.go
Tal-or: OK let me revaluate

Tal-or: I hope now it's more explicit

test/utils/padder/padder.go
fromanirh: relevant snippet I mentioned in the above comment:
```
klog.Infof("e2e infra setup begin")

	mcps, err := machineconfigpools.GetNodeGroupsMCPs(context.TODO(), fxt.Client, nodeGroups)

	klog.Infof("setting e2e infra for %d MCPs", len(mcps))

	var dss []*appsv1.DaemonSet
	for _, mcp := range mcps {
		if mcp.Spec.NodeSelector == nil {
			klog.Warningf("the machine config pool %q does not have node selector", mcp.Name)
			continue
		}

		dsName := objectnames.GetComponentName(numacellmanifests.Prefix, mcp.Name)
		klog.Infof("setting e2e infra for %q: daemonset %q", mcp.Name, dsName)

		ds := numacellmanifests.DaemonSet(mcp.Spec.NodeSelector.MatchLabels, fxt.Namespace.Name, dsName, numaCellDPImage)
		err = fxt.Client.Create(context.TODO(), ds)
		Expect(err).ToNot(HaveOccurred(), "cannot create the numacell daemonset %q in the namespace %q", ds.Name, ds.Namespace)

		dss = append(dss, ds)
	}

	klog.Infof("daemonsets created")

	var wg sync.WaitGroup
	for _, ds := range dss {
		wg.Add(1)
		go func(ds *appsv1.DaemonSet) {
			defer GinkgoRecover()
			defer wg.Done()

			klog.Infof("waiting for daemonset %q to be ready", ds.Name)

			// TODO: what if timeout < period?
			err := e2ewait.ForDaemonSetReady(fxt.Client, ds, 10*time.Second, timeout)
			Expect(err).ToNot(HaveOccurred(), "DaemonSet %q failed to go running")
		}(ds)
	}
	wg.Wait()

	klog.Infof("e2e infra setup completed")
```

Tal-or: We can't import the ```gomega``` package for the assertions, for two reasons:
1. we just decided to move padder under ```pkg``` so it's unusual to use to ```gomega``` outside ```test```
2. it's a bad habit to have assertions outside ```ginkgo```'s specs

fromanirh: Sure! I'm not against moving into `pkg`, but I wasn't suggesting (let alone requesting) to move it just yet. I'm actually fine both ways.
Also, the snippet above should not be taken verbatim, but rather as illustration as possible flow to make the code more robust.

Tal-or: ack. let me revalute

Tal-or: @fromanirh Please check the new commit. I hope it address the problem you were concerning about

test/utils/padder/padder.go
```diff
@@ -0,0 +1,212 @@
...
+//	}
+// pad.Nodes(2).UntilAvailableIsResourceList(rl).Pad()
+
+const PadderLabel = "nrop-test-pad-pod"
+
+type Padder struct {
+	// Client defines the API client to run CRUD operations, that will be used for testing
+	Client    client.Client
+	namespace string
+	*padRequest
```
fromanirh, line 58: so IIUC the intent is to track different pad requests for different node groups.
I think we can start addressing (and documenting) the problem of padding just ONE node group - not many, and extend later if needed

test/utils/padder/padder.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+type padRequest struct {
+	nNodes           int
+	allocationTarget corev1.ResourceList
```
fromanirh, line 63: OTOH we should allow since the beginning different allocation targets for each nodes, possibly with numa-zone granularity.

test/utils/padder/padder.go
```diff
@@ -0,0 +1,212 @@
...
+
+// UntilAvailableIsResource will pad pods into nodes until reach the desired available allocationTarget
+func (p *Padder) UntilAvailableIsResource(resName corev1.ResourceName, quantity string) *Padder {
+	quan := resource.MustParse(quantity)
+	p.allocationTarget[resName] = quan
+	return p
+}
+
+// UntilAvailableIsResourceList is like UntilAvailableIsResource but with a complete ResourceList as a parameter
+func (p *Padder) UntilAvailableIsResourceList(resources corev1.ResourceList) *Padder {
```
fromanirh, line 91: do we have a clear usecase for the two `UntilAvailableIs*` APIs or are we adding them "just in case"?

test/utils/padder/padder.go
```diff
@@ -0,0 +1,212 @@
...
+	return &Padder{
+		Client:     cli,
+		namespace:  ns,
+		padRequest: &padRequest{allocationTarget: make(corev1.ResourceList)},
+	}, nil
+}
+
+// Nodes number of nodes to pad
+func (p *Padder) Nodes(n int) *Padder {
+	p.padRequest.nNodes = n
```
fromanirh, line 79: this is suspicious. If we need to restrict the nodes to pad from the set we autodetect, we would need to restrict using a stronger criteria (node name? labels?). If this is just to signal how many nodes we are expected to pad, we can easily autodetect.

test/utils/padder/padder.go
```diff
@@ -0,0 +1,212 @@
...
+				padPod.Spec.Containers[0].Resources.Limits = diffList
+				padPod.Spec.Containers[0].Resources.Requests = diffList
+
+				// place the pod on the selected node
+				//padPod.Spec.NodeSelector = map[string]string{nodes.LabelHostname: nrt.Name}
+				padPod.Spec.NodeSelector = map[string]string{"kubernetes.io/hostname": nrt.Name}
+				if err := p.Client.Create(context.TODO(), padPod); err != nil {
+					return err
+				}
+				if _, err = wait.ForPodPhase(p.Client, p.namespace, padPod.Name, corev1.PodRunning, time.Minute); err != nil {
```
fromanirh, line 140: uhm, do we want to wait sequentially for each pod to go running (or fail)?

test/utils/padder/padder.go
```diff
@@ -0,0 +1,212 @@
...
+		if nodePadded {
+			p.nNodes--
+		}
+	}
+	return nil
+}
+
+// Clean can be called after test finished
+// in order to clean all the padding pod in an easier way
+func (p *Padder) Clean() error {
```
fromanirh, line 169: this looks good

test/utils/padder/padder.go
Tal-or: I thought about that but then you're making the API much more complex for the user.
For example, how the user should tell which nodes he wants? by name/label?
Why does the user even care about specific nodes? if the user wants different allocations, we can have multiple padding requests instead.

test/utils/padder/padder.go
Tal-or: We need at least one of them :) 
We have both of them for convenience

test/utils/padder/padder.go
fromanirh: that's a good  point. But some test cases will need uneven allocation across NUMA zones. Let me re-evaluate my comment.

test/utils/padder/padder.go
fromanirh: ack

test/utils/padder/padder.go
Tal-or: we can make that in parallel indeed. let me rewrite that part

test/utils/padder/padder.go
Tal-or: thanks

test/utils/padder/padder.go
Tal-or: we just need to know how many nodes are going to be padded we don't actually care from their names/label.
let me explain using an exmaple:
I have 4 nodes (Worker nodes with single-numa-node) I'm padding 3 out 4 nodes until available CPUs reaches 3.
Then I create a test pod which asks for 4 CPUs. 
And I know I only have one worker nodes which can handle that requests (because the 3 others have only 3 available CPUs)

test/utils/padder/padder.go
Tal-or: @fromanirh on second thought i'm not sure that having this function inside a go routine is a good idea.
How do we report an error? using a channel is overkill in my opinion. 
having panic or assertion outside of ginkgo spec is also a bad idea. WDYT?

fromanirh (commented):
/lgtm
/approve

PR #117, opened 2022-01-25
makefile: change default VERSION parameter value

Tal-or:
We want, by default, to always use the most updated version existed.
This is new value is a valid semantic version tag that pushed to quay for every new image that gets created.

Also see: https://github.com/openshift/release/pull/25141


fromanirh (commented):
/lgtm
/approve

Tal-or:
/retest
looks like infra issue

PR #118, opened 2022-01-25
improve verify-generated.sh error message

Tal-or:
When the verify-generated return with error, it tells the user to run \'make generate\' and commit results.
There are times when it's not enough and the user need to run other make targets.

This new message aims to resolve that.


fromanirh:
/lgtm
/approve
thanks, this is something I wanted to fix myself.

PR #119, opened 2022-01-25
Revert "gh actions: opt in the kind e2e flow"

Tal-or:
This reverts commit 629a7a0f4bbe2870bcd50f79eb05f30a0b0653bf.

this reverted commit not needed anymore due to following reasons:
1. It doesn't seem to work.
2. The original problem it intended to solve, already addressed at: https://github.com/openshift-kni/numaresources-operator/pull/98

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -55,7 +55,6 @@ jobs:
         IMG=${built_image} KUSTOMIZE_DEPLOY_DIR="config/kind-ci/" make deploy
 
     - name: E2E Tests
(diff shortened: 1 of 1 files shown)


fromanirh:
/lgtm
/approve
yes, it doesn't make sense to spend time debugging the making the magic annotation since you provided a better fix

Tal-or:
> it seems we will need to invest more time either way, as the lane on kind is now constantly failing

Yup. Working to figure that out.

Tal-or:
Depend on https://github.com/openshift-kni/numaresources-operator/pull/120 fix

PR #120, opened 2022-01-25
makefile: fix make deploy target

Tal-or:
We should actually apply the manifests that got built by Kustomize and not just output them to the console

Regression caused by: https://github.com/openshift-kni/numaresources-operator/pull/98

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,7 +196,7 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 	kubectl apply -f $(CRD_MACHINE_CONFIG_POOL_URL)
 	kubectl apply -f $(CRD_KUBELET_CONFIG_URL)
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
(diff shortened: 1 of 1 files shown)


fromanirh (commented):
the intent of the fix is right, but it seems weird that this target, which was automatically generated by `operator-sdk`, was wrong. Could you please check if a commit sneaked in and unintentionally broke this?

fromanirh:
/hold
I know what's going on and I'm monitoring this change. Should be OK with a retry, but let's serialize the changes.

Tal-or:
> /hold I know what's going on and I'm monitoring this change. Should be OK with a retry, but let's serialize the changes.

You're talking about the CI failures?

fromanirh:
> > /hold I know what's going on and I'm monitoring this change. Should be OK with a retry, but let's serialize the changes.
> 
> You're talking about the CI failures?

Yes. There is too much contention about the prow resources. It seems to be improving. If this is not the case we will need to help the tool by serializing changes (/hold-ing them is a clumsy way to do that, but it seems it's the only tool we have)

PR #122, opened 2022-01-26
manifests: set executable under the command stanza

Tal-or:
According to the line we took at: https://github.com/k8stopologyawareschedwg/deployer/pull/79
We should also have the same here.

### DIFF
--- a/pkg/numaresourcesscheduler/manifests/yaml/deployment.yaml
+++ b/pkg/numaresourcesscheduler/manifests/yaml/deployment.yaml
@@ -30,8 +30,9 @@ spec:
             requests:
               cpu: "100m"
               memory: "500Mi"
-          args:
+          command:
             - /bin/kube-scheduler
+          args:
             - --config=/etc/kubernetes/config.yaml
           volumeMounts:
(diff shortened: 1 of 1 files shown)


fromanirh:
/lgtm
/approve
let's make sure we have test coverage because while I think this fix is good and implements the best practices, I'm not 100% sure it plays well with the way the container is built (Dockerfile entrypoint)

Tal-or:
> /lgtm /approve let's make sure we have test coverage because while I think this fix is good and implements the best practices, I'm not 100% sure it plays well with the way the container is built (Dockerfile entrypoint)

Now that we have https://github.com/openshift-kni/numaresources-operator/pull/123 it will surface issues like that.
Unless you really want something more specific

fromanirh:
> > /lgtm /approve let's make sure we have test coverage because while I think this fix is good and implements the best practices, I'm not 100% sure it plays well with the way the container is built (Dockerfile entrypoint)
> 
> Now that we have #123 it will surface issues like that. Unless you really want something more specific

#123 it's a fine start. We will add more specific tests later/if needed.

PR #125, opened 2022-01-28
Check RTE daemonset is restarted

jlojosnegros:
We wanna check RTE daemonset is restarted and image updated whenever NRO is upgraded.
All these things were almost already done in this test case. I just added a new check to ensure the daemonset is a new one.

### DIFF
--- a/test/e2e/install/install_test.go
+++ b/test/e2e/install/install_test.go
@@ -24,7 +24,8 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
-	v1 "k8s.io/api/apps/v1"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -41,6 +42,11 @@ import (
 	e2eimages "github.com/openshift-kni/numaresources-operator/test/utils/images"
 	"github.com/openshift-kni/numaresources-operator/test/utils/machineconfigpools"
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
+	e2ewait "github.com/openshift-kni/numaresources-operator/test/utils/objects/wait"
+)
+
+const (
+	containerNameRTE = "resource-topology-exporter"
 )
 
 var _ = Describe("[Install] continuousIntegration", func() {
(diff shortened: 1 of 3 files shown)


test/e2e/install/install_test.go
```diff
@@ -252,8 +254,10 @@ var _ = Describe("[Install] durability", func() {
 					klog.Warningf("failed to get the RTE DaemonSet: %v", err)
 					return false
 				}
+				recreated := dsInitialTimeStamp.Before(&ds.CreationTimestamp)
```
fromanirh, line 257: better to check the Generation: https://github.com/openshift-kni/numaresources-operator/blob/e2e-infra-numacell/test/utils/objects/wait/daemonset.go#L60 because, well, this is the recommended k8s way to ensure an object was updated by somewhere in the cluster, and because this way should be (more) resistant against clock skew (if any) between nodes

jlojosnegros: ok, makes sense.

test/e2e/install/install_test.go
```diff
@@ -247,13 +248,16 @@ var _ = Describe("[Install] durability", func() {
 				err := e2eclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(nroObj), updatedNroObj)
 				Expect(err).ToNot(HaveOccurred())
 
-				ds, err := getDaemonSetByOwnerReference(updatedNroObj.GetUID())
+				updatedDs, err := getDaemonSetByOwnerReference(updatedNroObj.GetUID())
```
fromanirh, line 251: unnecessary (but no big deal)

test/utils/objects/wait/daemonset.go
```diff
@@ -0,0 +1,32 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+)
+
+func AreDaemonsetInstancesAvailable(newStatus *appsv1.DaemonSetStatus, instances int32) bool {
+	return newStatus.UpdatedNumberScheduled == instances &&
+		newStatus.NumberReady == instances &&
+		newStatus.NumberAvailable == instances
+}
+
+func IsDaemonsetComplete(ds *appsv1.DaemonSet, newStatus *appsv1.DaemonSetStatus) bool {
```
fromanirh, line 29: I'd call `IsDaemonSetReady` which seems more fitting than "Complete"

test/e2e/install/install_test.go
```diff
@@ -191,7 +192,7 @@ var _ = Describe("[Install] durability", func() {
 			}
 		})
 
-		It("should be able to delete NUMAResourceOperator CR and redeploy without polluting cluster state", func() {
+		It("[test_id: 47587] should be able to delete NUMAResourceOperator CR and redeploy without polluting cluster state", func() {
```
fromanirh, line 195: we can write an approxiimation of the upgrade test if we update the NROP deployment -not delete and recreate it.
In addition, the v0.1.0 is way too old. But this is another issue which should be tracked separately.

fromanirh:
to properly implement this testcase we can indeed build on top of the existing tests, but

1. we need to update the NROP deployment (not to delete/recreate) AND, most importantly
2. we need an auxiliary NROP image, always up to date. What we want to check is the image sha changes, hence we can do the simplest of things: create an additional NROP image with a extra layer on top, and use that to simulate a hotfix update.

jlojosnegros:
> to properly implement this testcase we can indeed build on top of the existing tests, but
> 
> 1. we need to update the NROP deployment (not to delete/recreate) AND, most importantly
> 2. we need an auxiliary NROP image, always up to date. What we want to check is the image sha changes, hence we can do the simplest of things: create an additional NROP image with a extra layer on top, and use that to simulate a hotfix update.

Does this means it is preferable to write another test where we update NRO instead of use this approach?

fromanirh:
> > to properly implement this testcase we can indeed build on top of the existing tests, but
> > 
> > 1. we need to update the NROP deployment (not to delete/recreate) AND, most importantly
> > 2. we need an auxiliary NROP image, always up to date. What we want to check is the image sha changes, hence we can do the simplest of things: create an additional NROP image with a extra layer on top, and use that to simulate a hotfix update.
> 
> Does this means it is preferable to write another test where we update NRO instead of use this approach?

yes, it is preferible add a new test (`It` block) which does update.

jlojosnegros:
> 2\. we need an auxiliary NROP image, always up to date. What we want to check is the image sha changes, hence we can do the simplest of things: create an additional NROP image with a extra layer on top, and use that to simulate a hotfix update.

After #130, we are using updated Ci-images, so I suppose that covers this second point, doesn't it?

fromanirh:
> > 2. we need an auxiliary NROP image, always up to date. What we want to check is the image sha changes, hence we can do the simplest of things: create an additional NROP image with a extra layer on top, and use that to simulate a hotfix update.
> 
> After #130, we are using updated Ci-images, so I suppose that covers this second point, doesn't it?

yes, correct.

Tal-or:
@jlojosnegros about the broken ```ci-install-e2e``` did you try rebase?

jlojosnegros:
> @jlojosnegros about the broken `ci-install-e2e` did you try rebase?

Yep. I have it rebased locally but still working on some issues with the test. Hope to push it asap

test/e2e/install/install_test.go
```diff
@@ -188,22 +194,78 @@ var _ = Describe("[Install] durability", func() {
 			}
 		})
 
-		It("should be able to delete NUMAResourceOperator CR and redeploy without polluting cluster state", func() {
-			var nname client.ObjectKey
-			for _, obj := range deployedObj {
-				if nroObj, ok := obj.(*nropv1alpha1.NUMAResourcesOperator); ok {
-					nname = client.ObjectKeyFromObject(nroObj)
+		It("should restart RTE DaemonSet when image is updated in NUMAResourcesOperator", func() {
```
fromanirh, line 197: missing test id?

test/e2e/install/install_test.go
```diff
@@ -188,22 +194,78 @@ var _ = Describe("[Install] durability", func() {
...
+			dsCreatedPollPeriod := 10 * time.Second
+			Eventually(func() error {
+				var err error
+				ds, err = getDaemonSetByOwnerReference(uid)
+				return err
+			}, dsCreatedTimeOut, dsCreatedPollPeriod).Should(BeNil())
+
+			dsReadyTimeOut := 5 * time.Minute
+			dsReadyPollPeriod := 10 * time.Second
+			By("... and ready")
```
fromanirh, line 221: nice, but you can probably coalesce the two waiting loops. We don't care too much about created but undready DS

test/e2e/install/install_test.go
```diff
@@ -261,6 +323,27 @@ var _ = Describe("[Install] durability", func() {
 	})
 })
 
+func findContainerByName(daemonset appsv1.DaemonSet, containerName string) (*corev1.Container, error) {
+
+	//shortcut
+	containers := daemonset.Spec.Template.Spec.Containers
+
+	if len(containers) == 0 {
```
fromanirh, line 331: the `range` loop handle the case of empty slice nicely. But no big deal

fromanirh (commented):
please add the test_id and this should be ready. Other comments are minor/possible improvements

fromanirh:
/retest
totally infra issue

test/e2e/install/install_test.go
fromanirh: true! let's keep it as it is

fromanirh (commented):
/approve
/lgtm

PR #126, opened 2022-01-28
e2e infra: support to pin padding pods to specific numa zones

fromanirh:
Add device plugin to expose the numa cells of a machine as resources, through the device plugin API of the kubelet.
Includes the related manifests + go packages to interact with them.

Enable the e2e tests to pin specific numa cells on specific machines, greatly sinplifying the test fixture setup.
This approach breaks badly the design of kubernetes and its spirit, so it should NEVER EVER be used outside e2e tests and on any cluster outside CI and (early) development stage.


fromanirh:
depends on https://github.com/openshift-kni/numaresources-operator/pull/127

fromanirh:
/test ci-install-e2e
definitely infra issue

test/e2e/serial/test_suite_serial_test.go
```diff
@@ -34,8 +59,105 @@ import (
...
+	var err error
+
+	__fxt, err = e2efixture.Setup("e2e-test-infra")
+	Expect(err).ToNot(HaveOccurred(), "unable to setup infra test fixture")
+
+	nroSchedObj = &nropv1alpha1.NUMAResourcesScheduler{}
+	err = __fxt.Client.Get(context.TODO(), client.ObjectKey{Name: nrosched.NROSchedObjectName}, nroSchedObj)
+	Expect(err).ToNot(HaveOccurred(), "cannot get %q in the cluster", nrosched.NROSchedObjectName)
+
+	nroOperObj = &nropv1alpha1.NUMAResourcesOperator{}
```
Tal-or, line 78: At what stage during the serial test we are supposed to deploy the ```NUMAResourcesScheduler``` and ```NUMAResourcesOperator```? is the ```install``` suit going to be a prerequisite for the serial suit?

test/e2e/serial/test_suite_serial_test.go
```diff
@@ -34,8 +59,105 @@ import (
...
+
+	setupInfra(__fxt, nroOperObj.Spec.NodeGroups, 3*time.Minute)
+})
+
+var _ = AfterSuite(func() {
+	if _, ok := os.LookupEnv("E2E_INFRA_NO_TEARDOWN"); ok {
+		return
+	}
+
+	// numacell daamonset automatically cleaned up when we remove the namespace
```
Tal-or, line 96: nit: daemonset

Tal-or (commented):
Overall it seems good to me, but due to the size of this PR I would prefer additional reviewer input

test/e2e/serial/test_suite_serial_test.go
fromanirh: This is something I need to document better: the `serial` suite is expected to run against a fully configured TAS-enabled cluster (no matter how) and to validate it. We don't need an explicit dep on the `install` suite, as `install` suite is just one of the many vehicles to deploy TAS on a cluster.

test/e2e/serial/test_suite_serial_test.go
fromanirh: thanks, fixed

test/e2e/serial/test_suite_serial_test.go
Tal-or: OK but it's something we should bear in mind, especially because I guess we'll have a dedicated lane for this type of test.

fromanirh: absolutely, added a README to document the use case and the goals of the lane

test/e2e/serial/README.md
```diff
@@ -0,0 +1,20 @@
+# e2e serial testsuite
+
+The "serial" e2e test suite holds e2e tests covering topology-aware scheduling as a whole.
+
+For this purposes "end to end" means the cluster, taken as black box, with minimal breakages to this rule
+only iff and when these breakages are extremely helpful (point in case: the numacell plugin).
```
jlojosnegros, line 6: typo?

PR #128, opened 2022-01-31
rte: exit on config changes if the file exists

fromanirh:
the `--exit-on-config-changes` should make the RTE process fail
only if the config file which should be monitored actually exists.
Otherwise the behaviour is ill-defined to begin with.

### DIFF
--- a/rte/main.go
+++ b/rte/main.go
@@ -39,6 +39,7 @@ import (
 type localArgs struct {
 	SysConf             sysinfo.Config
 	ConfigPath          string
+	HasConfigFile       bool
 	ExitOnConfigChanges bool
 }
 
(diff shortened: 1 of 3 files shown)


fromanirh:
/hold
let's clarify the behaviour first

fromanirh:
> wrong approach. We should react also to config file creation and deletion.

Thinking more about it: yes, the current implementation is incomplete, but even watching the parent directory won't solve the issue at hand because the config file is expected to be provided using a configMap, so there is the volume mount interaction and the controller interaction.

Having more elaborate fs watching won't solve the issue at hand, we would need some help from the controller and/or the daemonset configuration

fromanirh:
/hold cancel
rationale in https://github.com/openshift-kni/numaresources-operator/pull/128#issuecomment-1025921312

rte/main.go
```diff
@@ -97,15 +98,19 @@ func main() {
...
 
 	if parsedArgs.LocalArgs.ExitOnConfigChanges {
-		cw, err := config.NewWatcher(parsedArgs.LocalArgs.ConfigPath, func() error {
-			klog.Infof("configuration file %q changed: exit", parsedArgs.LocalArgs.ConfigPath)
-			os.Exit(0)
-			return nil
-		})
-		if err != nil {
-			klog.Fatalf("cannot watch the configuration file %q: %v", parsedArgs.LocalArgs.ConfigPath, err)
+		if !parsedArgs.LocalArgs.HasConfigFile {
```
Tal-or, line 101: What if the ConfigMap is created later on? is it for the operator to handle?

fromanirh: this is exactly the point. The operator should react somehow to this event. AFAICT, nothing we can fix at RTE level. Perhaps we would need to change the way configMaps are handled (e.g., throwing ideas: makes them immutable so a change of ConfigMap triggers a deamonset change).

I'm not actually sure this PR is an actual improvement over the current state, because now the only way to learn the `exit-on-config-change` feature is auto-disabled is to look into the logs.

fromanirh:
this change would make the `exit-on-config-changes` break silently if no config file is present. ATM we prefer to let the RTE pod be in `CrashLoop`, because at least it's evident something is wrong in the system.

PR #131, opened 2022-02-02
Ensure install test are executed

jlojosnegros:
Some test were not executed and fail due to some timing issues.
Create a new github action to execute them at least as frecuently as e2e
tests

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -87,3 +87,76 @@ jobs:
       with:
         name: kind_logs
         path: /tmp/kind_logs
+  e2e-install:
+    runs-on: ubuntu-20.04
+    env:
+      built_image: "numaresources-operator:ci" # Arbitrary name
+      E2E_NAMESPACE_NAME: numaresources
+      E2E_TOPOLOGY_MANAGER_POLICY: SingleNUMANodeContainerLevel
+      LOG_DIR: /tmp/test_e2e_intall_logs
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0 # Fetch all history for all tags and branches
+
+    - name: Set up golang
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.17
+
+    - name: Verify modules
+      run: go mod verify
+
+    - name: build test binary
+      run: |
+        make binary-e2e-rte binary-e2e-install binary-e2e-uninstall
+
+    - name: Build image
+      run: |
+        docker build . -t ${built_image}
+
+    - name: Create K8s Kind Cluster
+      run: |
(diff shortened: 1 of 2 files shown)


test/e2e/install/install_test.go
```diff
@@ -201,9 +201,14 @@ var _ = Describe("[Install] durability", func() {
...
+			By("waiting for the DaemonSet to be created..")
 			uid := nroObj.GetUID()
-			ds, err := getDaemonSetByOwnerReference(uid)
-			Expect(err).ToNot(HaveOccurred())
+			ds := &v1.DaemonSet{}
+			Eventually(func() error {
+				var err error
+				ds, err = getDaemonSetByOwnerReference(uid)
+				return err
+			}, 5*time.Minute, 10*time.Second).Should(BeNil())
```
fromanirh, line 211: this is a good change we should merge anyway. Better to split on a separate PR we can merge quickly

jlojosnegros: The problem I see with splitting this in another PR is that if this test is not being executed we cannot be sure it is really ok

.github/workflows/e2e.yml
```diff
@@ -87,3 +87,76 @@ jobs:
...
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.17
+
+    - name: Verify modules
+      run: go mod verify
+
+    - name: build test binary
+      run: |
+        make binary-e2e-rte binary-e2e-install binary-e2e-uninstall
```
fromanirh, line 113: The problem is our github actions infra is already overtaxed, and I'm not convinced here that the benefits are greater than the costs. Let's keep the k8s (through kind) coverage minimal (but sufficient) until there is a crystal clear case to enhance it. Our main focus is OCP anyway.

test/e2e/install/install_test.go
Tal-or: The test is executed under the ```ci-e2e-install``` lane.
Why do you think it's not?

test/e2e/install/install_test.go
```diff
@@ -201,9 +201,14 @@ var _ = Describe("[Install] durability", func() {
...
 			Expect(err).ToNot(HaveOccurred())
 
+			By("waiting for the DaemonSet to be created..")
 			uid := nroObj.GetUID()
-			ds, err := getDaemonSetByOwnerReference(uid)
-			Expect(err).ToNot(HaveOccurred())
+			ds := &v1.DaemonSet{}
+			Eventually(func() error {
+				var err error
+				ds, err = getDaemonSetByOwnerReference(uid)
```
Tal-or, line 209: I would add a ```klog.Warningf``` here which says something like: "ds is not ready/found yet"

jlojosnegros:
No need to change GH actions.
Code change already covered by #132

PR #132, opened 2022-02-02
Wait for DaemonSet to be created by NRO

jlojosnegros:
Sometimes NUMAResourcesOperator takes some time creating RTE DaemonSet
so it is better to wait for it in testing to avoid failures.

Split of #131


fromanirh (commented):
/approve
/lgtm

PR #133, opened 2022-02-02
e2e: fix potential race between controller and assertion check

Tal-or:
In the current behavior, we're changing the ```LogLevel``` field under the CR and then checking the change results under the container.
There might be times when the reconciliation loop hasn't occurred yet (thus the pod didn't get updated) and we're already checking for the pod changes under the test.

With the new behavior using Eventually we'll have until timeout to check if the test condition were met.

### DIFF
--- a/test/e2e/rte/rte_test.go
+++ b/test/e2e/rte/rte_test.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"github.com/ghodss/yaml"
 	"github.com/onsi/ginkgo"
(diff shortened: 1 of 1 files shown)


fromanirh (commented):
using Eventually is more correct anyway, so I'm all for this change. Please make CI happy and ping me for the final review (spoiler: will likely LGTM immediately unless there are some major issues)

test/e2e/rte/rte_test.go
```diff
@@ -73,21 +74,37 @@ var _ = ginkgo.Describe("with a running cluster with all the components", func()
...
+
+					// TODO better match by name than assumes container #0 is the right one
+					rteCnt := &ds.Spec.Template.Spec.Containers[0]
+					found, match := matchLogLevelToKlog(rteCnt, nropObj.Spec.LogLevel)
+					if !found {
+						klog.Warningf("DaemonSet: %s --v flag doesn't exist in container %s args under ds", ds.Name, rteCnt.Name)
+					}
+					isMatchedInAllDss = isMatchedInAllDss && (found && match)
+				}
+				return isMatchedInAllDss
```
fromanirh, line 105: would this work?
```
for _, ds := range rteDss {
	// TODO better match by name than assumes container #0 is the right one
	rteCnt := &ds.Spec.Template.Spec.Containers[0]
	found, match := matchLogLevelToKlog(rteCnt, nropObj.Spec.LogLevel)
	if !found {
		klog.Warningf("DaemonSet: %s --v flag doesn't exist in container %s args under ds", ds.Name, rteCnt.Name)
		return false
	}
	if !match {
        	return false
	}
}
return true
```

test/e2e/rte/rte_test.go
```diff
@@ -98,17 +115,32 @@ var _ = ginkgo.Describe("with a running cluster with all the components", func()
...
+					rteCnt := &ds.Spec.Template.Spec.Containers[0]
+					found, match := matchLogLevelToKlog(rteCnt, nropObj.Spec.LogLevel)
+					if !found {
+						klog.Warningf("DaemonSet: %s --v flag doesn't exist in container %s args under ds", ds.Name, rteCnt.Name)
+						return false
+					}
+
+					isMatchedInAllDss = isMatchedInAllDss && (found && match)
+				}
+				return isMatchedInAllDss
```
fromanirh, line 141: same as above

fromanirh (commented):
looks good, maybe we can simplify a bit more

test/e2e/rte/rte_test.go
Tal-or: nice!

Tal-or:
/retest
infra issue

fromanirh (commented):
/approve
/lgtm

Tal-or:
I flipped the positions of the interval and timeout arguments under the ```Eventually``` function. it should be ok now

fromanirh (commented):
/lgtm
/approve

PR #134, opened 2022-02-04
Makefile: fix build process of binaries

fromanirh:
Ensure we disable CGO, to avoid depending on glibc,
and ensure we consume the local vendored deps.


jlojosnegros (approved):
/lgtm

PR #136, opened 2022-02-05
e2e: install: wait for objects to be deleted

fromanirh:
the install suite was flaking with errors similar to
```
Message: "object is being deleted: kubeletconfigs.machineconfiguration.openshift.io \"kc-test\" already exists",
```

This is because the test code was not waiting for the object
to be deleted from the cluster, but it was rather going along
if the deletion *request* was succesfully accepted by the system.

It's more correct to wait for objects to be deleted before to
keep going with the tests, so we do it in this patch, adding
the necessary utilities to enable this flow.

### DIFF
--- a/test/e2e/install/install_test.go
+++ b/test/e2e/install/install_test.go
@@ -19,6 +19,7 @@ package install
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	. "github.com/onsi/ginkgo"
@@ -31,6 +32,8 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 
 	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
 	"github.com/k8stopologyawareschedwg/deployer/pkg/manifests/rte"
@@ -62,12 +65,7 @@ var _ = Describe("[Install] continuousIntegration", func() {
 	Context("with a running cluster with all the components", func() {
 		It("[test_id: 47574] should perform overall deployment and verify the condition is reported as available", func() {
 			deployedObj := overallDeployment()
-			var nname client.ObjectKey
-			for _, obj := range deployedObj {
-				if nroObj, ok := obj.(*nropv1alpha1.NUMAResourcesOperator); ok {
-					nname = client.ObjectKeyFromObject(nroObj)
-				}
-			}
+			nname := client.ObjectKeyFromObject(deployedObj.nroObj)
 			Expect(nname.Name).ToNot(BeEmpty())
 
 			By("checking that the condition Available=true")
@@ -103,7 +101,6 @@ var _ = Describe("[Install] continuousIntegration", func() {
 			const DSCheckTimeout = 1 * time.Minute
 			const DSCheckPollingPeriod = 5 * time.Second
 			Eventually(func() bool {
-
 				ds, err := getDaemonSetByOwnerReference(updatedNROObj.UID)
 				if err != nil {
 					klog.Warningf("unable to get Daemonset  %v", err)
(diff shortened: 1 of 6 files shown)


fromanirh:
/retest
infra issue:
```
msg=Cluster operator etcd Degraded is True with ClusterMemberController_Error::DefragController_Error::EtcdMembers_UnhealthyMembers::StaticPods_Error: ClusterMemberControllerDegraded: unhealthy members found during reconciling members
```

fromanirh:
/retest
same as https://github.com/openshift-kni/numaresources-operator/pull/136#issuecomment-1030614236

test/e2e/install/install_test.go
```diff
@@ -399,6 +396,58 @@ func overallDeployment() []client.Object {
 	return deployedObj
 }
 
+func deleteObject(obj client.Object) {
+	err := e2eclient.Client.Delete(context.TODO(), obj)
+	if err != nil {
+		Expect(errors.IsNotFound(err)).To(BeTrue(), fmt.Sprintf("unexpected error: %v", err))
```
Tal-or, line 405: Assertion outside of the ginkgo spec?

test/e2e/install/install_test.go
```diff
@@ -402,6 +399,57 @@ func overallDeployment() []client.Object {
 	return deployedObj
 }
 
+// TODO: what if timeout < period?
+func teardownDeployment(nrod nroDeployment, timeout time.Duration) {
+	var wg sync.WaitGroup
+	if nrod.mcpObj != nil {
+		err := e2eclient.Client.Delete(context.TODO(), nrod.mcpObj)
+		Expect(err).ToNot(HaveOccurred())
```
Tal-or, line 407: Again the assertion is outside the ginkgo spec.
I was looking at the ginkgo doc, I remembered that it's a bad practice to have an assertion outside of the spec, but this is what i found https://onsi.github.io/ginkgo/#no-assertions-in-container-nodes
so I guess inside a function it's okish?

test/e2e/install/install_test.go
fromanirh: the `deleteObject` helper didn't make too much sense in the first place. Removed.

test/e2e/install/install_test.go
fromanirh: it depends on the call tree. In our case, both this and `overallDeployment` are being called inside a `Context` (`AfterEach` and `BeforeEach` respectively). So it works, but reusing the functions freely can end up to runtime failure.
We can't enforce requirements about when functions must be called. Things we can do:
- we remove all `Expect`s from both functions
- we document the requirements

I don't see many other ways out.
I do thing failing out in the helpers has some merit (https://about.sourcegraph.com/go/advanced-testing-in-go/#Test-helpers) even though in this case the cut isn't so clear

jlojosnegros:
> it depends on the call tree. In our case, both this and `overallDeployment` are being called inside a `Context` (`AfterEach` and `BeforeEach` respectively). So it works, but reusing the functions freely can end up to runtime failure.
We can't enforce requirements about when functions must be called. Things we can do:
- we remove all `Expect`s from both functions
- we document the requirements

I don't see many other ways out.
I do thing failing out in the helpers has some merit (https://about.sourcegraph.com/go/advanced-testing-in-go/#Test-helpers) even though in this case the cut isn't so clear 

Calling a function outside a container-node does not seem a good practice to begin with, so I do not see any problem with helper functions making asserttions. That said, maybe documenting this is not a bad idea either.

jlojosnegros (approved):
/lgtm

fromanirh:
> > it depends on the call tree. In our case, both this and `overallDeployment` are being called inside a `Context` (`AfterEach` and `BeforeEach` respectively). So it works, but reusing the functions freely can end up to runtime failure.
> > We can't enforce requirements about when functions must be called. Things we can do:
> 
>     * we remove all `Expect`s from both functions
> 
>     * we document the requirements
> 
> 
> I don't see many other ways out. I do thing failing out in the helpers has some merit (https://about.sourcegraph.com/go/advanced-testing-in-go/#Test-helpers) even though in this case the cut isn't so clear
> 
> Calling a function outside a container-node does not seem a good practice to begin with, so I do not see any problem with helper functions making asserttions. That said, maybe documenting this is not a bad idea either.

Fair enough. I'll file a followup doc PR. I wonder if docstrings can be sufficient or if we need something more structured (and if so, what)

Tal-or:
> Fair enough. I'll file a followup doc PR. I wonder if docstrings can be sufficient or if we need something more structured (and if so, what)

IMHO I wouldn't bother to document that. the ```overallDeployment``` function is a private one, inside one of our test packages.
No one is going to use it but us.
inline comment like _reusing the functions freely outside of the testing scope can end up with runtime failure_ is good enough I guess

fromanirh:
> > Fair enough. I'll file a followup doc PR. I wonder if docstrings can be sufficient or if we need something more structured (and if so, what)
> 
> IMHO I wouldn't bother to document that. the `overallDeployment` function is a private one, inside one of our test packages. No one is going to use it but us. inline comment like _reusing the functions freely outside of the testing scope can end up with runtime failure_ is good enough I guess

ok! I'll add comments as part as the upcoming PRs.

test/e2e/install/install_test.go
jlojosnegros: Calling a function outside a container-node does not seem a good practice to begin with, so I do not see any problem with helper functions making assertions. That said, maybe documenting this is not a bad idea either.

PR #137, opened 2022-02-06
vendor: bump k8swg/rte to latest version

Tal-or:
Latest version contains fixes for some flaky tests we had.
Maybe it is better to create a new tag and only then update the dependencies.


fromanirh:
Thanks!
I don't think there is anough material on k8swg/rte yet to justify a new tag, so let's bump our deps

PR #138, opened 2022-02-07
e2e: fix master node affinity test

Tal-or:
When running in an OCP environment, the label "node-role.kubernetes.io/control-plane" doesn't exist for the master nodes.
We shall add this label only when running in a K8S environment

### DIFF
--- a/test/e2e/sched/utils/node.go
+++ b/test/e2e/sched/utils/node.go
@@ -22,17 +22,30 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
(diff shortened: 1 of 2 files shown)


test/e2e/sched/utils/node.go
```diff
@@ -22,17 +22,30 @@ import (
...
+func ListMasterNodes(aclient client.Client) ([]corev1.Node, error) {
 	nodeList := &corev1.NodeList{}
 	labels := metav1.LabelSelector{
 		MatchLabels: map[string]string{
-			"node-role.kubernetes.io/control-plane": "",
-			"node-role.kubernetes.io/master":        "",
+			LabelMasterRole: "",
 		},
 	}
+	if configuration.Platform == platform.Kubernetes {
```
fromanirh, line 45: this is going to change when OCP catches up with kube, but it's good enough for now

fromanirh (commented):
/lgtm
/approve

PR #141, opened 2022-02-08
e2e: wait for MCP UPDATED status

Tal-or:
After a teardown of the environment at the end of each test, we should wait for all mcps` ```UPDATED``` status to be ``True`` , in order to avoid conflicts before applying new MachineConfig data on the next tests.


Tal-or:
/retest
notification file test failed again - known issue

PR #143, opened 2022-02-10
e2e: serial: implement test id 47594

Tal-or:
The test creates a guaranteed pod with toleration.
The cluster has two feasible nodes with taint matching the toleration of the workload, but only one of them has all the requested resources on a single NUMA.
The test verifies the pod land on the correct node, otherwise, it failed

depend on #116

### DIFF
--- a/test/e2e/serial/workload_placement_test.go
+++ b/test/e2e/serial/workload_placement_test.go
@@ -27,23 +27,29 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
+	"k8s.io/kubernetes/pkg/util/taints"
 
+	nrtv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
+	numacellapi "github.com/openshift-kni/numaresources-operator/test/deviceplugin/pkg/numacell/api"
 	schedutils "github.com/openshift-kni/numaresources-operator/test/e2e/sched/utils"
 	e2efixture "github.com/openshift-kni/numaresources-operator/test/utils/fixture"
 	e2enrt "github.com/openshift-kni/numaresources-operator/test/utils/noderesourcetopologies"
+	e2enodes "github.com/openshift-kni/numaresources-operator/test/utils/nodes"
 	"github.com/openshift-kni/numaresources-operator/test/utils/nrosched"
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
 	e2ewait "github.com/openshift-kni/numaresources-operator/test/utils/objects/wait"
-
-	nrtv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
-
-	numacellapi "github.com/openshift-kni/numaresources-operator/test/deviceplugin/pkg/numacell/api"
+	e2epadder "github.com/openshift-kni/numaresources-operator/test/utils/padder"
 )
+
+const testKey = "testkey"
 
 var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 	var fxt *e2efixture.Fixture
+	var padder *e2epadder.Padder
 	var nrtList nrtv1alpha1.NodeResourceTopologyList
 	var nrts []nrtv1alpha1.NodeResourceTopology
 
@@ -51,6 +57,9 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 		var err error
 		fxt, err = e2efixture.Setup("e2e-test-workload-placement")
 		Expect(err).ToNot(HaveOccurred(), "unable to setup test fixture")
+
+		padder, err = e2epadder.New(fxt.Client, fxt.Namespace.Name)
+		Expect(err).ToNot(HaveOccurred())
 
 		err = fxt.Client.List(context.TODO(), &nrtList)
 		Expect(err).ToNot(HaveOccurred())
@@ -66,7 +75,9 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 	})
 
 	AfterEach(func() {
-		err := e2efixture.Teardown(fxt)
+		err := padder.Clean()
+		Expect(err).NotTo(HaveOccurred())
+		err = e2efixture.Teardown(fxt)
 		Expect(err).NotTo(HaveOccurred())
 	})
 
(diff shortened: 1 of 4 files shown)


Tal-or:
/hold
depend on #116

Tal-or:
CI doesn't exercise the test. it was tested locally.

test/utils/nodes/nodes.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/k8stopologyawareschedwg/deployer/pkg/clientutil"
```
fromanirh, line 27: if we want to go with `deployer/pkg/clientutil` (no strong feelings in either direction) we go full swing and just use
https://github.com/fromanirh/deployer/blob/main/pkg/clientutil/nodes/nodes.go -> `GetByRole`

fromanirh: or actually `GetWorkers` :)

test/e2e/serial/workload_placement_test.go
```diff
@@ -240,6 +250,83 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+				corev1.ResourceCPU:    resource.MustParse("4"),
+				corev1.ResourceMemory: resource.MustParse("100Mi"),
+			}
+			cnt.Resources.Requests = requiredRes
+			cnt.Resources.Limits = requiredRes
+
+			err := fxt.Client.Create(context.TODO(), testPod)
+			Expect(err).ToNot(HaveOccurred())
+
+			_, err = e2ewait.ForPodPhase(fxt.Client, testPod.Namespace, testPod.Name, corev1.PodRunning, timeout)
```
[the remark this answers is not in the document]
Tal-or: >shouldn't we check the pod lands on the node we expect?

I had that in mind, but how do I know which is the targeted node?
In order for that to be fulfilled, we need to mention explicitly (by name or label) which nodes are going to be saturated, making sure to exclude at least one (depending on the requirement of the test) and mark it as the targeted node.

It's definitely possible, but then, we're throwing the responsibility of verifying if a node is applicable to be a targeted node or a saturated node, on the user. 
IOW, the user will have to verify that node X has enough resources in order to declare it as the targeted node, and explicitly mention all the other nodes (by name or label) to get saturated.

OTOH the current approach is just mentioning how many nodes we want to saturate: 
https://github.com/openshift-kni/numaresources-operator/pull/143/commits/7c0e60a8e88cf10844e1b1c0b59ec69da6d3283d#diff-68784af7379bbc72fe22f9f4df729d2f7c6c4b564a4709c5768adf1a2ce5f378R263 

A combination of both approaches will be as follow:
1. We'll just mention how many nodes we want to saturate - As we're doing today.
2. The padder package will make sure to find the applicable nodes and saturate them according to the user request -  As  we're doing todayץ
3. The padder package will return the names of the saturate nodes - API change in the padder package.
4. The user will deduce from the list, which node isn't saturated and mark it as the targeted node - Change in the test logic.

WDYT?

>we should also probably check the pod was scheduled with the NUMA-aware scheduler

Will do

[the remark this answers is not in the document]
fromanirh: > > Can't the test do that on its own?
> 
> From my perspective test is the user. I want to save the user from dealing with all these checks.

Well I'd be happy if we can, but I believe we can be a little more demanding if our user is a machine (running a test) vs a human. So the test code can (and probably should, because there's hardly a simpler way) do a bit more homework here.

> Let me see what I can do. One thing I'm concerned about is, that atm skipping the test in case of an incompatible cluster (not having 2 multi-NUMA nodes) is a real issue which is not related to the test itself, but to our internal resources issues (which you're aware of)

Skipping the test is a perfectly valid flow if the cluster isn't up to the expectations. OTOH not running the testsuite at all -never- it's a completely different story. But we can (and we actually should) address both things in parallel to cover more ground in less time.

test/utils/nodes/nodes.go
Tal-or: I'm fully aware of this function, but it initializes the Client by itself and not using the one already provided by the ```Fixture``` so I wasn't sure it's such a good idea.

Tal-or: There is also this annoying problem:
`Cannot use 'node' (type *"github.com/k8stopologyawareschedwg/deployer/vendor/k8s.io/api/core/v1".Node) as the type *"github.com/openshift-kni/numaresources-operator/vendor/k8s.io/api/core/v1".Node`

Which happens because two packages that import a dependecny with different names

Tal-or:
/hold
Needs to be tested locally.

test/utils/nodes/nodes.go
fromanirh: fair points. We can keep as it is.

test/utils/padder/padder.go
```diff
@@ -109,12 +109,28 @@ func (p *Padder) Pad(timeout time.Duration) error {
...
+		return fmt.Errorf("not enough nodes were found for padding. requested: %d, got: %d", p.nNodes, len(singleNumaNrt))
+	}
+
 	nNodes := p.nNodes
 	var pods []*corev1.Pod
+	candidateNrts := nrtutil.AccumulateNames(singleNumaNrt)
 
-	for i := 0; i < len(singleNumaNrt) && nNodes > 0; i++ {
-		nrt := singleNumaNrt[i]
+	for ok := true; ok; ok = nNodes > 0 {
```
fromanirh, line 121: maybe just
```
for nNodes > 0 {
```
?

test/utils/padder/padder.go
```diff
@@ -204,10 +206,14 @@ func (p *Padder) Clean() error {
 	if errors.IsNotFound(err) {
 		err = nil
 	}
-	p.targetedNodes = []string{}
+	p.PaddedNodes = []string{}
 	return err
 }
 
+func (p *Padder) GetPaddedNodes() []string {
```
fromanirh, line 230: why, considering the field is already public?

test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -256,6 +256,16 @@ func ZoneResourcesMatchesRequest(resources []nrtv1alpha1.ResourceInfo, requests
 	return true
 }
 
+func FilterSingleNumaNodePolicyNrts(list []nrtv1alpha1.NodeResourceTopology) []nrtv1alpha1.NodeResourceTopology {
```
fromanirh, line 259: it's ok for now, but I'd have extended `FilterTopologyManagerPolicy` to take a set of policy into account.

test/utils/padder/padder.go
```diff
@@ -197,17 +197,34 @@ func (p *Padder) Pad(timeout time.Duration) error {
...
-	if errors.IsNotFound(err) {
-		err = nil
+	if err := p.Client.DeleteAllOf(context.TODO(), pod, delOpts...); err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	podList := &corev1.PodList{}
+	listOpts := []client.ListOption{
```
fromanirh, line 212: nit: this could be inlined in the call of `List` right below. It is not obvious which value does this temporary provide (readability-wise)

test/utils/padder/padder.go
```diff
@@ -197,17 +197,34 @@ func (p *Padder) Pad(timeout time.Duration) error {
...
+		client.InNamespace(p.namespace),
+		client.MatchingLabels{PadderLabel: ""},
+	}
+	if err := p.Client.List(context.TODO(), podList, listOpts...); err != nil {
+		return err
 	}
+
+	for _, padPod := range podList.Items {
```
fromanirh, line 220: minor, but important: wait should always be done in parallel unless there is a very strong (and documented) reason not to.

test/e2e/serial/workload_placement_test.go
```diff
@@ -507,6 +518,135 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+
+			for i := range nodes {
+				node := &nodes[i]
+				updatedNode, updated, err := taints.AddOrUpdateTaint(node, &t[0])
+				Expect(err).ToNot(HaveOccurred())
+				if updated {
+					node = updatedNode
+					klog.Infof("adding taint: %q to node: %q", t[0].String(), node.Name)
+					err = fxt.Client.Update(context.TODO(), node)
```
fromanirh, line 548: is not clear why if `AddOrUpdateTaint` succeeded... we need to update the node again?

test/e2e/serial/workload_placement_test.go
```diff
@@ -507,6 +518,135 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 			Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s", updatedPod.Namespace, updatedPod.Name, schedulerName)
 		})
 	})
+
+	When("cluster has two feasible nodes with taint but only one has the requested resources on a single NUMA zone", func() {
+		timeout := 5 * time.Minute
+		BeforeEach(func() {
```
fromanirh, line 524: it seems we're not using `e2efixture`, is it the case? if so, why?

test/e2e/serial/workload_placement_test.go
```diff
@@ -507,6 +518,135 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+
+			t, _, err := taints.ParseTaints([]string{testTaint()})
+			Expect(err).ToNot(HaveOccurred())
+
+			By("removing taints from the nodes")
+			for i := range nodes {
+				node := &nodes[i]
+				updatedNode, updated, err := taints.RemoveTaint(node, &t[0])
+				Expect(err).ToNot(HaveOccurred())
+				if updated {
```
fromanirh, line 566: same as per line 548 above

test/e2e/serial/workload_placement_test.go
```diff
@@ -507,6 +518,135 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			targetNodeName, ok := targetNodeNameSet.PopAny()
+			Expect(ok).To(BeTrue())
+
+			testPod := objects.NewTestPodPause(fxt.Namespace.Name, "testpod")
+			pSpec := &testPod.Spec
+			if pSpec.Tolerations == nil {
+				pSpec.Tolerations = testToleration()
+			} else {
+				pSpec.Tolerations = append(pSpec.Tolerations, testToleration()...)
+			}
```
fromanirh, line 611: more idiomatic:
1. create empty `tolerations` if `nil`
2. always add the tolerations we want (now it's safe to do because guaranteed non-nil)

test/e2e/serial/workload_placement_test.go
```diff
@@ -507,6 +518,135 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+
+			By(fmt.Sprintf("checking the pod landed on the target node %q vs %q", updatedPod.Spec.NodeName, targetNodeName))
+			Expect(updatedPod.Spec.NodeName).To(Equal(targetNodeName),
+				"node landed on %q instead of on %v", updatedPod.Spec.NodeName, targetNodeName)
+
+			By(fmt.Sprintf("checking the pod was scheduled with the topology aware scheduler %q", schedulerName))
+			schedOK, err := nrosched.CheckPODWasScheduledWith(fxt.K8sClient, updatedPod.Namespace, updatedPod.Name, schedulerName)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s", updatedPod.Namespace, updatedPod.Name, schedulerName)
```
fromanirh, line 635: please check the NRT resources are consumed according to the requirements

test/e2e/serial/workload_placement_test.go
```diff
@@ -507,6 +518,135 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			By("deleting the test pod")
+			if err := fxt.Client.Delete(context.TODO(), testPod); err != nil {
+				if !errors.IsNotFound(err) {
+					Expect(err).ToNot(HaveOccurred())
+				}
+			}
+
+			By("checking the test pod is removed")
+			err = e2ewait.ForPodDeleted(fxt.Client, testPod.Namespace, testPod.Name, 3*time.Minute)
+			Expect(err).ToNot(HaveOccurred())
```
fromanirh, line 646: please check the NRT resources are released

fromanirh (commented):
the direction seem good, quite some comments inside. We basically miss only a couple extra checks, then we can move forward and merge (and re-iterate once we get the testing cluster reliably)

test/utils/padder/padder.go
Tal-or: Google and his example :)

test/utils/padder/padder.go
Tal-or: The field shouldn't be public. my bad

test/e2e/serial/workload_placement_test.go
Tal-or: ```AddOrUpdateTaint``` update the given object in memory. It doesn't perform a call for the API server.

test/e2e/serial/workload_placement_test.go
Tal-or: same answer

test/e2e/serial/workload_placement_test.go
fromanirh: ok, makes sense

test/e2e/serial/workload_placement_test.go
Tal-or: Why not? we do, check l line 535 we're using the ```fxt.Client```

test/e2e/serial/workload_placement_test.go
Tal-or: nice!

test/e2e/serial/workload_placement_test.go
fromanirh: right, my bad. Reading to hastily. It seems good.

test/utils/noderesourcetopologies/noderesourcetopologies.go
Tal-or: Refactored.

fromanirh:
/test ci-index
the `lint` failure is real

test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -256,6 +256,15 @@ func ZoneResourcesMatchesRequest(resources []nrtv1alpha1.ResourceInfo, requests
 	return true
 }
 
+func FilterByPolicies(list []nrtv1alpha1.NodeResourceTopology, policies []nrtv1alpha1.TopologyManagerPolicy) []nrtv1alpha1.NodeResourceTopology {
+	var filteredNrts []nrtv1alpha1.NodeResourceTopology
+	for _, policy := range policies {
```
fromanirh, line 261: we can refactor more in the future to iterate over the NRT slice just once. But this can wait.

fromanirh (commented):
/lgtm
/approve
thanks for the updates. Good enough for now, let's move forward.

fromanirh (commented):
/lgtm
/approve
thanks for fixing the conflict

Tal-or:
Do I need to merge https://github.com/openshift-kni/numaresources-operator/pull/157 first and only then cherry-pick this one?

Tal-or:
> 

@fromanirh not good. I think https://github.com/openshift-kni/numaresources-operator/pull/149 is missing from 4.10 @jlojosnegros  do you mind having a second look?

Tal-or:
> @Tal-or: #143 failed to apply on top of branch "release-4.10":
> 
> ```
> Applying: padder: picks nodes for padding randomly
> Applying: padder: add a function to return the padded nodes
> Applying: utils: rename filterSingleNumaNodePolicyNrts to FilterByPolicies
> Applying: padder: wait for pods deletion
> Applying: e2e: serial: implement test id 47594
> Using index info to reconstruct a base tree...
> M	test/e2e/serial/workload_placement_test.go
> Falling back to patching base and 3-way merge...
> Auto-merging test/e2e/serial/workload_placement_test.go
> CONFLICT (content): Merge conflict in test/e2e/serial/workload_placement_test.go
> error: Failed to merge in the changes.
> hint: Use 'git am --show-current-patch=diff' to see the failed patch
> Patch failed at 0005 e2e: serial: implement test id 47594
> When you have resolved this problem, run "git am --continue".
> If you prefer to skip this patch, run "git am --skip" instead.
> To restore the original branch and stop patching, run "git am --abort".
> ```

@fromanirh not good. I think https://github.com/openshift-kni/numaresources-operator/pull/149 is missing from 4.10 @jlojosnegros do you mind having a second look?

fromanirh:
> > @Tal-or: #143 failed to apply on top of branch "release-4.10":
> > ```
> > Applying: padder: picks nodes for padding randomly
> > Applying: padder: add a function to return the padded nodes
> > Applying: utils: rename filterSingleNumaNodePolicyNrts to FilterByPolicies
> > Applying: padder: wait for pods deletion
> > Applying: e2e: serial: implement test id 47594
> > Using index info to reconstruct a base tree...
> > M	test/e2e/serial/workload_placement_test.go
> > Falling back to patching base and 3-way merge...
> > Auto-merging test/e2e/serial/workload_placement_test.go
> > CONFLICT (content): Merge conflict in test/e2e/serial/workload_placement_test.go
> > error: Failed to merge in the changes.
> > hint: Use 'git am --show-current-patch=diff' to see the failed patch
> > Patch failed at 0005 e2e: serial: implement test id 47594
> > When you have resolved this problem, run "git am --continue".
> > If you prefer to skip this patch, run "git am --skip" instead.
> > To restore the original branch and stop patching, run "git am --abort".
> > ```
> 
> @fromanirh not good. I think #149 is missing from 4.10 @jlojosnegros do you mind having a second look?

we're in a bit of a rush, so I took the liberty: https://github.com/openshift-kni/numaresources-operator/pull/149#issuecomment-1042666963

PR #145, opened 2022-02-11
Remove requests from SS deployment in kind-ci.

jlojosnegros:
Secondary scheduler pod sometimes fails to run on master node due to
lack of resources.
Secondary scheduler deployment is created by NROSchedReconciler which
loads a yaml file as template, so we use `yq` tool to delete the
`requests` from that file so the pod could be scheduled.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -27,6 +27,12 @@ jobs:
       with:
         fetch-depth: 0 # Fetch all history for all tags and branches
 
+    # Secondary scheduler pod sometimes fails to run on master node due to lack of resources.
+    # Secondary scheduler deployment is created by NROSchedReconciler which loads this yaml file
+    # as template, so we use `yq` tool to delete the `requests` so the pod could be scheduled.
+    - name: Remove Secondary Scheduler Deployment Requests
+      run: yq eval -i 'del(.spec.template.spec.containers[select(.spec.template.containers[].name=="secondary-scheduler")].resources.requests)' pkg/numaresourcesscheduler/manifests/yaml/deployment.yaml
+
     - name: Set up golang
       uses: actions/setup-go@v2
       with:


jlojosnegros:
> changing the deployment like we do in https://github.com/openshift-kni/numaresources-operator/blob/v4.10.0-pre2022012501/.github/workflows/e2e.yml#L55 was already borderline, but changing the source code is too much IMO. This doesn't look like the best direction.


Surely not the best solution, but we did not find any other so the pod could be scheduled on master node.
I tried with kustomize, but this deployment is not deployed directly but created by the scheduler reconciler so kustomize did not seem to work.
I think the need of this kind of workaround will be eliminated when NROS would be deployed by a dedicated operator, but in the meanwhile ci is failing most of the time due to this, so I think we should find a way to mitigate the problem.

Tal-or:
> changing the deployment like we do in https://github.com/openshift-kni/numaresources-operator/blob/v4.10.0-pre2022012501/.github/workflows/e2e.yml#L55 was already borderline, but changing the source code is too much IMO. This doesn't look like the best direction.

We can't use kustomize as we did in previous changes, so it looked like the most painless option.

Moreover, the scheduler is temporary anyway, so when we will move into SSOP we can come up with more elegant solution.

fromanirh:
> > changing the deployment like we do in https://github.com/openshift-kni/numaresources-operator/blob/v4.10.0-pre2022012501/.github/workflows/e2e.yml#L55 was already borderline, but changing the source code is too much IMO. This doesn't look like the best direction.
> 
> Surely not the best solution, but we did not find any other so the pod could be scheduled on master node.

Sure, the problem here is the fact that the free tier of github actions provides us machines too weak for our use case.
Which is fine, when we get things for free we can't complain too much.
So better use these resources in a different way. We can either reorganize the kind lanes or prioritize the tests we want to run. For the purposes of this operator, k8s/kind is a tier-2 platform anyway, so it's OK to run only a subset of tests in there.
OTOH we should not purposefully exclude k8s, but that's a far cry from what we need to do here.

> I tried with kustomize, but this deployment is not deployed directly but created by the scheduler reconciler so kustomize did not seem to work.

Indeed, kustomize can't work here. The internal manifests are embedded into the operator. We are effectively changing the source code here, and that's why this change is excessive.

> I think the need of this kind of workaround will be eliminated when NROS would be deployed by a dedicated operator, but in the meanwhile ci is failing most of the time due to this, so I think we should find a way to mitigate the problem.

I agree, and I believe the best mitigation is rethink how we run the tests, and which of them we run.

fromanirh:
> > changing the deployment like we do in https://github.com/openshift-kni/numaresources-operator/blob/v4.10.0-pre2022012501/.github/workflows/e2e.yml#L55 was already borderline, but changing the source code is too much IMO. This doesn't look like the best direction.
> 
> We can't use kustomize as we did in previous changes, so it looked like the most painless option.
> 
> Moreover, the scheduler is temporary anyway, so when we will move into SSOP we can come up with more elegant solution.

Thing is the previous approach was borderline already. CI should test an artifact as close as possible (ideally identical) to the shipped version. The saving grace here was k8s is a tier-2 platform, and we have some slack space wrt the OCP recommended settings. But we were bending the rules a bit already. Let's not move too forward in that direction.

jlojosnegros:
This is not the correct approach. We should find another way to solve this issue without modifying the production code.

PR #149, opened 2022-02-14
e2e: deploy with gu pod scheduled in NUMA zone

jlojosnegros:
Using a deployment with a guaranteed pod as workload, ensure that
if there is one node with at least one NUMA zone with
enough resources to have a pod scheduled into it,
the scheduler will use that NUMA zone even if there are
some other nodes with enough resources at node level to schedule the pod

Checking for at least two nodes with two NUMA zones with enough
resources and padding all but one of those nodes
so they would not have enough resources in only one NUMA zoone
to handle the pod

----
This is similar to #124 but using a deployment as workload

### DIFF
--- a/test/e2e/serial/workload_placement_test.go
+++ b/test/e2e/serial/workload_placement_test.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/klog/v2"
 
+	schedutils "github.com/openshift-kni/numaresources-operator/test/e2e/sched/utils"
 	e2efixture "github.com/openshift-kni/numaresources-operator/test/utils/fixture"
 	e2enrt "github.com/openshift-kni/numaresources-operator/test/utils/noderesourcetopologies"
 	"github.com/openshift-kni/numaresources-operator/test/utils/nrosched"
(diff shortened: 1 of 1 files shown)


test/e2e/serial/workload_placement_test.go
```diff
@@ -506,6 +511,41 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s", updatedPod.Namespace, updatedPod.Name, schedulerName)
 		})
+
+		It("[test_id: 47583] a deployment with a guaranteed pod with one container should be scheduled into one NUMA zone", func() {
```
jlojosnegros, line 515: Not sure if this is the proper test_id

fromanirh: it's a label so should not be a big deal. I'd just writr `test_id:47583` (no space in between) for starters

fromanirh:
/retest
infra issue:
```
  * could not run steps: step src failed: could not get build src: builds.build.openshift.io "src" not found 
```

fromanirh (commented):
/lgtm
/approve

PR #150, opened 2022-02-14
WIP: e2e: serial: implement test id 47591

Tal-or:
Test creates a deployment with 1 replica, 1 guaranteed pod, two containers, requesting available resources, verifies that pod scheduled successfully and NRT get updated accordingly.
Test update the resources request of the pod under the deployment (post scheduling) and make sure that NRT gets updated accordingly.

depend on: #143


test/utils/padder/padder.go
```diff
@@ -61,6 +62,13 @@ type Padder struct {
 	*padRequest
 }
 
+// PaddingOptions is some configuration that modifies options for a pad request.
+type PaddingOptions struct {
+	// LabelSelector pad nodes with a given label.
+	// the number of padded nodes is still limit by padRequest.nNodes value
```
fromanirh, line 68: which is kinda redundant with the label selector, then.
No need to change this now, but we will need to revisit the API later.

fromanirh (commented):
initial review, looks nice so far

test/e2e/serial/workload_placement_test.go
```diff
@@ -715,6 +717,305 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+				Skip(fmt.Sprintf("not enough nodes with 2 NUMA Zones: found %d", len(nrtCandidates)))
+			}
+
+			singleNUMAPolicyNrts := e2enrt.FilterByPolicies(nrtInitialList.Items, []nrtv1alpha1.TopologyManagerPolicy{nrtv1alpha1.SingleNUMANodePodLevel, nrtv1alpha1.SingleNUMANodeContainerLevel})
+			nodesNameSet := e2enrt.AccumulateNames(singleNUMAPolicyNrts)
+
+			// the only node which was not padded is the targetedNode
+			// since we know exactly how the test setup looks like we expect only targeted node here
+			targetNodeNameSet := nodesNameSet.Difference(paddedNodesSet)
+			Expect(targetNodeNameSet.Len()).To(Equal(1))
```
fromanirh, line 762: minor: let's use a more descriptive error message here, like (random rushed example):
```
Expect(targetNodeNameSet.Len()).To(Equal(1), "cannot find the target node")
```

test/e2e/serial/workload_placement_test.go
```diff
@@ -715,6 +717,305 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			rl = corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("2"),
+				corev1.ResourceMemory: resource.MustParse("8G"),
+			}
+
+			err = padder.Nodes(1).UntilAvailableIsResourceList(rl).Pad(timeout, e2epadder.PaddingOptions{LabelSelector: sel})
+			Expect(err).ToNot(HaveOccurred())
+
+			// we reorganize the cluster state, so we need to get an updated NRTs which will be treated as the initial ones
+			nrtInitialList, err = e2enrt.GetUpdated(fxt.Client, nrtv1alpha1.NodeResourceTopologyList{}, time.Second*10)
```
fromanirh, line 928: I don't like reusing variables, but let's fix as followup

test/e2e/serial/workload_placement_test.go
```diff
@@ -715,6 +717,305 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+				"pod should not landed on node %q", targetNodeName)
+
+			By(fmt.Sprintf("checking the pod was scheduled with the topology aware scheduler %q", schedulerName))
+			schedOK, err = nrosched.CheckPODWasScheduledWith(fxt.K8sClient, updatedPod.Namespace, updatedPod.Name, schedulerName)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s", updatedPod.Namespace, updatedPod.Name, schedulerName)
+
+			rl = e2enrt.ResourcesFromGuaranteedPod(updatedPod)
+
+			nrtPostUpdateNo2, err := e2enrt.FindFromList(nrtPostUpdateNo2DeploymentList.Items, updatedPod.Spec.NodeName)
```
fromanirh, line 984: "No2" depends too much on the flow, let's pick a better name

test/e2e/serial/workload_placement_test.go
```diff
@@ -715,6 +717,305 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+
+			By(fmt.Sprintf("checking NRT for target node %q updated correctly", targetNodeName))
+			// TODO: this is only partially correct. We should check with NUMA zone granularity (not with NODE granularity)
+			_, err = e2enrt.CheckZoneConsumedResourcesAtLeast(*nrtPostUpdate, *nrtPostUpdateNo2, rl)
+			Expect(err).ToNot(HaveOccurred())
+
+			err = unlabel()
```
fromanirh, line 992: let's make sure we unlabel inconditionally - e.g. even if the tests fail with error, so we don't leave leftovers around

fromanirh (commented):
nice work. This was a very complex test. Besides minor comment, let's have another pass to make sure (or at least spend more effort) not to leave the cluster polluted if the test fails.
Other than that, it's ready to merge

test/e2e/serial/workload_placement_test.go
Tal-or: sure I'll pass it to the ```AfterEach()```

fromanirh: Adding it on `AfterEach` to remove any labels we may have left over seems like a good approach. Works for me.

test/utils/padder/padder.go
fromanirh: ok, not urgent anyway.

Tal-or:
> no need to change it now, but let's generalize only when we are sure we need to (it seems we need exactly one `cleanFunc`), let's not build infra "just in case", especially in the context of e2e tests.

ack!

fromanirh:
> > `defer` won't be called in case of assertion as far as I remember
> 
> My bad it does get called. I'll address in later PR

no need to rush. It's low priority.

PR #151, opened 2022-02-15
bundle: examples: update CR examples

fromanirh:
copy data from doc/examples, which still is our authoritative source.

### DIFF
--- a/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
@@ -8,20 +8,28 @@ metadata:
           "apiVersion": "nodetopology.openshift.io/v1alpha1",
           "kind": "NUMAResourcesOperator",
           "metadata": {
-            "name": "numaresourcesoperator-sample"
+            "name": "numaresourcesoperator"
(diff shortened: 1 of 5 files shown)


bundle/manifests/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -8,20 +8,28 @@ metadata:
...
         {
           "apiVersion": "nodetopology.openshift.io/v1alpha1",
           "kind": "NUMAResourcesScheduler",
           "metadata": {
-            "name": "numaresourcesscheduler-sample"
+            "name": "numaresourcesscheduler"
           },
           "spec": {
-            "foo": "bar"
+            "imageSpec": "quay.io/openshift-kni/scheduler-plugins:4.10-snapshot"
```
cynepco3hahue, line 32: probably should be 4.11-snapshot because we already branched, and the release-4.10 should have 4.10-snapshot

fromanirh: right, fixing. This however will mean we cannot easily cherry-pick the change. No big deal though.

PR #162, opened 2022-02-17
Add container including e2e tests

ffromani:
Add container which includes all our e2e suites. Will be used for integration of the serial lane in CI envs.

### DIFF
--- a/Dockerfile.tests
+++ b/Dockerfile.tests
@@ -1,0 +1,12 @@
+# Build all the e2e suites
+FROM golang:1.17 as builder
+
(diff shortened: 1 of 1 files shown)


fromanirh:
/retest
```
could not get build root: builds.build.openshift.io "root" not found
```
obvious infra issue

Dockerfile.openshift.tests
```diff
@@ -0,0 +1,3 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal
+COPY e2e-*.test /usr/local/bin/
```
jlojosnegros, line 2: what if the test are not yet built?

PR #163, opened 2022-02-17
e2e-workload: Workload requests guaranteed pod resources available on one node but not on a single numa

marioferh:
Ensure that there is at least one node with guaranteed resources available but cannot be scheduled in a single numa zone.

### DIFF
--- a/test/e2e/serial/workload_placement_test.go
+++ b/test/e2e/serial/workload_placement_test.go
@@ -576,6 +576,151 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 		})
 	})
 
+	Context("with no suitable node", func() {
+		var requiredRes corev1.ResourceList
+		BeforeEach(func() {
+			requiredNUMAZones := 2
+			By(fmt.Sprintf("filtering available nodes with at least %d NUMA zones", requiredNUMAZones))
+			nrtCandidates := e2enrt.FilterZoneCountEqual(nrts, requiredNUMAZones)
+
+			neededNodes := 1
+			if len(nrtCandidates) < neededNodes {
+				Skip(fmt.Sprintf("not enough nodes with 2 NUMA Zones: found %d, needed %d", len(nrtCandidates), neededNodes))
+			}
+
+			nrtCandidateNames := e2enrt.AccumulateNames(nrtCandidates)
+
+			//TODO: we should calculate requiredRes from NUMA zones in cluster nodes instead.
+			requiredRes = corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("4"),
+				corev1.ResourceMemory: resource.MustParse("4Gi"),
+			}
+
+			By("Padding selected node")
+			// TODO This should be calculated as 3/4 of requiredRes
+			paddingRes := corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("3"),
+				corev1.ResourceMemory: resource.MustParse("3Gi"),
+			}
+
+			var paddingPods []*corev1.Pod
+			for _, nodeName := range nrtCandidateNames.List() {
+
+				nrtInfo, err := e2enrt.FindFromList(nrtCandidates, nodeName)
+				Expect(err).NotTo(HaveOccurred(), "missing NRT info for %q", nodeName)
+
+				for idx, zone := range nrtInfo.Zones {
+					podName := fmt.Sprintf("padding%s-%d", nodeName, idx)
+					padPod, err := makePaddingPod(fxt.Namespace.Name, podName, zone, paddingRes)
+					Expect(err).NotTo(HaveOccurred(), "unable to create padding pod %q on zone", podName, zone.Name)
+
+					padPod, err = pinPodTo(padPod, nodeName, zone.Name)
+					Expect(err).NotTo(HaveOccurred(), "unable to pin pod %q to zone", podName, zone.Name)
+
+					err = fxt.Client.Create(context.TODO(), padPod)
+					Expect(err).NotTo(HaveOccurred(), "unable to create pod %q on zone", podName, zone.Name)
+
+					paddingPods = append(paddingPods, padPod)
+				}
+			}
+
+			failedPods := e2ewait.ForPodListAllRunning(fxt.Client, paddingPods)
+			for _, failedPod := range failedPods {
+				_ = objects.LogEventsForPod(fxt.K8sClient, failedPod.Namespace, failedPod.Name)
+			}
+			Expect(failedPods).To(BeEmpty(), "some padding pods have failed to run")
+		})
+
+		It("[test_id:47617] workload requests guaranteed pod resources available on one node but not on a single numa", func() {
+
+			By("Scheduling the testing pod")
+			pod := objects.NewTestPodPause(fxt.Namespace.Name, "testPod")
+			pod.Spec.SchedulerName = schedulerName
+			pod.Spec.Containers[0].Resources.Limits = requiredRes
+
+			err := fxt.Client.Create(context.TODO(), pod)
+			Expect(err).NotTo(HaveOccurred(), "unable to create pod %q", pod.Name)
+
+			err = e2ewait.WhileInPodPhase(fxt.Client, pod.Namespace, pod.Name, corev1.PodPending, 10*time.Second, 3)
+			if err != nil {
+				_ = objects.LogEventsForPod(fxt.K8sClient, pod.Namespace, pod.Name)
+			}
+			Expect(err).ToNot(HaveOccurred())
+
+			By(fmt.Sprintf("checking the pod was scheduled with the topology aware scheduler %q", schedulerName))
+			schedOK, err := nrosched.CheckPODWasScheduledWith(fxt.K8sClient, pod.Namespace, pod.Name, schedulerName)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s", pod.Namespace, pod.Name, schedulerName)
+		})
+
+		It("a deployment with a guaranteed pod resources available on one node but not on a single numa", func() {
+
+			By("Scheduling the testing deployment")
+			deploymentName := "test-dp"
+			var replicas int32 = 1
+
+			podLabels := map[string]string{
+				"test": "test-dp",
+			}
+			nodeSelector := map[string]string{}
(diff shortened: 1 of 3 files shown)


test/e2e/serial/workload_placement_test.go
```diff
@@ -561,6 +561,85 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 		})
 	})
 
+	Context("with a node no suitable", func() {
```
fromanirh, line 564: better: "with no suitable node"

test/e2e/serial/workload_placement_test.go
```diff
@@ -561,6 +561,85 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			pod := objects.NewTestPodPause(fxt.Namespace.Name, "testPod")
+			pod.Spec.SchedulerName = schedulerName
+			pod.Spec.Containers[0].Resources.Limits = requiredRes
+
+			err := fxt.Client.Create(context.TODO(), pod)
+			Expect(err).NotTo(HaveOccurred(), "unable to create pod %q", pod.Name)
+
+			By("waiting for node to be up&running")
+			podRunningTimeout := 1 * time.Minute
+			failedPod, err := e2ewait.ForPodPhase(fxt.Client, pod.Namespace, pod.Name, corev1.PodRunning, podRunningTimeout)
```
fromanirh, line 631: here we expect the pod to be pending, so it cannot go running, so this will fail 100% of times, doesn't it?

test/e2e/serial/workload_placement_test.go
```diff
@@ -561,6 +561,85 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			}
+			Expect(failedPods).To(BeEmpty(), "some padding pods have failed to run")
+		})
+
+		It("[test_id:47617] Workload requests guaranteed pod resources available on one node but not on a single numa", func() {
+
+			By("Scheduling the testing pod")
+			pod := objects.NewTestPodPause(fxt.Namespace.Name, "testPod")
+			pod.Spec.SchedulerName = schedulerName
+			pod.Spec.Containers[0].Resources.Limits = requiredRes
```
fromanirh, line 624: this covers the guaranteed pod, but we need to run the same test for deployment (with guaranteed pods) and daemonsets (with guaranteed pods)

fromanirh (commented):
initial review

test/e2e/serial/workload_placement_test.go
marioferh: But the condition is expected error:
`Expect(err).To(HaveOccurred(), "Pod %q not up&running after %v", pod.Name, podRunningTimeout)
 Expect(failedPod).To(BeFalse())`
 
 Anyway is redundant with next lines that check pending status. Remove

test/e2e/serial/workload_placement_test.go
marioferh: Sure. Just want to be sure that this is ok, then extend to deployment and daemonset.

test/e2e/serial/workload_placement_test.go
```diff
@@ -561,6 +561,79 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+
+			err := fxt.Client.Create(context.TODO(), pod)
+			Expect(err).NotTo(HaveOccurred(), "unable to create pod %q", pod.Name)
+
+			err = e2ewait.WhileInPodPhase(fxt.Client, pod.Namespace, pod.Name, corev1.PodPending, 10*time.Second, 3)
+			if err != nil {
+				_ = objects.LogEventsForPod(fxt.K8sClient, pod.Namespace, pod.Name)
+			}
+			Expect(err).ToNot(HaveOccurred())
```
fromanirh, line 633: let's check the pod was scheduled with the NUMA aware scheduler

test/e2e/serial/workload_placement_test.go
fromanirh: I'll need to review the padding logic but the general direction seems good

test/e2e/serial/workload_placement_test.go
```diff
@@ -561,6 +561,79 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 		})
 	})
 
+	Context("with a no suitable node", func() {
+		var requiredRes corev1.ResourceList
+		BeforeEach(func() {
+			const requiredNUMAZones = 2
```
fromanirh, line 567: still not a fan of `const` used like this, but is not wrong either

test/e2e/serial/workload_placement_test.go
fromanirh: yup, is redundant (and a tiny bit of unclear)

test/e2e/serial/workload_placement_test.go
```diff
@@ -561,6 +561,84 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 		})
 	})
 
+	Context("with a no suitable node", func() {
```
fromanirh, line 564: good! better `with no suitable node`

test/e2e/serial/workload_placement_test.go
```diff
@@ -561,6 +561,84 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+		var requiredRes corev1.ResourceList
+		BeforeEach(func() {
+			requiredNUMAZones := 2
+			By(fmt.Sprintf("filtering available nodes with at least %d NUMA zones", requiredNUMAZones))
+			nrtCandidates := e2enrt.FilterZoneCountEqual(nrts, requiredNUMAZones)
+
+			const neededNodes = 1
```
fromanirh, line 571: let's either use `const` also above on line 567 or drop it from here.
I for myself I still have a funky feeling about using `const` this way :\

test/e2e/serial/workload_placement_test.go
fromanirh: was this addressed?

test/e2e/serial/workload_placement_test.go
```diff
@@ -576,6 +576,149 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 		})
 	})
 
+	Context("with no suitable node", func() {
```
fromanirh, line 579: missing `test_id`(s)

test/e2e/serial/workload_placement_test.go
```diff
@@ -576,6 +576,149 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			deployment.Spec.Template.Spec.SchedulerName = schedulerName
+			deployment.Spec.Template.Spec.Containers[0].Resources.Limits = requiredRes
+
+			err := fxt.Client.Create(context.TODO(), deployment)
+			Expect(err).NotTo(HaveOccurred(), "unable to create deployment %q", deployment.Name)
+
+			By("waiting for deployment to be up&running")
+			dpRunningTimeout := 1 * time.Minute
+			dpRunningPollInterval := 10 * time.Second
+			err = e2ewait.ForDeploymentComplete(fxt.Client, deployment, dpRunningPollInterval, dpRunningTimeout)
```
fromanirh, line 676: if the underlying pod stay in pending (and they should per test description), the deployment should actually complete

test/e2e/serial/workload_placement_test.go
```diff
@@ -576,6 +576,149 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			Expect(err).NotTo(HaveOccurred(), "Deployment %q not up&running after %v", deployment.Name, dpRunningTimeout)
+
+			By(fmt.Sprintf("checking deployment pods have been scheduled with the topology aware scheduler %q ", schedulerName))
+			pods, err := schedutils.ListPodsByDeployment(fxt.Client, *deployment)
+			Expect(err).To(HaveOccurred(), "Unable to get pods from Deployment %q:  %v", deployment.Name, err)
+
+			pod := pods[0]
```
fromanirh, line 683: just iterate over all pods you found, even if we expect just one. It's a generalization coming free.
If you want to be stricter (up to you) add another Expect we get exactly one pod.

test/e2e/serial/workload_placement_test.go
```diff
@@ -576,6 +576,149 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			ds.Spec.Template.Spec.SchedulerName = schedulerName
+			ds.Spec.Template.Spec.Containers[0].Resources.Limits = requiredRes
+
+			err := fxt.Client.Create(context.TODO(), ds)
+			Expect(err).NotTo(HaveOccurred(), "unable to create deployment %q", ds.Name)
+
+			By("waiting for deployment to be up&running")
+			dsRunningPollInterval := 10 * time.Second
+			dsRunningTimeout := 1 * time.Minute
+			ds, err = e2ewait.ForDaemonSetReady(fxt.Client, ds, dsRunningPollInterval, dsRunningTimeout)
```
fromanirh, line 708: if the underlying pod stay in pending (and they should per test description), the daemonset should not actually become ready

test/e2e/serial/workload_placement_test.go
```diff
@@ -576,6 +576,149 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			Expect(err).NotTo(HaveOccurred(), "Deployment %q not up&running after %v", ds.Name, dsRunningTimeout)
+
+			By(fmt.Sprintf("checking deployment pods have been scheduled with the topology aware scheduler %q ", schedulerName))
+			pods, err := schedutils.ListPodsByDaemonset(fxt.Client, *ds)
+			Expect(err).To(HaveOccurred(), "Unable to get pods from Deployment %q:  %v", ds.Name, err)
+
+			pod := pods[0]
```
fromanirh, line 715: just iterate over all pods you found, even if we expect just one. It's a generalization coming free.
If you want to be stricter (up to you) add another Expect we get exactly one pod.

fromanirh (commented):
almost there!

test/e2e/serial/workload_placement_test.go
marioferh: then it would be Expect(err).To(HaveOccurred()

test/e2e/serial/workload_placement_test.go
fromanirh: I guess you mean `Expect(err).To(HaveOccurred())` (remove `Not`). I think we can start like this and optimize later, yes

test/e2e/serial/workload_placement_test.go
fromanirh: yes, we can start this way and optimize later

fromanirh:
/lgtm
/approve
thanks for the updates

PR #165, opened 2022-02-18
Makefile: reorder all targets

fromanirh:
reorganize the '*-all' targets for better clarity,
and make sure we have the 'binary-e2e-all' target.

Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #166, opened 2022-02-18
e2e: serial: revamp the runner script

fromanirh:
reorganize the e2e serial runner script and make sure
it can be shipped in the tests container image.
This means fixing the paths.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,7 +165,12 @@ binary-e2e-sched:
 binary-e2e-serial:
 	go test -c -v -o bin/e2e-serial.test ./test/e2e/serial
 
-binary-e2e-all: binary-e2e-install binary-e2e-rte binary-e2e-sched binary-e2e-uninstall binary-e2e-serial
+binary-e2e-all: binary-e2e-install binary-e2e-rte binary-e2e-sched binary-e2e-uninstall binary-e2e-serial runner-e2e-serial
+
+runner-e2e-serial:
+	@grep -v 'source hack/common.sh' hack/run-test-e2e-serial.sh | \
+		env BIN_DIR=/usr/local/bin envsubst > bin/run-e2e-serial.sh && \
+		chmod 0755 bin/run-e2e-serial.sh
 
 build: generate fmt vet binary
 
--- a/hack/run-test-e2e-serial.sh
+++ b/hack/run-test-e2e-serial.sh
@@ -2,6 +2,20 @@
 
 source hack/common.sh
 
+# we expect the lane to run against a already configured cluster
+SETUP="${SETUP:-false}"
+TEARDOWN="${TEARDOWN:-false}"
+RUN_TESTS="${RUN_TESTS:-true}"
+
+# so few arguments is no bother enough for getopt
+for arg in "$@"; do
+	case "${arg}" in
+		--setup) SETUP="true";;
+		--teardown) TEARDOWN="true";;
+		--no-run-tests) RUN_TESTS="false";;
+	esac
+done
+
 NO_COLOR=""
 if ! which tput &> /dev/null 2>&1 || [[ $(tput -T$TERM colors) -lt 8 ]]; then
   echo "Terminal does not seem to support colored output, disabling it"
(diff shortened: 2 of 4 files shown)


fromanirh:
depends on https://github.com/openshift-kni/numaresources-operator/pull/165

fromanirh:
/hold

because https://github.com/openshift-kni/numaresources-operator/pull/166#issuecomment-1046178640

Makefile
```diff
@@ -167,6 +167,11 @@ binary-e2e-serial:
 
 binary-e2e-all: binary-e2e-install binary-e2e-rte binary-e2e-sched binary-e2e-uninstall binary-e2e-serial
 
+runner-e2e-serial:
+	@grep -v 'source hack/common.sh' hack/run-test-e2e-serial.sh | \
```
jlojosnegros, line 171: a little bit hackey, isn't it?

jlojosnegros (approved):
/lgtm

Makefile
fromanirh: yes, it is. Question is how we can reuse the same codebase for both use cases.

hack/run-test-e2e-serial.sh
```diff
@@ -12,5 +26,37 @@ if [ -n "${E2E_SERIAL_FOCUS}" ]; then
...
 fi
 
-echo "Running Serial, disruptive E2E Tests"
-${BIN_DIR}/e2e-serial.test ${NO_COLOR} --ginkgo.v --ginkgo.failFast --ginkgo.reportFile=/tmp/artifacts/e2e-serial --test.parallel=1 ${FOCUS}
+function setup() {
+	if [[ x"${SETUP}" != "xtrue" ]]; then
+		return 0
+	fi
+	echo "Running NRO install test suite"
+	${BIN_DIR}/e2e-install.test ${NO_COLOR} --ginkgo.v --ginkgo.failFast --ginkgo.reportFile=/tmp/artifacts/e2e-serial-install --ginkgo.focus='\[Install\] continuousIntegration'
```
Tal-or, line 34: This is not enough. it's only the NROP CR installation.
We also need the ```NUMAResourcesScheduler``` CR from here:
https://github.com/openshift-kni/numaresources-operator/blob/main/hack/run-test-e2e.sh#L8-L26

hack/run-test-e2e-serial.sh
```diff
@@ -12,5 +26,37 @@ if [ -n "${E2E_SERIAL_FOCUS}" ]; then
...
+}
+
+function teardown() {
+	if [[ x"${TEARDOWN}" != "xtrue" ]]; then
+		return 0
+	fi
+        echo "Running NRO uninstall test suite";
+        ${BIN_DIR}/e2e-uninstall.test ${NO_COLOR} --ginkgo.v --ginkgo.reportFile=/tmp/artifacts/e2e-serial-uninstall;
```
Tal-or, line 42: Also adding teardown for the ```NUMAResourcesScheduler```

hack/run-test-e2e-serial.sh
```diff
@@ -12,5 +26,37 @@ if [ -n "${E2E_SERIAL_FOCUS}" ]; then
...
+function setup() {
+	if [[ x"${SETUP}" != "xtrue" ]]; then
+		return 0
+	fi
+	echo "Running NRO install test suite"
+	${BIN_DIR}/e2e-install.test ${NO_COLOR} --ginkgo.v --ginkgo.failFast --ginkgo.reportFile=/tmp/artifacts/e2e-serial-install --ginkgo.focus='\[Install\] continuousIntegration'
+}
+
+function teardown() {
+	if [[ x"${TEARDOWN}" != "xtrue" ]]; then
```
Tal-or, line 38: ```xtrue```?

hack/run-test-e2e-serial.sh
```diff
@@ -2,6 +2,20 @@
 
 source hack/common.sh
 
+# we expect the lane to run against a already configured cluster
+SETUP="${SETUP:-false}"
+TEARDOWN="${TEARDOWN:-false}"
```
Tal-or, line 7: I think the default should be keeping the cluster intact, so ```true```

Tal-or:
/hold
Please address my comments

hack/run-test-e2e-serial.sh
fromanirh: good point, addressing

hack/run-test-e2e-serial.sh
fromanirh: yes, I need to handle the sched part in this script

hack/run-test-e2e-serial.sh
fromanirh: you mean they should go in pairs? if we don't do setup (and we don't want to do that _by default_), we should not do teardown either

Tal-or: My bad, I Missed the ```SETUP``` variable above

hack/run-test-e2e-serial.sh
fromanirh: old habits die hard: https://stackoverflow.com/questions/174119/why-do-shell-script-comparisons-often-use-xvar-xyes

Tal-or: TIL

fromanirh: in this case I'm actually misleading because we explicitely require bash (not any `sh` compatible shell) so the conditions which will make this idiom useful just can't trigger. Will remove in the next upload.

fromanirh:
/hold cancel
lgtm is gone anyway :)

Tal-or:
/lgtm
/approve
Thank you for the changes

PR #168, opened 2022-02-21
Conditions: avoid unnecessary updates due to wrong timestamp

jlojosnegros:
We create a new `default` set of conditions to compare with the current
ones. These conditions are created with `now()`	as value for
`LastTransitionTime` field.
This way comparison always return there are differences between
conditions and therefore they are always updated even if it is not
needed.

Update `LastTransitionTime` to have the same value as the current
conditions so it would not modify the comparison.

Note: everything in condition handling is based on the assumption that
the conditions order in the conditions array is always the same.

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -22,7 +22,9 @@ import (
 
 	"github.com/pkg/errors"
 
-	"k8s.io/apimachinery/pkg/api/equality"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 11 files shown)


jlojosnegros:
/cc @fromanirh 
BTW thanks for the test.

pkg/status/status.go
```diff
@@ -42,17 +42,21 @@ const (
 	ConditionTypeIncorrectNUMAResourcesOperatorResourceName = "IncorrectNUMAResourcesOperatorResourceName"
 )
 
-func Update(ctx context.Context, client k8sclient.Client, rte *nropv1alpha1.NUMAResourcesOperator, condition string, reason string, message string) error {
+func Update(ctx context.Context, client k8sclient.Client, rte *nropv1alpha1.NUMAResourcesOperator, condition string, reason string, message string) (bool, error) {
 	conditions := NewConditions(condition, reason, message)
+	for index := 0; index < len(rte.Status.Conditions); index++ {
```
fromanirh, line 47: since we need to iterate over all the conditions anyway, and since the `semantic.DeepEqual` has no choice but do the same (in a generic way, but still), I'm wondering if we should not just add an explicit equality check which does NOT consider timestamps - but does consider everything else.
That will also enable us to reuse the `conditions` we build on line 46 instead of re-creating on line 54

fromanirh (commented):
thanks, this is a good catch! let's reason a bit about the implementation

pkg/status/status.go
jlojosnegros: Yep, I was also thinking about that . I  went with this one to keep using `semantic.DeepEqual` but I can give it a try...

jlojosnegros: looking around ... google go-cmp package has an [`IgnoreField`](https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#IgnoreFields) option that could be useful here. 

Code will get something like 
`ret := cmp.Equal(conditions, rte.Status.Conditions, cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime"))
`

More info: https://github.com/google/go-cmp

fromanirh: we already depend on `go-cmp`, so this is a good option.
Looking at `metav1.Condition`, the struct is pretty simple:
```
type Condition struct {
	Type string `json:"type" protobuf:"bytes,1,opt,name=type"`
	Status ConditionStatus `json:"status" protobuf:"bytes,2,opt,name=status"`
	ObservedGeneration int64 `json:"observedGeneration,omitempty" 
	LastTransitionTime Time `json:"lastTransitionTime" protobuf:"bytes,4,opt,name=lastTransitionTime"`
	Reason string `json:"reason" protobuf:"bytes,5,opt,name=reason"`
	Message string `json:"message" protobuf:"bytes,6,opt,name=message"`
}
```
we should also ignore `ObservedGeneration`.
An explicit function to check equality seems also still very much on the table.

jlojosnegros: Ok. Both ways seem ok to me so if any have a preference between go-cmp and a explicit function I will take it.

fromanirh: let's start with `go-cmp`

jlojosnegros:
https://github.com/openshift-kni/numaresources-operator/blob/f52e0bc30807e431b3cd2bf82769225308fea6d5/controllers/numaresourcesscheduler_controller.go#L192

Seems NumaResourcersSchedulerOperator has the same problem ... and threre are also some code repetition that could be handled

fromanirh:
> https://github.com/openshift-kni/numaresources-operator/blob/f52e0bc30807e431b3cd2bf82769225308fea6d5/controllers/numaresourcesscheduler_controller.go#L192
> 
> Seems NumaResourcersSchedulerOperator has the same problem ... and threre are also some code repetition that could be handled

can't remember why we went with inlined code in here. Using `status` package won't hurt I believe.

fromanirh:
ah, right. because `status` takes a NROP (not sched) object. We can use some little cleanup in this area.

jlojosnegros:
yep ... will open a new PR once this is merged to try to solve it if you agree

fromanirh:
> yep ... will open a new PR once this is merged to try to solve it if you agree

go for it!

Issue #169, opened 2022-02-21
Conditions: Avoid unnecessary updates in NRSchedOperator


jlojosnegros:
NUMAResourcesSchedulerOperator has the same minor error comparing Conditions as NRO had and was solved in #168
This come from some code duplication in both components that should be approached

https://github.com/openshift-kni/numaresources-operator/blob/f52e0bc30807e431b3cd2bf82769225308fea6d5/controllers/numaresourcesscheduler_controller.go#L192

jlojosnegros:
solved by #170

PR #170, opened 2022-02-21
Update SS Conditions cmp to avoid unnnedded update

jlojosnegros:
Use status package to properly compare conditions.
Also avoid code repetition between NRO and SS as much as possible.

handing #169

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -157,14 +157,27 @@ func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctr
 func (r *NUMAResourcesOperatorReconciler) updateStatus(ctx context.Context, instance *nropv1alpha1.NUMAResourcesOperator, condition string, reason string, message string) (ctrl.Result, error) {
 	klog.Error(message)
 
-	if _, err := status.Update(ctx, r.Client, instance, condition, reason, message); err != nil {
+	if _, err := updateStatus(ctx, r.Client, instance, condition, reason, message); err != nil {
 		klog.InfoS("Failed to update numaresourcesoperator status", "Desired condition", status.ConditionDegraded, "error", err)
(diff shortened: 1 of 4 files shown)


controllers/numaresourcesoperator_controller.go
```diff
@@ -157,16 +157,29 @@ func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctr
...
 		klog.InfoS("Failed to update numaresourcesoperator status", "Desired condition", status.ConditionDegraded, "error", err)
 		return ctrl.Result{}, err
 	}
-
 	// we do not return an error here because to pass the validation error a user will need to update NRO CR
 	// that will anyway initiate to reconcile loop
 	return ctrl.Result{}, nil
 }
 
+func updateNROConditions(ctx context.Context, cli client.Client, rte *nropv1alpha1.NUMAResourcesOperator, condition string, reason string, message string) (bool, error) {
```
fromanirh, line 169: I'd call it `updateStatus` for symmetry with the scheduler controller

pkg/status/status.go
```diff
@@ -44,23 +37,19 @@ const (
 	ConditionTypeIncorrectNUMAResourcesOperatorResourceName = "IncorrectNUMAResourcesOperatorResourceName"
 )
 
-func Update(ctx context.Context, client k8sclient.Client, rte *nropv1alpha1.NUMAResourcesOperator, condition string, reason string, message string) (bool, error) {
+func UpdateConditions(currentConditions []metav1.Condition, condition string, reason string, message string) ([]metav1.Condition, bool) {
```
fromanirh, line 40: this is nice, but doesn't really update conditions, does it? A name like `GetUpdatedConditions` seems a bit more fitting. Besides the name, the rest seems just fine.

fromanirh (commented):
nice, this is very similar to what I was thinking

PR #172, opened 2022-02-22
e2e test: Burstable pod not fitting any NUMA zone

jlojosnegros:
Test to check burstable pods with too much required resources to be
scheduled on any NUMA zone but fitting on at least one node, are still
scheduled using Secondary Scheduler with no change in
NumaResourceTopology for that node.

### DIFF
--- a/test/e2e/serial/workload_placement_test.go
+++ b/test/e2e/serial/workload_placement_test.go
@@ -18,6 +18,7 @@ package serial
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
@@ -29,7 +30,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	nodev1 "k8s.io/api/node/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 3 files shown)


test/e2e/serial/workload_placement_test.go
```diff
@@ -429,6 +429,124 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+				}
+
+				// Wait for all the padding pods to be up&running
+				failedPods := e2ewait.ForPodListAllRunning(fxt.Client, paddingPods)
+				for _, failedPod := range failedPods {
+					_ = objects.LogEventsForPod(fxt.K8sClient, failedPod.Namespace, failedPod.Name)
+				}
+				Expect(failedPods).To(BeEmpty())
+			})
+			When("Trying to schedule a deployment with burstable nodes", func() {
```
jlojosnegros, line 507: typo nodes -> pods

test/e2e/serial/workload_placement_test.go
```diff
@@ -429,6 +429,124 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 			}, time.Minute, time.Second*5).Should(BeTrue(), "resources not restored on %q", updatedPod.Spec.NodeName)
 		})
 	})
+	Context("with at least one node with at least two numa zones", func() {
+		var nrtCandidates []nrtv1alpha1.NodeResourceTopology
+
+		BeforeEach(func() {
+			const requiredNUMAZones = 2
+			By(fmt.Sprintf("filtering available nodes with at least %d NUMA zones", requiredNUMAZones))
+			nrtCandidates := e2enrt.FilterZoneCountEqual(nrts, requiredNUMAZones)
```
yanirq, line 438: will FilterZoneCountEqual get all nodes exactly with 2 numa zones ? or above 2  ?  if the answer is the former then the candidates will not fit the test description

jlojosnegros: `FilterZoneCountEqual` gets those nodes with exactly the number of zones given. For the test any node with at least two numa zones will fit ... but it seems we have some restrictions right now 

https://github.com/jlojosnegros/numaresources-operator/blob/8718cfa681abd706e56cf55bec8ef7f457d5342b/test/e2e/serial/workload_placement_test.go#L105

I do not know if it is worthy to repeat the comment on each call or it is better to make a general TODO comment in the file to avoid future problems.

test/e2e/serial/workload_placement_test.go
```diff
@@ -429,6 +429,124 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+				for _, nodeName := range nrtCandidateNames.List() {
+					podName := fmt.Sprintf("padding%s-full", nodeName)
+
+					node := &corev1.Node{}
+					nodeKey := client.ObjectKey{Name: nodeName}
+					err := fxt.Client.Get(context.TODO(), nodeKey, node)
+					Expect(err).NotTo(HaveOccurred())
+
+					// calculate resources to fill node
+					podRes := node.Status.Allocatable
```
Tal-or, line 483: Is having a pod that asks for the complete amount of allocatable resources is possible?

test/e2e/serial/workload_placement_test.go
```diff
@@ -429,6 +429,124 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			By(fmt.Sprintf("filtering available nodes with at least %d NUMA zones", requiredNUMAZones))
+			nrtCandidates := e2enrt.FilterZoneCountEqual(nrts, requiredNUMAZones)
+
+			const neededNodes = 1
+			if len(nrtCandidates) < neededNodes {
+				Skip(fmt.Sprintf("not enough nodes with at least %d NUMA Zones: found %d, needed %d", requiredNUMAZones, len(nrtCandidates), neededNodes))
+			}
+		})
+
+		Context("there are enough resources on one node but not in any numa zone", func() {
```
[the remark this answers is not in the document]
jlojosnegros: What I understand is that `Describe`, `Context` and `When` are synonymous and are though as ways to build a hierarchy where you can setup your SUT one condition at a time so you could reuse those setups for different tests...

test/e2e/serial/workload_placement_test.go
```diff
@@ -429,6 +429,124 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+					deployment = objects.NewTestDeployment(replicas, podLabels, nodeSelector, fxt.Namespace.Name, deploymentName, objects.PauseImage, []string{objects.PauseCommand}, []string{})
+					deployment.Spec.Template.Spec.SchedulerName = schedulerName
+					// make it burstable
+					deployment.Spec.Template.Spec.Containers[0].Resources.Requests = reqResources
+
+					err := fxt.Client.Create(context.TODO(), deployment)
+					Expect(err).NotTo(HaveOccurred(), "unable to create deployment %q", deployment.Name)
+
+					By("waiting for deployment to be up&running")
+					dpRunningTimeout := 1 * time.Minute
```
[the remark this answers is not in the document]
jlojosnegros: Is there any benefit in that? I mean, using variables make code more readable IMO so if there are no drawbacks I would prefer to keep them

test/e2e/serial/workload_placement_test.go
fromanirh: The best option seems to fix the text in the `By` above (remove "at least")

test/e2e/serial/workload_placement_test.go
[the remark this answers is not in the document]
fromanirh: a possible downside is that is uncommon (and, I'd say, a possible smell) to reuse values across `Wait*` or `Eventually`. Even if the actual value can be the same, each is functionally independent. So each of them would need its own set of variables. Which is not wrong either, but... uncommon.

test/e2e/serial/workload_placement_test.go
jlojosnegros: After checking ... no, it is not. So I would have to figure out another way to be sure there is no resources on the other nodes ...

test/e2e/serial/workload_placement_test.go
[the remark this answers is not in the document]
jlojosnegros: Ok that makes sense because you couple different Waits if you use the same variables ...

test/e2e/serial/workload_placement_test.go
jlojosnegros: Leaving alone the big error of assign directly one map to the other .... 
if k8s docs said "The allocatable block indicates the amount of resources on a Node that is available to be consumed by normal Pods." and there are no quotas ... using Allocatable Memory and CPU values should give a valid ResourceList for a pod , isn't it?

test/e2e/serial/workload_placement_test.go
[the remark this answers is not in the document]
Tal-or: Usually, that's what we're doing.
In addition to that, we also have here 3 layers of ```BeforeEach``` which is confusing either IMO. 
If we're not planning to have multiple ```It```'s in between each context then we don't need to split the ```BeforeEach```'s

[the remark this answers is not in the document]
jlojosnegros: Mmm I was not able to find a best-practice regarding where to use each of the different container nodes neither in the docs nor in our code, so maybe I am not using them properly. 

Regarding the `BeforeEach` hierarchy ... I though one of the main advantages to use Ginkgo was to be able to build test using a hierarchy were you can setup your SUT one action at a time, so you could reuse the setup code and also to see easily which states you are covering and which ones no ... 
On the clarity point ... I find easier to read a small, to the point, `BeforeEach`/`AfterEach` function implementing the action commented in the `Context`/`When`/`Describe` than a larger one with all the operations ... 

But I don't have a strong opinion so, let me change it

test/e2e/serial/workload_placement_test.go
```diff
@@ -443,6 +444,128 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 			}, time.Minute, time.Second*5).Should(BeTrue(), "resources not restored on %q", updatedPod.Spec.NodeName)
 		})
 	})
+	Context("cluster with node/s having two numa zones, and there are enough resources on one node but not in any numa zone when trying to schedule a deployment with burstable pods", func() {
+		var nrtCandidates []nrtv1alpha1.NodeResourceTopology
+		var targetNodeName string
+		var targetNodeNRTInitial *nrtv1alpha1.NodeResourceTopology
+		var deployment *appsv1.Deployment
+
+		BeforeEach(func() {
```
fromanirh, line 453: I don't see a real benefit in using a `BeforeEach` for exactly one `It` block. Maybe move this code inside the `It`?

test/e2e/serial/workload_placement_test.go
```diff
@@ -443,6 +444,128 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			// add a mim ammount of resources just in case all the zones
+			// are equal so the pod wont fit on any of them
+			excessRes := corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("100m"),
+				corev1.ResourceMemory: resource.MustParse("100M"),
+			}
+			reqResources := maxResources.DeepCopy()
+
+			reqCpu := reqResources[corev1.ResourceCPU]
+			reqCpu.Add(excessRes[corev1.ResourceCPU])
```
fromanirh, line 489: it seems to me we can inline the `resourceMustParse` above and remove `excessRes` entirely

test/e2e/serial/workload_placement_test.go
```diff
@@ -443,6 +444,128 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			reqResources[corev1.ResourceCPU] = reqCpu
+
+			reqMem := reqResources[corev1.ResourceMemory]
+			reqMem.Add(excessRes[corev1.ResourceMemory])
```
fromanirh, line 493: same

test/e2e/serial/workload_placement_test.go
```diff
@@ -443,6 +444,128 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+				node := &corev1.Node{}
+				nodeKey := client.ObjectKey{Name: nodeName}
+				err := fxt.Client.Get(context.TODO(), nodeKey, node)
+				Expect(err).NotTo(HaveOccurred())
+
+				// Padding all the other nodes until only maxResources is free
+				// remember reqResources = maxReources + excessResources
+				// so we ensure out pod will not be scheduled on any other node.
+				padPod, err := makeNodePaddingPod(fxt.Namespace.Name, *node, maxResources)
+				if err != nil && strings.HasPrefix(err.Error(), "NOT_ENOUGH_RES") {
```
fromanirh, line 508: uhm this is a smell. We should really try hard to make logic on the error text

fromanirh (commented):
looks pretty good, some questions inside. I need to review the allocation logic, will have another pass later

test/e2e/serial/workload_placement_test.go
jlojosnegros: Here the point is `SaturateZoneUntilLeft` returns an error when there is no enough resources in the Node to create a pod with the parameters ... but in this case that is not an error because if the node has not enough resources it is ok ...
I was thinking about creating an Error type with some kind of "type" to be able to differentiate between the different causes of errors ... but in the end it felt like too much.

test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -123,10 +128,31 @@ func SaturateZoneUntilLeft(zone nrtv1alpha1.Zone, requiredRes corev1.ResourceLis
 	return paddingRes, nil
 }
 
+func SaturateNodeUntilLeft(node corev1.Node, requiredRes corev1.ResourceList) (corev1.ResourceList, error) {
```
fromanirh, line 131: It seems we will need to rename functions here in a future cleanup PR, because the `Saturate*` function family is not saturating a node, but rather computing the resources needed to saturate a node.

test/e2e/serial/workload_placement_test.go
```diff
@@ -1129,3 +1278,20 @@ func labelNode(cli client.Client, label, nodeName string) (func() error, error)
...
+	var max resource.Quantity
+
+	for _, zone := range nrtInfo.Zones {
+		zoneQty, ok := e2enrt.FindResourceAvailableByName(zone.Resources, resName.String())
+		if !ok {
+			continue
+		}
+
+		if zoneQty.Cmp(max) > 1 {
+			max = zoneQty.DeepCopy()
```
fromanirh, line 1293: or just assign, which should be sufficient to avoid side effects (see comment below) and do `DeepCopy` just one time at return time

test/e2e/serial/workload_placement_test.go
```diff
@@ -1129,3 +1278,20 @@ func labelNode(cli client.Client, label, nodeName string) (func() error, error)
 
 	return unlabel, nil
 }
+
+func maxResourceType(nrtInfo nrtv1alpha1.NodeResourceTopology, resName corev1.ResourceName) resource.Quantity {
+
```
fromanirh, line 1283: nit: we usually avoid spaces at the beginning of a function body

test/e2e/serial/workload_placement_test.go
```diff
@@ -1054,19 +1177,45 @@ func makePaddingPod(namespace, nodeName string, zone nrtv1alpha1.Zone, podReqs c
...
+	klog.Infof("want to have node %q with allocatable: %s", node.Name, e2ereslist.ToString(podReqs))
+
+	paddingReqs, err := e2enrt.SaturateNodeUntilLeft(node, podReqs)
+	if err != nil {
+		return nil, err
+	}
+
+	klog.Infof("padding resource to saturate %q: %s", node.Name, e2ereslist.ToString(paddingReqs))
+
+	padPod := objects.NewTestPodPause(namespace, fmt.Sprintf("padpod-%s-full", node.Name))
```
fromanirh, line 1190: turns out this way of creating pod names is wrong, can lead to spurious rejections for name too long. 
See: https://github.com/openshift-kni/numaresources-operator/commit/41460a9d5eee4c7a70631a65215a0248c0a40339
We can fix this and all the occurrences in a later PR though

test/utils/noderesourcetopologies/noderesourcetopologies.go
jlojosnegros: makes sense.

test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -123,10 +128,31 @@ func SaturateZoneUntilLeft(zone nrtv1alpha1.Zone, requiredRes corev1.ResourceLis
...
+		if !ok {
+			return nil, fmt.Errorf("resource %q not found in node %q", string(resName), node.Name)
+		}
+
+		if nodeQty.Cmp(resQty) < 0 {
+			klog.Errorf("resource %q already too scarce in zone %q (target %v amount %v)", resName, node.Name, resQty, nodeQty)
+			return nil, ErrNotEnoughResources
+		}
+		klog.Infof("node %q resource %q available %s allocation target %s", node.Name, resName, nodeQty.String(), resQty.String())
+		paddingQty := nodeQty.DeepCopy()
```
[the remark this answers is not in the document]
jlojosnegros: No. I did not experienced any side effect I just did it the same way `SaturateZoneUntilLeft` was doing it.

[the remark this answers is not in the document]
fromanirh: No need, we can keep it like this way.

fromanirh (commented):
/approve
/lgtm
thanks for the updates

jlojosnegros:
rebase to main to solve conflicts

fromanirh:
conflicts again
time has come for us to split `workload_placement_test.go` (obviously not in this PR)

jlojosnegros:
rebasing...

PR #176, opened 2022-02-23
WIP: controllers: exclude memory reporting

fromanirh:
Exclude the memory reporting temporarily.


fromanirh:
better to fix on the scheduler side

PR #184, opened 2022-02-25
e2e: serial: fix pod generation

fromanirh:
Fix pod naming issues emerged during the fine tuining of the e2e serial suite.
Please check the individual commits for details.

### DIFF
--- a/test/e2e/serial/workload_placement_test.go
+++ b/test/e2e/serial/workload_placement_test.go
@@ -275,7 +275,7 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 				corev1.ResourceCPU:    resource.MustParse("6"),
 				corev1.ResourceMemory: resource.MustParse("6Gi"),
 			}
-			pod.Spec.Containers[0].Name = "testcnt-1"
+			pod.Spec.Containers[1].Name = "testcnt-1"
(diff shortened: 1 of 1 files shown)


fromanirh:
/hold
let me have a test run myself

fromanirh:
/retest
```
error updating LB Target Group
```
infra issue

fromanirh:
/hold
running last tests

Tal-or (commented):
Sorry for my ignorance, but what does it mean from the PR's comment:
_the hostname can make the pod name exceed the limit_

fromanirh:
> Sorry for my ignorance, but what does it mean from the PR's comment: _the hostname can make the pod name exceed the limit_

Pod name can't be longer than 63 chars. If we construct the name using prefix (`padpod-` in this case) + hostname + extra (zone name like `zone-$NUM`) it becomes quite easy to end up with constructed names longer than 63 chars.
This will cause the apiserver to reject the pod creation right from the start.

fromanirh:
/hold cancel
good enough for now, we will add targeted fixes (if needed) when enabling the tests to scope=pod in future PRs.

Tal-or:
/lgtm
/approve
We'll probably need to do the same for the padder package

PR #185, opened 2022-02-27
e2e: serial: Modify configurable values

Tal-or:
The test modifies every configurable value under the ```NUMAResourcesOperator``` and ```NUMAResourcesScheduler``` CR's and checks that the controller applied the configuration as expected.

The test requires MCP changes, hence nodes reboot, and therefore it's label with a special ```[reboot_required]``` label to denote that.
We also add the ```[slow]``` label to denote this test might take more time than usual.

### DIFF
--- a/test/e2e/serial/workload_placement_test.go
+++ b/test/e2e/serial/workload_placement_test.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/ghodss/yaml"
@@ -38,9 +39,16 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/util/taints"
 
+	"github.com/google/go-cmp/cmp"
 	nrtv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
+	nropv1alpha1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1alpha1"
+	"github.com/openshift-kni/numaresources-operator/pkg/flagcodec"
+	"github.com/openshift-kni/numaresources-operator/pkg/loglevel"
+	nropmcp "github.com/openshift-kni/numaresources-operator/pkg/machineconfigpools"
 	numacellapi "github.com/openshift-kni/numaresources-operator/test/deviceplugin/pkg/numacell/api"
 	schedutils "github.com/openshift-kni/numaresources-operator/test/e2e/sched/utils"
+	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
+	"github.com/openshift-kni/numaresources-operator/test/utils/configuration"
 	e2efixture "github.com/openshift-kni/numaresources-operator/test/utils/fixture"
 	e2enrt "github.com/openshift-kni/numaresources-operator/test/utils/noderesourcetopologies"
 	e2enodes "github.com/openshift-kni/numaresources-operator/test/utils/nodes"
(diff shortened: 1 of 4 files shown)


Tal-or:
A real bug was found in the business logic during the test implementation. 
Working to address it.
This PR will be dependent on the PR with the solution.

test/e2e/serial/workload_placement_test.go
```diff
@@ -1037,6 +1045,174 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 				return ok
 			}, time.Minute, time.Second*5).Should(BeTrue(), "resources not restored on %q", updatedPod.Spec.NodeName)
 		})
+
+		It("[test_id:47674][rebootRequired] should be able to modify the configurable values under the NUMAResourcesOperator and NUMAResourcesScheduler CR", func() {
```
fromanirh, line 1049: let's avoid `snakeCase` or `CamelCase` in the tags name (convention). `reboot_required` is better. Also please add a `slow` tag, so it reads like
`[test_id:47674][reboot_required][slow] should be...`

test/e2e/serial/workload_placement_test.go
```diff
@@ -1657,6 +1667,231 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+
+			nroSchedObj := &nropv1alpha1.NUMAResourcesScheduler{}
+			err := fxt.Client.Get(context.TODO(), client.ObjectKey{Name: nrosched.NROSchedObjectName}, nroSchedObj)
+			Expect(err).ToNot(HaveOccurred(), "cannot get %q in the cluster", nrosched.NROSchedObjectName)
+			initialNroSchedObj := nroSchedObj.DeepCopy()
+
+			nroOperObj := &nropv1alpha1.NUMAResourcesOperator{}
+			err = fxt.Client.Get(context.TODO(), client.ObjectKey{Name: objects.NROName()}, nroOperObj)
+			Expect(err).ToNot(HaveOccurred(), "cannot get %q in the cluster", objects.NROName())
+			initialNroOperObj := nroOperObj.DeepCopy()
```
fromanirh, line 1681: we already to this in `test_suite_serial_test.go` but this test wants to mutate these values, so it's safer to redo that here.

test/e2e/serial/workload_placement_test.go
```diff
@@ -1657,6 +1667,231 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			targetedNode := workers[0]
+
+			unlabelFunc, err := labelNode(fxt.Client, e2enodes.GetLabelRoleMCPTest(), targetedNode.Name)
+			Expect(err).ToNot(HaveOccurred())
+
+			labelFunc, err := unlabelNode(fxt.Client, e2enodes.GetLabelRoleWorker(), "", targetedNode.Name)
+			Expect(err).ToNot(HaveOccurred())
+
+			defer func() {
+				By(fmt.Sprintf("unlabling node: %q", targetedNode.Name))
```
fromanirh, line 1695: typo: "unlabeling"

fromanirh: right below similar typo "labeling"

test/e2e/serial/workload_placement_test.go
```diff
@@ -1657,6 +1667,231 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 				return ok
 			}, time.Minute, time.Second*5).Should(BeTrue(), "resources not restored on %q", targetedNodeName)
 		})
+
+		It("[test_id:47674][reboot_required][slow] should be able to modify the configurable values under the NUMAResourcesOperator and NUMAResourcesScheduler CR", func() {
```
fromanirh, line 1671: please also add a label to signal this test depends on some images, like `[images]`

fromanirh (commented):
nice work, some minor comments inside, please check anyway. The biggest thing is we need a new label. Can be done in a followup PR.

fromanirh:
/lgtm
/approve
actually let's add the label in a followup PR, this is big enough already

fromanirh:
label added in https://github.com/openshift-kni/numaresources-operator/pull/205

PR #187, opened 2022-02-28
rte: don't use google pause container

fromanirh:
u/s RTE wants to use the google pause container for the
shared pool container. We need to minimize the container image,
so we replace it reusing the same RTE image with a different
entry point.

Since we require anyway the RTE image to be based on a ubi-minimal
(or equivalent) image, and since the shared-pool-container
does not need to reap pids but just to sleep forever peacefully,
we can use just the simplest sleep command.


fromanirh:
/retest
seems infra issue

fromanirh:
/hold cancel
local test passing (e2e test pending)

jlojosnegros (approved):
/lgtm

marioferh (approved):
/lgtm

PR #188, opened 2022-03-01
ConfigMap hash annotation in workload resources

Tal-or:
We're adding an annotation with the ConfigMap's hash in workload resources like deployment and daemonset ```.spec.template ``` filed.
This change allows us to track changes/updates done to the ConfigMap objects. 
A new hash will get calculated for every ConfigMap change/update which in turn causes the workload resource to restart the relevant pod and read the updated ConfigMap.
This change is important in order to make sure that our pods (RTE and the scheduler) are not running with a stale ConfigMap configuration.


fromanirh:
we will need a similar change for RTE - and to disable the `exit-on-config-change` flag as well.

controllers/numaresourcesscheduler_controller.go
```diff
@@ -194,3 +201,23 @@ func (r *NUMAResourcesSchedulerReconciler) updateStatus(ctx context.Context, sch
 	}
 	return nil
 }
+
+func (r *NUMAResourcesSchedulerReconciler) syncConfigMapHash(ctx context.Context) (string, error) {
```
fromanirh, line 205: please add a reference as comment about an article/blog post which illustrates this technique

Tal-or:
> we will need a similar change for RTE - and to disable the `exit-on-config-change` flag as well.


This solution will work only when RTE is deployed with the controller.
Is that good enough to justify the removal of the current solution which works (although not perfectly) with the deployer as well?

fromanirh:
> > we will need a similar change for RTE - and to disable the `exit-on-config-change` flag as well.
> 
> This solution will work only when RTE is deployed with the controller. Is that good enough to justify the removal of the current solution which works (although not perfectly) with the deployer as well?

we will not remove the fix from the deployer, but we should use the same approach for the RTE in the numaresourcesoperator_controller, and we should stop enabling the `exit-on-config-changes` flag (https://github.com/openshift-kni/numaresources-operator/blob/v4.10.0-rc0/pkg/objectstate/rte/rte.go#L345)

controllers/numaresourcesscheduler_controller.go
```diff
@@ -194,3 +200,24 @@ func (r *NUMAResourcesSchedulerReconciler) updateStatus(ctx context.Context, sch
...
+
+// a reference which illustrates such technique can be found at https://blog.questionable.services/article/kubernetes-deployments-configmap-change/
+func (r *NUMAResourcesSchedulerReconciler) syncConfigMapHash(ctx context.Context) (string, error) {
+	mfConfigMap := r.SchedulerManifests.ConfigMap // shortcut
+	updatedConfigMap := &corev1.ConfigMap{}
+	key := client.ObjectKey{
+		Namespace: mfConfigMap.Namespace,
+		Name:      mfConfigMap.Name,
+	}
+	if err := r.Get(ctx, key, updatedConfigMap); err != nil {
```
fromanirh, line 212: minor: I'd have used `client.ObjectKeyFromObject` but not worth a resubmit for this

pkg/numaresourcesscheduler/objectstate/sched/sched.go
```diff
@@ -220,6 +226,10 @@ func UpdateSchedulerName(cm *corev1.ConfigMap, name string) error {
 
 }
 
+func GetConfigMapHash(cm *corev1.ConfigMap) string {
+	return fmt.Sprintf("SHA256:%x", sha256.Sum256([]byte(cm.String())))
```
fromanirh, line 230: so we want hash ALL the objects, metadata included (and not the data), right?

controllers/numaresourcesscheduler_controller.go
```diff
@@ -194,3 +200,24 @@ func (r *NUMAResourcesSchedulerReconciler) updateStatus(ctx context.Context, sch
 	}
 	return nil
 }
+
+// a reference which illustrates such technique can be found at https://blog.questionable.services/article/kubernetes-deployments-configmap-change/
+func (r *NUMAResourcesSchedulerReconciler) syncConfigMapHash(ctx context.Context) (string, error) {
```
fromanirh, line 205: this function is not syncing, it is getting the hash of the current config map. Which is a pre-requisite for syncing, but it is not sending updates to the apiserver. I'd rename like `computeCurrentConfigMapHash`

main.go
```diff
@@ -321,6 +321,6 @@ func renderSchedulerManifests(schedManifests schedmanifests.Manifests, imageSpec
 	klog.InfoS("Updating scheduler manifests")
 	mf := schedManifests.Clone()
 	schedstate.UpdateDeploymentImageSettings(mf.Deployment, imageSpec)
-	schedstate.UpdateDeploymentConfigMapSettings(mf.Deployment, schedManifests.ConfigMap.Name)
+	schedstate.UpdateDeploymentConfigMapSettings(mf.Deployment, schedManifests.ConfigMap.Name, schedstate.GetConfigMapHash(schedManifests.ConfigMap))
```
fromanirh, line 324: this is one of the cases on which we can use a temporary `cmHash` like you do in `numaresourcesscheduler_controller.go`

fromanirh (commented):
looks nice!
A key question inside about what should be hashed

pkg/numaresourcesscheduler/objectstate/sched/sched.go
Tal-or: Indeed. any change to the ConfigMap will trigger a restart to the pod. 
Also from a computational perspective, it's much easier. IOW we don't need to invent our own function to return a string from a subset of fields in the ConfigMap.

main.go
Tal-or: This is exactly the same to what I'm doing in ```numaresourcesscheduler_controller.go``` - using compute the hash from the manifests instead of the object from the API server.

pkg/numaresourcesscheduler/objectstate/sched/sched.go
fromanirh: ok, works for me

Tal-or: On second thought, I'm not sure this is such a good idea. why is a new label to the ConfigMap supposed to restart the scheduler pod? 
WDYT?

fromanirh: well, this is the reason I was inquiring. My initial idea was we should only hash the configmap `Spec`, and not the metadata. But I didn't spent too much time thinking about the scenario, so I'm open to suggestions (and not _requiring_ to just hash the `Spec`)

controllers/numaresourcesscheduler_controller.go
Tal-or: blessing change, resubmiting anyway

pkg/numaresourcesscheduler/objectstate/sched/sched.go
Tal-or: let me reevaluate. I inclined to go with hash the ```Spec``` but let's see the cost of such thing

controllers/numaresourcesscheduler_controller.go
```diff
@@ -194,3 +200,20 @@ func (r *NUMAResourcesSchedulerReconciler) updateStatus(ctx context.Context, sch
...
+	updatedConfigMap := &corev1.ConfigMap{}
+	if err := r.Get(ctx, client.ObjectKeyFromObject(mfConfigMap), updatedConfigMap); err != nil {
+		// ConfigMap not created yet, use the data from the manifests
+		if apierrors.IsNotFound(err) {
+			updatedConfigMap = mfConfigMap
+		} else {
+			return "", errors.Wrapf(err, "could not calculate ConfigMap %q hash", key.String())
+		}
+	}
+	cmHash := schedstate.GetConfigMapHash(updatedConfigMap)
```
fromanirh, line 216: let's move this function in a common package

fromanirh (commented):
let's retarget this PR and add the similar change also for the `numaresourcesoperator` controller

Tal-or:
> let's retarget this PR and add the similar change also for the `numaresourcesoperator` controller

OK i'll start working on it

pkg/hash/configmap.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	cmHash := ConfigMapData(updatedConfigMap)
+	klog.InfoS("configmap hash calculated", "hash", cmHash)
+	return cmHash, nil
+}
+
+func ConfigMapData(cm *corev1.ConfigMap) string {
+	var dataAsString string
+
+	if cm.Data != nil {
```
fromanirh, line 59: this is correct, but if `BinaryData` is != nil, this function will silently break. We should at least detect and error out accordingly, because we don't need `BinaryData` now or in the near future, but breaking silently is just too bad

controllers/numaresourcesoperator_controller.go
```diff
@@ -319,6 +320,15 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 		return daemonSetsNName, err
 	}
 
+	// ConfigMap should be provided by the kubeletconfig reconciliation loop
+	if r.RTEManifests.ConfigMap != nil {
+		cmHash, err := hash.ComputeCurrentConfigMap(ctx, r.Client, r.RTEManifests.ConfigMap)
+		if err != nil {
+			return daemonSetsNName, err
+		}
+		rtestate.UpdateDaemonSetHashAnnotation(r.RTEManifests.DaemonSet, cmHash)
```
fromanirh, line 329: in the future (ofc unrelated to this PR) we will need to move `Update*` utilities in their own package. Perhaps like
```
pkg/manifests/rte
pkg/manifests/sched
```
mimiching (and extending) what `deployer` does

pkg/objectstate/rte/rte.go
```diff
@@ -361,7 +361,6 @@ func UpdateDaemonSetUserImageSettings(ds *appsv1.DaemonSet, userImageSpec, built
 		klog.Warningf("Cannot modify the command line arguments %v", cnt.Args)
 		return nil
 	}
-	fl.SetToggle("--exit-on-conf-change")
```
fromanirh, line 363: this is correct, but we can and should remove all the `flagcodec` processing here, lines [359, 365]. In any case, let's keep the `flagcodec` package around for some time still (I don't remember if this is the only usage of that package; likely not).

fromanirh (commented):
very nice. Some minor but important fixes and we're good to go.

pkg/hash/configmap.go
Tal-or: > If BinaryData is != nil, this function will silently break.

Sorry for my ignorance, but why that should happen?

fromanirh (commented):
we're almost ready. Let's solve the `BinaryData` thingy, let's add some unit tests for the newly added `hash` package and we're good.

fromanirh:
> we're almost ready. Let's solve the `BinaryData` thingy, let's add some unit tests for the newly added `hash` package and we're good.

`BinaryData` can be postponed, but we need _some_ unit test coverage for the `hash` package

pkg/hash/configmap.go
Tal-or: no, I mean why if in case that  ```BinaryData != nil``` the code will break? I'm not saying that it should never happen that ``BinaryData != nil```, I'm asking why if it's happening it will break the code

fromanirh: ah, sorry. Perhaps "break" is too strong of a statement. And also this comment is maybe misplaced, should have been in ComputeCurrentConfigMap

The fact is that if`BinaryData` is not taken into account in the hash, then different ConfigMaps (because BinaryData is different) can have the same hash, which is not what we want.
The problem is the ComputeCurrentConfigMap function whant to compute the hash of a `ConfigMap`, which in general can have a BinaryData, so we either restrict ComputeCurrentConfigMap to consider only the `Data` (maybe renaming it?) or  we need to handle `BinaryData` somehow. Erroring out if `!= nil` is an option, and so is adding the `BinaryData` in the hash. I'm fine with all the approaches, as long as it is handled in `ComputeCurrentConfigMap` - `ConfigMapData` is indeed fine.

Tal-or: Now it's clear, thanks!

IMHO ```BinaryData``` is part of data so it should be taken under consideration as well  (ofc when it's not ```nil```) this will make the function more generic and will allow support in other ConfigMap types which do contain ```BinaryData```

Tal-or:
> `BinaryData` can be postponed, but we need _some_ unit test coverage for the `hash` package

On the grill

pkg/hash/configmap.go
fromanirh: That's the best solution IMO!

pkg/hash/configmap.go
```diff
@@ -0,0 +1,66 @@
...
+		dataAsString = fmt.Sprintf("%v", cm.Data)
```
fromanirh, line 60: I'm not sure this representation is stable, meaning that we always get the same string across different golang versions. But let's address in a followup PR

pkg/hash/configmap_test.go
```diff
@@ -0,0 +1,169 @@
+package hash
```
fromanirh, line 1: missing copyright boilerplate, but not critical now

fromanirh:
the lint failures need attention but the lint lane is non-blocking, so let me re-add the hold

Tal-or:
> PTAL to the lint failures the gh actions E2E flow error is non-blocking

Fixed the lint issue

pkg/hash/configmap.go
Tal-or: hmm, I'm open to suggestions in a follow-up PR.
Besides manually serializing the data I don't see any other easy option here.

fromanirh: for similar needs, kube uses `spew`, which was later obsoleted and replaced by https://pkg.go.dev/github.com/kr/pretty
see: https://github.com/kubernetes/kubernetes/pull/102882/commits/2f426fdba6fa0458c4a47dd22b37cc920d9f2a98#diff-3b7b6a45fe39aa8a907c6a88b4f067e54c82461ac12da73c0c92d678274b5493R50

The golang formatting per se is as much as stable as third pary packages, but we need to make sure we traverse maps always in the same ordering (= sort keys), which is solved for free by packages like `pretty` (but let's doublecheck)

fromanirh:
@Tal-or could you please take care of the cherry-pick to the `release-4.10` branch? we don't need 1:1 commit fidelity so a manual backport is ok (we avoid because automatick cherry-picks are more convenient)

Tal-or:
> @Tal-or could you please take care of the cherry-pick to the `release-4.10` branch? we don't need 1:1 commit fidelity so a manual backport is ok (we avoid because automatick cherry-picks are more convenient)

Sure I'll handle it

PR #194, opened 2022-03-01
e2e: make it easier to integrate the NROP testsuite

fromanirh:
In order to make it easier to integrate the e2e suite in other tests images (other than numaresources-operator-tests), rename all the executables and entry points to include "nrop" in order to reduce  the probability of name clash.
Allow to ovverride the `numacell` image, so we can also include the numacell dp in a single test image.


fromanirh:
/hold
need to add the numacell plugin in the -tests image, which depend on https://github.com/openshift/release/pull/26670

marioferh (approved):
/lgtm

PR #198, opened 2022-03-02
e2e: make tests pod's image configurable

shajmakh:
Running the tests on disconnected environments might lead to pod crash while pulling images from external registries.

Keep using Pause image as the default test image and allow configuring the image via an environment variable TEST_IMAGE.

### DIFF
--- a/test/utils/images/images.go
+++ b/test/utils/images/images.go
@@ -16,13 +16,11 @@ limitations under the License.
 
 package images
 
-const (
-	RTETestImageCI   = "quay.io/openshift-kni/resource-topology-exporter:test-ci"
-	SchedTestImageCI = "quay.io/openshift-kni/scheduler-plugins:test-ci"
+import "os"
 
-	// NEVER EVER USE THIS OUTSIDE CI or (early) DEVELOPMENT ENVIRONMENTS
-	NUMACellDevicePluginTestImageCI = "quay.io/openshift-kni/numacell-device-plugin:test-ci"
-
-	//the default image used for test pods
-	PauseImage = "gcr.io/google_containers/pause-amd64:3.0"
-)
+func GetPauseImage() string {
+	if pullSpec, ok := os.LookupEnv("E2E_PAUSE_IMAGE_URL"); ok {
+		return pullSpec
+	}
+	return PauseImage
+}
--- a/test/utils/objects/pod.go
+++ b/test/utils/objects/pod.go
@@ -24,6 +24,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
+
+	"github.com/openshift-kni/numaresources-operator/test/utils/images"
 )
 
 func NewTestPodPause(namespace, name string) *corev1.Pod {
@@ -38,7 +40,7 @@ func NewTestPodPause(namespace, name string) *corev1.Pod {
 			Containers: []corev1.Container{
 				{
 					Name:    name + "-cnt",
-					Image:   PauseImage,
+					Image:   images.GetPauseImage(),
 					Command: []string{PauseCommand},
 				},
 			},


test/e2e/serial/workload_placement_test.go
```diff
@@ -158,7 +158,7 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 				Expect(err).ToNot(HaveOccurred(), "cannot detect the zone ID from %q", zone.Name)
 				// now put a minimal pod (1 cpu 1Gi) on both zones. Now the target node as whole will still have the
 				// required resources, but no NUMA zone individually will
-				targetedPaddingPod := objects.NewTestPodPause(fxt.Namespace.Name, fmt.Sprintf("tgtpadpod-%s", zone.Name))
+				targetedPaddingPod := objects.NewTestPod(fxt.Namespace.Name, fmt.Sprintf("tgtpadpod-%s", zone.Name))
```
fromanirh, line 161: renaming this function is unnecessary - and I'd said wrong in this context

test/utils/images/images.go
```diff
@@ -0,0 +1,37 @@
...
+
+package images
+
+import (
+	"os"
+
+	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
+)
+
+var testImage string
```
fromanirh, line 21: besides my general dislike for globals and `init` (unless where is clearly needed, and here is not), we can simply have something like
```
func GetPauseImage() string {
       if pullSpec, ok := os.LookupEnv("E2E_PAUSE_IMAGE_URL"); ok {
               return pullSpec
       }
       return objects.PauseImage
}
```
and remove everything else

test/utils/objects/pod.go
```diff
@@ -37,9 +38,16 @@ func NewTestPodPause(namespace, name string) *corev1.Pod {
 			TerminationGracePeriodSeconds: &zero,
 			Containers: []corev1.Container{
 				{
-					Name:    name + "-cnt",
-					Image:   PauseImage,
-					Command: []string{PauseCommand},
+					Name:  name + "-cnt",
+					Image: images.Test(),
+					Command: []string{
```
fromanirh, line 43: this change breaks the compatibility with the `pause` image, which is something I'd avoid just yet. Let's just revert this change. The function should read like
```
func NewTestPodPause(namespace, name string) *corev1.Pod {
	var zero int64
	return &corev1.Pod{
			Name:      name,
			Namespace: namespace,
		},
		Spec: corev1.PodSpec{
			TerminationGracePeriodSeconds: &zero,
			Containers: []corev1.Container{
				{
					Name:    name + "-cnt",
					Image:   images.GetPauseImage(),
					Command: []string{PauseCommand},
				},
			},
		},
	}
}
```

test/utils/objects/pod.go
```diff
@@ -20,13 +20,14 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/openshift-kni/numaresources-operator/test/utils/images"
```
fromanirh, line 23: the conventional import grouping is
```
stdlib

3rd parties (most known, usually kube)

3rd parties (less known, usually github packages)

local
```
so please move this import in its own group at the bottom of the `import` block

fromanirh (commented):
thanks for the PR!

test/e2e/serial/workload_placement_test.go
shajmakh: thanks for your comments.
why do we want to limit the use only for Pause image in the name of the function?

test/utils/objects/pod.go
shajmakh: thank you, I'll note this for next time. updated

test/e2e/serial/workload_placement_test.go
fromanirh: Because I think is not time yet to generalize the function. We still want a container which does `pause` and sits idle

shajmakh: ok, I see. updated. thanks!

test/utils/images/images.go
```diff
@@ -0,0 +1,26 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package images
+
+import "os"
```
fromanirh, line 19: ok, so you moved the `PauseImage` const in the `test/utils/images` package. I don't feel strongly about that. But then you need to either revert the change or to import the package here...

test/utils/images/images.go
```diff
@@ -0,0 +1,26 @@
...
+
+func GetPauseImage() string {
+	if pullSpec, ok := os.LookupEnv("E2E_PAUSE_IMAGE_URL"); ok {
+		return pullSpec
+	}
+	return PauseImage
```
fromanirh, line 25: ...and fix the name here and in all the relevant places as reported by go vet/lint.
Up to you, I'm fine both ways.

test/utils/images/images.go
shajmakh: importing `test/utils/objects` caused cyclic import which isn't allowed in Go.

fromanirh (commented):
/lgtm
/approve
on a followup PR, let's make sure we can use the -tests image (either independently or integrated into something else) as pause image

PR #202, opened 2022-03-03
e2e:serial: use a lowercase letters in pod name

Tal-or:
Pod name must be consist of the following letters only: ```'[a-z0-9]'```
The serial test is falling without this change.


fromanirh (commented):
/approve
/lgtm

PR #203, opened 2022-03-03
e2e:serial: schedule guaranteed pod in selective way

Tal-or:
The test creates a pod, pin it selectively using the nodeName stanza and makes sure that NRT objects are updating respectivley

### DIFF
--- a/test/e2e/serial/workload_placement_test.go
+++ b/test/e2e/serial/workload_placement_test.go
@@ -1037,6 +1037,82 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 				return ok
 			}, time.Minute, time.Second*5).Should(BeTrue(), "resources not restored on %q", updatedPod.Spec.NodeName)
 		})
+
+		It("[test_id:47584] should be able to schedule guarnteed pod in selective way", func() {
+			nrtList := nrtv1alpha1.NodeResourceTopologyList{}
+			nrtListInitial, err := e2enrt.GetUpdated(fxt.Client, nrtList, time.Minute)
+			Expect(err).ToNot(HaveOccurred())
+
+			workers, err := e2enodes.GetWorkerNodes(fxt.Client)
+			Expect(err).ToNot(HaveOccurred())
+
+			// TODO choose randomly
+			targetedNodeName := workers[0].Name
+
+			nrtInitial, err := e2enrt.FindFromList(nrtListInitial.Items, targetedNodeName)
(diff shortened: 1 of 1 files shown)


test/e2e/serial/workload_placement_test.go
```diff
@@ -1037,6 +1037,82 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+			cnt.Resources.Limits = requiredRes
+
+			By(fmt.Sprintf("creating pod %s/%s", testPod.Namespace, testPod.Name))
+			err = fxt.Client.Create(context.TODO(), testPod)
+			Expect(err).ToNot(HaveOccurred())
+
+			updatedPod, err := e2ewait.ForPodPhase(fxt.Client, testPod.Namespace, testPod.Name, corev1.PodRunning, timeout)
+			Expect(err).ToNot(HaveOccurred())
+
+			By(fmt.Sprintf("checking the pod landed on the target node %q vs %q", updatedPod.Spec.NodeName, targetedNodeName))
```
fromanirh, line 1080: maybe also check the pod was NOT scheduled by any scheduler after it got running?

fromanirh (commented):
/lgtm
/approve
good enough for now, possible non-blocking improvement inside

test/e2e/serial/workload_placement_test.go
Tal-or: I thought about that, but it's a bit complex to check.
The thing is that this information mentioned implicitly under the events.

For example, if the pod was scheduled by a scheduler we'll see the following under the pod's events:
```
Events:
  Type    Reason          Age   From                  Message
  ----    ------          ----  ----                  -------
  Normal  Scheduled       24s   topo-aware-scheduler  Successfully assigned default/pod1 to <node-name>
    Normal  AddedInterface  23s   multus                Add eth0 [ip] from ovn-kubernetes
  Normal  Pulling         23s   kubelet               Pulling image "quay.io..."
  Normal  Pulled          22s   kubelet               Successfully pulled image "quay.io..." in 1.025265706s
  Normal  Created         22s   kubelet               Created container c1
  Normal  Started         22s   kubelet               Started container c1

```
But, if the pod is scheduled manually you'll see no line with reason named ```Scheduled```
But I'm not sure search for no line named ```Scheduled``` is robust enough.

Instead, I assigned a fake scheduler name. if the pod was relying on the scheduler, it would stay in ```Pending``` state since such scheduler doesn't exist.

PR #205, opened 2022-03-04
e2e: serial: start the non-regression and unschedulable test groups

fromanirh:
Start to split the `workload_placement_test.go` behemoth.

split the non-regression tests into their own go source.
Non-regression tests are tests which guarantee the numa-aware scheduler
behaves like the original scheduler in cases on which numa-awareness
is not needed or relevant.

split the "unschedulable" tests into their own go source.
"unschedulable" tests are tests which wat to check a given workload is
NOT scheduled by the numa-aware scheduler.

As side benefit of this split, we de-entangle tests from a unwarranted
requirement inherited from the previous layout, enabling it to run
against any TM scope.


shajmakh:
Thanks for the patch. /lgtm

test/utils/nrosched/nrosched.go
```diff
@@ -52,18 +56,32 @@ func CheckPODWasScheduledWith(k8sCli *kubernetes.Clientset, podNamespace, podNam
...
+		if evCheck(item) {
+			klog.Infof("-> found relevant scheduling event for pod %s/%s: %v", podNamespace, podName, item)
 			return true, nil
 		}
 	}
 	klog.Warningf("Failed to find relevant scheduling event for pod %s/%s", podNamespace, podName)
 	return false, nil
 }
 
+func CheckPODSchedulingFailed(k8sCli *kubernetes.Clientset, podNamespace, podName, schedulerName string) (bool, error) {
```
Tal-or, line 71: This function doesn't cover for pod which stuck ```Pending``` state which is also included under the "scheduling failed" category.

test/e2e/serial/workload_unschedulable_test.go
```diff
@@ -196,7 +196,7 @@ var _ = Describe("[serial][disruptive][scheduler] workload unschedulable", func(
 			dsName := "test-ds"
 
 			podLabels := map[string]string{
-				"test": "test-dp",
+				"test": "test-daemonset",
```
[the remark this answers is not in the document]
fromanirh: That's correct. Here we're in the `It` block concerning a daemonset (see lines 194-195), so we're fixing a typo, should have been `test-ds` since the beginning. I'm not sure we should highlight more the typo fix.

test/utils/nrosched/nrosched.go
fromanirh: Good point! that's intentional. Pod in `Pending` state are'nt technically failed from scheudling POV. So I think we should handle them separately with a new helper.

test/e2e/serial/workload_unschedulable_test.go
```diff
@@ -0,0 +1,214 @@
...
+			}
+			Expect(err).ToNot(HaveOccurred())
+
+			By(fmt.Sprintf("checking the pod was scheduled with the topology aware scheduler %q", schedulerName))
+			schedOK, err := nrosched.CheckPODSchedulingFailed(fxt.K8sClient, pod.Namespace, pod.Name, schedulerName)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s", pod.Namespace, pod.Name, schedulerName)
+		})
+
+		It("a deployment with a guaranteed pod resources available on one node but not on a single numa", func() {
```
shajmakh, line 159: I should probably generate TC ids for the dp & ds tests.

test/e2e/serial/workload_unschedulable_test.go
```diff
@@ -0,0 +1,214 @@
...
+			err = e2ewait.WhileInPodPhase(fxt.Client, pod.Namespace, pod.Name, corev1.PodPending, 10*time.Second, 3)
+			if err != nil {
+				_ = objects.LogEventsForPod(fxt.K8sClient, pod.Namespace, pod.Name)
+			}
+			Expect(err).ToNot(HaveOccurred())
+
+			By(fmt.Sprintf("checking the pod was scheduled with the topology aware scheduler %q", schedulerName))
+			schedOK, err := nrosched.CheckPODSchedulingFailed(fxt.K8sClient, pod.Namespace, pod.Name, schedulerName)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s", pod.Namespace, pod.Name, schedulerName)
```
shajmakh, line 156: > Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s"

in the test context, if schedOK was false, it means that the event we searched for was not found, & in this context, we search for the event that tells that the SS failed to schedule the pod. so the failure msg here is inaccurate. so if schedOK is false, it could mean that the pod was actually scheduled by SS or the event was not found for some other reason.

shajmakh: the same would apply to the ds & dp cases

fromanirh: right, this need to be clarified

test/e2e/serial/workload_unschedulable_test.go
shajmakh: below are the ids:
daemonset 48962
deplyment 48963

fromanirh:
no need to retest after we just merged a `OWNERS`

PR #207, opened 2022-03-06
[release-4.10][manual] ConfigMap hash annotation in workload resources

Tal-or:
We're adding an annotation with the ConfigMap's hash in workload resources like deployment and daemonset .spec.template filed.
This change allows us to track changes/updates done to the ConfigMap objects.
A new hash will get calculated for every ConfigMap change/update which in turn causes the workload resource to restart the relevant pod and read the updated ConfigMap.
This change is important in order to make sure that our pods (RTE and the scheduler) are not running with a stale ConfigMap configuration.


Tal-or:
We agreed to stop with cherry-picking for release-4.10 until the serial tests stabilization.
But this PR is an exception due to the following:
1. It required a manual backporting intervention which I am concerned will grow bigger and become more surgical when main branch move forward
2. It doesn't relate to the serial tests, but to the controller business logic itself.

fromanirh:
I agree this PR is an exception of the current slow-down proposal we are considering and I agree we should have this PR sooner than later

fromanirh (commented):
/lgtm
/approve

PR #208, opened 2022-03-07
e2e: serial: split more tests form `workload_placement_test.go`

fromanirh:
We keep extracting tests from the `workload_placement_test.go` behemoth to make the source code more manageable; this also allows us to enable the tests to run with scope=pod where already possible, or to make the future enablement easier.

Depends on https://github.com/openshift-kni/numaresources-operator/pull/205


test/e2e/serial/workload_overhead_test.go
```diff
@@ -0,0 +1,249 @@
...
+					Expect(err).ToNot(HaveOccurred())
+					nrtPostCreate, err := e2enrt.FindFromList(nrtListPostCreate.Items, pod.Spec.NodeName)
+					Expect(err).ToNot(HaveOccurred())
+
+					_, err = e2enrt.CheckZoneConsumedResourcesAtLeast(*nrtInitial, *nrtPostCreate, podResources)
+					Expect(err).ToNot(HaveOccurred())
+
+					// Here error have to Ocurr because pod overhead resources should not be count
+					_, err = e2enrt.CheckZoneConsumedResourcesAtLeast(*nrtInitial, *nrtPostCreate, podResourcesWithOverhead)
+					Expect(err).To(HaveOccurred())
```
shajmakh, line 243: The test looks good. 
some checks are missing though as per the description in the TP: 
- Total pod resources ( + overhead) should be reflected under the node description. (I am not sure if there is an API to this) 
- there should be an additional step: modify the pod overhead in such a way that the total (pod  +overhead) resources don't fit on any node, thus the pod should stay pending.

shajmakh: Jira ticket to track this CNF-4572

test/e2e/serial/workload_overhead_test.go
```diff
@@ -0,0 +1,249 @@
...
+						APIVersion: "node.k8s.io/vi",
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "test-rtclass",
+					},
+					Handler: "runc",
+					Overhead: &nodev1.Overhead{
+						PodFixed: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("500m"),
+							corev1.ResourceMemory: resource.MustParse("500M"),
```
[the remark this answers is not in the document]
fromanirh: it will work anyway, but better be explicit indeed.

test/e2e/serial/workload_overhead_test.go
fromanirh: the `node.Status.Allocatable` is NOT updated when the resource are consumed (that's a common gotcha) so the first point needs to be postponed/ removed

the second point is fair, will add the additional step in a followup PR, I'd prefer to keep this one as simple as possible (ideally code movement + trivial fixes, no new logic if we can help it)

shajmakh: thanks for your reply.
point 1: I meant that the resources consumption per pod should be updated under the node description under the `Non-terminated Pods` list, those will include both pod +overhead resources:

```
Non-terminated Pods:                      (50 in total)
  Namespace                               Name                                                 CPU Requests  CPU Limits  Memory Requests  Memory Limits  Age
  ---------                               ----                                                 ------------  ----------  ---------------  -------------  ---
  default                                 helper-deploymentoops-again-5869455848-2ctbg         3 (2%)        3 (2%)      3G (3%)          3G (3%)        4d3h
  default                                 helper-deploymentoops-again-5869455848-8dfhw         3 (2%)        3 (2%)      3G (3%)          3G (3%)        4d3h
  default                                 helper-deploymentoops-again-5869455848-rvxz9         3 (2%)        3 (2%)      3G (3%)          3G (3%)        4d3h

```



I agree on point 2.

fromanirh: I need to look deeper into that to assess if/how this is feasible (that's a good point I didn't consider before!) so I'll need to defer to a new PR same as for point 2

shajmakh: ok np, I believe this is expected behavior regardless of the scheduler or any other NRO component. Plus, this is the only guarantee that the overhead was actually applied to the pod & yet the pod was scheduled as anticipated.

PR #211, opened 2022-03-08
e2e: serial: split tests form `workload_placement_test.go` - take 2

fromanirh:
split the "accounting" tests into their own go source.
"accounting" tests are tests which want to check how the numa-aware stack
deals with resource accounting, without yet involving the node selection
process.

Split off the tests which make use of nodeselector to their own golang source.
Split off the tests which verify how NROP handle config changes to their own golang source.

As side benefit of this move, we de-entangle a test by a unwarranted
requirement inherited from the previous layout, enabling it to run
against any TM scope.


test/e2e/serial/resource_accounting_test.go
```diff
@@ -0,0 +1,258 @@
...
+			// TODO: we need AT LEAST 2 (so 4, 8 is fine...) but we hardcode the padding logic to keep the test simple,
+			// so we can't support ATM zones > 2. HW with zones > 2 is rare anyway, so not to big of a deal now.
+			By(fmt.Sprintf("filtering available nodes with at least %d NUMA zones", requiredNumaZones))
+			nrtTwoZoneCandidates = e2enrt.FilterZoneCountEqual(nrts, requiredNumaZones)
+			if len(nrtTwoZoneCandidates) < requiredNodeNumber {
+				Skip(fmt.Sprintf("not enough nodes with 2 NUMA Zones: found %d", len(nrtTwoZoneCandidates)))
+			}
+		})
+
+		It("[placement][case:1] should keep the pod pending if not enough resources available, then schedule when resources are freed", func() {
```
shajmakh, line 95: since we didn't have this test in the TP, I added it to track its results & this is the test ID: 49068

shajmakh: I guess tier is critical. wdyt?

fromanirh: I think this is `tier2` at most

PR #213, opened 2022-03-09
e2e: serial: split tests form workload_placement_test.go - take 4

shajmakh:
Split off test that verifies a burstable pod is properly scheduled and doesn't expect for resources to be accounted in NRT. The test is relevant for both TM policies.

### DIFF
--- a/test/e2e/serial/workload_placement_test.go
+++ b/test/e2e/serial/workload_placement_test.go
@@ -18,7 +18,6 @@ package serial
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"strconv"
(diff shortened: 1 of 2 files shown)


shajmakh:
/assign @fromanirh 
may you have a look please?

shajmakh:
fixing lint ..

test/e2e/serial/workload_placement_test.go
```diff
@@ -71,10 +70,15 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 		err = fxt.Client.List(context.TODO(), &nrtList)
 		Expect(err).ToNot(HaveOccurred())
 
-		tmPolicy := nrtv1alpha1.SingleNUMANodeContainerLevel
-		nrts = e2enrt.FilterTopologyManagerPolicy(nrtList.Items, tmPolicy)
+		// we're ok with any TM policy as long as the updater can handle it,
+		// we use this as proxy for "there is valid NRT data for at least X nodes
+		policies := []nrtv1alpha1.TopologyManagerPolicy{
```
fromanirh, line 75: why we need this change in this split?

test/e2e/serial/workload_placement_test.go
```diff
@@ -310,7 +193,7 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 			Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s", updatedPod.Namespace, updatedPod.Name, schedulerName)
 		})
 
-		It("[test_id:47583][tier1] a deployment with a guaranteed pod with one container should be scheduled into one NUMA zone", func() {
+		It("[test_id: 47583][tier1] a deployment with a guaranteed pod with one container should be scheduled into one NUMA zone", func() {
```
fromanirh, line 196: no spaces between `test_id:` and the number, please

test/e2e/serial/workload_placement_test.go
```diff
@@ -374,7 +257,7 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 			}
 		})
 
-		It("[test_id:47591][tier1] should modify workload post scheduling while keeping the resource requests available", func() {
+		It("[test_id: 47591][tier1] should modify workload post scheduling while keeping the resource requests available", func() {
```
fromanirh, line 260: same

test/e2e/serial/workload_placement_test.go
fromanirh: still here, I don't think we need this change. Could you please elaborate?

test/e2e/serial/workload_placement_test.go
shajmakh: right, my mistake didn't pay much attention to previous fixing PRs. fixed. thanks.

test/e2e/serial/workload_placement_test.go
shajmakh: fixed. thanks.

test/e2e/serial/workload_placement_test.go
shajmakh: we don't need it in this split, after a second thought better to have it in a separate patch. I'll update it. thanks.

fromanirh (commented):
/lgtm
/approve
thanks for the updates

PR #171, opened 2022-02-22
e2e: serial: generalize test multi-container placement

fromanirh:
Generalize the existing e2e test which verifies verify the behaviour of the NUMA aware scheduler
by creating a testcase which is expected to foul the default, NUMA-unaware scheduler:

- 2+ nodes have all the required resources availaible at *node* level
- exactly 1 node have all the required resources available at *NUMA
  zone* level.

Hence we expect here the NUMA-aware scheduler to pick the only right
spot every time, at the first try.

In this specific test we expect the same behaviour regardless of the TM policy,
so we generalize the existing test by making it tabular.

We intentionally just check if the TM policy matches expectations, avoiding implicit
restart: is up to the test runner to setup the config accordingly, orchestrating reboots,
and perhaps running the same suite twice.

Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #214, opened 2022-03-10
e2e: serial: enable TM scope pod for tests in workload_placement_test_go

shajmakh:
Tests in `workload_placement_test.go` run only if TM scope is container and skipped if the scope is pod, while they can support both scopes.

Enable tests in workload_placement_test.go to run whether TM policy is container or pod.

### DIFF
--- a/test/e2e/serial/workload_placement_test.go
+++ b/test/e2e/serial/workload_placement_test.go
@@ -71,10 +71,15 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
 		err = fxt.Client.List(context.TODO(), &nrtList)
 		Expect(err).ToNot(HaveOccurred())
 
-		tmPolicy := nrtv1alpha1.SingleNUMANodeContainerLevel
-		nrts = e2enrt.FilterTopologyManagerPolicy(nrtList.Items, tmPolicy)
+		// we're ok with any TM policy as long as the updater can handle it,
+		// we use this as proxy for "there is valid NRT data for at least X nodes
+		policies := []nrtv1alpha1.TopologyManagerPolicy{
+			nrtv1alpha1.SingleNUMANodeContainerLevel,
+			nrtv1alpha1.SingleNUMANodePodLevel,
+		}
+		nrts = e2enrt.FilterByPolicies(nrtList.Items, policies)
 		if len(nrts) < 2 {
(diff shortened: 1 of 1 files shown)


test/e2e/serial/workload_placement_test.go
```diff
@@ -71,10 +71,15 @@ var _ = Describe("[serial][disruptive][scheduler] workload placement", func() {
...
+		// we're ok with any TM policy as long as the updater can handle it,
+		// we use this as proxy for "there is valid NRT data for at least X nodes
+		policies := []nrtv1alpha1.TopologyManagerPolicy{
+			nrtv1alpha1.SingleNUMANodeContainerLevel,
+			nrtv1alpha1.SingleNUMANodePodLevel,
+		}
+		nrts = e2enrt.FilterByPolicies(nrtList.Items, policies)
 		if len(nrts) < 2 {
-			Skip(fmt.Sprintf("not enough nodes with policy %q - found %d", string(tmPolicy), len(nrts)))
+			Skip(fmt.Sprintf("not enough nodes with valid policy - found %d", len(nrts)))
```
jlojosnegros, line 82: minor: maybe I will print which are the "valid" policies and how many nodes are needed but those are just minor adjustments.

jlojosnegros (approved):
/lgtm

test/e2e/serial/workload_placement_test.go
shajmakh: thank you for your review. good point, I will adjust that in the later patch as there are multiple instances for this part of code.

shajmakh:
/hold 
let's verify with @fromanirh

fromanirh:
I think this is too broad. Some tests on `workload_placement` need to do different things depending on the scope.
Some others intentionally restrict themselves to a specific TM scope.
A good example is https://github.com/openshift-kni/numaresources-operator/pull/171 - we required nontrivial logic changes (and a new test_id) to actually enable `scope=pod`.

To move forward, we need either:
1. to check all the remaining test one by one in `workload_placement_test.go`, and to (breifly!) document if and why the are expected to work nice with `scope=pod`
2. more invasive changes on a test-by-test basis to each remaining test to enable `scope=pod`

In general, tests which *both* use a single container *and* don't care specifically for the scope are ok to be run in either scope.
If one of the former preconditions is void, then we cannot just run them on any scope, and we need to tune the test behaviour depending on the scope.

PR #216, opened 2022-03-11
e2e: serial: make it easier to integrate

fromanirh:
The previous reorganization of the e2e tests was
made into the attempt to follow more closely the
go conventions, adding the `_test` suffix for source
files containing test code.

This worked well but made hard to import the testsuite
into external projects (cnf-tests).
Trying to import the existing suite leads to
```
no non-test files found
```
which is not good.

Hence we do another reorganization for the serial suite
only, moving the actual test definition into a utility
package.

No expected change in behaviour for the NROP testsuite
proper (nor test image).

### DIFF
--- a/test/e2e/serial/e2e_serial_test.go
+++ b/test/e2e/serial/e2e_serial_test.go
@@ -1,0 +1,31 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package serial
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	_ "github.com/openshift-kni/numaresources-operator/test/e2e/serial/tests"
+)
+
+func TestSerial(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "serial")
+}
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -48,7 +48,7 @@ import (
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 )
 
-var _ = Describe("[serial][disruptive][slow] configuration management", func() {
+var _ = Describe("[serial][disruptive][slow] numaresources configuration management", func() {
 	var fxt *e2efixture.Fixture
 	var padder *e2epadder.Padder
 	var nrtList nrtv1alpha1.NodeResourceTopologyList
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package serial
+package tests
 
 import (
 	"context"
--- a/test/utils/configuration/configuration.go
+++ b/test/utils/configuration/configuration.go
@@ -19,6 +19,7 @@ package configuration
 import (
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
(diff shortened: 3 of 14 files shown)


fromanirh:
rationale for this PR is to enable https://github.com/openshift-kni/cnf-features-deploy/pull/1010

test/utils/configuration/configuration.go
```diff
@@ -56,8 +60,26 @@ func init() {
 
 	Platform, err = detect.Detect()
 	if err != nil {
-		panic(fmt.Errorf("failed to detect a platform: %w", err))
+		Platform = getPlatformFromEnv(envVarPlatform)
+	}
+	if Platform == platform.Unknown {
```
shajmakh, line 65: why is it important to check the platform?

fromanirh: it's because the install/uninstall suites. the `test/utils/configuration` package is used in these but also in the `serial` suite, so we need this fix. Without, we get
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-kni_cnf-features-deploy/1010/pull-ci-openshift-kni-cnf-features-deploy-master-ci/1502239721626013696
```
running bin/cnftests -junit _cache/junit -ginkgo.dryRun -ginkgo.v
I0311 11:14:02.441403   22177 request.go:665] Waited for 1.046519169s due to client-side throttling, not priority and fairness, request: GET:[https://172.30.0.1:443/apis/console.openshift.io/v1?timeout=32s](https://172.30.0.1/apis/console.openshift.io/v1?timeout=32s)
panic: failed to detect a platform: clusterversions.config.openshift.io is forbidden: User "system:serviceaccount:ci-op-x0q88vkn:default" cannot list resource "clusterversions" in API group "config.openshift.io" at the cluster scope
goroutine 1 [running]:
github.com/openshift-kni/numaresources-operator/test/utils/configuration.init.0()
	/go/src/github.com/openshift-kni/cnf-features-deploy/vendor/github.com/openshift-kni/numaresources-operator/test/utils/configuration/configuration.go:59 +0x105
/go/src/github.com/openshift-kni/cnf-features-deploy
```

test/e2e/serial/e2e_serial_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+package serial
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	_ "github.com/openshift-kni/numaresources-operator/test/e2e/serial/tests"
```
shajmakh, line 25: good to know this trick with "_"

test/utils/configuration/configuration.go
shajmakh: ok thanks for clarifying that.

test/e2e/serial/e2e_serial_test.go
fromanirh: from https://go.dev/ref/spec#Import_declarations:
```
An import declaration declares a dependency relation between the importing and imported package. It is illegal for a package to import itself, directly or indirectly, or to directly import a package without referring to any of its exported identifiers. To import a package solely for its side-effects (initialization), use the [blank](https://go.dev/ref/spec#Blank_identifier) identifier as explicit package name:
```

this will trigger the `Describe` and create the specs to be used by gingko (it's more complex than this but that's roughly the idea)

fromanirh: also: https://onsi.github.io/ginkgo/#adding-specs-to-a-suite

shajmakh: thanks for the links

fromanirh:
flake (before) + infra issue (last failure)
```
level=error msg=Error: error reading Route Table Association (rtbassoc-0064583be71308361): empty result
```

fromanirh:
ok, the failure happened too many times in a row and needs to be investigated; still I don't understand how this change may have caused these failures.

fromanirh:
/retest
I still suspect a infra issue

test/e2e/serial/tests/configuration.go
```diff
@@ -48,7 +48,7 @@ import (
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 )
 
-var _ = Describe("[serial][disruptive][slow] configuration management", func() {
+var _ = Describe("[serial][disruptive][slow] numaresources configuration management", func() {
```
Tal-or, line 51: If this should be used as a label maybe it's better to set it as a label format, i.e. ```[numaresources]```

test/utils/configuration/configuration.go
```diff
@@ -28,6 +29,7 @@ import (
 const (
 	envVarMCPUpdateTimeout  = "NROP_E2E_MCP_UPDATE_TIMEOUT"
 	envVarMCPUpdateInterval = "NROP_E2E_MCP_UPDATE_INTERVAL"
+	envVarPlatform          = "NROP_E2E_PLATFORM"
```
Tal-or: Since the tests are going to be consumed by external projects, are all these environment variables are documented somewhere?
We can put them [here](https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/serial/README.md) for start

test/e2e/serial/tests/configuration.go
fromanirh: this is what I would like to do myself, but it seems to be going against the standard practice in cnf-tests. For the time being, I'd rather try to blend in their ecosystem.

PR #217, opened 2022-03-11
WIP DNM Test ci 20220311

fromanirh:
/hold
test CI


fromanirh:
turns out bugs were real, being fixed as part of https://github.com/openshift-kni/numaresources-operator/pull/216 (maybe we should split them off?)

PR #220, opened 2022-03-15
WIP: sched: better interoperability with the noderesourcefit plugin

fromanirh:
Since the enabling of memory resource alignment, we started to observe misscheduling due to the fact that the NRT filter plugin allows nodes which the builtin noderesources fit plugin rejects for lack of memory.

This PR reorganizes the code to remove this misbehaviour, and to make these conditions easier to be troubleshooted in the future.


Tal-or (commented):
I think we should have the same change under the deployer's manifests.

fromanirh:
> I think we should have the same change under the deployer's manifests.

possibly. We need to see if this is a RH-specific change or a change the general community can be interested into. I'm still investigating.

fromanirh:
postponed

PR #221, opened 2022-03-15
fix and unconditionally enable the workload placement by tm scope tests

fromanirh:
the interaction between the noderesourcetopology and the noderesource filter plugins highlighted quite few complications.

by design, NRT objects do NOT account for non-esclusive resources; this mean burstable pods do take node resources, but these are not reflected into NRT available fields. This is unfortunate, but due to the fact that non-esclusive resources are not pinned to a NUMA zone, hence are difficult to track and account.

This means the NRT and the noderesource plugin have a different perspective about the available resources.
Note the kubelet has the same picture as the noderesource plugin (bar network delays) hence - noderesource can cause a node to be filtered out "unexpectedly" - just disabling noderesource (debatable) can lead to pod rejections for `OutOfCpu` or `OutOfMemory` errors, postponing but not solving the issue.

All the above means that tests which want to saturate a node must take the burstable pod into account and not assume eviction will solve all the problems, which is what we do in this PR by adding (estimate of) node base load and using these values when doing the node padding.

Additionally, to improve the troubleshooting experience, we can and should improve the scheduling configuration, but that's
being handled in different PRs (e.g. https://github.com/openshift-kni/numaresources-operator/pull/220 )

### DIFF
--- a/test/e2e/serial/tests/e2e_suite.go
+++ b/test/e2e/serial/tests/e2e_suite.go
@@ -29,7 +29,6 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -46,11 +45,6 @@ import (
 	"github.com/openshift-kni/numaresources-operator/test/utils/nrosched"
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
 	e2ewait "github.com/openshift-kni/numaresources-operator/test/utils/objects/wait"
-)
-
-const (
-	OCPBaseLoadMemory = "4608Mi" // estimate; OCP 4.10
-	OCPBaseLoadCores  = "2"      // estimate; OCP 4.10 actually roughly 1.3 cores, but let's take HT pairs
 )
 
 const (
@@ -212,15 +206,4 @@ func unlabelNodes(cli client.Client, nrtList nrtv1alpha1.NodeResourceTopologyLis
 		err = cli.Update(context.TODO(), &node)
 		Expect(err).ToNot(HaveOccurred())
 	}
-}
-
-func applyBaseload(res corev1.ResourceList) corev1.ResourceList {
-	adjustedCPU := res.Cpu()
-	adjustedCPU.Add(resource.MustParse(OCPBaseLoadCores))
-	res[corev1.ResourceCPU] = *adjustedCPU
-
-	adjustedMemory := res.Memory()
-	adjustedMemory.Add(resource.MustParse(OCPBaseLoadMemory))
-	res[corev1.ResourceMemory] = *adjustedMemory
-	return res
 }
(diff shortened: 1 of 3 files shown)


test/e2e/serial/tests/e2e_suite.go
```diff
@@ -47,6 +48,11 @@ import (
 	e2ewait "github.com/openshift-kni/numaresources-operator/test/utils/objects/wait"
 )
 
+const (
```
Tal-or, line 51: Why not read these values from the nodes themselves under the node ```.status.allocatable``` and ```.status.capacity```

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -246,22 +248,22 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			setupPaddingPodLevel,
 			[]corev1.ResourceList{
 				{
-					corev1.ResourceCPU:    resource.MustParse("4"),
+					corev1.ResourceCPU:    resource.MustParse("6"),
```
Tal-or, line 249: Maybe I'm missing something, but why not using the ```applyBaseload``` here as well?

test/e2e/serial/tests/e2e_suite.go
fromanirh: because it will not there. We need to perform the same logic that is done when we run `oc described node` to get the stats at the end; IOW, we need to sum the `requests` of all the pods running on a node, is not a simple difference.
It is surely something we can improve later though.

test/e2e/serial/tests/workload_placement_tmpol.go
fromanirh: baseload is per-node, here we hare defining the container requests which we expect to be aligned on a NUMA cell. Hence is not exactly the same thing (I agree this code can be more explicit)

Tal-or: OK it's a table-driven test, I missed that

test/e2e/serial/tests/e2e_suite.go
fromanirh: fixed in the last commit

test/utils/nodes/baseload.go
```diff
@@ -0,0 +1,95 @@
...
+				case corev1.ResourceMemory:
+					mem.Add(resQty)
+				}
+			}
+		}
+	}
+	// get full cpus, and always take even number of CPUs
+	cpu.RoundUp(resource.Kilo)
+	cpuMillis, _ := cpu.AsInt64()
+	cpuAmount := roundUp(cpuMillis/1000, 2)
```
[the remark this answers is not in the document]
fromanirh: good catch! I wrote it with not enough caffeine and this is what you get when you don't test the tests (!!! - mostly joking, let's not get down into this rabbit hole just yet)

fromanirh:
/override ci/prow/ci-e2e
infra issue, CI passed before, and this suite isn't exercised anyay; past result are good enough.

Tal-or:
/lgtm
/approve
Thanks for all the fixes

PR #222, opened 2022-03-15
e2e: skip with call stack offset

Tal-or:
When the test calls outputed by the Skip function, it points to the code location where its caller is located which is ```skipUnlessEnvVar``` function (i.e. ```.../github.com/openshift-kni/numaresources-operator/test/e2e/serial/tests/workload_placement.go:760```)

We less care about the ```skipUnlessEnvVar``` function and more about it's caller which is the actual test code.
Let's add caller skip in order to get more useful code location.


fromanirh (commented):
/lgtm
/approve

PR #224, opened 2022-03-22
e2e: serial: add reporter

fromanirh:
Add polarion and junit reporters like we do for the
performance-addon-operators functests.


shajmakh:
thank you for the updates.
/lgtm 
one question regarding all the vendor files, what caused them to change and why?

fromanirh:
> thank you for the updates. /lgtm one question regarding all the vendor files, what caused them to change and why?

added `kubevirt.io/qe-tools` and bumped ginkgo/gomega deps. I'm not completely sure about the latter, but lagging too much behind is not risk-free either. So let's try this out, I'll be on the lookout for issues/regressions.

PR #225, opened 2022-03-22
Dockerfile: fix envsubst dependency

fromanirh:
fix non-OCP builds by bundling a local and specialized version of envsubst, which can do partial variable substitution.

### DIFF
--- a/hack/envsubst.go
+++ b/hack/envsubst.go
@@ -1,0 +1,27 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/drone/envsubst"
+)
+
+func main() {
+	stdin := bufio.NewScanner(os.Stdin)
+
+	for stdin.Scan() {
+		line, err := envsubst.Eval(stdin.Text(), func(env string) string {
+			if val, ok := os.LookupEnv(env); ok {
+				return val
+			}
+			return fmt.Sprintf("${%s}", env)
+		})
+		if err != nil {
+			log.Fatalf("Error while envsubst: %v", err)
+		}
+		fmt.Println(line)
+	}
+}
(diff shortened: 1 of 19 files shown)


fromanirh:
/hold
actually bad idea, need more refinement

fromanirh:
/hold cancel
I'm now happy enough with this PR

hack/run-test-serial-e2e.sh
```diff
@@ -29,6 +27,67 @@ if [ -n "${E2E_SERIAL_SKIP}" ]; then
...
+		*)
+			echo "unrecognized option: $1"
+			echo "use the form --opt val and not the form --opt=val"
+			exit 1
+			;;
+	esac
+done
+
+# mandatory
+if [ -z "$REPORT_DIR" ]; then
```
shajmakh, line 81: I am not familiar with shell lang, but I see all the vars are used within {}, do we need them to wrap REPORT_DIR?

fromanirh:
infra issue (limits exceeded)

jlojosnegros (approved):
/lgtm

hack/envsubst.go
```diff
@@ -0,0 +1,27 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/drone/envsubst"
```
cynepco3hahue, line 9: Why can not we just use shell `envsubst`?

fromanirh: It's an option. We will need this in the builder image (which is debian based though), and we need a specific behaviour (= replace only the env vars which have an actual value set) which is a bit harder to guarantee using shell tools.
I think all in all this approach has enough merit to be proposed, but I'm not pushing for it so if we all agree to use the shell variant (provided the above requirements are met) we can go this way.

PR #226, opened 2022-03-22
serial: add baseload to maxResources before padding

shajmakh:
Creating padding pods without considering the current baseload on the node causes scheduling those pods to fail due to insufficient resources on that node.

Account for the base load in the padding resources to avoid such cases.

### DIFF
--- a/test/e2e/serial/tests/resource_accounting.go
+++ b/test/e2e/serial/tests/resource_accounting.go
@@ -18,7 +18,6 @@ package tests
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"time"
 
@@ -30,11 +29,13 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/klog/v2"
+	resourcehelper "k8s.io/kubernetes/pkg/api/v1/resource"
 
 	nrtv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
 	schedutils "github.com/openshift-kni/numaresources-operator/test/e2e/sched/utils"
 	e2efixture "github.com/openshift-kni/numaresources-operator/test/utils/fixture"
 	e2enrt "github.com/openshift-kni/numaresources-operator/test/utils/noderesourcetopologies"
+	"github.com/openshift-kni/numaresources-operator/test/utils/nodes"
 	"github.com/openshift-kni/numaresources-operator/test/utils/nrosched"
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
 	e2ewait "github.com/openshift-kni/numaresources-operator/test/utils/objects/wait"
@@ -269,6 +270,12 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
 		var targetNodeNRTInitial *nrtv1alpha1.NodeResourceTopology
 		var deployment *appsv1.Deployment
 
+		/*
+			1. fully pad non-target nodesmake pod per numa considering the baseload: saturate the numa leaving only the baseload
+			2. create burstable pod consuming max resources of target node
+			Note: the test pod should request an amount of resources that are available on compute node level but not on a single numa,
+			   thus need to account for the baseload when running the test pod on the target node
+		*/
 		BeforeEach(func() {
 			const requiredNUMAZones = 2
 			By(fmt.Sprintf("filtering available nodes with %d NUMA zones", requiredNUMAZones))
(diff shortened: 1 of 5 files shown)


fromanirh:
> the padding pod still requests total available resources on **all** numas. this wouldn't work for pod TM scope thus need to split the load into multiple pods.

true, we need to revisit the amount of resources request after we took the baseload into account. So it is like
1. take into account the baseload when do padding
2. take into account the free resources **including the baseload** when running the testing pod

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -296,56 +297,80 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
+				baseload, err := nodes.GetLoad(fxt.K8sClient, nodeName)
+				ExpectWithOffset(1, err).ToNot(HaveOccurred(), "missing node load info for %q", nodeName)
+				By(fmt.Sprintf("computed base load: %s", baseload))
+
+				//get nrt info of the node
+				By(fmt.Sprintf("preparing target node %q to fit the test case", nodeName))
+				nrtInfo, err := e2enrt.FindFromList(nrtCandidates, nodeName)
+				ExpectWithOffset(1, err).ToNot(HaveOccurred(), "missing NRT info for %q", nodeName)
+
+				//split the base laod on all numa zones
```
fromanirh, line 340: the code seems to be contradicting the comment. What's the intention here?

test/e2e/serial/tests/workload_placement.go
```diff
@@ -746,7 +746,7 @@ func maxResourceType(nrtInfo nrtv1alpha1.NodeResourceTopology, resName corev1.Re
 			continue
 		}
 
-		if zoneQty.Cmp(max) > 1 {
+		if zoneQty.Cmp(max) == 1 {
```
fromanirh, line 749: why?

test/utils/nodes/baseload.go
```diff
@@ -48,31 +49,13 @@ func GetLoad(k8sCli *kubernetes.Clientset, nodeName string) (Load, error) {
...
-			for resName, resQty := range cnt.Resources.Requests {
-				switch resName {
-				case corev1.ResourceCPU:
-					cpu.Add(resQty)
-				case corev1.ResourceMemory:
-					mem.Add(resQty)
-				}
-			}
-		}
+		req, _ := resourcehelper.PodRequestsAndLimits(&pod)
```
fromanirh, line 53: let's handle the error, especially if unlikely

test/utils/nodes/baseload.go
```diff
@@ -48,31 +49,13 @@ func GetLoad(k8sCli *kubernetes.Clientset, nodeName string) (Load, error) {
...
-	// we round the CPU consumption as expressed in millicores (not entire cores)
-	// in order to (try to) avoid bugs related to integer division
-	// int64(2900 / 1000) -> 2 -> roundUp(2, 2) -> 2 (correct, but unexpected!)
-	// OTOH
-	// roundUp(2900, 2000) -> 4000 -> 4000/1000 -> 4 (intended behaviour)
-	// TODO: we can use some testing of the test utilities here (!)
-	cpuAmount := roundUp(cpuMillis, 2000)
-	nl.CPU = *resource.NewQuantity(cpuAmount/1000, resource.DecimalSI)
 
+	nl.CPU = *resource.NewQuantity(cpu.Value(), resource.DecimalSI)
```
fromanirh, line 58: why you simplified away the code?

fromanirh (commented):
initial review, need to deep dive in the logic proper

test/e2e/serial/tests/resource_accounting.go
shajmakh: not relevant anymore, was considering splitting the baseload between all numa-zones, not sure if that is valid especially when numas > 2 and baseload is relatively little. So I'll adjust that comment. Thanks for your note.

test/e2e/serial/tests/workload_placement.go
shajmakh: because resource.cmp return -1, 0, 1 when zoneQty is less, equal, greater than max. having it > 1 is always false thus max will always be 0. 
resource:
https://github.com/openshift-kni/numaresources-operator/blob/3289fb4fdb391b018fd98ec2fe2c5f95030d7c63/vendor/k8s.io/apimachinery/pkg/api/resource/quantity.go#L590
did I miss something?

test/utils/nodes/baseload.go
shajmakh: also it seems that quantity.Value() does the work for us of rounding up the cpus:
https://github.com/openshift-kni/numaresources-operator/blob/3289fb4fdb391b018fd98ec2fe2c5f95030d7c63/vendor/k8s.io/apimachinery/pkg/api/resource/quantity.go#L737

test/utils/nodes/baseload.go
shajmakh: `PodRequestsAndLimits` returns requests and limits, in our case we don't care about the limits

fromanirh: my bad, I should have checked the function signature. In this case it is fine as it is indeed.

test/e2e/serial/tests/workload_placement.go
fromanirh: Nope, actually now I miss how this could have possibly worked before :) good catch!

test/utils/nodes/baseload.go
fromanirh: uhm, something doesn't click here.
First of all, the error doesn't add up: in the code, we do
```
	nl.CPU = *resource.NewQuantity(cpuAmount/1000, resource.DecimalSI)
```
but cpuAmount is a int64, so from where this error is coming? `990m` is a string representation, which can't be hold in `cpuAmount`.
Second, the `Value()` function does indeed the round up, butin this code we do want to round up to cpu *pairs*, not to single integral CPU (this could have been more explicit in the comment, it is just hinted)

shajmakh: > but cpuAmount is a int64, so from where this error is coming?
you are right, we don't have the error in the original version, so to make it clear why we got 0 CPU value after `AsInt64` I added a report based on the result of this function, which seems to return false as a second arg:
```
cpuMillis, ok := cpu.AsInt64()
	if !ok {
		return nl, fmt.Errorf("could not parse cpu=%s as int64", cpu)
	}
```
https://github.com/openshift-kni/numaresources-operator/blob/3289fb4fdb391b018fd98ec2fe2c5f95030d7c63/vendor/k8s.io/apimachinery/pkg/api/resource/quantity.go#L480



> Second, the Value() function does indeed the round up, butin this code we do want to round up to cpu pairs, not to single integral CPU (this could have been more explicit in the comment, it is just hinted)
do we want that in case we are dealing with SMT level 2? I may add the value to hit the upper even value of the cpus

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -269,6 +270,12 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
 		var targetNodeNRTInitial *nrtv1alpha1.NodeResourceTopology
 		var deployment *appsv1.Deployment
 
+		/*
+			1. fully pad non-target nodesmake pod per numa considering the baseload: saturate the numa leaving only the baseload
+			2. create burstable pod consuming max resources of target node
+			Note: the test pod should request an amount of resources that are available on compute node level but not on a single numa,
+			   thus need to account for the baseload when running the test pod on the target node
+		*/
```
fromanirh, line 278: the sentences here are a bit confusing, please consider a rephrasing

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -284,65 +291,80 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
 
-			// add a min ammount of resources just in case all the zones
-			// are equal so the pod wont fit on any of them
-			excessRes := corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("100m"),
-				corev1.ResourceMemory: resource.MustParse("100M"),
+			//get maximum node CPU and Memory
+			maxResources := corev1.ResourceList{
+				corev1.ResourceCPU:    availableResourceType(*targetNodeNRTInitial, corev1.ResourceCPU),
+				corev1.ResourceMemory: availableResourceType(*targetNodeNRTInitial, corev1.ResourceMemory),
```
fromanirh, line 308: can we use node allocatable? we should document why/why not

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -284,65 +291,80 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
+				baseload, err := nodes.GetLoad(fxt.K8sClient, nodeName)
+				ExpectWithOffset(1, err).ToNot(HaveOccurred(), "missing node load info for %q", nodeName)
+				By(fmt.Sprintf("computed base load: %s", baseload))
+
+				//get nrt info of the node
+				By(fmt.Sprintf("preparing node %q to fit the test case", nodeName))
+				nrtInfo, err := e2enrt.FindFromList(nrtCandidates, nodeName)
+				ExpectWithOffset(1, err).ToNot(HaveOccurred(), "missing NRT info for %q", nodeName)
+
+				//TODO: currently, the all the padding pods account for the total baseload, need to enhance distributing it on all numa zones equally (not sure if valid though)
```
fromanirh, line 339: you can either account the baseload all on a (any) numa zone, or split it evenly. The former is probably simpler. I don't see atm a clearly better option.

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -284,65 +291,80 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
 				}
-				Expect(err).NotTo(HaveOccurred())
 
-				pinnedPadPod, err := pinPodToNode(padPod, nodeName)
-				Expect(err).NotTo(HaveOccurred())
+				//pas each zone with a pod so the test would be valid for both TM scopes: container & pod levels
+				By("pad all numa zones on the node,each with different pod.")
+				for _, zone := range nrtInfo.Zones {
+					By(fmt.Sprintf("padding node %q zone %q to fit only %s", nrtInfo.Name, zone.Name, e2ereslist.ToString(baseloadNumaRes)))
+					padPod, err := makePaddingPod(fxt.Namespace.Name, nrtInfo.Name, zone, baseloadNumaRes)
```
fromanirh, line 349: this doesn't seem correct though, seel line 339 above

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -284,65 +291,80 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
+				}
 			}
 
 			// Wait for all the padding pods to be up&running
 			failedPods := e2ewait.ForPodListAllRunning(fxt.Client, paddingPods)
 			for _, failedPod := range failedPods {
 				_ = objects.LogEventsForPod(fxt.K8sClient, failedPod.Namespace, failedPod.Name)
+				//note that this test does not use podOverhead thus pod req and lim would be the pod's resources as set upon creating
+				req, lim := resourcehelper.PodRequestsAndLimits(failedPod)
+				klog.Infof("Resources for pod %q: requests: %s ; limits: %s", failedPod.Name, e2ereslist.ToString(req), e2ereslist.ToString(lim))
```
fromanirh, line 367: this won't hurt, but I recommend to always log namespace + name (not just name)

test/e2e/serial/tests/workload_placement.go
```diff
@@ -545,19 +545,20 @@ func makePaddingPod(namespace, nodeName string, zone nrtv1alpha1.Zone, podReqs c
...
 
-	klog.Infof("padding resource to saturate %q: %s", node.Name, e2ereslist.ToString(paddingReqs))
+// 	klog.Infof("padding resource to saturate %q: %s", node.Name, e2ereslist.ToString(paddingReqs))
 
-	padPod := newPaddingPod(node.Name, "", namespace, paddingReqs)
-	return padPod, nil
-}
+// 	padPod := newPaddingPod(node.Name, "", namespace, paddingReqs)
+// 	return padPod, nil
+// }
```
fromanirh, line 561: please delete commented code

test/e2e/serial/tests/workload_placement.go
```diff
@@ -737,20 +738,34 @@ func matchLogLevelToKlog(cnt *corev1.Container, level operatorv1.LogLevel) (bool
...
+// 		if !ok {
+// 			continue
+// 		}
+
+// 		if zoneQty.Cmp(max) == 1 {
+// 			max = zoneQty
+// 		}
+// 	}
+// 	return max.DeepCopy()
+// }
```
fromanirh, line 755: ditto

test/utils/nodes/baseload.go
fromanirh: ok, so we need to be more careful in the conversion. Fixes to `baseload.go` need to be moved in their own commit (if not PR).
Rounding to HT pairs is meant to prevent any issue with HT alignment in cpumanager. Is not strictly needed, but should help. With modern servers with >= 32 cores per NUMA zone, should be no bother.

fromanirh: all x86_64 cpus on the market or on the horizon all have even number of cores, so if we take care to reserve a even number, everything is nice and consistent. It's probably easier to just round up the baseload to multiples of 2

test/e2e/serial/tests/resource_accounting.go
shajmakh: ok I'll go with the baseload on one numa because splitting may result in fractions of CPU that cannot be reflected in NRT. thanks for clarifying this point!

shajmakh: @fromanirh  can we assume the base load of a node is not greater than the available topology on one numa? asking that to know if I need to tackle a case where the base load is not available on one numa

fromanirh: if the detected baseload seems greater than the available resources on any NUMA zone, then the test should fail immediately, because it's very likely something is (very) wrong

shajmakh:
closing the PR, need more modifications and better to split commits into single PRs

test/e2e/serial/tests/resource_accounting.go
shajmakh: node total allocatable might be good in case we certainly know that there is no pods scheduled on the nodes. I wouldn't prefer setting it to allocatable just to avoid noisy pods leftovers.

PR #227, opened 2022-03-23
add e2e serial config suite

fromanirh:
Move the configuration steps we used to do in the {Before,After}Suite{,Helper} in their own subpackage.

This way, we can keep the current suite behaving like before (bar the necessary mechanical API and import changes) but we
enable better integration with external projects (= cnf-tests)

It goes like this:
cnf-tests want to run the configuration steps in a different suite (configsuite) separated by the test suite proper (e2esuite).

In this model, we will add all the configuration steps in the config suite; in the test suite, we will integrate the tests proper and the minimal configuration support (our own fixture).

### DIFF
--- a/test/e2e/serial/config/fixture.go
+++ b/test/e2e/serial/config/fixture.go
@@ -1,0 +1,72 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"context"
+
+	. "github.com/onsi/gomega"
+
+	"k8s.io/klog/v2"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	nrtv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
+
+	nropv1alpha1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1alpha1"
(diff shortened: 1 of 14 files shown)


test/e2e/serial/config/fixture.go
```diff
@@ -0,0 +1,72 @@
...
+	var cfg E2EConfig
+
+	cfg.Fixture, err = e2efixture.Setup("e2e-test-infra")
+	Expect(err).ToNot(HaveOccurred(), "unable to setup infra test fixture")
+
+	err = cfg.Fixture.Client.List(context.TODO(), &cfg.NRTList)
+	Expect(err).ToNot(HaveOccurred())
+
+	err = cfg.Fixture.Client.Get(context.TODO(), client.ObjectKey{Name: nrosched.NROSchedObjectName}, cfg.NROSchedObj)
+	Expect(err).ToNot(HaveOccurred(), "cannot get %q in the cluster", nrosched.NROSchedObjectName)
```
shajmakh, line 55: do we configure the components or do we assume they are already on the cluster and just verify that before proceeding?

fromanirh: for the purpose of this suite and this subpkg, we assume the operator is already installed, but we need to do some configuration steps. The `install` suite is meant to install the operator.

test/e2e/serial/tests/configuration.go
```diff
@@ -257,11 +259,11 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
 				return true, nil
 			}, 10*time.Minute, 30*time.Second).Should(BeTrue())
 
-			By(fmt.Sprintf("modifing the NUMAResourcesOperator ExporterImage filed to %q", nropTestCIImage))
+			By(fmt.Sprintf("modifing the NUMAResourcesOperator ExporterImage filed to %q", serialconfig.NropTestCIImage))
 			err = fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(nroOperObj), nroOperObj)
```
shajmakh, line 263: can we use serialconfig.Config NROOperObj and NROSchedObj instead?

fromanirh: we can evaluate (in a later PR) but in this case the test wants to alter the config mutating the objects, thus getting a local copy seems safer

shajmakh:
thanks for the patch!
/lgtm

fromanirh:
closed by mistake (`git push -d` instead `git push -f`), let's continue https://github.com/openshift-kni/numaresources-operator/pull/228

PR #228, opened 2022-03-23
e2e: add serial config suite

fromanirh:
Move the configuration steps we used to do in the {Before,After}Suite{,Helper} in their own subpackage.

This way, we can keep the current suite behaving like before (bar the necessary mechanical API and import changes) but we
enable better integration with external projects (= cnf-tests)

It goes like this:
cnf-tests want to run the configuration steps in a different suite (configsuite) separated by the test suite proper (e2esuite).

In this model, we will add all the configuration steps in the config suite; in the test suite, we will integrate the tests proper and the minimal configuration support (our own fixture).

### DIFF
--- a/test/e2e/serial/tests/sched_removal.go
+++ b/test/e2e/serial/tests/sched_removal.go
@@ -33,22 +33,26 @@ import (
 	"github.com/openshift-kni/numaresources-operator/test/utils/nrosched"
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
 	e2ewait "github.com/openshift-kni/numaresources-operator/test/utils/objects/wait"
+
+	serialconfig "github.com/openshift-kni/numaresources-operator/test/e2e/serial/config"
 )
 
 var _ = Describe("[serial][disruptive][scheduler] numaresources scheduler removal on a live cluster", func() {
 	var fxt *e2efixture.Fixture
 
 	BeforeEach(func() {
+		Expect(serialconfig.Config).ToNot(BeNil())
+
 		var err error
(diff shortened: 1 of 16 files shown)


fromanirh:
/retest
```
level=error msg=Error: [ERR]: Error building changeset: InvalidChangeBatch:
```
infra issue

test/e2e/serial/tests/sched_removal.go
```diff
@@ -81,11 +85,11 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources scheduler remova
...
 			var err error
 
-			By(fmt.Sprintf("deleting the NRO Scheduler object: %s", nroSchedObj.Name))
-			err = fxt.Client.Delete(context.TODO(), nroSchedObj)
+			By(fmt.Sprintf("deleting the NRO Scheduler object: %s", serialconfig.Config.NROSchedObj.Name))
+			err = fxt.Client.Delete(context.TODO(), serialconfig.Config.NROSchedObj)
 			Expect(err).ToNot(HaveOccurred())
 
-			dp := createDeployment(fxt, "testdp", schedulerName)
+			dp := createDeployment(fxt, "testdp", serialconfig.Config.SchedulerName)
```
shajmakh, line 93: why not use serialconfig schedobj in line 112 (and all its instances)? I see it is used similarly there as well

fromanirh: this is a good point. At this point in time I'm doing 1:1 replacement, in the future we would want to simplify/reorganize in the future

fromanirh:
trivial rebase after LGTM

PR #229, opened 2022-03-24
deps: update pcidb to 1.0.0, ghw to 0.9.0

fromanirh:
This version of pcidb switched the default: now the automatic
fetch of pcidb from the network is opt-in.


fromanirh:
xref: https://github.com/jaypipes/pcidb/issues/28
note we have caching in place (https://github.com/openshift-kni/numaresources-operator/blob/main/Dockerfile.openshift#L7) which make unlikely for this project to have contributed to the request storm; still, having this change is good per se.

PR #230, opened 2022-03-24
e2e:serial: modify kubeletconfig while TAS is running

Tal-or:
This test ensure that TAS knows how to handle and respond to live kubeletconfig changes while it's already running

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -29,12 +29,16 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/klog/v2"
 
+	"github.com/ghodss/yaml"
 	"github.com/google/go-cmp/cmp"
 	nrtv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
 	nropv1alpha1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1alpha1"
+	"github.com/openshift-kni/numaresources-operator/pkg/kubeletconfig"
 	nropmcp "github.com/openshift-kni/numaresources-operator/pkg/machineconfigpools"
+	rteconfig "github.com/openshift-kni/numaresources-operator/rte/pkg/config"
 	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
 	"github.com/openshift-kni/numaresources-operator/test/utils/configuration"
 	e2efixture "github.com/openshift-kni/numaresources-operator/test/utils/fixture"
@@ -44,6 +48,7 @@ import (
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
 	e2ewait "github.com/openshift-kni/numaresources-operator/test/utils/objects/wait"
 	e2epadder "github.com/openshift-kni/numaresources-operator/test/utils/padder"
+	e2ereslist "github.com/openshift-kni/numaresources-operator/test/utils/resourcelist"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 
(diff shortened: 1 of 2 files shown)


test/e2e/serial/tests/configuration.go
```diff
@@ -343,5 +345,137 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
...
+					if kcMcpSel.Matches(labels.Set(mcp.Labels)) {
+						targetedKCs = append(targetedKCs, kc)
+					}
+				}
+			}
+			Expect(len(targetedKCs)).ToNot(BeZero(), "there should be at least one kubeletconfig.machineconfiguration object")
+
+			By("modifying reserved CPUs under kubeletconfig")
+			initialRsvCPUs := make(map[string]string)
+			for i := 0; i < len(targetedKCs); i++ {
```
fromanirh, line 381: do we want to change all the kubeletconfig we find or just one?
meaning, what's the upside of changing all of them?

test/e2e/serial/tests/configuration.go
```diff
@@ -343,5 +345,137 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
 			Expect(err).ToNot(HaveOccurred())
 			Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s", updatedPod.Namespace, updatedPod.Name, schedulerTestName)
 		})
+
+		It("[test_id:47585][reboot_required] can change kubeletconfig and controller should adapt", func() {
```
fromanirh, line 349: let's also add the `[slow]` tag

test/e2e/serial/tests/configuration.go
```diff
@@ -343,5 +345,137 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
...
+						// the initial and updated NRT object derived from the change in the reserved CPUs under the
+						// kubeletconfig. we don't care from the exact accounting, we just want to make sure that there is
+						// a difference.
+						equal, err := e2enrt.CheckEqualAvailableResources(*nrt, *updatedNrt)
+						if err != nil {
+							klog.Warningf("failed to check equal available resources of NodeResourceTopology %q", updatedNrt.Name)
+							return false, err
+						}
+						if equal == true {
+							klog.Warningf("NodeResourceTopology %q has not been updated after kubelet modification", updatedNrt.Name)
```
fromanirh, line 429: This works but seems a bit too fragile. I wonder if it is more explicit/safer to check the `configmap`.
Or, we can add a new `status` field in the operator to report which version of the `configmap` the operator is using.
Let's think about options.

test/e2e/serial/tests/configuration.go
```diff
@@ -343,5 +345,137 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
...
+			}()
+		})
 	})
 })
+
+// we want to make sure that we're changing the value to be something different from the existing one.
+// since we don't know the current ReservedSystemCPUs values, let's pick something random, make sure it's different and apply it.
+func applyNewReservedSystemCPUsValue(oldRsvCPUs *string) {
+	newRsvCPUs := "3-4"
+	// if it happens to be the same, pick something else
```
fromanirh, line 476: correct, but we should iterate a bit over it and no just once, should'nt we?
also, this deserves a log

fromanirh (commented):
good start here, but let's think about making the test more robust

test/e2e/serial/tests/configuration.go
Tal-or: > we can add a new status field in the operator to report which version of the configmap the operator is using.
Let's think about options.

That sounds like a good suggestion, but there are some open questions:
1. How do we track the version number? 
2. What is the version convention?
3. What happens if the operator gets restarted? or if the configmap gets restarted? (without changes)

test/e2e/serial/tests/configuration.go
Tal-or: > but we should iterate a bit over it and no just once

Not sure I understand. why?

test/e2e/serial/tests/configuration.go
Tal-or: ack

test/e2e/serial/tests/configuration.go
Tal-or: We probably can pick one randomly and wait for the relevant MCPs.
But in this specific setup, I presume there will be only one so it doesn't really matter.

Tal-or:
> /retest

Let's address the open questions from the previous comments first.
There will be some additional iterations anyway

test/e2e/serial/tests/configuration.go
fromanirh: true, when running in our CI. But we want to make preparations for the test to be _possibly_ used as validation tool. So let's not make too many implicit assumptions over the environment, and let's be explicit about test requirements and scenarios

test/e2e/serial/tests/configuration.go
fromanirh: it could be sufficient to expose the hash signature of the configmap the operator is using. This acts as the simplest validation mechanism, and sidesteps all the (real) issues outlined above.

test/e2e/serial/tests/configuration.go
fromanirh: in the code above we try a couple CPUs to reserve. If this fails, we try another set (determined how?) and then we give up.
Why only two attempts? We may want to either iterate more over CPU pairs to use as reserved, or to document why only two attempts is enough.

Tal-or: Each KubeletConfig has a (single) field if the ```ReservedSystemCPUs```
In runtime, we don't know the current value of this field, but actually, we don't care what the actual value is. All we want it to assign a value that is different from the existing one, in order to trigger a machine-config change.

Since we don't care about the value, we just pick something totally random (```newRsvCPUs := "3-4"```) now since we don't know the current value, there are two options here:
1. the current value is equal to the random value we choose
2. the current value is **not** equal to the random value we choose.

In option number 2 we are good to go, but if happened and we land on option number one, it won't trigger a machine-config change (because the value has left the same) so we just pick another random value, which now we are certain that it is different from the existing one.

To conclusion, the maximum attempts is 2.

test/e2e/serial/tests/configuration.go
Tal-or: Actually, it is much easier to look for the new ```SystemCPUsReserved``` value in the ```configmap``` 
But adding the configmap hash as one of the status is a good idea in general

fromanirh: true, but looking into that value in the `configmap` will tell us what the operator is _supposed_ to use, not what is actually using

test/e2e/serial/tests/configuration.go
fromanirh: ok, makes sense even if it still feels clunky somehow. But I don't mind. Please distill this comment into a source code comment, and we can move forward

test/e2e/serial/tests/configuration.go
Tal-or: The same applies for checking the ```configmap``` hash under the ```status```

test/e2e/serial/tests/configuration.go
Tal-or: Is the comment above the function signature isn't enough?

test/e2e/serial/tests/configuration.go
fromanirh: how? `status` report what the operator (/controller) is using, so it reflects somehow its internal state. OTOH we can also dump the content of the configmap in the status, but if we go this way I'd just start with the signature

test/e2e/serial/tests/configuration.go
fromanirh: you can just extend that comment explaining why trying twice is sufficient

test/e2e/serial/tests/configuration.go
Tal-or: The ```configmap``` gets **created** by the operator, then a fortiori is being used by it :)

Tal-or: > but if we go this way I'd just start with the signature

In progress

fromanirh: right, got sidetracked into a tunnel vision. Will re-evaluate this whole subthread. Feel free to pause this work.

fromanirh:
> ```
> Restored cache for golangci-lint from key 'golangci-lint.cache-2725-370749122e24374d4d68aa6b6d90c25bedda7bc7' in 55[79](https://github.com/openshift-kni/numaresources-operator/runs/5735116565?check_suite_focus=true#step:3:79)0ms
> Prepared env in 55790ms
> 
> run golangci-lint
>   Running [/home/runner/golangci-lint-1.41.1-linux-amd64/golangci-lint run --out-format=github-actions --timeout=15m0s --verbose] in [] ...
>   Error: undeclared name: `schedulerName` (typecheck)
>   Error: RandomName not declared by package fixture (typecheck)
>   Error: undeclared name: `schedulerName` (typecheck)
>   Error: undeclared name: `schedulerName` (typecheck)
> ```
> 
> @fromanirh Very strange, in my local environment everything is ok and compiled.

please rebase on top of the main branch and let's see

test/e2e/serial/tests/configuration.go
```diff
@@ -345,5 +350,196 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
...
+
+			var targetedKC *machineconfigv1.KubeletConfig
+			for _, mcp := range mcps {
+				for i := 0; i < len(kcList.Items); i++ {
+					kc := &kcList.Items[i]
+					kcMcpSel, err := metav1.LabelSelectorAsSelector(kc.Spec.MachineConfigPoolSelector)
+					Expect(err).ToNot(HaveOccurred())
+
+					if kcMcpSel.Matches(labels.Set(mcp.Labels)) {
+						// pick the first one you find
```
fromanirh, line 378: so `break` just here I guess?

test/e2e/serial/tests/configuration.go
```diff
@@ -345,5 +350,196 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
...
+			wg.Wait()
+
+			By("checking that NUMAResourcesOperator's ConfigMap has changed")
+			cmList := &corev1.ConfigMapList{}
+			err = fxt.Client.List(context.TODO(), cmList)
+			Expect(err).ToNot(HaveOccurred())
+
+			var nropCm *corev1.ConfigMap
+			for i := 0; i < len(cmList.Items); i++ {
+				if objects.IsOwnedBy(cmList.Items[i].ObjectMeta, nroOperObj.ObjectMeta) {
```
fromanirh, line 417: this likely means we will need to report the `configMap` namespace + name in the operator status, like we do for deployment. Good enough for now, though.

fromanirh (commented):
/lgtm
/approve

PR #233, opened 2022-03-29
hack: deploy MCO CRDs only if NOT on openshift

fromanirh:
we have the tooling to detect the platform, so let's
be more future proof and let's deploy the MCO CRDs
we need only if NOT running on openshift.


Makefile
```diff
@@ -174,7 +174,7 @@ binary-e2e-serial:
 
 binary-e2e-all: binary-e2e-install binary-e2e-rte binary-e2e-sched binary-e2e-uninstall binary-e2e-serial runner-e2e-serial build-pause
 
-runner-e2e-serial: bin/envsubst
+runner-e2e-serial: $(shell pwd)/bin/envsubst
```
fromanirh, line 177: probably not needed now

hack/lsplatform/lsplatform.go
```diff
@@ -0,0 +1,36 @@
...
+
+	if platformName != "" {
+		ret := 1
+		userPlatform, ok := platform.FromString(platformName)
+		if !ok {
+			log.Fatalf("error parsing the user platform: %q", platformName)
+		}
+
+		if userPlatform == clusterPlatform {
+			ret = 0
```
shajmakh, line 31: shouldn't it return 1 if the platforms are matching?

fromanirh: we're following the bash (shell) conventions here.
I conditionals, the unix shell interprets 0 as success, because 0 means success, meaning the command executed without errors: https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html

shajmakh: ok good to know that, thanks for the information!

shajmakh:
thanks for the patch
/lgtm

fromanirh:
/override ci-e2e
well known flake, not relevant for this PR. Let's save some CI time.

PR #234, opened 2022-03-30
serial: fix resource_accounting test padding

shajmakh:
The test was failing to fully pad the nodes due to not correctly accounting for the base load on the nodes. 

fix the test by adjusting the way of padding so the test now accounts for the existed load on the nodes and is relevant for both TM scopes (pod and container).


test/e2e/serial/tests/resource_accounting.go
```diff
@@ -288,65 +294,108 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
 			Expect(err).NotTo(HaveOccurred())
 
-			//get maximum zone CPU and Memory to be sure it wont fit on any zone
-			maxResources := corev1.ResourceList{
-				corev1.ResourceCPU:    maxResourceType(*targetNodeNRTInitial, corev1.ResourceCPU),
-				corev1.ResourceMemory: maxResourceType(*targetNodeNRTInitial, corev1.ResourceMemory),
-			}
+			//calculate base load on the target node
+			baseload, err := nodes.GetLoad(fxt.K8sClient, targetNodeName)
+			ExpectWithOffset(1, err).ToNot(HaveOccurred(), "missing node load info for %q", targetNodeName)
```
fromanirh, line 305: here we should not use `ExpectWithOffset` because this is a top-level test. We should use `WithOffset` inside common helpers.

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -288,65 +294,108 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
+			By(fmt.Sprintf("computed base load: %s", baseload))
```
fromanirh, line 306: `By` is meant to report the partial stages in the tests. We can somehow stretch the definition, but this specific message seems better served using a log

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -288,65 +294,108 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
 
-			// add a min ammount of resources just in case all the zones
-			// are equal so the pod wont fit on any of them
-			excessRes := corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("100m"),
-				corev1.ResourceMemory: resource.MustParse("100M"),
+			//get maximum available node CPU and Memory
+			reqResources := corev1.ResourceList{
+				corev1.ResourceCPU:    availableResourceType(*targetNodeNRTInitial, corev1.ResourceCPU),
```
fromanirh, line 310: but `availableResourceType` is defined later in this series, so this code won't compile :\ Let's reorder the patch sequence to make this compile

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -288,65 +294,108 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
-
-			reqCpu := reqResources[corev1.ResourceCPU]
-			reqCpu.Add(excessRes[corev1.ResourceCPU])
-			reqResources[corev1.ResourceCPU] = reqCpu
 
-			reqMem := reqResources[corev1.ResourceMemory]
-			reqMem.Add(excessRes[corev1.ResourceMemory])
-			reqResources[corev1.ResourceMemory] = reqMem
+			By("prepare the test's pod resources as maximum available resources on the target node with the baselaod deducted")
+			baseload.Deduct(reqResources)
```
fromanirh, line 315: ditto for the `Deduct` function

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -288,65 +294,108 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
-				// remember reqResources = maxReources + excessResources
-				// so we ensure out pod will not be scheduled on any other node.
-				padPod, err := makeNodePaddingPod(fxt.Namespace.Name, *node, maxResources)
-				if errors.Is(err, e2enrt.ErrNotEnoughResources) {
-					klog.Infof("Node %q has not enough resources without padding", nodeName)
-					continue
+				//calculate base load on the node
+				baseload, err := nodes.GetLoad(fxt.K8sClient, nodeName)
+				ExpectWithOffset(1, err).ToNot(HaveOccurred(), "missing node load info for %q", nodeName)
+				By(fmt.Sprintf("computed base load: %s", baseload))
```
fromanirh, line 328: same as before

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -288,65 +294,108 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
+
+				//get nrt info of the node
+				By(fmt.Sprintf("preparing node %q to fit the test case", nodeName))
+				nrtInfo, err := e2enrt.FindFromList(nrtCandidates, nodeName)
+				ExpectWithOffset(1, err).ToNot(HaveOccurred(), "missing NRT info for %q", nodeName)
```
fromanirh, line 333: same as above

fromanirh (commented):
initial review, mostly minor stuff. I need to still grok the allocation logic here

test/e2e/serial/tests/resource_accounting.go
shajmakh: okay good to know that, fixing it

test/e2e/serial/tests/resource_accounting.go
shajmakh: 100% right, fixing

test/e2e/serial/tests/resource_accounting.go
shajmakh: I missed this would cause a compilation error, I thought it is enough to have all the correlated changes in one patch, thanks for highlighting that, fixing it.

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -288,65 +294,108 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
-			reqResources := maxResources.DeepCopy()
-
-			reqCpu := reqResources[corev1.ResourceCPU]
-			reqCpu.Add(excessRes[corev1.ResourceCPU])
-			reqResources[corev1.ResourceCPU] = reqCpu
 
-			reqMem := reqResources[corev1.ResourceMemory]
-			reqMem.Add(excessRes[corev1.ResourceMemory])
-			reqResources[corev1.ResourceMemory] = reqMem
+			By("prepare the test's pod resources as maximum available resources on the target node with the baselaod deducted")
```
fromanirh, line 314: what's the benefit of doing like this (which seems OK) over `e2enrt.SaturateNodeUntilLeft` ?

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -288,65 +294,108 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
+					Expect(err).ToNot(HaveOccurred())
+					paddingPods = append(paddingPods, padPod)
+				}
 			}
 
 			// Wait for all the padding pods to be up&running
 			failedPods := e2ewait.ForPodListAllRunning(fxt.Client, paddingPods)
 			for _, failedPod := range failedPods {
 				_ = objects.LogEventsForPod(fxt.K8sClient, failedPod.Namespace, failedPod.Name)
+				//note that this test does not use podOverhead thus pod req and lim would be the pod's resources as set upon creating
```
fromanirh, line 396: fine from now (included) on

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -288,65 +294,108 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
-				Expect(err).NotTo(HaveOccurred())
+				numaCpu, _ := e2enrt.FindResourceAvailableByName(baseloadZone.Resources, corev1.ResourceCPU.String())
+				numaMem, _ := e2enrt.FindResourceAvailableByName(baseloadZone.Resources, corev1.ResourceMemory.String())
+				availableOnNuma := corev1.ResourceList{
+					corev1.ResourceCPU:    numaCpu,
+					corev1.ResourceMemory: numaMem,
+				}
+
+				for resName, resQty := range availableOnNuma {
+					Expect(resQty.Cmp(baseloadRes[resName])).Should(BeNumerically(">=", 0), "resource %q in base load is %s and is greater than the avialable quantity on zone %q, ", resName, baseloadRes[resName], baseloadZone.Name, resQty)
```
fromanirh, line 352: this is a good check, but feels misplaced. It should be enforced elsewhere. Let's remove it for now.

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -288,65 +294,108 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
+
+				padPod, err = pinPodTo(padPod, nrtInfo.Name, baseloadZone.Name)
+				ExpectWithOffset(1, err).ToNot(HaveOccurred(), "unable to pin pod %q to zone %q", padPod.Name, baseloadZone.Name)
 
-				paddingPods = append(paddingPods, pinnedPadPod)
+				err = fxt.Client.Create(context.TODO(), padPod)
+				Expect(err).ToNot(HaveOccurred())
+				paddingPods = append(paddingPods, padPod)
+
+				//pad each zone with a pod so the test would be valid for both TM scopes: container & pod levels
```
fromanirh, line 369: let's focus on `scope=pod` now and let's make changes all across the suite to support `scope=container` later when it's available.
Also, I fail to see the benefit of the code here (lines 369-389)

test/e2e/serial/tests/resource_accounting.go
shajmakh: good suggestion, I wasn't aware such a function already exists. Yet using `SaturateNodeUntilLeft` is not accurate with its current implementation, not particularly in this context but in general, because it returns the max resources to be padded on a node, but this actually unachievable because one container pod cannot be placed on 2 nodes (since we are in a single-numa-node mode). But I believe `SaturateNodeUntilLeft` can be adjusted by internally using `Saturate**Zone**UntilLeft`. note also that `SaturateNodeUntilLeft` is used by the deprecated makeNodePaddingPod

shajmakh: now I wonder why lint didn't fail on leaving `SaturateNodeUntilLeft` while it is not used elsewhere

test/e2e/serial/tests/resource_accounting.go
shajmakh: readjusted the flow a bit, may you have a look when possible? thanks!

fromanirh:
> /hold cancel

this PR is marked as draft, un-draft it when you think is ready

fromanirh (commented):
ok, good enough! please fix the last pending comment and we can merge

internal/baseload/baseload.go
```diff
@@ -94,6 +94,19 @@ func (nl Load) Apply(res corev1.ResourceList) corev1.ResourceList {
 	return res
 }
 
+// Deduct subtract the current node load from the given ResourceList, mutating
+// the parameter in place and also returning the updated value.
+func (nl Load) Deduct(res corev1.ResourceList) corev1.ResourceList {
```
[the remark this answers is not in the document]
shajmakh: right, thanks for pointing to this! fixed.

shajmakh:
> /lgtm /approve thanks for your work!

thanks for your valuable reviews!

PR #235, opened 2022-03-30
utils: baseload: fix GetLoad() to correctly reflect the node load

shajmakh:
AsInt64 failed to parse cpus millis like 800m, which caused to return 0 value as the CPU consumption on the node.

Use Value() to round up the milli CPUs to complete CPU then continue rounding it up to multiples of 2.

### DIFF
--- a/test/utils/nodes/baseload.go
+++ b/test/utils/nodes/baseload.go
@@ -62,17 +62,16 @@ func GetLoad(k8sCli *kubernetes.Clientset, nodeName string) (Load, error) {
 			}
 		}
 	}
+
 	// get full cpus, and always take even number of CPUs
-	cpuMillis, _ := cpu.AsInt64()
 	// we round the CPU consumption as expressed in millicores (not entire cores)
 	// in order to (try to) avoid bugs related to integer division
 	// int64(2900 / 1000) -> 2 -> roundUp(2, 2) -> 2 (correct, but unexpected!)
 	// OTOH
-	// roundUp(2900, 2000) -> 4000 -> 4000/1000 -> 4 (intended behaviour)
+	// roundUp(2900, 2000) -> 4000 -> 4000/1000 -> 4 (intended behaviour).
+	// Value() round up the millis and roundUp rounds it up to multiples of 2 if needed.
 	// TODO: we can use some testing of the test utilities here (!)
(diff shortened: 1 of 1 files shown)


test/utils/nodes/baseload.go
```diff
@@ -62,18 +63,19 @@ func GetLoad(k8sCli *kubernetes.Clientset, nodeName string) (Load, error) {
...
 	// int64(2900 / 1000) -> 2 -> roundUp(2, 2) -> 2 (correct, but unexpected!)
 	// OTOH
-	// roundUp(2900, 2000) -> 4000 -> 4000/1000 -> 4 (intended behaviour)
+	// roundUp(2900, 2000) -> 4000 -> 4000/1000 -> 4 (intended behaviour).
+	// Value() round up the millis and roundUp rounds it up to multiples of 2 if needed.
 	// TODO: we can use some testing of the test utilities here (!)
-	cpuAmount := roundUp(cpuMillis, 2000)
-	nl.CPU = *resource.NewQuantity(cpuAmount/1000, resource.DecimalSI)
-
+	klog.Infof(fmt.Sprintf("Total resources' requests by pods on node %s: CPU=%s ; mem=%s ", nodeName, cpu, mem))
```
fromanirh, line 75: we're too deep in the flow for this kind of logging, better to move it elsewhere

test/utils/nodes/baseload.go
```diff
@@ -62,18 +63,19 @@ func GetLoad(k8sCli *kubernetes.Clientset, nodeName string) (Load, error) {
...
+	nl.CPU = *resource.NewQuantity(roundUp(cpu.Value(), 2), resource.DecimalSI)
 	mem.RoundUp(resource.Giga)
+	klog.Infof(fmt.Sprintf("Rounding up CPU millis %s equals to %d and memory to %s", cpu, cpu.Value(), mem))
```
fromanirh, line 78: same

fromanirh: there is good content in this PR, but the sad truth is this code should have had tests when it was introduced, because bugs here can cause subtle bugs. The very comment block above should have been unit tests.
So, let me try to make amends: https://github.com/openshift-kni/numaresources-operator/pull/237

PR #236, opened 2022-03-30
WIP: e2e: sched: install better error reporting

fromanirh:
Improving the error reporting when the test fails,
adding the event dump of the relevant object.
The test proper should be equivalent, but we want
to make the troubleshooting easier.


fromanirh:
become part of #250

PR #237, opened 2022-03-30
project: add the internal/ hierarchy

fromanirh:
Our e2e tests are complex beasts which require a good amount of support
code. Some of this support code could in turn use some unit testing
to reduce the testing tuning churn, but it's awkward to have tests
inside tests.
On the other hand, most of this code, albeit useful, feel awkward to
be placed under pkg/ because this is not code we really want to export
and support.

To move forward, we introduce the internal/ hierarchy which sits in
a pretty nice middle ground. We can have in here utility packages which
should be almost pkg/ quality, but which we don't need to support -
the tooling will prevent external users.

See: https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface

We may want to do more code movement from/to internal/ in the future,
and we may want to revisit our pkg/ tree.


fromanirh:
**future** candidate to be moved in `internal/`:
- `pkg/testutils`
- `test/utils/padder`
- ... ?

Tal-or:
/approve
/lgtm

Nice idea, the padder package can move into ```internal/``` as well, but can be addressed in a separate PR together with some unit-tests

internal/resourcelist/resourcelist_test.go
```diff
@@ -0,0 +1,77 @@
...
+package resourcelist
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func TestRound(t *testing.T) {
```
shajmakh, line 26: should be TestToString

fromanirh: Right, copy/paste mistake. Will fix in a future PR

internal/baseload/baseload.go
```diff
@@ -35,21 +32,15 @@ type Load struct {
 	Memory resource.Quantity
 }
 
-func GetLoad(k8sCli *kubernetes.Clientset, nodeName string) (Load, error) {
+func FromPods(nodeName string, pods []corev1.Pod) Load {
```
shajmakh, line 35: if this method is used outside https://github.com/openshift-kni/numaresources-operator/pull/237/files#diff-25c2d12ac74aa576033eb8a0dafef636e5a9a67f5ce312a3a2fa5f8c72a8a8a2R63 , we cannot guarantee that the pods list is only for the node name. maybe we need to filter the pods list to get the pods that are on the nodeName.

shajmakh: an example of misuse would be passing a list of all the pods and specifying a nodeName, then the baseload would be the total cluster baseload and not that node baselaod.

fromanirh: that's true. We should improve in a future PR. That's also one of the reasons this sits in `internal/`, we have some slack space :)

PR #238, opened 2022-03-31
e2e: serial: fix rebase glitch

fromanirh:
this was caused by a forced rushed merge.

PR #239, opened 2022-03-31
e2e: serial: generalize tmscope-agnostic tests

fromanirh:
Some workload placement test do not really depend on the Topology Manager Scope (tmscope).
So we generalize them to gain some more coverage for cheap.


shajmakh:
Thanks for the PR.
/lgtm
I'll consider adding test ids in another PR

fromanirh:
trivially rebased on top of #238

PR #241, opened 2022-04-01
serial: add events reporting for failing pods

shajmakh:
Add events reporting of the failing pods to simplify debugging test failures when needed.


jlojosnegros (approved):
/lgtm

PR #243, opened 2022-04-04
utils, e2e: fix ListPodsByDeployment to get only deployment pods

shajmakh:
The old version of ListPodsByDeployment() was returning a list of all the pods that have the same label as the passed deployment. This is misleading because there might be other deployments that use the same label, which will result in listing all pods of multiple deployments, hence flaking upcoming test validations.
    
Fix the method by getting only that specific deployment's pods via verifying the replicasets.

### DIFF
--- a/test/e2e/sched/install/install_test.go
+++ b/test/e2e/sched/install/install_test.go
@@ -92,7 +92,7 @@ var _ = Describe("[Scheduler] install", func() {
 			}, deploymentCheckTimeout, deploymentCheckPollPeriod).Should(BeTrue(), "Deployment Status not OK")
 
 			By("Check secondary scheduler pod is scheduled on a control-plane node")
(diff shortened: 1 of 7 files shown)


fromanirh:
I see the point here, but:

1. each test MUST run on its own namespace (handled by the test Fixture object). So listing for pods pertaining to deployments in a given namespace is sufficient: this guarantee we don't accidentally include leftovers (which we should NOT have in the first place, but this is another story) and
2. if a test needs to run multiple deployments, it must ensure they use different labels for matching. This cannot be fixed in the utility functions

fromanirh:
please ignore the `CI Go / lint` lane failure, known issue.

shajmakh:
closing this PR, because the actual need is to prevent leaving the cluster with any tests' leftovers.
will have a later PR tacking this

PR #245, opened 2022-04-05
[release-4.10.0][manual] update pcidb 4.10.0

fromanirh:
Consume the pcidb release 1.0.0.
The key feature of pcidb 1.0.0 is it doesn't fetch from the network the missing pci.ids db anymore, the behaviour is now opt-in (and     we do NOT want it!)


fromanirh:
the bot did the trick, let's use #244

PR #242, opened 2022-04-04
e2e: remove opt-in skip in the node-selector workload placement test

fromanirh:
Add the fixes which allow us to always run this test, so removing the opt-in skip.
See individual patches for details.

PR #248, opened 2022-04-05
gh actions: split the CI static check lanes

fromanirh:
First and foremost, we do this to get rid of the annoying cache corruption issue we hit lately.
Second, until we fix the annoying caching issue, we can at least disable only the lint lane.
The end result should be the same, no intended changes in behaviour.


fromanirh:
xref: https://github.com/openshift-kni/numaresources-operator/pull/242#issuecomment-1088493505
xref: https://github.com/golangci/golangci-lint-action#how-to-use

PR #249, opened 2022-04-05
e2etest: TopoManager scope=pod two containers not sched

jlojosnegros:
Given topology-manager-scope=pod
When trying to schedule a pod with two containers and each of them fits on different numa zones
then pod is not scheduled.

### DIFF
--- a/test/e2e/serial/tests/workload_unschedulable.go
+++ b/test/e2e/serial/tests/workload_unschedulable.go
@@ -229,7 +229,6 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 			}
 		})
 	})
-
 	Context("with at least two nodes with two numa zones and enough resources in one numazone", func() {
 		It("[test_id:47592][tier2][unsched] a daemonset with a guaranteed pod resources available on one node/one single numa zone but not in any other node", func() {
 			requiredNUMAZones := 2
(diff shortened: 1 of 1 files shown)


test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -220,4 +220,133 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 			}
 		})
 	})
+
+	Context("with at least one node", func() {
+		It("[test_id:47616][tier2][tmscope:pod] pod with two containers each on one numa zone can NOT be scheduled", func() {
+
+			// Filter by number of numa zones
+			const requiredNUMAZones = 2
```
fromanirh, line 228: This still feels a bit awkward in go. Maybe just
```
By(fmt.Sprintf("filtering available nodes with at least %d NUMA zones", 2))
nrtCandidates := e2enrt.FilterZoneCountEqual(nrtList.Items, 2)
```
(same for the other `const` right below)

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -220,4 +220,133 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+
+			var paddingPods []*corev1.Pod
+			for _, nodeName := range nrtCandidateNames.List() {
+
+				nrtInfo, err := e2enrt.FindFromList(nrtCandidates, nodeName)
+				Expect(err).NotTo(HaveOccurred(), "missing NRT info for %q", nodeName)
+
+				for idx, zone := range nrtInfo.Zones {
+					podName := fmt.Sprintf("padding%s-%d", nodeName, idx)
+					padPod, err := makePaddingPod(fxt.Namespace.Name, podName, zone, paddingRes)
```
fromanirh, line 281: We need to take into account the node base load, or the pod may be rejected for the wrong reason, which is a nasty false positive. Example: https://github.com/openshift-kni/numaresources-operator/pull/242

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -220,4 +220,133 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+			paddingForTagetNode := corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("6"),
+				corev1.ResourceMemory: resource.MustParse("6Gi"),
+			}
+
+			targetNrt, err := e2enrt.FindFromList(nrtCandidates, targetNodeName)
+			Expect(err).NotTo(HaveOccurred(), "missing NRT info for targetNode %q", targetNodeName)
+			for idx, zone := range targetNrt.Zones {
+				podName := fmt.Sprintf("padding%s-%d", targetNodeName, idx)
+				padPod, err := makePaddingPod(fxt.Namespace.Name, podName, zone, paddingForTagetNode)
```
fromanirh, line 305: ditto about taking the node base load into consideration

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -220,4 +220,133 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+			interval := 10 * time.Second
+			By(fmt.Sprintf("Checking pod %q keeps in %q state for at least %v seconds ...", pod.Name, string(corev1.PodPending), interval*3))
+			err = e2ewait.WhileInPodPhase(fxt.Client, pod.Namespace, pod.Name, corev1.PodPending, interval, 3)
+			if err != nil {
+				_ = objects.LogEventsForPod(fxt.K8sClient, pod.Namespace, pod.Name)
+			}
+			Expect(err).ToNot(HaveOccurred())
+
+			By(fmt.Sprintf("checking the pod was scheduled with the topology aware scheduler %q", serialconfig.Config.SchedulerName))
+			isFailed, err := nrosched.CheckPODSchedulingFailed(fxt.K8sClient, pod.Namespace, pod.Name, serialconfig.Config.SchedulerName)
```
fromanirh, line 347: we should probably refine this function to have a stricter check (ofc in another PR)

test/e2e/serial/tests/workload_unschedulable.go
jlojosnegros: what about using 
`requiredNUMAZones := 2` 
to avoid repeating "magic" numbers?

fromanirh: yes, we can do this

test/e2e/serial/tests/workload_unschedulable.go
fromanirh: here: https://github.com/openshift-kni/numaresources-operator/pull/251

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -220,4 +220,133 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+				corev1.ResourceMemory: resource.MustParse("5Gi"),
+			}
+
+			By("filtering available nodes with allocatable resources on at least one NUMA zone that can match request")
+			nrtCandidates = e2enrt.FilterNRTsEachRequestOnADifferentZone(nrtCandidates, requiredResCnt1, requiredResCnt2)
+			if len(nrtCandidates) < neededNodes {
+				Skip(fmt.Sprintf("not enough nodes with NUMA zones each of them can match requests: found %d, needed: %d", len(nrtCandidates), neededNodes))
+			}
+
+			// filter by policy
```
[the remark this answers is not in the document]
jlojosnegros: done

test/e2e/serial/tests/workload_unschedulable.go
jlojosnegros: done

test/e2e/serial/tests/workload_unschedulable.go
jlojosnegros: done

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -229,4 +229,156 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+				corev1.ResourceMemory: resource.MustParse("4Gi"),
+			}
+
+			requiredResCnt2 := corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("5"),
+				corev1.ResourceMemory: resource.MustParse("5Gi"),
+			}
+
+			By("filtering available nodes with allocatable resources on at least one NUMA zone that can match request")
+			nrtCandidates = e2enrt.FilterNRTsEachRequestOnADifferentZone(nrtCandidates, requiredResCnt1, requiredResCnt2)
```
fromanirh, line 269: I'd prefer this helper to be moved in `internal/` (to be decided where) and gain unit test coverage. To move forward with this PR, we can (and probably should) have a super simple local helper which checks both the combinations: cnt1/zone1 and cont2/zone2 vs cnt1/zone2 and cnt2/zone1.

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -229,4 +229,156 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+			// Pad the zones so no one could handle both containers
+			paddingResources := corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("6"),
+				corev1.ResourceMemory: resource.MustParse("6Gi"),
+			}
+			//calculate base load on the target node
+			targetNodeBaseLoad, err := nodes.GetLoad(fxt.K8sClient, targetNodeName)
+			Expect(err).ToNot(HaveOccurred(), "missing node load info for %q", targetNodeName)
+
+			targetNodeBaseLoad.Deduct(paddingResources)
```
fromanirh, line 322: why deduct here?

jlojosnegros: I thought I should take into consideration the resources needed for the node base load, but maybe I should use "deduct" just for one of the numa zones instead of on all of them.

fromanirh: Wait, I didn't mean this is wrong. I just feel this is a bit hard to understand, so if you can clarify a bit the flow this can be enough to move forward (or at least to have a better review next cycle :) )

jlojosnegros: But it is wrong ... 
At this point we know `targetNode` have at least two zones which can accommodate each one at least one of the requirements (containers) but maybe one of the zones has enough room for both requests, so here I want to left `paddingResources` at most of each NUMA zone so each zone can accommodate one request (container) but not both of them in the same zone.

baseLoad is calculated at node level ...so I do not know how should I handle it.
If we can assume that already running pods in the node where scheduled by TAS using numa-aware scheduling those resources should already be reflected into the node NRT, so there is no need to subtract any baseload, right?

fromanirh: ok, concerning the base load. On each worker node we have a bunch of infra containers which are either best effort or burstable, and whose resources are not pinned on NUMA zones. And that's fine per se. The TAS accounting, however, is done considering guaranteed pods, because these are the only pods (actually containers in them, but still) which can get NUMA-pinned resources.

So we know the node has less resources then accounted, but we cannot account them properly to any NUMA zone, because of the awkward contractiction we have above.

There is not a right solution atm. The most conservative approach would be to deduct the base load from _all_ zones, but we don't do that. We usually account all these resources on a random NUMA zone.

jlojosnegros: Ok that's what I thought ... Ok I will deduce it from one of the numa zones ... but if understand it properly, this still could make the test fail due to wrong padding if the zone we deduce the baseload from ends having enough room for the two containers.

jlojosnegros:
rebase to solve conflicts

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -340,4 +339,182 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+				baseloadResources := corev1.ResourceList{
+					corev1.ResourceCPU:    baseload.CPU,
+					corev1.ResourceMemory: baseload.Memory,
+				}
+
+				paddingResources, err := e2enrt.SaturateNodeUntilLeft(*nrtInfo, baseloadResources)
+				Expect(err).ToNot(HaveOccurred(), "could not get padding resources for node %q", nrtInfo.Name)
+
+				for idx, zone := range nrtInfo.Zones {
+					podName := fmt.Sprintf("padding%s-%d", nodeName, idx)
```
fromanirh, line 408: let's not use the `nodeName` to construct pod names, because the node name length alone can exceed the maximum allowed length for pod names (!!)

jlojosnegros: sorry it seems I didn't save the file before commiting ... need more coffee .

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -340,4 +339,182 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+			By(fmt.Sprintf("Checking pod %q keeps in %q state for at least %v seconds ...", pod.Name, string(corev1.PodPending), interval*3))
+			err = e2ewait.WhileInPodPhase(fxt.Client, pod.Namespace, pod.Name, corev1.PodPending, interval, 3)
+			if err != nil {
+				_ = objects.LogEventsForPod(fxt.K8sClient, pod.Namespace, pod.Name)
+			}
+			Expect(err).ToNot(HaveOccurred())
+
+			By(fmt.Sprintf("checking the pod was scheduled with the topology aware scheduler %q", serialconfig.Config.SchedulerName))
+			isFailed, err := nrosched.CheckPODSchedulingFailedForAlignment(fxt.K8sClient, pod.Namespace, pod.Name, serialconfig.Config.SchedulerName)
+			Expect(err).ToNot(HaveOccurred())
```
fromanirh, line 491: let's add a event dump here:
```
if !isFailed {
}
```

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -340,4 +339,182 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+			Expect(err).ToNot(HaveOccurred(), "could not get padding resources for node %q", targetNrt.Name)
+
+			for idx, zone := range targetNrt.Zones {
+				paddingRes := paddingResourcesTargetNodePerZone[zone.Name]
+
+				if idx == 0 { //any zone would do it, we just choose one.
+					targetNodeBaseLoad.Deduct(paddingRes)
+				}
+
+				podName := fmt.Sprintf("padding%s-%d", targetNodeName, idx)
```
fromanirh, line 448: let's avoid node name also here. We can use
```
podName := fmt.Sprintf("padding-tgt-%d", targetNodeName, idx)
```

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -340,4 +339,182 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+			}
+
+			targetNrt, err := e2enrt.FindFromList(nrtCandidates, targetNodeName)
+			Expect(err).NotTo(HaveOccurred(), "missing NRT info for targetNode %q", targetNodeName)
+
+			paddingResourcesTargetNodePerZone, err := e2enrt.SaturateNodeUntilLeft(*targetNrt, paddingResources)
+			Expect(err).ToNot(HaveOccurred(), "could not get padding resources for node %q", targetNrt.Name)
+
+			for idx, zone := range targetNrt.Zones {
+				paddingRes := paddingResourcesTargetNodePerZone[zone.Name]
```
fromanirh, line 442: there actually are multiple valid ways to compute the padding/target resources. And we're still forming our own best practices, so there is a variety showcased in the existing tests. I think the following is clearer and a bit more explicit, but I won't mind alternative (readable :) ) ways as long as they are verified:
```
			By("Padding target node")
			//calculate base load on the target node
			targetNodeBaseLoad, err := nodes.GetLoad(fxt.K8sClient, targetNodeName)
			Expect(err).ToNot(HaveOccurred(), "missing node load info for %q", targetNodeName)

			// Pad the zones so no one could handle both containers
			// so we should left enought resources on each zone to accomodate
			// the "biggest" ( in term of resources) container but not the
			// sum of both
			paddingResources := corev1.ResourceList{
				corev1.ResourceCPU:    // FILLME
				corev1.ResourceMemory: // FILLME
			}

			targetNrt, err := e2enrt.FindFromList(nrtCandidates, targetNodeName)
			Expect(err).NotTo(HaveOccurred(), "missing NRT info for targetNode %q", targetNodeName)

			for idx, zone := range targetNrt.Zones {
				zoneRes := paddingResources.DeepCopy() // extra safety
				if idx == 0 {                          //any zone would do it, we just choose one.
					targetNodeBaseLoad.Apply(zoneRes)
				}

				paddingRes, err := e2enrt.SaturateZoneUntilLeft(zone, zoneRes)
				Expect(err).ToNot(HaveOccurred(), "could not get padding resources for node %q", targetNrt.Name)

				podName := fmt.Sprintf("padding-tgt-%d", targetNodeName, idx)
				padPod := newPaddingPod(targetNodeName, zone.Name, fxt.Namespace.Name, paddingRes)

				padPod, err = pinPodTo(padPod, targetNodeName, zone.Name)
				Expect(err).NotTo(HaveOccurred(), "unable to pin pod %q to zone %q", podName, zone.Name)

				err = fxt.Client.Create(context.TODO(), padPod)
				Expect(err).NotTo(HaveOccurred(), "unable to create pod %q on zone %q", podName, zone.Name)

				paddingPods = append(paddingPods, padPod)
			}
```

fromanirh (commented):
managed to finally review properly the allocation logic

fromanirh:
/lgtm
/approve
thanks for the quick updates!

PR #255, opened 2022-04-07
e2e:serial: unschedulable numa level test

Tal-or:
This test creates a deployment with multiple replicas with a pod that is not fit into a node at the numa-level, meanning that there are enough resources in all numa nodes together, but not enough in each one of them, so all pods` replica scheduling must be failed.

In the next step we're reducing the amount of resources requests by the pod and expect at least some of the replicas to be running.

### DIFF
--- a/test/e2e/serial/tests/workload_unschedulable.go
+++ b/test/e2e/serial/tests/workload_unschedulable.go
@@ -25,6 +25,8 @@ import (
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	nrtv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
 	schedutils "github.com/openshift-kni/numaresources-operator/test/e2e/sched/utils"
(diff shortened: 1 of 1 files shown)


test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -229,4 +231,138 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+			Expect(allFailed).To(BeTrue(), "some pods are running, but we expect all of them to fail")
+
+			By("updating deployment in such way that some pods will fit into NUMA nodes")
+			err = fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(dp), dp)
+			Expect(err).ToNot(HaveOccurred())
+
+			// 6 pods in total (replica is 6)
+			// we should expect 'expectedReadyReplicas' out of 6 pods to be running
+			expectedReadyReplicas := calcExpectedReadyReplicas(numOfnrtCandidates, len(nrts)-numOfnrtCandidates)
+			Expect(err).ToNot(HaveOccurred())
```
fromanirh, line 320: this seems stale. Also, we probably should `klog.Infof` the number of expected replicas

fromanirh:
Had a pass, looks good! let's make sure this works as intended on a multi-numa multi-worker cluster, address the inline comments and we should be able to merge

test/e2e/serial/tests/workload_unschedulable.go
Tal-or: What do you mean by stale? this is a new code

fromanirh: I mean that on line 315 you check for the error not to have occurred (fine). Then we don't write to this value again, and here above we run the same check which is pointless because we didn't write to this value again. Hence this check seems stale (copy/paste mistake?)

Tal-or:
/hold 
Need to verify locally

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -229,4 +231,140 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 			}
 		})
 	})
+
+	// other than the other tests, here we expect all the worker nodes (including none-bm hosts) to be padded
+	Context("with zero suitable nodes", func() {
+		var numOfnrtCandidates int
+		BeforeEach(func() {
```
fromanirh, line 238: I'm not sold about new tests which have exactly one `It` block but a `BeforeEach` block. I think this is backwards: we should extract common code once we have it, not adding it proactively: it's too hard to predict the future and commit into a direction (any direction).
Anyway, I'm not blocking about this either.

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -229,4 +231,140 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+			numOfnrtCandidates = len(nrtCandidates)
+			if numOfnrtCandidates < neededNodes {
+				Skip(fmt.Sprintf("not enough nodes with 2 NUMA Zones: found %d, needed %d", len(nrtCandidates), neededNodes))
+			}
+		})
+		It("[test_id:47615][tier2][unsched] a deployment with multiple guaranteed pods resources that doesn't fit at the NUMA level", func() {
+			By("padding all the nodes")
+			requiredRes := corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("4"),
+				corev1.ResourceMemory: resource.MustParse("4G"),
```
fromanirh, line 253: `G` (= 1000) or `Gi` (= 1024) ?

fromanirh (commented):
thanks for the update, looks good now. let's wait for the verification

test/e2e/serial/tests/workload_unschedulable.go
Tal-or: I thought it affect some test results, hence I changed it to 4G, but it turns out it didn't, I'll change it back.

Tal-or:
/hold cancel
Verified locally

fromanirh:
needs rebase :\

fromanirh:
/lgtm
thanks for the very quick rebase

PR #256, opened 2022-04-08
e2e: serial: remove skip if `E2E_SERIAL_STAGING` variable set

fromanirh:
fix the pod overhead test. This also allows to remove the last `E2E_SERIAL_STAGING` instance.

### DIFF
--- a/test/e2e/serial/tests/workload_overhead.go
+++ b/test/e2e/serial/tests/workload_overhead.go
@@ -135,8 +135,6 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
 			})
 			It("[test_id:47582][tier2] schedule a guaranteed Pod in a single NUMA zone and check overhead is not accounted in NRT", func() {
 
-				skipUnlessEnvVar("E2E_SERIAL_STAGING", "FIXME: NRT filter clashes with the noderesources fit plugin")
-
 				// even if it is not a hard rule, and even if there are a LOT of edge cases, a good starting point is usually
 				// in the ballpark of 5x the base load. We start like this
 				podResources := corev1.ResourceList{
--- a/test/e2e/serial/tests/workload_overhead.go
+++ b/test/e2e/serial/tests/workload_overhead.go
@@ -31,6 +31,7 @@ import (
 
 	nrtv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
 	"github.com/openshift-kni/numaresources-operator/internal/resourcelist"
+	e2ereslist "github.com/openshift-kni/numaresources-operator/internal/resourcelist"
 	schedutils "github.com/openshift-kni/numaresources-operator/test/e2e/sched/utils"
 	e2efixture "github.com/openshift-kni/numaresources-operator/test/utils/fixture"
 	e2enrt "github.com/openshift-kni/numaresources-operator/test/utils/noderesourcetopologies"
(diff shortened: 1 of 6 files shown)


test/e2e/serial/tests/workload_overhead.go
```diff
@@ -147,14 +147,11 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
 
 				// need a zone with resources for overhead, pod and a little bit more to avoid zone saturation
 				zoneRequiredResources := rtClass.Overhead.PodFixed.DeepCopy()
-				zoneRequiredResources.Cpu().Add(*podResources.Cpu())
-				zoneRequiredResources.Cpu().Add(*minRes.Cpu())
-
-				zoneRequiredResources.Memory().Add(*podResources.Memory())
-				zoneRequiredResources.Memory().Add(*minRes.Memory())
+				resourcelist.AddCoreResources(zoneRequiredResources, *podResources.Cpu(), *podResources.Memory())
+				resourcelist.AddCoreResources(zoneRequiredResources, *minRes.Cpu(), *minRes.Memory())
```
shajmakh, line 151: adding minRes as is now will result in a total of CPUs with portions, and due to a gap of not accounting for CPU portions in NRT, the consumed resources will not be reflect in NRT. thus need to round this up, either the after all addition (resourcelist.RoundUp) or nested round up in  AddCoreResources.

fromanirh: good point! let's however not add unrelated logic to `AddCoreResources` and let's just bump the requirements here.

test/e2e/serial/tests/workload_overhead.go
```diff
@@ -225,8 +230,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
 				Expect(err).NotTo(HaveOccurred(), "Unable to get pods from Deployment %q:  %v", deployment.Name, err)
 
 				podResourcesWithOverhead := podResources.DeepCopy()
-				podResourcesWithOverhead.Cpu().Add(*rtClass.Overhead.PodFixed.Cpu())
-				podResourcesWithOverhead.Memory().Add(*rtClass.Overhead.PodFixed.Memory())
+				resourcelist.AddCoreResources(podResourcesWithOverhead, *rtClass.Overhead.PodFixed.Cpu(), *rtClass.Overhead.PodFixed.Memory())
```
shajmakh, line 233: note that the pod should be guaranteed, yet in line 213 it sets only the limits

fromanirh: this is fine: https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed
```
Note: If a Container specifies its own memory limit, but does not specify a memory request, Kubernetes automatically assigns a memory request that matches the limit. Similarly, if a Container specifies its own CPU limit, but does not specify a CPU request, Kubernetes automatically assigns a CPU request that matches the limit. 
```

test/e2e/serial/tests/workload_overhead.go
shajmakh: same for the overhead, it is currently 500m cpus. we want to test this is not reflected in the NRT so better that we have 1 cpu

fromanirh: another good point! Let's round up the value after the addition.

test/e2e/serial/tests/workload_overhead.go
shajmakh: good to know that, thanks for the link!

test/e2e/serial/tests/workload_overhead.go
```diff
@@ -243,9 +251,14 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
 					_, err = e2enrt.CheckZoneConsumedResourcesAtLeast(*nrtInitial, *nrtPostCreate, podResources)
```
shajmakh, line 251: don't we need to check that the output value (here is ignored "_") is not empty because of the same reason that we check it is empty in line 254?

fromanirh: yep, I should have fixed also this. Thanks!

fromanirh: we do, revisited the checks in the last upload

shajmakh:
thanks for the patch
/lgtm

test/e2e/serial/tests/workload_overhead.go
```diff
@@ -240,12 +260,16 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
...
 					Expect(err).ToNot(HaveOccurred())
+					// If the pods are running, and they are because we reached this far, then the resources must have been accounted SOMEWHERE!
+					Expect(match).ToNot(Equal(""), "inconsistent accounting: no resources consumed by deployment running")
 
-					// Here error have to Ocurr because pod overhead resources should not be count
-					_, err = e2enrt.CheckZoneConsumedResourcesAtLeast(*nrtInitial, *nrtPostCreate, podResourcesWithOverhead)
-					Expect(err).To(HaveOccurred())
+					matchWithOverhead, err := e2enrt.CheckZoneConsumedResourcesAtLeast(*nrtInitial, *nrtPostCreate, podResourcesWithOverhead)
+					Expect(err).ToNot(HaveOccurred())
+					// OTOH if we add the overhead no zone is expected to have allocated the EXTRA resources - exactly because the overhead
```
yanirq, line 270: I might loose the acronyms in the comments but up to you

test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -232,7 +232,7 @@ func FilterAnyZoneMatchingResources(nrts []nrtv1alpha1.NodeResourceTopology, req
 	for _, nrt := range nrts {
 		matches := 0
 		for _, zone := range nrt.Zones {
-			klog.Infof(" ----> node %q zone %q provides %s requrst %s", nrt.Name, zone.Name, e2ereslist.ToString(AvailableFromZone(zone)), reqStr)
+			klog.Infof(" ----> node %q zone %q provides %s request %s", nrt.Name, zone.Name, e2ereslist.ToString(AvailableFromZone(zone)), reqStr)
```
yanirq, line 235: arrows is a new format for us to observe things more easily ?   maybe this should be consistent, just a thought, not for this PR.

fromanirh: it was mostly me trying to give a structure to the output, and this somehow stuck, probably because we lack better alternatives.

yanirq:
some minor nits (not even for this PR) but overall looks good to me

PR #262, opened 2022-04-12
e2e_test Daemonset partial available resources

jlojosnegros:
Given a cluster with multiple nodes and multiple numa zones per node
When a Daemonset, with guaranteed pods, ask for resources only available
    on some of the nodes
Then all the daemonset pods are scheduled by Secondary Scheduler but
    only those in the nodes with resources end in  Running state while the
    other keep Pending.

### DIFF
--- a/test/e2e/serial/tests/workload_unschedulable.go
+++ b/test/e2e/serial/tests/workload_unschedulable.go
@@ -229,4 +229,115 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 			}
 		})
 	})
+
+	Context("with at least two nodes with two numa zones and enough resources in one numazone", func() {
+		It("[test_id:47592][tier2][unsched] a daemonset with a guaranteed pod resources available on one node/one single numa zone but not in any other node", func() {
+			requiredNUMAZones := 2
+			By(fmt.Sprintf("filtering available nodes with at least %d NUMA zones", requiredNUMAZones))
+			nrtCandidates := e2enrt.FilterZoneCountEqual(nrts, requiredNUMAZones)
+
+			neededNodes := 2
+			if len(nrtCandidates) < neededNodes {
+				Skip(fmt.Sprintf("not enough nodes with 2 NUMA Zones: found %d, needed %d", len(nrtCandidates), neededNodes))
+			}
+
+			nrtCandidateNames := e2enrt.AccumulateNames(nrtCandidates)
+
+			targetNodeName, ok := nrtCandidateNames.PopAny()
+			Expect(ok).To(BeTrue(), "unable to get targetNodeName")
+
+			//TODO: we should calculate requiredRes from NUMA zones in cluster nodes instead.
+			requiredRes := corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("16"),
+				corev1.ResourceMemory: resource.MustParse("16Gi"),
+			}
+
+			By("Padding non selected nodes")
+			// TODO This should be calculated as 3/4 of requiredRes
+			paddingRes := corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("12"),
+				corev1.ResourceMemory: resource.MustParse("12Gi"),
+			}
+
+			var paddingPods []*corev1.Pod
+			for nodeIdx, nodeName := range nrtCandidateNames.List() {
+
+				nrtInfo, err := e2enrt.FindFromList(nrtCandidates, nodeName)
+				Expect(err).NotTo(HaveOccurred(), "missing NRT info for %q", nodeName)
+
+				baseload, err := nodes.GetLoad(fxt.K8sClient, nodeName)
+				Expect(err).NotTo(HaveOccurred(), "cannot get base load for %q", nodeName)
+
+				for zoneIdx, zone := range nrtInfo.Zones {
+					zoneRes := paddingRes.DeepCopy() // extra safety
+					if zoneIdx == 0 {                // any zone is fine
+						baseload.Apply(zoneRes)
+					}
+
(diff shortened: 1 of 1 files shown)


test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -229,4 +229,118 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 			}
 		})
 	})
+
+	Context("with at least two nodes with two numa zones and enough resources in one numazone", func() {
+		var requiredRes corev1.ResourceList
+		var targetNodeName string
+		BeforeEach(func() {
```
fromanirh, line 236: for this specific case I'd just inline the `BeforeEach` in the `It` block. Or, if we want to keep it, we would need a bit stronger rationale

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -229,4 +229,118 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+
+			neededNodes := 2
+			if len(nrtCandidates) < neededNodes {
+				Skip(fmt.Sprintf("not enough nodes with 2 NUMA Zones: found %d, needed %d", len(nrtCandidates), neededNodes))
+			}
+
+			nrtCandidateNames := e2enrt.AccumulateNames(nrtCandidates)
+
+			var ok bool
+			targetNodeName, ok = nrtCandidateNames.PopAny()
```
fromanirh, line 249: inlining in the `It` block would allow us to simplify the handling of these variables (`targetNodeName`...)

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -229,4 +229,118 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+				baseload, err := nodes.GetLoad(fxt.K8sClient, nodeName)
+				Expect(err).NotTo(HaveOccurred(), "cannot get base load for %q", nodeName)
+
+				for idx, zone := range nrtInfo.Zones {
+					zoneRes := paddingRes.DeepCopy() // extra safety
+					if idx == 0 {                    // any zone is fine
+						baseload.Apply(zoneRes)
+					}
+
+					podName := fmt.Sprintf("padding%s-%d", nodeName, idx)
```
fromanirh, line 280: let's not use `nodeName` when building a pod name (if we do elsewhere, we need to fix in other PRs) because it can easily lead to names longer than the allowed limit, creating alse negatives

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -229,4 +229,118 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+			ds.Spec.Template.Spec.Containers[0].Resources.Limits = requiredRes
+
+			err := fxt.Client.Create(context.TODO(), ds)
+			Expect(err).NotTo(HaveOccurred(), "unable to create deployment %q", ds.Name)
+
+			By(fmt.Sprintf("checking daemonset pods have been scheduled with the topology aware scheduler %q ", serialconfig.Config.SchedulerName))
+			pods, err := schedutils.ListPodsByDaemonset(fxt.Client, *ds)
+			Expect(err).ToNot(HaveOccurred(), "Unable to get pods from daemonset %q:  %v", ds.Name, err)
+
+			//TODO: should not we wait until pods have at least been scheduled? how?
```
fromanirh, line 321: we need to wait for the DaemonSet to be ready. If the tests must succeed, the daemonset must go ready, or we can fail right at that pojt

fromanirh (commented):
initial review

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -229,4 +229,118 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+
+			By(fmt.Sprintf("checking only daemonset pod in targetNode:%q is up and running", targetNodeName))
+			podRunningTimeout := 3 * time.Minute
+			for _, pod := range pods {
+				if pod.Spec.NodeName == targetNodeName {
+					scheduledWithTAS, err := nrosched.CheckPODWasScheduledWith(fxt.K8sClient, pod.Namespace, pod.Name, serialconfig.Config.SchedulerName)
```
fromanirh, line 327: let's add 
if err != nil {
}
```
also here

fromanirh (commented):
the overall logic looks OK, once we have the inline comments handled (and once it's verified on a cluster) we're good to go.

test/e2e/serial/tests/workload_unschedulable.go
jlojosnegros: Yeah, my problem here was that our daemonset will hopefully never be "ready" as pods should be executed just on a subset of the desired nodes. 
Maybe, as we know the number of nodes with enough resources to accommodate daemonset pods, we could wait for a specific number of pods to be "Ready", as a special case for the `ForDaemonSetReady` function we already have in place.

fromanirh: Unless the current approach is flaky, I think it's good enough and we can iterate/improve later. For extra points, we may want to file an issue to track this case.

jlojosnegros: Oook ... sorry somehow I miss your second comment ... 
It could be flaky if the pod running in the node with enough resources reaches Running state faster then the other pods are scheduled into the other nodes and fail due to lack of resources, but if that case is possible (not sure) it seems very unlikely to me.

fromanirh:
/hold cancel
had a chance to test it myself

PR #265, opened 2022-04-12
serial: add tests for scheduling pod with nodeAffinity

shajmakh:
Add new tests checking that deployment pods with nodeAffinity specified are scheduled properly on the feasible single numa node even though the resources are available on node level on two compute nodes.


test/e2e/serial/tests/workload_placement_nodelabel.go
```diff
@@ -82,19 +88,27 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
 	// This is ugly, but automatically computing the values is not straightforward
 	// and will we want to start lean and mean.
-	Context("with two nodes with two NUMA zones", func() {
-		It("[test_id:47598][tier2] should place the pod in the node with available resources in one NUMA zone and fulfilling node selector", func() {
+	Context("with two labeled nodes with two NUMA zones", func() {
+		labelName := "size"
+		labelValueMedium := "medium"
+		labelValueLarge := "large"
+		var targetNodeName, alternativeNodeName string
+		var ok bool
```
fromanirh, line 96: do we really need to make this a context variable?

test/e2e/serial/tests/workload_placement_nodelabel.go
```diff
@@ -111,45 +125,20 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			}
 			nrtCandidateNames := e2enrt.AccumulateNames(nrtCandidates)
 
-			// we need to label two of the candidates nodes
-			// one of them will be the targetNode where we expect the pod to be scheduled
-			// and the other one will not have enough resources on only one numa zone
-			// but will fulfill the node selector filter.
-			targetNodeName, ok := nrtCandidateNames.PopAny()
+			targetNodeName, ok = nrtCandidateNames.PopAny()
```
fromanirh, line 128: what about
```
var ok bool
targetNodeName, ok = nrtCandidateNames.PopAny()
```
?

test/e2e/serial/tests/workload_placement_nodelabel.go
```diff
@@ -111,45 +125,20 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
 			Expect(ok).To(BeTrue(), "cannot select a target node among %#v", nrtCandidateNames.List())
 			By(fmt.Sprintf("selecting target node we expect the pod will be scheduled into: %q", targetNodeName))
 
-			alternativeNodeName, ok := nrtCandidateNames.PopAny()
-			Expect(ok).To(BeTrue(), "cannot select a target node among %#v", nrtCandidateNames.List())
+			alternativeNodeName, ok = nrtCandidateNames.PopAny()
```
fromanirh, line 132: same?

test/e2e/serial/tests/workload_placement_nodelabel.go
```diff
@@ -211,6 +224,193 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+
+			By("Verifing the NRT statistics are updated")
+			targetNodeNRTCurrent, err := e2enrt.FindFromList(nrtCandidates, targetNodeName)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(e2enrt.CheckEqualAvailableResources(*targetNodeNRTInitial, *targetNodeNRTCurrent)).To(BeTrue())
+		})
+
+		Context("label two nodes with different label values but both matching the node affinity of the deployment pod of the test", func() {
+			var err error
+			var unlabelTarget, unlabelAlternative func() error
```
fromanirh, line 236: let's make sure we call these functions if the tests fails for error, and thus exits before we call these functions in the expected place

test/e2e/serial/tests/workload_placement_nodelabel.go
```diff
@@ -211,6 +224,193 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+
+			It("[test_id:47597][tier2] should be able to schedule pod with affinity property requiredDuringSchedulingIgnoredDuringExecution on the available node with feasible numa zone", func() {
+				By("create a deployment with one guaranteed pod with node affinity property: requiredDuringSchedulingIgnoredDuringExecution")
+				deploymentName := "test-dp"
+				var replicas int32 = 1
+
+				podLabels := map[string]string{
+					"test": "test-dp",
+				}
+				nodeSelector := map[string]string{}
```
fromanirh, line 256: this seems unneeded

test/e2e/serial/tests/workload_placement_nodelabel.go
```diff
@@ -211,6 +224,193 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+				deployment := objects.NewTestDeployment(replicas, podLabels, nodeSelector, fxt.Namespace.Name, deploymentName, objects.PauseImage, []string{objects.PauseCommand}, []string{})
+				deployment.Spec.Template.Spec.SchedulerName = serialconfig.Config.SchedulerName
+				deployment.Spec.Template.Spec.Containers[0].Resources.Limits = requiredRes
+				affinity := createNodeAffinityRequiredDuringSchedulingIgnoredDuringExecution(labelName, []string{labelValueLarge, labelValueMedium}, corev1.NodeSelectorOpIn)
+				deployment.Spec.Template.Spec.Affinity = affinity
```
fromanirh, line 261: nit: can we avoid the temporary value?

test/e2e/serial/tests/workload_placement_nodelabel.go
```diff
@@ -211,6 +224,193 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+
+			It("[test_id:49843][tier2] should be able to schedule pod with affinity property prefferdDuringSchedulingIgnoredDuringExecution on the available node with feasible numa zone", func() {
+				By("create a deployment with one guaranteed pod with node affinity property: prefferdDuringSchedulingIgnoredDuringExecution")
+				deploymentName := "test-dp"
+				var replicas int32 = 1
+
+				podLabels := map[string]string{
+					"test": "test-dp",
+				}
+				nodeSelector := map[string]string{}
```
fromanirh, line 316: same

fromanirh (commented):
initial review, looks nice! some minor(-ish) comments inside

fromanirh:
to move forward, let's make sure we unlabel on error path (if it is already the case, let's explain in the comments); after that, let's review the code organization proper, which bite us already. Everything else is minor and not blocking.

test/e2e/serial/tests/workload_placement_nodelabel.go
```diff
@@ -0,0 +1,416 @@
...
+				Eventually(func() bool {
+					updatedDp := &appsv1.Deployment{}
+					err = fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(deployment), updatedDp)
+					Expect(err).ToNot(HaveOccurred())
+					return e2ewait.IsDeploymentComplete(deployment, &updatedDp.Status)
+				}, time.Second*30, time.Second*5).Should(BeTrue(), "deployment %q became unready", deployment.Name)
+
+			})
+
+			It("[test_id:49843][tier2] should be able to schedule pod with affinity property prefferdDuringSchedulingIgnoredDuringExecution on the available node with feasible numa zone", func() {
```
shajmakh, line 308: this test case and 47597 can be part of Describe Table since both are performing same checked but different node affinity configurations.

fromanirh: they can. This will probably require some test code refactoring, though.

test/e2e/serial/tests/workload_placement_nodelabel.go
shajmakh: changed the code a bit so we'll need the temp value to show what affinity property we are working with in each test

test/e2e/serial/tests/workload_placement_nodelabel.go
shajmakh: there is no need, in fact, I just put it up here so that the second declared variable in some assignments can be updated properly without considering it a new var inside the `It` scope. I updated the code as per your suggestion. Thanks for raising this!

test/e2e/serial/tests/workload_placement_nodelabel.go
shajmakh: updated, thanks for highlighting that.

fromanirh (approved):
/lgtm
/hold
thanks for the updates! has this been verified against a cluster? if so, feel free to remove the hold

shajmakh:
> /lgtm /hold thanks for the updates! has this been verified against a cluster? if so, feel free to remove the hold

yes, the tests were tested and they passed as expected

fromanirh:
> > /lgtm /hold thanks for the updates! has this been verified against a cluster? if so, feel free to remove the hold
> 
> yes, the tests were tested and they passed as expected

awesome
/hold cancel

PR #266, opened 2022-04-12
e2e: serial: add hugepages test cases

fromanirh:
we need to have hugepages coverage. We start with a subset of tests which is
1. tier1and
2. already tabular
so we can quickly add these tests and start get moving.

We start with 2Mi hugepages just because they are a bit easier to manage (less risk of memory fragmentation), but some 1i coverage is also needed and should be added in a future PR.


fromanirh:
ping (@shajmakh)

shajmakh:
Thanks for the useful patch!
I'll handle adding the tests' ids in a different PR since we have similar cases in the project.
/lgtm

fromanirh:
trivial rebase

PR #267, opened 2022-04-13
numaresourcesscheduler: respond to owned objects` modifications

Tal-or:
Prior to this change, every external/manual tamper to the generated objects wouldn't be detected by the controller, hence it keeps the objects in an undesired state.

### DIFF
--- a/controllers/numaresourcesscheduler_controller.go
+++ b/controllers/numaresourcesscheduler_controller.go
@@ -23,12 +23,17 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/pkg/errors"
@@ -110,8 +115,22 @@ func (r *NUMAResourcesSchedulerReconciler) Reconcile(ctx context.Context, req ct
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *NUMAResourcesSchedulerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// we want to initiate reconcile loop only on changes under labels, annotations or spec of the object
+	p := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return e.ObjectNew.GetGeneration() != e.ObjectOld.GetGeneration() ||
+				!apiequality.Semantic.DeepEqual(e.ObjectNew.GetLabels(), e.ObjectOld.GetLabels()) ||
+				!apiequality.Semantic.DeepEqual(e.ObjectNew.GetAnnotations(), e.ObjectOld.GetAnnotations())
+		},
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&nrsv1alpha1.NUMAResourcesScheduler{}).
+		Owns(&rbacv1.ClusterRole{}, builder.WithPredicates(p)).
+		Owns(&rbacv1.ClusterRoleBinding{}, builder.WithPredicates(p)).
+		Owns(&corev1.ServiceAccount{}, builder.WithPredicates(p)).
+		Owns(&corev1.ConfigMap{}, builder.WithPredicates(p)).
+		Owns(&appsv1.Deployment{}, builder.WithPredicates(p)).
 		Complete(r)
 }
 
--- a/controllers/numaresourcesscheduler_controller_test.go
+++ b/controllers/numaresourcesscheduler_controller_test.go
@@ -18,6 +18,8 @@ package controllers
 
 import (
 	"context"
+	"k8s.io/klog/v2"
+	"time"
 
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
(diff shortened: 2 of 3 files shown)


controllers/numaresourcesscheduler_controller.go
```diff
@@ -109,8 +114,21 @@ func (r *NUMAResourcesSchedulerReconciler) Reconcile(ctx context.Context, req ct
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *NUMAResourcesSchedulerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// we want to initiate reconcile loop only on change under labels or spec of the object
```
fromanirh, line 117: should we ignore annotations them? why or why not?

fromanirh:
This seems like a good improvement, thanks for starting this work.

controllers/numaresourcesscheduler_controller.go
Tal-or: I thought it might somehow reduce reconciliation loops, but maybe let's stay with the basics and reconcile for every change.

Tal-or: I followed the controller code, there is no reason to act upon other changes because according to the code we won't revert metadata changes.
For ```Labels``` and ```Annotations``` changes, the controller will merge old and new values.
For changes under ```spec``` it will force the desired state.

Tal-or:
I'm happy with the PR, lifting WIP

ffromani (commented):
/lgtm

test/e2e/sched/sched_test.go
```diff
@@ -68,6 +72,88 @@ var _ = Describe("[Scheduler] imageReplacement", func() {
...
+			for i := 0; i < len(cmList.Items); i++ {
+				if e2eobjects.IsOwnedBy(cmList.Items[i].ObjectMeta, nroSchedObj.ObjectMeta) {
+					nroCM = &cmList.Items[i]
+				}
+			}
+
+			initialCM := nroCM.DeepCopy()
+			nroCM.Data["somekey"] = "somevalue"
+
+			err = e2eclient.Client.Update(context.TODO(), nroCM)
```
fromanirh, line 113: update can fail on a live cluster. We either need to use something smarter (patch?) or wrap the (get+)update in a loop

ffromani (commented):
/lgtm cancel

test/e2e/sched/sched_test.go
Tal-or: Why that update would failed?

fromanirh: I don't know the exact reason, but this is something we may happen any time in a cluster (part of how kube works) so robust code should tolerate it

Tal-or: It's a bit worrying that a global wide-used function can fail without any good reason.
Does it mean we should stop using it? what about all the other places and production code we have which contain this function?

Of course, I can wrap it in a loop. But the general statement you raised here makes me question.

fromanirh: Well this is kubernetes operates since day one pretty much. Retrying loop (which is what ultimately a control loop is about) and optimistic concurrency are the cornerstones of the platform. I'm not sure I understand the concern :\

controllers/numaresourcesscheduler_controller_test.go
```diff
@@ -158,5 +158,52 @@ var _ = ginkgo.Describe("Test NUMAResourcesScheduler Reconcile", func() {
...
+
+			initialCM := cm.DeepCopy()
+			cm.Data["somekey"] = "somevalue"
+
+			initialDP := dp.DeepCopy()
+			dp.Spec.Template.Spec.Hostname = "newname"
+			c := corev1.Container{Name: "newcontainer"}
+			dp.Spec.Template.Spec.Containers = append(dp.Spec.Template.Spec.Containers, c)
+
+			err = reconciler.Client.Update(context.TODO(), cm)
```
fromanirh, line 187: this has to be done in a `Eventually` loop

controllers/numaresourcesscheduler_controller_test.go
```diff
@@ -158,5 +158,52 @@ var _ = ginkgo.Describe("Test NUMAResourcesScheduler Reconcile", func() {
...
+
+			err = reconciler.Client.Update(context.TODO(), dp)
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+			key = client.ObjectKeyFromObject(nrs)
+			_, err = reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: key})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+			key = client.ObjectKeyFromObject(cm)
+			err = reconciler.Client.Get(context.TODO(), key, cm)
```
fromanirh, line 198: probably same as above - I'm not sure why we need to retry a `Get` though

test/e2e/sched/sched_test.go
```diff
@@ -68,6 +73,90 @@ var _ = Describe("[Scheduler] imageReplacement", func() {
...
 				return deploy.Spec.Template.Spec.Containers[0].Image == e2eimages.SchedTestImageCI
 			}, time.Minute, time.Second*10).Should(BeTrue())
+
+			By("reverting NROS changes")
+			err = e2eclient.Client.Get(context.TODO(), client.ObjectKeyFromObject(nroSchedObj), nroSchedObj)
+			Expect(err).ToNot(HaveOccurred())
+
+			nroSchedObj.Spec = objects.TestNROScheduler().Spec
+
+			err = e2eclient.Client.Update(context.TODO(), nroSchedObj)
```
fromanirh, line 83: let's do this inside a `Eventually` loop

test/e2e/sched/sched_test.go
```diff
@@ -68,6 +73,90 @@ var _ = Describe("[Scheduler] imageReplacement", func() {
...
+			dp, err := schedutils.GetDeploymentByOwnerReference(nroSchedObj.GetUID())
+			Expect(err).ToNot(HaveOccurred())
+
+			initialDP := dp.DeepCopy()
+
+			dp.Spec.Template.Spec.Hostname = "newhostname"
+			c := objects.NewTestPodPause("", "newcontainer").Spec.Containers[0]
+			dp.Spec.Template.Spec.Containers = append(dp.Spec.Template.Spec.Containers, c)
+
+			err = e2eclient.Client.Update(context.TODO(), dp)
```
fromanirh, line 143: should probably be done inside a `Eventually` loop

ffromani (commented):
we need to wrap updates inside `Eventually` loops, and we're good

controllers/numaresourcesscheduler_controller_test.go
Tal-or: I'm not sure it's necessary, there are plenty of examples in the code of ```Get``` but none of them is in ```Eventually``` loop except ones that are expected to be transitioned.

ffromani (commented):
/lgtm
thanks for the updates

PR #269, opened 2022-04-20
runner: fix render process

fromanirh:
The entry point of the numaresources-operator-tests image is a runner
script derived from `hack/run-test-serial-e2e.sh`. This script can
consume helpers (say from `hack/`), but the entry point must be self
sufficient and contained.

In commit 2fbcd8653b0d6d8878ddd666feacce7096a15e4f we factored all
the code to use the `runcmd` helper. This is a good change per se,
but breaks the entry point isolation requirement.

The process which creates the (allegedly) self-sufficient entry point
from the helper script is called `rendering`.
To move forward, we make the rendering process capable of (a trivial
and simplistic form of) inlining.

Note this make our bundled version of `envsubst` now redundant
Later, we will need to revisit the rendering process from the ground up.
For this reason we don't remove the `envsubst` code just yet.


fromanirh:
/hold
this flow is untested

fromanirh:
/hold cancel
we have now at least some form of testing

shajmakh:
thanks for the patch
/lgtm

PR #276, opened 2022-04-28
e2e: serial: log pullSpec actually used

fromanirh:
We don't ever log the pullSpecs we are about to use,
which makes troubleshooting unnecessarily hard.
Fix this adding the logs.


shajmakh:
/lgtm
thanks for the patch

PR #280, opened 2022-04-29
e2e: serial: support NRT check after scheduling workloads

shajmakh:
TBD


shajmakh:
#282 is the updated patch (better branch name)

PR #282, opened 2022-05-02
serial: add NRT check after scheduling workloads

shajmakh:
Add the missing NRT checks to the tests to verify the NRT changes after scheduling workloads or modifying resources.

(Another PR is planned for supporting this step for after deleting workloads.)

### DIFF
--- a/test/e2e/serial/tests/resource_accounting.go
+++ b/test/e2e/serial/tests/resource_accounting.go
@@ -21,14 +21,15 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/ghodss/yaml"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	nrtv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
 
@@ -128,7 +129,6 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
 
 			candidateNodeNames := e2enrt.AccumulateNames(nrtCandidates)
 			// nodes we have now are all equal for our purposes. Pick one at random
-			// TODO: make sure we can control this randomness using ginkgo seed or any other way
 			targetNodeName, ok := candidateNodeNames.PopAny()
 			Expect(ok).To(BeTrue(), "cannot select a target node among %#v", candidateNodeNames.List())
 			unsuitableNodeNames := candidateNodeNames.List()
@@ -159,9 +159,18 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
 			failedPodIds := e2ewait.ForPaddingPodsRunning(fxt, paddingPods)
 			Expect(failedPodIds).To(BeEmpty(), "some padding pods have failed to run")
 
-			for _, zone := range nrtInfo.Zones {
+			var targetNrtBefore *nrtv1alpha1.NodeResourceTopology
+			var targetNrtListBefore nrtv1alpha1.NodeResourceTopologyList
+			for idx, zone := range nrtInfo.Zones {
+				if idx == len(nrtInfo.Zones)-1 {
+					// store the NRT of the target node before scheduling the last placeholder pod,
+					// later we'll compare this when we delete of of those pods
+					targetNrtListBefore, err := e2enrt.GetUpdated(fxt.Client, nrtList, 1*time.Minute)
+					Expect(err).ToNot(HaveOccurred())
+					targetNrtBefore, err = e2enrt.FindFromList(targetNrtListBefore.Items, targetNodeName)
+					Expect(err).NotTo(HaveOccurred())
+				}
 				By(fmt.Sprintf("making node %q zone %q unsuitable with a placeholder pod", nrtInfo.Name, zone.Name))
-				Expect(err).ToNot(HaveOccurred(), "cannot detect the zone ID from %q", zone.Name)
 				// now put a minimal pod (1 cpu 1Gi) on both zones. Now the target node as whole will still have the
 				// required resources, but no NUMA zone individually will
 				targetedPaddingPod := objects.NewTestPodPause(fxt.Namespace.Name, fmt.Sprintf("tgtpadpod-%s", zone.Name))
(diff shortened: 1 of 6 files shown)


test/e2e/serial/tests/resource_accounting.go
```diff
@@ -128,7 +129,6 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
 
 			candidateNodeNames := e2enrt.AccumulateNames(nrtCandidates)
 			// nodes we have now are all equal for our purposes. Pick one at random
-			// TODO: make sure we can control this randomness using ginkgo seed or any other way
```
fromanirh, line 131: why this comment was removed?

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -261,21 +277,37 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
 		var reqResources corev1.ResourceList
 		var err error
 
 		/*
-			1. choose a target node on which the test's burstable pod will run
-			2. fully pad the non-target nodes
-			3. test step: create a workload with burstable pod
+		 1. choose a target node on which the test's burstable pod will run
+		 2. fully pad the non-target nodes
+		 3. test step: create a workload with burstable pod and check which scheduler took charge and NRT
```
fromanirh, line 310: "and NRT" ? seems the sentence is truncated

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -294,7 +326,9 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
 			Expect(ok).To(BeTrue(), "cannot select a node among %#v", nrtCandidateNames.List())
 			By(fmt.Sprintf("selecting node to schedule the test pod: %q", targetNodeName))
 
-			targetNodeNRTInitial, err = e2enrt.FindFromList(nrtCandidates, targetNodeName)
+			targetNrtListInitial, err = e2enrt.GetUpdated(fxt.Client, nrtList, 1*time.Minute)
```
fromanirh, line 329: why do we need a fresh copy of the data?

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -261,21 +277,37 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
...
+			targetNrtAfter, err := e2enrt.FindFromList(targetNrtListAfter.Items, targetNodeName)
+			Expect(err).NotTo(HaveOccurred())
+
+			dataBefore, err := yaml.Marshal(targetNrtBefore)
+			Expect(err).ToNot(HaveOccurred())
+			dataAfter, err := yaml.Marshal(targetNrtAfter)
+			Expect(err).ToNot(HaveOccurred())
+			match, err := e2enrt.CheckZoneConsumedResourcesAtLeast(*targetNrtBefore, *targetNrtAfter, requiredRes)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(match).ToNot(Equal(""), "inconsistent accounting: no resources consumed by the running pod,\nNRT before test's pod: %s \nNRT after: %s \npod resources: %v", dataBefore, dataAfter, e2ereslist.ToString(requiredRes))
```
fromanirh, line 294: better: `Expect(match).ToNot(BeEmpty())` (here and below)

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -159,9 +159,18 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload resourc
 			failedPodIds := e2ewait.ForPaddingPodsRunning(fxt, paddingPods)
 			Expect(failedPodIds).To(BeEmpty(), "some padding pods have failed to run")
 
-			for _, zone := range nrtInfo.Zones {
+			var targetNrtBefore *nrtv1alpha1.NodeResourceTopology
+			var targetNrtListBefore nrtv1alpha1.NodeResourceTopologyList
+			for idx, zone := range nrtInfo.Zones {
+				if idx == len(nrtInfo.Zones)-1 {
```
fromanirh, line 165: this works. A better approach:

1. select the target node randomly (see note about reproducibility, however) earlier, around line 161
2. save the NRT counters state in this loop when `idx` matches the target node
3. delete the selected node using the pre-prepared `idx` down below around line 254

test/e2e/serial/tests/resource_accounting.go
fromanirh: Additionally, lines 288-294 seems a good candidate for a test helper

fromanirh (commented):
looks reasonnable, some suggestions inside, mostly around clarity.
We should extract helpers carefully, but here it seems we have a real opportunity about creating a new helper to check the NRT expectations.

test/e2e/serial/tests/resource_accounting.go
fromanirh: actually no. This needs to be re-evaluated carefully, perhaps rethinking `CheckZoneConsumedResourcesAtLeast` along the way

fromanirh:
/lgtm
good enough. The minor improvements listed here can be tracked separately with cleanup PRs which we should be doing regularly
https://github.com/openshift-kni/numaresources-operator/pull/282#discussion_r863482007
https://github.com/openshift-kni/numaresources-operator/pull/282#discussion_r863482907
https://github.com/openshift-kni/numaresources-operator/pull/282#discussion_r863485100

test/e2e/serial/tests/resource_accounting.go
shajmakh: because we have the ginkgo seed here:
https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/serial/e2e_serial_test.go#L63

fromanirh: we do, and we cannot control the seed :) unless we do crazy hacks with the machine clock, which we totally don't want to do.

fromanirh: besides, seeding ginkgo repeatedly (which we don't do yet) won't solve the other randomness usage, like iterating over a map like we do in `PopAny`

shajmakh: ok thanks for elaborating, I'll add such comments back, it'll be interesting to achieve this randomness

test/e2e/serial/tests/resource_accounting.go
shajmakh: IIUC and the target node you meant here is the target numa zone (because nrtInfo is the nrt of the target node), this scenario does not fit, because we don't want the selection of the numa to be random especially here, instead, we want it to be the last numa zone to hold the last placeholder pod in order to compare the NRTs before and after deleting that pod. saving NRTs upon scheduling earlier placeholder pods (hence numas) would lead to having an old version of the NRT, that when comparing it with the targetNrtAfter the result won't be accurate.

what can be done additionally is to compare the matching zone of which the NRT was updated (instead of expecting the match to be notEmpty), so I'll support this.

PR #284, opened 2022-05-06
Numaresource Operator Serial Test Improvements

swatisehgal:
This patch is make the test suite more robust.
Rather than using hardcoded scheduler name, we obtain it from
the numaresourcesscheduler CR.

In addition to that, we perform a check to make sure that in case
pod event checks return no events, we print a warning as that could
indicate misconfiguration in the cluster.


test/utils/nrosched/nrosched.go
```diff
@@ -59,6 +57,9 @@ func checkPODEvents(k8sCli *kubernetes.Clientset, podNamespace, podName, schedul
 		klog.ErrorS(err, "cannot get events for pod %s/%s", podNamespace, podName)
 		return false, err
 	}
+	if len(events.Items) == 0 {
+		klog.Warningf("No events received for %s/%s!", podNamespace, podName)
```
fromanirh, line 61: and we should likely `return false` here. Or we may go as far as return `false, fmt.Errorf("no event received for %s/%s", podNamespace, podName)`

test/utils/nrosched/nrosched.go
```diff
@@ -59,6 +57,9 @@ func checkPODEvents(k8sCli *kubernetes.Clientset, podNamespace, podName, schedul
 		klog.ErrorS(err, "cannot get events for pod %s/%s", podNamespace, podName)
 		return false, err
 	}
+	if len(events.Items) == 0 {
```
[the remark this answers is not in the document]
fromanirh: Yes, this is possible if we are looking in the right place, but the cluster didn't generate events yet. For example when we pick the wrong scheduler name. No scheduler will engage -> no events

test/utils/nrosched/nrosched.go
```diff
@@ -119,3 +120,25 @@ func CheckNROSchedulerAvailable(cli client.Client, nroSchedName string) *nropv1a
 func IsSchedulingErrorCannotAlign(msg string) bool {
 	return strings.Contains(msg, ErrorCannotAlignPod)
 }
+
+func GetNROSchedulerName(cli client.Client) string {
```
fromanirh, line 104: I think we have a duplicate of this code, don't we?

test/utils/nrosched/nrosched.go
```diff
@@ -93,42 +93,29 @@ func CheckPODWasScheduledWith(k8sCli *kubernetes.Clientset, podNamespace, podNam
...
-			klog.Warningf("missing conditions in %v", nroSchedObj)
-			return false
-		}
-
-		klog.Infof("condition: %v", cond)
-
-		return cond.Status == metav1.ConditionTrue
-	}, 5*time.Minute, 10*time.Second).Should(BeTrue(), "Scheduler condition did not become available")
-	return nroSchedObj
+func CheckNROSchedulerAvailable(cli client.Client) *nropv1alpha1.NUMAResourcesScheduler {
```
fromanirh, line 96: what's the benefit of removing the explicit scheduler name?

test/utils/nrosched/nrosched.go
```diff
@@ -93,42 +93,29 @@ func CheckPODWasScheduledWith(k8sCli *kubernetes.Clientset, podNamespace, podNam
...
 
 func IsSchedulingErrorCannotAlign(msg string) bool {
 	return strings.Contains(msg, ErrorCannotAlignPod)
 }
 
 func GetNROSchedulerName(cli client.Client) string {
+	return isSchedulerAvailable(cli).Spec.SchedulerName
+}
+
+func isSchedulerAvailable(cli client.Client) *nropv1alpha1.NUMAResourcesScheduler {
```
fromanirh, line 108: but this is not telling if a scheduler (which one?) is available, it is getting the scheduler object data. So the name doesn't seem really fitting.

test/utils/nrosched/nrosched.go
swatisehgal: Ack

test/utils/nrosched/nrosched.go
swatisehgal: Yeah, this certainly needs a change. My initial intention was to return a boolean corresponding to a check if scheduler is available but later decided to encapsulate the common code between CheckNROSchedulerAvailable and GetNROSchedulerName. Anyway, will fix it.

test/utils/nrosched/nrosched.go
swatisehgal: It seemed a bit counter intuitive to me that we are trying to obtain the scheduler object and subsequently the scheduler name (GetNROSchedulerName) by accessing the API but already have a scheduler Name.

test/utils/nrosched/nrosched.go
swatisehgal: Hmm, I might have missed that. Let me look in the code again.

test/utils/nrosched/nrosched.go
fromanirh: true, to some extent. But here we managed to mislead ourselves (!!!). `nroSchedName` is the name of the NUMAResourcesScheduler object, which includes the scheduler name. The name of the object and the name of the scheduler (the latter to be used in podSpec) can very well differ.

swatisehgal:
Thanks for the review @fromanirh and @shajmakh. Moving the PR to WIP as I am working on addressing the review comments.

test/utils/nrosched/nrosched.go
swatisehgal: Ah, okay. I understand. I will make the relevant changes in the next iteration.

swatisehgal:
@fromanirh @shajmakh This PR is ready for another round of review.

shajmakh:
thanks for the changes, looks good to me but let's wait for Francesco's review

test/utils/nrosched/nrosched.go
```diff
@@ -92,10 +92,10 @@ func CheckPODWasScheduledWith(k8sCli *kubernetes.Clientset, podNamespace, podNam
 	return checkPODEvents(k8sCli, podNamespace, podName, schedulerName, isScheduledWith)
 }
 
-func CheckNROSchedulerAvailable(cli client.Client, nroSchedName string) *nropv1alpha1.NUMAResourcesScheduler {
+func CheckNROSchedulerAvailable(cli client.Client, NUMAResourcesSchedObjName string) *nropv1alpha1.NUMAResourcesScheduler {
```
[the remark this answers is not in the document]
swatisehgal: Would you prefer `nroSchedObjName` here? I was trying to capture NUMAResourcesScheduler object name explicitly to prevent anyone confusing it with NUMAResourcesOperator object name.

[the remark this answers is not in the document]
fromanirh: let's try this name for some time and let's see how we feel using it for a while

PR #285, opened 2022-05-09
e2e: serial cleanup 

shajmakh:
Commits: 
- serial: return TODO comment of controlling the randomness selection 
- e2e: minor improvments following PR #282 review comments
- e2e: wrap the api client operations in Eventually()


test/e2e/serial/tests/configuration.go
```diff
@@ -201,8 +203,9 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
 
 				testMcp := objects.TestMCP()
 				By(fmt.Sprintf("deleting mcp: %q", testMcp.Name))
-				err = fxt.Client.Delete(context.TODO(), testMcp)
-				Expect(err).ToNot(HaveOccurred())
+				Eventually(func() error {
+					return fxt.Client.Delete(context.TODO(), testMcp)
```
fromanirh, line 207: same goes for delete:

1. the deletion request is accepted/rejected synchronously
2. then we need to wait for the object (s) to be gone, and if we don't do it, that's the bug we need to solve

test/e2e/serial/tests/configuration.go
```diff
@@ -400,8 +409,9 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
 			err = kubeletconfig.KubeletConfToMCKubeletConf(kcObj, targetedKC)
 			Expect(err).ToNot(HaveOccurred())
 
-			err = fxt.Client.Update(context.TODO(), targetedKC)
-			Expect(err).ToNot(HaveOccurred())
+			Eventually(func() error {
+				return fxt.Client.Update(context.TODO(), targetedKC)
```
fromanirh, line 413: wrapping the update like this can be not enough: we need to make sure we get a new copy of the object we want to update, we apply our changes on top and we sent an update at the end, all wrapped in a retry loop

fromanirh (commented):
thanks for the fixes! few comments inside. I think we can fix the updates outside retry loop incrementally

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -195,7 +197,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 				Expect(err).ToNot(HaveOccurred())
 				Expect(match).ToNot(Equal(""), "inconsistent accounting: no resources consumed by the running pod,\nNRT before test's pod: %s \nNRT after: %s \npod resources: %v", dataBefore, dataAfter, e2ereslist.ToString(requiredRes))
 			},
-			Entry("[test_id:48713][tmscope:cnt] with topology-manager-scope: container",
```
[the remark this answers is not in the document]
shajmakh: no, this is not expected, I'll check maybe I missed a rebase, thanks for raising this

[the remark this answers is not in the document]
shajmakh: fixing it

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -683,12 +683,14 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 
 			By("waiting for some of the pods to be running")
 			dpKey := client.ObjectKeyFromObject(dp)
+			time.Sleep(2 * time.Minute) //DELETE_ME
```
swatisehgal, line 697: Given that we are already waiting 5 minutes for the replicas to be created later, this delay seems unnecessary here. Also, I see a comment DELETE_ME here. Was this was added for testing purposes and the intention was to remove it or there is any other reason behind it that I am missing here?

shajmakh: need to be removed, it was just for debugging purposes, thanks for pointing this, will address it as well

PR #286, opened 2022-05-09
WIP: introduce the scheduler profile -sorted

fromanirh:
TBD: run NRT match *before* NodeResourcesFit


fromanirh:
xref: https://kubernetes.io/docs/reference/scheduling/config/#multiple-profiles

pkg/numaresourcesscheduler/objectstate/sched/sched.go
```diff
@@ -205,13 +207,20 @@ func UpdateSchedulerName(cm *corev1.ConfigMap, name string) error {
...
+		for j := range schedProf.PluginConfig {
+			plugin := &schedProf.PluginConfig[j]
+			if plugin.Name != SchedulerPluginName {
+				continue
 			}
+			schedName := name
+			if schedProf.SchedulerName != nil && strings.HasSuffix(*schedProf.SchedulerName, SchedulerProfileSortedSuffix) {
+				schedName = name + SchedulerProfileSortedSuffix
+			}
+			schedProf.SchedulerName = &schedName
```
yanirq, line 223: SchedulerName is a pointer ?

Tal-or: According to the doc, it will be set to the default scheduler in case it's empty.

Tal-or (commented):
We should keep the ```topo-aware-scheduler``` for backward compatibility, there's no doubt about that.
But If we think that the correct way for this profile is the ```-sorted``` profile, I don't see any reason to have them both.
Let's keep the old name, without the ```-sorted``` suffix, but set the plugins order as the sorted one.

fromanirh:
> We should keep the `topo-aware-scheduler` for backward compatibility, there's no doubt about that. But If we think that the correct way for this profile is the `-sorted` profile, I don't see any reason to have them both. Let's keep the old name, without the `-sorted` suffix, but set the plugins order as the sorted one.

I tend to agree, but we need to carefully evaluate the dependencies - explicit and implicit - between NRT and noderesourcefit. To which extent the former depend on the latter for prefiltering? By switching the order, are we exposing bugs?

fromanirh:
Also, can we claim NRT functionally replaces noderesourcefit, so the latter is actually... redundant?

Tal-or:
I'm not fully aware of the benefits we gain by switching the order,  and how ```noderesourcefit``` works, so I can't tell whether it's redundant or not.

Tal-or:
>  By switching the order, are we exposing bugs?

Even if we do, the solution won't be switching back to the former order, because it has a bug as well right? (otherwise, why do we need this change in the first place)

IMVH, the solution for that is fixing these bugs (when and if they are found) and not keeping both versions.

fromanirh:
Let's explore the other option, aka just enforcing the explicit ordering, on a sibling PR.

fromanirh:
here: https://github.com/openshift-kni/numaresources-operator/pull/293

fromanirh:
https://github.com/openshift-kni/numaresources-operator/pull/293 is a better approach

PR #288, opened 2022-05-10
log: Improve log message to clearly indicate the wait

swatisehgal:
This patch updates a log message to clearly indicate that
we are waiting for the deployment to reach a certain number
of replicas. The previous log message gave an impression
of a failure causing confusion and making the person running
the test suite question why the test suite passed even though
the log messages indicated failures


shajmakh:
Thanks for clarifying that
/lgtm

PR #290, opened 2022-05-16
utils: fix fetching the scheduler name

shajmakh:
Fetch the scheduler name from the scheduler object from under the Status instead of Specs.

### DIFF
--- a/test/utils/nrosched/nrosched.go
+++ b/test/utils/nrosched/nrosched.go
@@ -123,5 +123,5 @@ func IsSchedulingErrorCannotAlign(msg string) bool {
 
 func GetNROSchedulerName(cli client.Client, NUMAResourcesSchedObjName string) string {
 	nroSchedObj := CheckNROSchedulerAvailable(cli, NUMAResourcesSchedObjName)
-	return nroSchedObj.Spec.SchedulerName
+	return nroSchedObj.Status.SchedulerName
 }


test/utils/nrosched/nrosched.go
```diff
@@ -123,5 +123,5 @@ func IsSchedulingErrorCannotAlign(msg string) bool {
 
 func GetNROSchedulerName(cli client.Client, NUMAResourcesSchedObjName string) string {
 	nroSchedObj := CheckNROSchedulerAvailable(cli, NUMAResourcesSchedObjName)
-	return nroSchedObj.Spec.SchedulerName
+	return nroSchedObj.Status.SchedulerName
```
fromanirh, line 126: this is technically correct, but should matter rarely because the status is expected to match the spec. On which case(s) the two fields were different?

fromanirh:
/lgtm
please remove hold when happy

test/utils/nrosched/nrosched.go
fromanirh: right, this is the only case on which they can legitimately differ. Good catch!
It's always more correct (or just plainly correct) because the Spec fields are for the user to tell the cluster the expectations about the state of an object. The Status fields are for the system to tell back the user what the system is able to provide (perhaps temporarily). The expectation, because of the reconciliation model, is that the Status will thus match the Spec, eventually.

shajmakh: ok, i see this is the situation here indeed:
https://github.com/openshift-kni/numaresources-operator/blob/4333f17f9948bce938252920502eb12a7b3ea96f/api/numaresourcesoperator/v1alpha1/numaresourcesscheduler_types.go#L26

swatisehgal:
/lgtm

Thanks for the fix, makes sense to use `Status` instead of `Spec` here.

shajmakh:
thanks folks for the review

PR #293, opened 2022-05-17
scheduler: force order of filters

fromanirh:
In order to clarify and make explicit the interplay between NRT filter and the noderesourcefit filter, we now explicit spell out the filter chain, instead of relying on the implicit default behaviour.

### DIFF
--- a/pkg/numaresourcesscheduler/manifests/yaml/configmap.nrt.yaml
+++ b/pkg/numaresourcesscheduler/manifests/yaml/configmap.nrt.yaml
@@ -13,8 +13,26 @@ data:
       - schedulerName: topo-aware-scheduler
         plugins:
           filter:
+            disabled:
+              - name: '*'
             enabled:
+              - name: NodeUnschedulable
+              - name: NodeName
+              - name: TaintToleration
+              - name: NodeAffinity
+              - name: NodePorts
+              - name: KNIDebug
+              - name: NodeResourcesFit
               - name: NodeResourceTopologyMatch
+              - name: VolumeRestrictions
+              - name: EBSLimits
+              - name: GCEPDLimits
+              - name: NodeVolumeLimits
+              - name: AzureDiskLimits
+              - name: VolumeBinding
+              - name: VolumeZone
+              - name: PodTopologySpread
+              - name: InterPodAffinity
           score:
             enabled:
               - name: NodeResourceTopologyMatch


pkg/numaresourcesscheduler/manifests/yaml/configmap.nrt.yaml
```diff
@@ -22,7 +22,6 @@ data:
               - name: NodeAffinity
               - name: NodePorts
               - name: NodeResourceTopologyMatch
-              - name: NodeResourcesFit
```
fromanirh: no, this is wrong. The overlap is not complete, we need to keep them both.

fromanirh (commented):
/hold

pkg/numaresourcesscheduler/manifests/yaml/configmap.nrt.yaml
swatisehgal: Yeah, we need to have both the plugins enabled.

pkg/numaresourcesscheduler/manifests/yaml/configmap.nrt.yaml
```diff
@@ -13,8 +13,25 @@ data:
...
+            disabled:
+              - name: '*'
             enabled:
+              - name: NodeUnschedulable
+              - name: NodeName
+              - name: TaintToleration
+              - name: NodeAffinity
+              - name: NodePorts
+              - name: NodeResourcesFit
               - name: NodeResourceTopologyMatch
```
Tal-or, line 25: Please keep me honest here but the original proposal (i.e. the sorted one)  was to have ```NodeResourceTopologyMatch``` prior to ```NodeResourcesFit```  is it?

https://github.com/openshift-kni/numaresources-operator/pull/286/commits/c03f5cb2105a0196b364d715169d3616628023e3#diff-18edee9604a942c68e91625bd378355cdaf900fb7f3bb0c9363f01d0981e5b8fR38

fromanirh: correct. I'm changing it intentionally, I'll update the commit message to provide the rationale.

fromanirh:
/hold
let's wait for https://github.com/openshift-kni/scheduler-plugins/pull/39 to be sorted out

PR #300, opened 2022-06-07
WIP: Ensure CPU alignment takes place appropriately

swatisehgal:


fromanirh:
> /cc @fromanirh

thanks, will review shortly

doc/examples/sched.yaml
```diff
@@ -5,3 +5,4 @@ metadata:
 spec:
   imageSpec: "quay.io/openshift-kni/scheduler-plugins:4.11-snapshot"
   logLevel: "Trace"
+  schedulerName: topo-aware-scheduler
```
fromanirh, line 8: what's the benefit of this change?

shajmakh:
@swatisehgal Thank you for the patch! 
was it tested on a cluster?

doc/examples/sched.yaml
swatisehgal: Mostly for the sake of consistency. In the example directory, we have a [pod manifest ](https://github.com/openshift-kni/numaresources-operator/blob/main/doc/examples/pod.numalign.yaml#L6) that is supposed to be scheduled with `topo-aware-scheduler ` so thought it made sense to add the scheduler name here.

swatisehgal:
> @swatisehgal Thank you for the patch! was it tested on a cluster?

Yes, this has been tested on a cluster.

doc/examples/sched.yaml
fromanirh: this value enforce the default already taken from the [configmap](https://github.com/openshift-kni/numaresources-operator/blob/v4.11.0-rc0/pkg/numaresourcesscheduler/manifests/yaml/configmap.nrt.yaml#L13), and the name is available in the [status](https://github.com/openshift-kni/numaresources-operator/blob/v4.11.0-rc0/api/numaresourcesoperator/v1alpha1/numaresourcesscheduler_types.go#L47) for introspection anyway, so I see little value in adding it explicitely here

swatisehgal: Ah I see. Going to remove this commit.

fromanirh:
/lgtm
on second thought, being explicit and a tiny bit more consistent at manifest level won't hurt

swatisehgal:
> on second thought, being explicit and a tiny bit more consistent at manifest level won't hurt

Okay, will add it back :)

fromanirh:
> > on second thought, being explicit and a tiny bit more consistent at manifest level won't hurt
> 
> Okay, will add it back :)

well, not very urgent. Whatever is easier for you now, I'm fine and I'll approve both versions, with and without

doc/examples/sched.yaml
fromanirh: for the next time:
`sched-name: Add scheduler name in the NUMAResourcesScheduler manifest `

this is not adding much to the commit proper. A better commit message could be

```
scheduler: explicit name in the NUMAResourcesScheduler manifest 

the recommended flow for consumers of the operator is to inspect the status of the CR object to learn about the scheduler name of the deployed numa-aware scheduler. However, in the manifests we use an explicit hardcoded name - not discovered at runtime. Hence, for consistency we use the same explicit name in all the manifests.

Note this name is also the same default name, but this is an implementation detail; we aim to improve the consistency among example manifests.
```

PR #302, opened 2022-06-10
build: introduce buildhelper replacing git-semver

fromanirh:
Replace git-semver, which we were using in the build step,
with another tailor-made tool superset of it.
What we gain is: we can now report the git commit from which
the build is done, and we can inject the values from outside,
bypassing the autodetection, to be friendlier with isolated build
systems, which would otherwise fail the autodetection.


fromanirh:
should be cherry-picked to `release-4.10` and `release-4.11`

PR #303, opened 2022-06-13
[release-4.10][manual] fix versioning 4.10

fromanirh:
Replace git-semver, which we were using in the build step, with another tailor-made tool superset of it.
What we gain is: we can now report the git commit from which the build is done, and we can inject the values from outside,
bypassing the autodetection, to be friendlier with isolated build systems, which would otherwise fail the autodetection.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,14 +131,14 @@ binary: build-tools
 	CGO_ENABLED=0 go build \
 		-mod=vendor \
 		-o bin/manager \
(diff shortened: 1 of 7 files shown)


fromanirh:
manual back port of https://github.com/openshift-kni/numaresources-operator/pull/302

fromanirh:
/hold cancel
verified in the isolated build system

marioferh (approved):

PR #305, opened 2022-06-15
RTE: update to consume pods fingerprinting support

fromanirh:
enable pods-fingerprinting by default on RTE, but enable cluster admins to *disable* it with per-node-group granularity.


fromanirh:
/hold
let's see if we can vendor in the RTE e2e tests

fromanirh:
/hold cancel
we import the u/s e2e suites automatically, no extra work needed

fromanirh:
/hold
we need a e2e test (new or changed)

fromanirh:
/hold cancel
we now have a e2e test

pkg/objectupdate/rte/rte.go
```diff
@@ -109,6 +110,20 @@ func DaemonSetHashAnnotation(ds *appsv1.DaemonSet, cmHash string) {
 	template.Annotations[hash.ConfigMapAnnotation] = cmHash
 }
 
+func DaemonSetArgs(ds *appsv1.DaemonSet) error {
```
jlojosnegros, line 113: minor comment: maybe it is because I am a little bit out of context but function name could be a little more expressive.

pkg/objectupdate/rte/rte.go
```diff
@@ -109,6 +110,20 @@ func DaemonSetHashAnnotation(ds *appsv1.DaemonSet, cmHash string) {
...
+	cnt, err := FindContainerByName(&ds.Spec.Template.Spec, MainContainerName)
+	if err != nil {
+		return err
+	}
+	flags := flagcodec.ParseArgvKeyValue(cnt.Args)
+	if flags == nil {
+		return fmt.Errorf("cannot modify the arguments for container %s", cnt.Name)
+	}
+	flags.SetToggle("--pods-fingerprint")
```
swatisehgal, line 122: This would mean that if we have the `pods-fingerprint` flag enabled in the RTE daemonset, it would be disabled. Seems a bit odd to me. Why don't we add an explicit flag for the podfingerprint in the NROP spec that gets passed to RTE?

Also the other way round, if I am a consumer of this with a not very dynamic cluster and want to disable this capability. Does this mean that the way I have to do that is go [here](https://github.com/openshift-kni/numaresources-operator/blob/92957803de8e8180824ae0761ba870850144bb6b/rte/main.go#L123-L124) and change the flags `true`?

fromanirh: I didn't though about possibly disabling the fingerprint, which is anyway a good point. I wonder what could be a possible way to expose this option, which I expect to be left as default (thus enabled) >= 90% of the times

fromanirh: ack, we can add a per-nodegroup disable flag for `pods-fingerprint` and tune the per-nodegroup daemonsets accordingly. I'm not sure we need to add `expose-timing` however. U/S added ( :) ) to make the e2e tests more robust and easier to write, I don't see much value in enabling this flag in our case.

swatisehgal: Also, I noticed that `expose-timing` flag here was not toggled. Was that intentional?

fromanirh: yes, `expose-timing` is supported but not enabled by default.

swatisehgal: Thanks, ignore my last comment. You already answered the question. I think github didn't refresh and show your latest comment.

swatisehgal: > we can add a per-nodegroup disable flag for `pods-fingerprint` and tune the per-nodegroup daemonsets accordingly.
I was actually thinking about enabling/disabling this feature for all the nodegroups but tuning per-nodegroup daemonsets is taking it a step further and better than I thought originally :)

fromanirh:
The PR is reviewable again

fromanirh:
/hold cancel
comments addressed

Tal-or (commented):
Is there any specific case when such a requirement is needed?

Tal-or:
Please ignore the above comment. it was misplaced

api/numaresourcesoperator/v1alpha1/numaresourcesoperator_types.go
```diff
@@ -43,6 +43,8 @@ type NUMAResourcesOperatorSpec struct {
 type NodeGroup struct {
 	// MachineConfigPoolSelector defines label selector for the machine config pool
 	MachineConfigPoolSelector *metav1.LabelSelector `json:"machineConfigPoolSelector,omitempty"`
+	// DisablePodsFingerprinting defines if pod fingerprint should be omitted for the machines belonging to this group
+	DisablePodsFingerprinting *bool `json:"disablePodsFingerprinting,omitempty"`
```
Tal-or, line 47: Is there a specific case where such a requirement is needed?
Having the ability to globally disable the feature isn't enough?

controllers/numaresourcesoperator_controller.go
```diff
@@ -351,6 +351,19 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 	return daemonSetsNName, nil
 }
 
+func daemonsetUpdater(gdm *rtestate.GeneratedDesiredManifest) error {
+	// on kubernetes we can just mount the kubeletconfig (no SCC/Selinux),
+	// so handling the kubeletconfig configmap is not needed at all.
+	// We cannot do this at GetManifests time because we need to mount
+	// a specific configmap for each daemonset, whose nome we know only
```
Tal-or: nit: s/nome/name

controllers/numaresourcesoperator_controller.go
```diff
@@ -603,3 +590,39 @@ func validateMachineConfigLabels(mc client.Object, trees []machineconfigpools.No
 	}
 	return nil
 }
+
+func daemonsetUpdater(gdm *rtestate.GeneratedDesiredManifest) error {
```
Tal-or, line 594: Maybe we can add some logs in this function (which can be enabled only when verbosity is higher) to report if we return earlier than expected in the flow.

Tal-or (commented):
Just some minor comment, but overall LGTM

controllers/numaresourcesoperator_controller.go
fromanirh: that's a nice idea, will add

api/numaresourcesoperator/v1alpha1/numaresourcesoperator_types.go
fromanirh: The unit of cluster partitioning for NROP is the `NodeGroup`. When we started the NROP, we wanted to support (but to some extent to discourage) having non-homogeneous config per cluster, so the worker nodes can be partitioned somehow (e.g. with different TM policies).
Up until now, we had an opinionated setup which unconditionally set the configuration, but now we want to allow to disable this setting - which should be enabled by default.
Disabling per-cluster is surely easier, but it goes against the original design.

fromanirh:
needs https://github.com/openshift-kni/numaresources-operator/pull/315

PR #306, opened 2022-06-15
miscelanous code cleanup - 20220615

fromanirh:
grab bag of cleanups accumulated of the past few weeks. Aggregate in one single PR to reduce the review overhead (hopefully)


Makefile
```diff
@@ -128,31 +128,40 @@ test-install-e2e: build-e2e-all
...
+	LDFLAGS+="-X github.com/openshift-kni/numaresources-operator/pkg/version.gitcommit=$(shell bin/buildhelper commit)"; \
 	CGO_ENABLED=0 go build \
 		-mod=vendor \
 		-o bin/manager \
-		-ldflags "-s -w -X github.com/openshift-kni/numaresources-operator/pkg/version.version=$(shell bin/buildhelper version) -X github.com/openshift-kni/numaresources-operator/pkg/version.gitcommit=$(shell bin/buildhelper commit)" \
+		-ldflags "$$LDFLAGS" \
 		main.go
 
 binary-rte: build-tools
+	LDFLAGS="-s -w "; \
```
yanirq, line 141: if LDFLAGS are always the same I assume we can extract them

fromanirh: we can. But each target will need to modify that value (`+=`), so we will need a local copy anyway. At this point the value of having it the value factored out becomes debatable.

PR #319, opened 2022-07-05
e2e:serial: check scheduling error per topology scope

Tal-or:
When the scheduler plugin failed to align a pod into a node it reports an error.
The error message is varies upon the `TopologyManagerScope` configuration.

We should support a check of the scheduling error for both container and pod scope configurations.
This patch adds the support for the container scope (pod scope support is already exists)

### DIFF
--- a/test/e2e/serial/tests/workload_unschedulable.go
+++ b/test/e2e/serial/tests/workload_unschedulable.go
@@ -49,6 +49,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 	var padder *e2epadder.Padder
 	var nrtList nrtv1alpha1.NodeResourceTopologyList
 	var nrts []nrtv1alpha1.NodeResourceTopology
+	var tmPolicy string
 
 	BeforeEach(func() {
 		Expect(serialconfig.Config).ToNot(BeNil())
@@ -73,6 +74,8 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 		if len(nrts) < 2 {
 			Skip(fmt.Sprintf("not enough nodes with valid policy - found %d", len(nrts)))
 		}
+		// we expect having the same policy across all NRTs
+		tmPolicy = nrts[0].TopologyPolicies[0]
 
 		// Note that this test, being part of "serial", expects NO OTHER POD being scheduled
 		// in between, so we consider this information current and valid when the It()s run.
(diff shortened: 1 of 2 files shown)


test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -239,7 +242,9 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 			podLabels := map[string]string{
 				"test": "test-daemonset",
 			}
-			nodeSelector := map[string]string{}
+			nodeSelector := map[string]string{
+				serialconfig.MultiNUMALabel: "2",
```
fromanirh, line 246: do we need this change in this patch?

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -334,7 +339,9 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
```
fromanirh, line 343: ditto

Tal-or:
Tested locally (only the unschedulable tests).
The tests were passed and not skipped.

test/utils/nrosched/nrosched.go
```diff
@@ -117,8 +119,12 @@ func CheckNROSchedulerAvailable(cli client.Client, NUMAResourcesSchedObjName str
 	return nroSchedObj
 }
 
-func IsSchedulingErrorCannotAlign(msg string) bool {
-	return strings.Contains(msg, ErrorCannotAlignPod)
+func IsSchedulingErrorCannotAlign(msg, policy string) bool {
```
fromanirh, line 122: let's inline this code in `IsSchedulingErrorCannotAlign` - in hindsight, this helper should never have been extracted (let alone made public) in the first place

ffromani (commented):
Nice, and we're almost there!
mostly questions and possible improvement inside

test/e2e/serial/tests/workload_unschedulable.go
Tal-or: Yes. In case there are none multi numa machines in the cluster (usually developer use-case) the daemonset will place a pod in there. 

Usually, the single NUMA machines are not strong enough and they got filtered by the `NodeResourceFit` plugin which in turn leads to test failure because we're looking for the `NodeTopologyMatch` error message and are unable to find it.

fromanirh: Uhm, makes sense but please split this change in its own commit (please copy/paste [a possibly enhanced version of] this explanation there)

Tal-or: Done.

test/utils/nrosched/nrosched.go
Tal-or: Done.

Tal-or:
> this seems legit. I'll check later.

The error or the code?

fromanirh:
so, the error is indeed legit and becomes evident if we rebase on top of the `main` branch. There are multiple way to fix, I don't have strong feelings. Shamelessly went for the easiest:
```
$ git diff
diff --git a/test/e2e/serial/tests/workload_unschedulable.go b/test/e2e/serial/tests/workload_unschedulable.go
index e489b919..39486671 100644
--- a/test/e2e/serial/tests/workload_unschedulable.go
+++ b/test/e2e/serial/tests/workload_unschedulable.go
@@ -384,7 +384,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
                                        Expect(err).ToNot(HaveOccurred(), "unable to get pod %s/%s to be Running after %v", pod.Namespace, pod.Name, podRunningTimeout)
 
                                } else {
-                                       isFailed, err := nrosched.CheckPODSchedulingFailedForAlignment(fxt.K8sClient, pod.Namespace, pod.Name, serialconfig.Config.SchedulerName, tmPolicy)
+                                       isFailed, err := nrosched.CheckPODSchedulingFailedForAlignment(fxt.K8sClient, pod.Namespace, pod.Name, serialconfig.Config.SchedulerName, string(tmPolicy))
                                        if err != nil {
                                        }

```

ffromani (commented):
/lgtm
/approve

PR #321, opened 2022-07-05
e2e: serial: tests for non-guaranteed pods

fromanirh:
verify that once a best-effort or burstable pod is scheduled with the NUMA-aware scheduler, no NRT resources are consumed.
This is again because the NRT only accounts exclusively allocated resources - to guaranteed pods.


swatisehgal:
Thanks for this PR!
/lgtm

PR #323, opened 2022-07-06
DNM: WIP: deflake 4.11

fromanirh:
Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #324, opened 2022-07-08
e2e fixes and cleanups

fromanirh:
Another grab-bag of mixed improvements and cleanups emerged during the deflaking attempt tracked inhttps://github.com/openshift-kni/numaresources-operator/pull/323. The real issue seems to be a selinux policy misalignment which will be tracked and fixed separately, but I believe there's work worth preserving in what we did in #323, so we go.


shajmakh:
thanks for the valuable improvements.
/lgtm

PR #325, opened 2022-07-11
e2e:serial: deflake podscope failures

Tal-or:
A couple of improvements for the serial e2e tests.
Some are intended to solve actual failures and some are just best-effort tuning.

The tests were performed on a cluster with nodes configured with a pod scope policy.

### DIFF
--- a/test/e2e/serial/tests/sched_removal.go
+++ b/test/e2e/serial/tests/sched_removal.go
@@ -25,7 +25,9 @@ import (
 	. "github.com/onsi/gomega"
 
 	appsv1 "k8s.io/api/apps/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	nropv1alpha1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1alpha1"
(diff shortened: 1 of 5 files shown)


test/e2e/serial/tests/sched_removal.go
```diff
@@ -137,10 +139,29 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources scheduler restar
 		It("[case:1][test_id:48069][tier2] should schedule any pending workloads submitted while the scheduler was unavailable", func() {
 			var err error
 
+			dps, err := getDPOwnedBy(fxt.Client, nroSchedObj.ObjectMeta)
```
fromanirh, line 142: this is fine. We can also use the scheduler object status field.

test/e2e/serial/tests/sched_removal.go
```diff
@@ -137,10 +139,29 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources scheduler restar
...
 			By(fmt.Sprintf("deleting the NRO Scheduler object: %s", nroSchedObj.Name))
 			err = fxt.Client.Delete(context.TODO(), nroSchedObj)
 			Expect(err).ToNot(HaveOccurred())
 
+			// make sure scheduler deployment is gone
+			Eventually(func() bool {
+				for _, dp := range dps {
+					if err := fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(dp), dp); err != nil {
+						if apierrors.IsNotFound(err) {
+							return true
```
fromanirh, line 154: it seems to me this means we bail out as soon as the first deployment is gone - but we should actually wait for the _last_ to be gone (meaning all of them are gone), right?

ffromani (approved):
/lgtm
all the changes look reasonnable to me

test/e2e/serial/tests/sched_removal.go
Tal-or: We should only have a single deployment for the scheduler.

We can actually assert there is only one and failed the test if it's different.

fromanirh: I'd prefer this way. `getDPOwnedBy` should either return 1 Deployment (success) or nil (on error).

fromanirh:
plase rebase on top of main branch

Tal-or:
/retest
looks like the ci stuck

PR #326, opened 2022-07-11
[WIP] Ginkgov2 upgrade

jlojosnegros:
It seems [Ginkgo v1 has a known problem with cleanup on interruptions](https://github.com/onsi/ginkgo/issues/222#issuecomment-183487276) that is not gonna be solved in v1 but [it is already solved in v2](https://github.com/onsi/ginkgo/issues/222#issuecomment-812966713) ([more info](https://onsi.github.io/ginkgo/MIGRATING_TO_V2#interrupt-behavior))

To be able to cleanup properly upon interruption we should first migrate to Ginkgo v2.

This is currently a work in progress.


jlojosnegros:
/hold

Just testing if all CI lanes works properly

PR #327, opened 2022-07-12
e2e: serial: placement testing of a deployment with 2 replicas

swatisehgal:
This test involves modification (reduction) of requested resources
where resource requests are available from 2 NUMA nodes and verification
that the NRT instances are updated correctly.

### DIFF
--- a/test/e2e/serial/tests/workload_placement.go
+++ b/test/e2e/serial/tests/workload_placement.go
@@ -31,10 +31,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 
 	nrtv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
 
+	"github.com/openshift-kni/numaresources-operator/internal/baseload"
 	e2ereslist "github.com/openshift-kni/numaresources-operator/internal/resourcelist"
 
 	"github.com/openshift-kni/numaresources-operator/pkg/flagcodec"
@@ -51,6 +53,7 @@ import (
 	e2ewait "github.com/openshift-kni/numaresources-operator/test/utils/objects/wait"
 	e2epadder "github.com/openshift-kni/numaresources-operator/test/utils/padder"
 	operatorv1 "github.com/openshift/api/operator/v1"
+	appsv1 "k8s.io/api/apps/v1"
 )
 
 var _ = Describe("[serial][disruptive][scheduler] numaresources workload placement", func() {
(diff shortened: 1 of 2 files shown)


test/e2e/serial/tests/workload_placement.go
```diff
@@ -413,6 +413,238 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+			// However, since deployment controller tries to strictly spread its replicas(pods) across multiple nodes, there is no way to override
+			// this behaviour and place two replicas that are part of a deployment on the same node. Even after padding the second node,
+			// the second replica is left in a pending state waiting for resources to be made available on the second node. Because of this, this test
+			// is downsized and we create a deployment with single replica and two containers to achieve the same behaviour.
+
+			hostsRequired := 2
+			paddedNodes := padder.GetPaddedNodes()
+			paddedNodesSet := sets.NewString(paddedNodes...)
+
+			nrtInitialList, err := e2enrt.GetUpdated(fxt.Client, nrtv1alpha1.NodeResourceTopologyList{}, time.Second*10)
```
fromanirh, line 427: minor: to list NRTs without any extra constraints, just use
```
err = fxt.Client.List(context.TODO(), &nrtList)
```
(yes, this means elsewhere in the code we do it wrongly)

test/e2e/serial/tests/workload_placement.go
```diff
@@ -413,6 +413,238 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+
+			targetNodeName, ok := targetNodeNameSet.PopAny()
+			Expect(ok).To(BeTrue())
+
+			targetNrtInitial, err := e2enrt.FindFromList(nrtInitialList.Items, targetNodeName)
+			Expect(err).NotTo(HaveOccurred())
+
+			var replicas int32 = 1
+			podLabels := map[string]string{
+				"test": "test-dp",
```
fromanirh, line 454: minor: in the future we will need to make sure each test use unique labels

test/e2e/serial/tests/workload_placement.go
```diff
@@ -413,6 +413,238 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+			}
+			nodeSelector := map[string]string{}
+
+			// We want a container to occupy as much resources from a single NUMA nodes as possible in order to prevent another
+			// container to be allocated resources from the same NUMA node. To determine the value of resources, we use the
+			// resource availablity of a NUMA node that has the least amount of resources out of all the NUMA nodes on that
+			// node and request that in the test-deployment.
+			cpus := leastAvailableResourceQtyInAllZone(*targetNrtInitial, corev1.ResourceCPU)
+			mem := leastAvailableResourceQtyInAllZone(*targetNrtInitial, corev1.ResourceMemory)
```
fromanirh, line 463: do we take into account the baseload here (should we?)

fromanirh (commented):
/lgtm
minor improvements inside but not enough to block this PR

test/e2e/serial/tests/workload_placement.go
swatisehgal: Ack, I will update.

test/e2e/serial/tests/workload_placement.go
swatisehgal: Sure, I will fix this.

test/e2e/serial/tests/workload_placement.go
swatisehgal: Yep, had some time to rethink this. It makes sense to consider the baseload here.

test/e2e/serial/tests/workload_placement.go
```diff
@@ -413,6 +415,305 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+			Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s", updatedPod1.Namespace, updatedPod1.Name, serialconfig.Config.SchedulerName)
+
+			rl1 := e2ereslist.FromGuaranteedPod(updatedPod1)
+			klog.Infof("post-create pod resource list: spec=[%s] updated=[%s]", e2ereslist.ToString(e2ereslist.FromContainers(podSpec.Containers)), e2ereslist.ToString(rl1))
+
+			nrtInitial1, err := e2enrt.FindFromList(nrtInitialList.Items, updatedPod1.Spec.NodeName)
+			Expect(err).ToNot(HaveOccurred())
+
+			nrtPostCreate1, err := e2enrt.FindFromList(nrtPostCreateDeploymentList.Items, updatedPod1.Spec.NodeName)
+			Expect(err).ToNot(HaveOccurred())
```
shajmakh, line 540: can we avoid duplicating the same checks for both replicas?

test/e2e/serial/tests/workload_placement.go
```diff
@@ -413,6 +415,305 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+
+			var checkConsumedRes checkConsumedResFunc
+			if nrtInitial0.TopologyPolicies[0] == string(nrtv1alpha1.SingleNUMANodePodLevel) {
+				// TODO: this is only partially correct. We should check with NUMA zone granularity (not with NODE granularity)
+				checkConsumedRes = e2enrt.CheckZoneConsumedResourcesAtLeast
+			} else {
+				checkConsumedRes = e2enrt.CheckNodeConsumedResourcesAtLeast
```
shajmakh, line 547: just to clarify: so if we are dealing with scope=container we check the consumedres on node level?

test/e2e/serial/tests/workload_placement.go
```diff
@@ -413,6 +415,305 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+			By(fmt.Sprintf("checking post-create NRT after pod: %q for target node %q updated correctly", updatedPod0.Name, targetNodeName))
+			dataBefore0, err := yaml.Marshal(nrtInitial0)
+			Expect(err).ToNot(HaveOccurred())
+			dataAfter0, err := yaml.Marshal(nrtPostCreate0)
+			Expect(err).ToNot(HaveOccurred())
+			match0, err := checkConsumedRes(*nrtInitial0, *nrtPostCreate0, rl0)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(match0).ToNot(BeEmpty(), "inconsistent accounting: no resources consumed by the running pod,\nNRT before test's pod: %s \nNRT after: %s \n total required resources: %s", dataBefore0, dataAfter0, e2ereslist.ToString(rl0))
+
+			if nrtInitial1.TopologyPolicies[0] == string(nrtv1alpha1.SingleNUMANodePodLevel) {
```
shajmakh, line 559: since both replicas will be on the same node, I am not confident this (& prev nrt check) will pass, because updatedPod1.Spec.NodeName & updatedPod0.Spec.NodeName points to the same node, so they'll have same nrts post deployment creation

test/e2e/serial/tests/workload_placement.go
```diff
@@ -413,6 +415,305 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+			Eventually(func() bool {
+				By(fmt.Sprintf("checking the resources are restored as expected on %q", updatedPod0.Spec.NodeName))
+
+				nrtListPostDelete, err := e2enrt.GetUpdated(fxt.Client, nrtPostUpdateDeploymentList, 1*time.Minute)
+				Expect(err).ToNot(HaveOccurred())
+
+				nrtPostDelete0, err := e2enrt.FindFromList(nrtListPostDelete.Items, updatedPod0.Spec.NodeName)
+				Expect(err).ToNot(HaveOccurred())
+
+				ok, err := e2enrt.CheckEqualAvailableResources(*nrtPostUpdate0, *nrtPostDelete0)
```
shajmakh, line 697: don't we want to check equality nrtPostDelete0 vs nrtInitial0? because nrtPostUpdate0 is nrt after applying the deployment update

test/e2e/serial/tests/workload_placement.go
```diff
@@ -413,6 +415,305 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+				nrtListPostDelete, err := e2enrt.GetUpdated(fxt.Client, nrtPostUpdateDeploymentList, 1*time.Minute)
+				Expect(err).ToNot(HaveOccurred())
+
+				nrtPostDelete1, err := e2enrt.FindFromList(nrtListPostDelete.Items, updatedPod1.Spec.NodeName)
+				Expect(err).ToNot(HaveOccurred())
+
+				ok, err := e2enrt.CheckEqualAvailableResources(*nrtPostUpdate1, *nrtPostDelete1)
+				Expect(err).ToNot(HaveOccurred())
+				return ok
+			}, time.Minute, time.Second*5).Should(BeTrue(), "resources not restored on %q", updatedPod1.Spec.NodeName)
```
shajmakh, line 714: also here, I think one-time check is enough because  nrtPostDelete1 & nrtPostDelete0 points to same target node.

test/e2e/serial/tests/workload_placement.go
```diff
@@ -600,6 +901,32 @@ func availableResourceType(nrtInfo nrtv1alpha1.NodeResourceTopology, resName cor
...
+
+		case corev1.ResourceMemory:
+			zoneQty.Sub(baseload.Memory)
+		}
+		if zoneQty.IsZero() || zoneQty.Cmp(zeroVal) < 0 {
+			zoneQty = res
+		}
+		res = zoneQty.DeepCopy()
+	}
+	return res
```
shajmakh, line 927: the name confuses me, can you tell please what's the goal of this function?

test/e2e/serial/tests/workload_placement.go
swatisehgal: Yep, that is correct. I will fix. Rather than checking after each replica we need to check once for the deployment as a whole.

test/e2e/serial/tests/workload_placement.go
swatisehgal: Yep, I think this was a copy paste mistake. Thanks for identifying this.

test/e2e/serial/tests/workload_placement.go
swatisehgal: Sure, I am open to changing the name to something more descriptive. I added a comment to explain the goal of the function and refactored it as well.
Just to summarize: we are trying to admit both the replicas of deployment on both NUMA nodes and need to determine the least available resource out of both the NUMA nodes. Here is what mean with an example:
Assuming resources after deducting baseload are:
NUMA 0: cpu: 2 mem: 5Gi
NUMA 1: cpu: 5 mem 3 Gi

When we specify the resource request for the deployment we request least resource across all zones for each resource, meaning we would like to request cpu=2 and mem=3 Gi. This is how we can be 100% sure that the resource request would be fulfilled for both the replicas on the their corresponding NUMA nodes.
Hope this helps.

test/e2e/serial/tests/workload_placement.go
swatisehgal: When we are dealing with scope=container, the containers that are part of a pod can be on any zone on that node so we have to check for resources consumed on the entire node when comparing initial and post create NRT objects. However for scope=pod, all the containers of that pod would be allocated resources from the same zone so we need to check for resources consumed at a zone level.

Hope this helps. Let me know if you need further clarification.

shajmakh: got it, thanks for the explanation

swatisehgal:
/hold
Haven't tested this locally.

TODOs

1. Need to rebase.
2. leastAvailableResourceQtyInAllZone logic needs to be doublechecked and might need some tweaking.
3. expose baseload CPU and memory through getters.

swatisehgal:
/unhold
This has been tested locally and passed.

Also, all the TODOs captured here: https://github.com/openshift-kni/numaresources-operator/pull/327#issuecomment-1199116955 have been addressed.

test/e2e/serial/tests/workload_placement.go
```diff
@@ -616,6 +619,209 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+			// resource availablity of a NUMA node that has the least amount of resources out of all the NUMA nodes on that
+			// node and request that in the test-deployment.
+			reqResources := corev1.ResourceList{
+				corev1.ResourceCPU:    cpus,
+				corev1.ResourceMemory: mem,
+			}
+
+			nodeSelector := map[string]string{
+				serialconfig.MultiNUMALabel: "2",
+			}
```
shajmakh, line 713: where was the target node labeled with this label?

fromanirh: the intent of the label (and nodeSelector) is to further restrict the worker node pool to nodes with exactly 2 NUMA cells. We should probably revisit this approach someday in the far future.

shajmakh: but we must leave the freedom for the scheduler to decide that, otherwise we are "helping" it with the filtering process

fromanirh: is not ideal indeed. This was introduced because we had clusters with workers with mixed NUMA nodes (e.g. 2 NUMA + 1 NUMA in the same pool) and, to a lesser extent, because most of our logic **in the tests** expects 2 NUMA nodes and can't work with 4 NUMA nodes or more.

On the other hand, we're forcing the scheduler to pick a node _among_ all the available nodes with 2 NUMA zones, which isn't too bad - the tests are still meaningful.

We should get rid of all the instances of `nodeSelector` someday in the future and be more restrictive in the test setup, yes.

swatisehgal: This was exactly the case for the cluster where I tested this and narrowing down to nodes that had 2 NUMAs made sense. Also in the context of this test, we want the two replicas to be placed on two NUMA nodes of the node so having nodes that don't have multiple NUMA nodes did not make much sense.

shajmakh: fair enough, thank you for elaborating.

test/e2e/serial/tests/workload_placement.go
```diff
@@ -802,10 +1008,78 @@ func allocatableResourceType(nrtInfo nrtv1alpha1.NodeResourceTopology, resName c
...
+
 func matchLogLevelToKlog(cnt *corev1.Container, level operatorv1.LogLevel) (bool, bool) {
 	rteFlags := flagcodec.ParseArgvKeyValue(cnt.Args)
 	kLvl := loglevel.ToKlog(level)
 
 	val, found := rteFlags.GetFlag("--v")
 	return found, val.Data == kLvl.String()
 }
+
+func checkReplica(pod corev1.Pod, targetNodeName string, K8sClient *kubernetes.Clientset, podSpec *corev1.PodSpec) {
```
fromanirh, line 1112: `podSpec` is unused (and should not probably be passed by pointer) but we can fix later.

test/e2e/serial/tests/workload_placement.go
```diff
@@ -616,6 +619,209 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+			cpus.Sub(CPUDiff)
+			mem.Sub(MemDiff)
+
+			Expect(cpus).ToNot(BeZero())
+			Expect(mem).ToNot(BeZero())
+
+			updatedRes := corev1.ResourceList{
+				corev1.ResourceCPU:    cpus,
+				corev1.ResourceMemory: mem,
+			}
```
shajmakh, line 783: nit: alternativly can use: https://github.com/openshift-kni/numaresources-operator/blob/main/internal/resourcelist/resourcelist.go#L70

fromanirh (commented):
/lgtm
/hold
let's move forward and have cleanups (if needed) later

test/e2e/serial/tests/workload_placement.go
swatisehgal: Sure, will update.

test/e2e/serial/tests/workload_placement.go
shajmakh: right

shajmakh:
@swatisehgal Thank you for all the additions and updates 
/lgtm

swatisehgal:
/unhold
Addressed all review comments and got ack from all the reviewers.

fromanirh:
/lgtm
thanks for the rebase! let's have this merged now, then I'll un-hold the other ready PRs.

PR #329, opened 2022-07-13
e2e:serial:cntscope: filter node based on single resource shortage

Tal-or:
A test to make sure that nodes get filtered out by the plugin, for any resource that nodes have a shortage with at the NUMA level.

The test setup all the nodes in such way that from the `NodeResourceFit` plugin perspective they do have all the requested resources by the pod.
But the `TopologyMatch` plugin which zoom-in and examine the node at the NUMA level, should find only one suitable node which is able to address the pod request.
All the other nodes have all the available resources except one which has a shortage on a specific NUMA and we expect the plugin to track that

### DIFF
--- a/test/e2e/serial/tests/workload_placement_tmpol.go
+++ b/test/e2e/serial/tests/workload_placement_tmpol.go
@@ -692,6 +692,217 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 				},
 			},
 		),
+		Entry("[tier1][testtype4][tmscope:container][cpu] pod with two gu cnt land on a node with enough resources, containers should be spread on a different zone",
+			nrtv1alpha1.SingleNUMANodeContainerLevel,
+			setupPaddingContainerLevel,
+			e2enrt.CheckNodeConsumedResourcesAtLeast,
+			podResourcesRequest{
+				appCnt: []corev1.ResourceList{
+					{
+						corev1.ResourceCPU:    resource.MustParse("4"),
+						corev1.ResourceMemory: resource.MustParse("4Gi"),
+						"hugepages-2Mi":       resource.MustParse("32Mi"),
+						"hugepages-1Gi":       resource.MustParse("1Gi"),
+					},
+					{
+						corev1.ResourceCPU:    resource.MustParse("4"),
+						corev1.ResourceMemory: resource.MustParse("4Gi"),
+						"hugepages-2Mi":       resource.MustParse("32Mi"),
+						"hugepages-1Gi":       resource.MustParse("1Gi"),
+					},
+				},
+			},
+			// we need keep the gap between Node level fit and NUMA level fit wide enough.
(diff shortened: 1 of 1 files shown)


test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -692,6 +692,217 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+						corev1.ResourceCPU:    resource.MustParse("4"),
+						corev1.ResourceMemory: resource.MustParse("4Gi"),
+						"hugepages-2Mi":       resource.MustParse("32Mi"),
+						"hugepages-1Gi":       resource.MustParse("1Gi"),
+					},
+				},
+			},
+			// we need keep the gap between Node level fit and NUMA level fit wide enough.
+			// for example if only 2 cpus are separating unsuitable node from becoming suitable,
+			// it's not good because the baseload should be added as well (which is around 2 cpus)
```
fromanirh, line 770: the baseload for memory is ~4/4.5G but the gist of the comment is still valid

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -692,6 +692,217 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+					},
+					{
+						corev1.ResourceCPU:    resource.MustParse("4"),
+						corev1.ResourceMemory: resource.MustParse("4Gi"),
+						"hugepages-2Mi":       resource.MustParse("32Mi"),
+						"hugepages-1Gi":       resource.MustParse("1Gi"),
+					},
+				},
+			},
+			// we need keep the gap between Node level fit and NUMA level fit wide enough.
```
fromanirh, line 821: this comments feels a bit out of place here, but no big deal

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -692,6 +692,217 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+			// for example if only 2 cpus are separating unsuitable node from becoming suitable,
+			// it's not good because the baseload should be added as well (which is around 2 cpus)
+			// and then the pod might land on the unsuitable node.
+			[]corev1.ResourceList{
+				{
+					corev1.ResourceCPU:    resource.MustParse("1"),
+					corev1.ResourceMemory: resource.MustParse("4Gi"),
+					"hugepages-2Mi":       resource.MustParse("32Mi"),
```
fromanirh, line 882: however this we don't have the resources at all

ffromani (commented):
/lgtm
we can improve later, good enough to move forward

test/e2e/serial/tests/workload_placement_tmpol.go
Tal-or: We should maybe tune the tests for a bit.
It depends on the tested cluster settings

PR #336, opened 2022-07-15
Dockerfile: fully qualified builder images

fromanirh:
Until we find better builder images, we actually want to
use the dockerhub images, so let's be explicit about that.


fromanirh:
we can make another exception here because of the nature of the change. If the containers are built, the change is good.

PR #342, opened 2022-07-18
e2e:serial: remove requirement for scope=pod in unsched tests

fromanirh:
Now that https://github.com/openshift-kni/numaresources-operator/pull/319 merged, we can test with both pod and container scope. Hence let's remove this restriction


fromanirh:
ok, let's make the last exception and let's address the test failure in https://github.com/openshift-kni/numaresources-operator/pull/324

PR #346, opened 2022-07-19
e2e: install: minimal fixes

fromanirh:
This e2e install test:
```
[Install] durability
/go/src/github.com/openshift-kni/numaresources-operator/test/e2e/install/install_test.go:138
  with a running cluster with all the components and overall deployment
  /go/src/github.com/openshift-kni/numaresources-operator/test/e2e/install/install_test.go:180
    should be able to delete NUMAResourceOperator CR and redeploy without polluting cluster state [It] 
```
started to fail reliably. This PR wants to add troubleshooting aids and, ultimately, the fix. Spawned off from #324

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -287,7 +287,10 @@ func (r *NUMAResourcesOperatorReconciler) syncMachineConfigPoolsStatuses(instanc
 				Conditions: mcp.Status.Conditions,
 			})
 
-			if !IsMachineConfigPoolUpdated(instance.Name, mcp) {
+			isUpdated := IsMachineConfigPoolUpdated(instance.Name, mcp)
+			klog.V(5).InfoS("Machine Config Pool state", "name", mcp.Name, "instance", instance.Name, "updated", isUpdated)
+
+			if !isUpdated {
 				return false
 			}
 		}
--- a/test/e2e/install/install_test.go
+++ b/test/e2e/install/install_test.go
@@ -328,6 +328,7 @@ var _ = Describe("[Install] durability", func() {
 				ds, err := getDaemonSetByOwnerReference(updatedNroObj.GetUID())
 				if err != nil {
 					klog.Warningf("failed to get the RTE DaemonSet: %v", err)
+					klog.Warningf("NRO:\n%s\n", objects.ToYAML(updatedNroObj))
 					return false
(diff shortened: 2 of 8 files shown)


fromanirh:
/cc @swatisehgal @Tal-or 
I think I found a good enough fix eventually. Got full green already, let's see if CI confirms

swatisehgal:
/lgtm

Thanks for working on this and unblocking the install tests.

/hold
in case others want to provide additional input on this. Feel free to unhold as you see fit.

Tal-or:
/lgtm
This PR looks great and has a lot of improvements, but the question is, which of the listed commits actually fixed the issue?
More importantly, do we know the root cause? and what caused the regression?

fromanirh:
> /lgtm This PR looks great and has a lot of improvements, but the question is, which of the listed commits actually fixed the issue? More importantly, do we know the root cause? and what caused the regression?

good point. AFAICT, https://github.com/openshift-kni/numaresources-operator/pull/346/commits/3c6e3cfe7c9a2047e8eea8723c154db4a92f7b4a is the fix. Everything else is cleanup and support code to deliver the actual fix.
I don't have a root cause but I think that the install test was broken because

1. 4.11 OCP - which we started to use in CI only recently takes longer to reboot, this combined to
2. the test was not correctly waiting for the MCP to be updated, leading to
3. timeout failed because the DS searching timeout actually covered MCP update (delete) + MCP update (recreate) + DS search, leading to timeout exhaustion and test failure.

Tal-or:
> > /lgtm This PR looks great and has a lot of improvements, but the question is, which of the listed commits actually fixed the issue? More importantly, do we know the root cause? and what caused the regression?
> 
> good point. AFAICT, [3c6e3cf](https://github.com/openshift-kni/numaresources-operator/commit/3c6e3cfe7c9a2047e8eea8723c154db4a92f7b4a) is the fix. Everything else is cleanup and support code to deliver the actual fix. I don't have a root cause but I think that the install test was broken because
> 
> 1. 4.11 OCP - which we started to use in CI only recently takes longer to reboot, this combined to
> 2. the test was not correctly waiting for the MCP to be updated, leading to
> 3. timeout failed because the DS searching timeout actually covered MCP update (delete) + MCP update (recreate) + DS search, leading to timeout exhaustion and test failure.

Make sense. Thank you for the explanation.

PR #347, opened 2022-07-19
e2e: install: skip broken test

ffromani:
We need more time and investment to fix this test, so let's
skip it meanwhile.


swatisehgal:
Given that we have an open PR investigating and fixing the issue, I think it is okay to skip for now.

/lgtm

fromanirh:
there is another investigation ongoing carried by @Tal-or . Let's wait a bit to see where this is heading and we will decide

fromanirh:
/hold cancel
we decided offline to move forward with the backlog while the proper investigation is ongoing

PR #348, opened 2022-07-20
WIP: hp setup draft, needs testing

shajmakh:
TBD


shajmakh:
the current PAO tag is v0.0.41001 which is aged and does not have the new hugepages tool changes needed in this batch.

PR #351, opened 2022-07-21
e2e: serial: schedule gu pod as part of a deployment to a specific node

swatisehgal:
The test creates a deployment, pins it selectively to a specific node. In addition to that,
it is checked that NRT objects are updated appropriately.

### DIFF
--- a/test/e2e/serial/tests/non_regression.go
+++ b/test/e2e/serial/tests/non_regression.go
@@ -21,19 +21,23 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/ghodss/yaml"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/klog/v2"
 
 	nrtv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
 
 	e2ereslist "github.com/openshift-kni/numaresources-operator/internal/resourcelist"
 
+	schedutils "github.com/openshift-kni/numaresources-operator/test/e2e/sched/utils"
 	serialconfig "github.com/openshift-kni/numaresources-operator/test/e2e/serial/config"
 	e2efixture "github.com/openshift-kni/numaresources-operator/test/utils/fixture"
 	e2enrt "github.com/openshift-kni/numaresources-operator/test/utils/noderesourcetopologies"
 	e2enodes "github.com/openshift-kni/numaresources-operator/test/utils/nodes"
+	"github.com/openshift-kni/numaresources-operator/test/utils/nrosched"
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
 	e2ewait "github.com/openshift-kni/numaresources-operator/test/utils/objects/wait"
 	e2epadder "github.com/openshift-kni/numaresources-operator/test/utils/padder"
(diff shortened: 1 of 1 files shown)


test/e2e/serial/tests/non_regression.go
```diff
@@ -191,5 +195,123 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+			}
+
+			By(fmt.Sprintf("creating a deployment with a guaranteed pod with two containers requiring total %s", e2ereslist.ToString(e2ereslist.FromContainers(podSpec.Containers))))
+			dp := objects.NewTestDeploymentWithPodSpec(replicas, podLabels, nodeSelector, fxt.Namespace.Name, "testdp", *podSpec)
+
+			err = fxt.Client.Create(context.TODO(), dp)
+			Expect(err).ToNot(HaveOccurred())
+
+			updatedDp, err := e2ewait.ForDeploymentComplete(fxt.Client, dp, time.Second*10, time.Minute)
+			Expect(err).ToNot(HaveOccurred())
```
fromanirh, line 250: if we select a non esistent scheduler, why the pods managed to land on a node? let's add a comment to clarify this

test/e2e/serial/tests/non_regression.go
```diff
@@ -191,5 +195,123 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 				return ok
 			}, time.Minute, time.Second*5).Should(BeTrue(), "resources not restored on %q", targetedNodeName)
 		})
+
+		It("[test_id:48964][tier2] should be able to schedule a guaranteed deployment pod in selective way", func() {
```
fromanirh, line 199: can we please rephrase what "selective way" means?

swatisehgal: Sure thing, will do now.

test/e2e/serial/tests/non_regression.go
swatisehgal: Yeah, good point. I think I referred to another test where something similar was done. I will double check this and make sure this does not cause unexpected behavior. Would probably be better to use the default scheduler here instead.

swatisehgal: I updated this and am no longer providing a `schedulerName` explicitly so that the pod is scheduled by the default scheduler.

test/e2e/serial/tests/non_regression.go
```diff
@@ -191,5 +195,120 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+			Expect(len(pods)).To(Equal(1))
+
+			updatedPod := pods[0]
+
+			By(fmt.Sprintf("checking the pod landed on the target node %q vs %q", updatedPod.Spec.NodeName, targetNodeName))
+			Expect(updatedPod.Spec.NodeName).To(Equal(targetNodeName),
+				"node landed on %q instead of on %v", updatedPod.Spec.NodeName, targetNodeName)
+
+			By(fmt.Sprintf("checking the pod was scheduled with the topology aware scheduler %q", serialconfig.Config.SchedulerName))
+			schedOK, err := nrosched.CheckPODWasScheduledWith(fxt.K8sClient, updatedPod.Namespace, updatedPod.Name, serialconfig.Config.SchedulerName)
```
fromanirh, line 263: let's make sure this is consistent: if we pick the default scheduler, this is going to fail

swatisehgal: Ack, fixed now. Moved back to a dummy scheduler name to indicate that the scheduler name does not matter. In addition to that, we check that we don't receive any scheduling event when pod events are inspected.

fromanirh:
```
 [test_id:47587][tier1] should restart RTE DaemonSet when image is updated in NUMAResourcesOperator [It]
```
New flake/breakage we need to investigate - but let's fix the serviceaccount issue before

fromanirh (commented):
/lgtm

PR #354, opened 2022-07-22
e2e: serial: daemonset with non-gu pod requesting resources on a node

swatisehgal:
Test case: Daemonset with a non-gu (burstable) pod requesting resources
available on one node but not on a single numa.

No changes in NRT objects occur.


fromanirh (commented):
/lgtm
/approve

PR #359, opened 2022-07-26
numaresourcesscheduler: fix service-account's merge function

Tal-or:
Due to inappropriate handling of ```ServiceAccount```'s update we endded up with race conditions between the Token controller: https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#token-controller and the NUMAResourcesScheduler controller which in turn leads to infinite UPDATEs which in turn overwhelmed the kube-api-server.

By using the correct merge function for the ```ServiceAccount``` object, we're getting the correct secret and avoiding the mentioned race condition.


fromanirh:
/lgtm
/approve
great catch and great investigation

PR #361, opened 2022-07-26
baseload: account for any resource type in baseload

shajmakh:
Till now we were accounting only for CPU and memory corev1 resources, while some tests verify the feature functionality with hugepages requests, and going forward we expect to have devices as part of the resource requests. Support accounting for any other resources while calculating the baseload.


internal/resourcelist/resourcelist.go
```diff
@@ -59,14 +60,24 @@ func FromContainers(containers []corev1.Container) corev1.ResourceList {
...
+	}
+}
 
-	adjustedMemory := res.Memory()
-	adjustedMemory.Add(mem)
-	res[corev1.ResourceMemory] = *adjustedMemory
+func SubCoreResources(res, resToSub corev1.ResourceList) {
+	for resName, resQty := range resToSub {
+		if resQty.Cmp(res[resName]) > 0 {
+			klog.Errorf("Cannot substract %q because it is not found in the current resources", resName)
```
fromanirh, line 74: let's actually bubble up this error. This will require returning error on `SubCoreResources` and `Deduct`, but it seems the best approach. Swallowing errors (albeit logging like we do here) it's always something to be very wary of.
`Deduct` is different from `Apply` (`Add`) after all, so let's make this evident in the function signature.

fromanirh (commented):
nice! let's fix the error handling in `SubCoreResources` and we're good to go

internal/resourcelist/resourcelist.go
shajmakh: I agree, will adjust it. thanks for highlighting that!

fromanirh (commented):
/lgtm
/approve
thanks!

PR #364, opened 2022-07-27
e2e: change image for init containers

shajmakh:
For disconnected environments (like we use in NROP CI) using quay images will cause the pod to fail to start running due to unreachable image url.since we mainly care about the amount of time the init container is alive, change the image to from test-ci to Pause image that would be replaced by the url passed into one of the following env vars: E2E_PAUSE_IMAGE_URL , E2E_NUMACELL_DEVICE_PLUGIN_URL.

### DIFF
--- a/test/e2e/serial/tests/workload_placement_tmpol.go
+++ b/test/e2e/serial/tests/workload_placement_tmpol.go
@@ -1151,7 +1151,7 @@ func makeInitTestContainers(pod *corev1.Pod, initCnt []corev1.ResourceList, time
 	for i := 0; i < len(initCnt); i++ {
 		pod.Spec.InitContainers = append(pod.Spec.InitContainers, corev1.Container{
 			Name:    fmt.Sprintf("inittestcnt-%d", i),
-			Image:   images.SchedTestImageCI,
+			Image:   images.GetPauseImage(),
 			Command: []string{"/bin/sleep"},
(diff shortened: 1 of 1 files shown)


test/e2e/serial/tests/workload_placement.go
```diff
@@ -255,6 +255,11 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			updatedDp, err = e2ewait.ForDeploymentComplete(fxt.Client, dp, time.Second*10, time.Minute)
 			Expect(err).ToNot(HaveOccurred())
 
+			//wait for old dp pod to be removed to get the updated list of pods of the updated deployment
```
Tal-or, line 258: I don't think we need that, because in line 270 we're checking that we only have a single pod. so even if we still have the old one, we would wait before proceeding.

shajmakh: you are right, I missed that line of code. will dig further.

shajmakh: removed this for now. the failure does not reproduce locally nor always on CI.The failure in CI does not reflect debug logging. I'll handle the fix in later PR when I get a clear reason for why this is happening.

Tal-or: Thank you for the change

PR #368, opened 2022-07-28
WIP: rte: securitycontext: enforce dropping of unneeded privileges

fromanirh:
To access the podresources socket, we need uid=0 - gid=0
is not enough, unfortunately - but we don't actually need
any real root privileges. Make sure we drop them.


fromanirh:
/hold
requires more work

fromanirh:
we will pursue a different approach

PR #370, opened 2022-07-28
e2e: serial: Resources shared by burstable and guaranteed pods

swatisehgal:
We create a burstable pod that occupies node resources.
Subsequently trying to schedule a guaranteed pod fails
as the pod doesn't have sufficient resources and stays
pending till burstable pod is deleting and enough resources
are available on the node to accomodate it.


swatisehgal:
Based on top of https://github.com/openshift-kni/numaresources-operator/pull/372

This PR has been tested locally and passes.

fromanirh (commented):
/lgtm
/hold
we need to make sure that pods are cleaned up on test failure - can be done in a followup PR

PR #371, opened 2022-07-28
e2e: print scheduler logs upon failure 

shajmakh:
The test fails in CI but passes locally. Print the logs of the scheduler pod upon failure to manage troubleshooting the issue causing this step to fail.

### DIFF
--- a/test/e2e/serial/tests/workload_placement.go
+++ b/test/e2e/serial/tests/workload_placement.go
@@ -42,6 +42,7 @@ import (
 	numacellapi "github.com/openshift-kni/numaresources-operator/test/deviceplugin/pkg/numacell/api"
 	schedutils "github.com/openshift-kni/numaresources-operator/test/e2e/sched/utils"
 	serialconfig "github.com/openshift-kni/numaresources-operator/test/e2e/serial/config"
+	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
 	e2efixture "github.com/openshift-kni/numaresources-operator/test/utils/fixture"
 	e2enrt "github.com/openshift-kni/numaresources-operator/test/utils/noderesourcetopologies"
 	"github.com/openshift-kni/numaresources-operator/test/utils/nodes"
@@ -120,7 +121,6 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 				Expect(err).ToNot(HaveOccurred())
 			}
 		})
-
 		It("[test_id:47591][tier1] should modify workload post scheduling while keeping the resource requests available", func() {
 			hostsRequired := 2
 			paddedNodes := padder.GetPaddedNodes()
@@ -367,8 +367,12 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 
 			updatedPod = pods[0]
 			By(fmt.Sprintf("checking the pod landed on a node which is different than target node %q vs %q", targetNodeName, updatedPod.Spec.NodeName))
-			Expect(updatedPod.Spec.NodeName).ToNot(Equal(targetNodeName),
-				"pod should not landed on node %q", targetNodeName)
+			if updatedPod.Spec.NodeName == targetNodeName {
+				_ = objects.LogEventsForPod(fxt.K8sClient, updatedPod.Namespace, updatedPod.Name)
+				//print the logs of the scheduler pod
+				logSchedulerPluginLogs(*fxt)
+			}
+			Expect(updatedPod.Spec.NodeName).ToNot(Equal(targetNodeName), "pod should not land on node %q", targetNodeName)
 
 			By(fmt.Sprintf("checking the pod was scheduled with the topology aware scheduler %q", serialconfig.Config.SchedulerName))
 			schedOK, err = nrosched.CheckPODWasScheduledWith(fxt.K8sClient, updatedPod.Namespace, updatedPod.Name, serialconfig.Config.SchedulerName)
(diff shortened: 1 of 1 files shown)


test/e2e/serial/tests/workload_placement.go
```diff
@@ -328,8 +328,8 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			// there are now no more than 2 available CPUs under the targeted node and our test pod under the deployment is asking for 5 CPUs
 			// so in order to be certain that the pod will land on different node we need to request more than 7 CPUs in total
 			requiredRes = corev1.ResourceList{
-				// 6 here + 2 on the second container
-				corev1.ResourceCPU:    resource.MustParse("6"),
+				// 8 here + 2 on the second container
+				corev1.ResourceCPU:    resource.MustParse("8"),
```
fromanirh, line 332: I'm not sure how this should help. Could you please elaborate about the math?

test/e2e/serial/tests/workload_placement.go
```diff
@@ -366,8 +366,10 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 
 			updatedPod = pods[0]
 			By(fmt.Sprintf("checking the pod landed on a node which is different than target node %q vs %q", targetNodeName, updatedPod.Spec.NodeName))
-			Expect(updatedPod.Spec.NodeName).ToNot(Equal(targetNodeName),
-				"pod should not landed on node %q", targetNodeName)
+			if updatedPod.Spec.NodeName == targetNodeName {
+				_ = objects.LogEventsForPod(fxt.K8sClient, updatedPod.Namespace, updatedPod.Name)
```
Tal-or, line 370: The idea of logging the test pod's logs is nice, but I don't expect something useful to be found here.
It is more useful to log the scheduler's pod logs, so we could see what leads to this decision.

shajmakh: here we actually log the pod events, the idea was to check if there is an event for this pod that placed it on the "invalid" node. if scheduler pod logs can help I'll can that.

Tal-or: Oh it's the pod events`, I thought it was the pod logs`.
yes, this might be helpful, explaining why the pod land where it landed.

test/e2e/serial/tests/workload_placement.go
fromanirh: I'm still not convinced by this specific change. I'm worried it can hide the real issue. Everything else LGTM.

shajmakh: I'll remove this then and let's see if we can get some info from the scheduler logs as Talor suggested

fromanirh: thanks!

test/e2e/serial/tests/workload_placement.go
```diff
@@ -809,3 +813,30 @@ func matchLogLevelToKlog(cnt *corev1.Container, level operatorv1.LogLevel) (bool
 	val, found := rteFlags.GetFlag("--v")
 	return found, val.Data == kLvl.String()
 }
+
+func logSchedulerPluginLogs(fxt e2efixture.Fixture) {
```
fromanirh, line 817: this reminds me that we need to review (someday in the future) the way we handle and pass around Fixture objects. Passing by value (not the pointer) seems most correct indeed.

test/e2e/serial/tests/workload_placement.go
```diff
@@ -809,3 +813,30 @@ func matchLogLevelToKlog(cnt *corev1.Container, level operatorv1.LogLevel) (bool
...
+	if err != nil {
+		klog.Warningf("error getting the scheduler deployment: %v", err)
+		return
+	}
+	schedPods, err := schedutils.ListPodsByDeployment(fxt.Client, *schedDp)
+	if err != nil {
+		klog.Warningf("error getting the scheduler pod: %v", err)
+		return
+	}
+	Expect(len(schedPods)).To(Equal(1))
```
fromanirh, line 834: the code above (never returning error, using klog and no expectations) suggests we want to tolerate failures in this function and not abort the test should this function fail. If so, why adding this only `Expect`?

fromanirh (commented):
just a question inside, overall LGTM

test/e2e/serial/tests/workload_placement.go
```diff
@@ -367,8 +367,12 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
 			updatedPod = pods[0]
 			By(fmt.Sprintf("checking the pod landed on a node which is different than target node %q vs %q", targetNodeName, updatedPod.Spec.NodeName))
-			Expect(updatedPod.Spec.NodeName).ToNot(Equal(targetNodeName),
-				"pod should not landed on node %q", targetNodeName)
+			if updatedPod.Spec.NodeName == targetNodeName {
+				_ = objects.LogEventsForPod(fxt.K8sClient, updatedPod.Namespace, updatedPod.Name)
+				//print the logs of the scheduler pod
+				logSchedulerPluginLogs(*fxt)
+			}
+			Expect(updatedPod.Spec.NodeName).ToNot(Equal(targetNodeName), "pod should not landed on node %q", targetNodeName)
```
fromanirh, line 375: "should not land" or "should not have landed"

test/e2e/serial/tests/workload_placement.go
```diff
@@ -398,7 +402,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			Expect(err).ToNot(HaveOccurred())
 
 			// the NRT updaters MAY be slow to react for a number of reasons including factors out of our control
-			// (kubelet, runtime). This is a known behaviour. We can only tolerate some delay in reporting on pod removal.
+			// (kubelet, runtime). This is a known behavior. We can only tolerate some delay in reporting on pod removal.
```
fromanirh, line 405: we really dislike the english spelling, don't we? :)

test/e2e/serial/tests/workload_placement.go
shajmakh: right, same I thought about, but having more than one scheduler pod can indicate a serious problem here.

shajmakh: Yet, it is not out goal in this function to check that so I'll make it and if with warning.

test/e2e/serial/tests/workload_placement.go
shajmakh: I wouldn't mind much but seeing a line under it irritates me :)

fromanirh (commented):
/lgtm

PR #375, opened 2022-08-01
e2e:install: wait for DaemonSet to show-up

### LINKED ISSUE #369: flake: fix install lane `should restart RTE DaemonSet when image is updated in NUMAResourcesOperator `
This test started to flake recently (week of 20220728 - and probably the week earlier):
```
 [Fail] [Install] durability with a running cluster with all the components and overall deployment [It] [test_id:47587][tier1] should restart RTE DaemonSet when image is updated in NUMAResourcesOperator 
```
we should fix the test

Tal-or:
We should wait, until timeout, for the DaemonSet count under NRO's status to be equal to 1, and not immediately return an error on the first falling try

Should fix https://github.com/openshift-kni/numaresources-operator/issues/369 

### DIFF
--- a/test/e2e/install/install_test.go
+++ b/test/e2e/install/install_test.go
@@ -196,7 +196,8 @@ var _ = Describe("[Install] durability", func() {
 					return false, err
 				}
 				if len(nroObj.Status.DaemonSets) != 1 {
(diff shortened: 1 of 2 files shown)


ffromani (commented):
/lgtm
even if this doesn't fix the linked issue, looks like a good change in general

PR #381, opened 2022-08-05
nrocli: validation fixes and retargeting

fromanirh:
enhance (and rename, since it's only supposed to validate) nrocli to detect NRT misconfiguration, even though only the most glaring.

Add along the way some fixes for the existing functionality, and options to be friendlier for integration and testing.
With these ehnancements, `nrovalidate` (formerly `nrocli`) can now detect and report incompatible environments, meaning environments on which NROP can be deployed but it will do nothing useful.


nrovalidate/main.go
```diff
@@ -70,15 +91,11 @@ func parseArgs(args ...string) (ProgArgs, error) {
...
-	if pArgs.Version {
-		return pArgs, err
-	}
-
 	return pArgs, nil
 }
 
 func validateCluster(args ProgArgs) error {
-	cli, err := clientutil.New()
+	cli, err := NewClientWithScheme(scheme)
```
shajmakh, line 98: what are the advantages of using the clientset with the scheme instead of clientset out of clientutil?

fromanirh: we need to have extra schemas registered in the client (NRO, MCP) which aren't registered by default in the clientutil one.

shajmakh:
/lgtm 
thanks!

.github/workflows/e2e-tools.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
     - name: Build tools
       run: |
-        make build-tools binary-nrocli
+        make build-tools binary-nrovalidate
```
Tal-or, line 48: it is a tool which supposes to validate, so maybe ```nrovalidator``` is more accurate here?

fromanirh: I don't have strong feelings. We're not following too closely the unix conventions anyway, so if you want it renamed I can do it.

Tal-or:
/lgtm
/approve
Thanks

PR #386, opened 2022-08-09
nrovalidate enhancements

fromanirh:
Finalize the `nrovalidate` tool to make it usable for integration and to check the cluster readiness.
Most notably, this include better input/output control and the ability to check the presence of NRT instances; without them the NUMA-aware scheduler will work like the core kubernetes scheduler.


pkg/validator/validator.go
```diff
@@ -0,0 +1,126 @@
...
+		return data, err
+	}
+
+	enabledNodeNames := sets.NewString()
+	for _, mcp := range nroMcps {
+		nodes, err := machineconfigpools.GetNodeListFromMachineConfigPool(ctx, cli, *mcp)
+		if err != nil {
+			return data, err
+		}
+		enabledNodeNames.Insert(getNodeNames(nodes)...)
```
shajmakh, line 184: so I understand that a node that is part of the mcp that is matching the nodeGroup mentioned in the NRO, is considered among the nodes that TAS is enabled for. don't we want to verify that RTE works && has pods for that node?

pkg/validator/validator.go
```diff
@@ -0,0 +1,126 @@
...
+
+type ValidatorData struct {
+	tasEnabledNodeNames sets.String
+	kConfigs            map[string]*kubeletconfigv1beta1.KubeletConfiguration
+	versionInfo         *version.Info
+}
+
+type collectHelper func(ctx context.Context, cli client.Client, data *ValidatorData) error
+
+func Collect(ctx context.Context, cli client.Client) (ValidatorData, error) {
```
shajmakh: this does not collect kConfigs for the nodes, how is it set?

pkg/validator/validator.go
fromanirh: I think this kind of validation is starting to be complex and I'd defer to a future stage. OTOH the existing validation should be sufficient to clearly detect incompatible clusters, on which the NUMA aware scheduler should perform like the builtin scheduler.

nrovalidate/main.go
```diff
@@ -100,12 +101,12 @@ func validateCluster(args ProgArgs) error {
...
 	}
 
-	data, err := nrovalidator.CollectData(context.TODO(), cli)
+	data, err := nrovalidator.Collect(context.TODO(), cli)
 	if err != nil {
 		return err
 	}
 
-	result, err := nrovalidator.Validate(*data)
+	result, err := nrovalidator.Validate(data)
```
shajmakh: nrovalidator.Collect sets the nodes names and the version only, we should also call CollectKubeletConfig() before Validate

fromanirh: please check https://github.com/openshift-kni/numaresources-operator/pull/386/commits/464fc94c0db25f404551c8263739d73c3cf322a2#diff-40da5a0178b34c207217154cc7fa3d432882f52681d9e262e33bce27f0156b36R84

fromanirh: Or also https://github.com/openshift-kni/numaresources-operator/pull/386/files/bff1ad7f297ca3bc4173540a252778a1f50086cc#diff-40da5a0178b34c207217154cc7fa3d432882f52681d9e262e33bce27f0156b36R80

shajmakh: I see, thanks for pointing that!

pkg/validator/noderesourcetopologies.go
```diff
@@ -0,0 +1,111 @@
...
+			return nil
+		}
+		return err
+	}
+
+	data.nrtList = nrtList
+	return nil
+}
+
+func ValidateNodeResourceTopologies(data ValidatorData) ([]deployervalidator.ValidationResult, error) {
```
shajmakh, line 59: the function returns the result within the if conditions, if we wish to return list of results, wouldn't it make sense to gather append all relevant results and then return the final results list, otherwise we don't need to return []deployervalidator.ValidationResult nor to append to ret, wdyt?

shajmakh: also we don't have any case the func sets err, why to return error?

fromanirh: I see your point here. The problem is that we are constraint by the functio signature - because validations want to be generic and return a list of validation errors, while in this specific case all the validations we implemented are mutually exclusive. So we will always either have zero or one validation errors, which we need to wrap in a slice anyway.
I don't see easy fixes without overhauling the API. Do you perhaps see a better approach here to solve this dilemma?

fromanirh: > also we don't have any case the func sets err, why to return error?

another case on which we are constrained by a more generic API, which doesn't make much sense for this specific set of validation implemented here. :\

nrovalidate/main.go
```diff
@@ -59,6 +61,7 @@ type ProgArgs struct {
 	Verbose bool
 	Quiet   bool
 	JSON    bool
+	What    sets.String
```
shajmakh, line 65: can we rename What to better word that reflects the use of the arg? maybe ValidatorType or similar

fromanirh: sure

pkg/validator/noderesourcetopologies.go
shajmakh: I don't have ideas for this, but anyway it is no big deal to keep it the way it is

.github/workflows/e2e-tools.yml
```diff
@@ -62,6 +62,8 @@ jobs:
     - name: E2E Tests
       run: |
         export KUBECONFIG=${HOME}/.kube/config
+        kubectl create -f doc/examples/mcp.yaml
+        kubectl create -f doc/examples/nrop.yaml
```
shajmakh, line 66: > These tests depend on running on a *un*configured cluster.

to clarify, the operator is installed but the crs aren't configured, is that right?

regarding lines 65-66 should we wait for the creation of the mcp and nrop?

fromanirh: correct. We need the CRD (definition) but no CR instances.
The creation of CRDs is usually very fast so I'm cutting a corner here ( :) ) but yes, the proper way would be to wait for it.
Do we want to have this change?

nrovalidate/main.go
```diff
@@ -100,36 +118,46 @@ func validateCluster(args ProgArgs) error {
 		return err
 	}
 
-	data, err := nrovalidator.CollectData(context.TODO(), cli)
+	if !args.Quiet {
+		fmt.Fprintf(os.Stderr, "INFO>>>>: enabled validators: %s\n", strings.Join(args.Validations.List(), ","))
+	}
+
+	data, err := nrovalidator.Collect(context.TODO(), cli, args.Labels, args.Validations)
```
shajmakh, line 125: I believe we can pass args as a whole instead of labels, and validation, this may enable better maintenance of the function in the future when there is a need to collect the data counting on other fields. what is the benefit of passing args in pieces instead of simply args ?

.github/workflows/e2e-tools.yml
shajmakh: I see. Then I agree on keeping it as is now.

shajmakh:
Thanks for the great work on this!
/lgtm

nrovalidate/main.go
fromanirh: good point. We should _not_ pass `args` because `args` is a type defined in `main` to hold the user-provided flags. So we would need a new structure in the `validator` pkg to collect the arguments. I'm 50/50 regarding this approach, sometimes is nice sometimes not really nice. In any case my threshold triggers around 2/3 arguments, so the next argument we add to `Collect` (if ever) would require us to evaluate this approach.

fromanirh:
> Thanks for the great work on this! /lgtm

Thank you for the comprehensive review!

PR #391, opened 2022-08-11
serial: check pod requesting unallocatable resources is unschedulable 

shajmakh:
Add new test that attempts to schedule a pod requesting resources that are greater than allocatable at numa level, and verifies the pod keeps on pending.

### DIFF
--- a/test/e2e/serial/tests/workload_placement.go
+++ b/test/e2e/serial/tests/workload_placement.go
@@ -1049,19 +1049,7 @@ func allocatableResourceType(nrtInfo nrtv1alpha1.NodeResourceTopology, resName c
 // NUMA nodes so that all the replicas can successfully obtain resources
 // from all the NUMA nodes.
 func leastAvailableResourceQtyInAllZone(nrtInfo nrtv1alpha1.NodeResourceTopology, baseload baseload.Load, resName corev1.ResourceName) resource.Quantity {
-	var maxResAllocatable resource.Quantity
-
-	// Finding the maximum allocatable resources of a resource type across all zones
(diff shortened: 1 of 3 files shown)


test/e2e/serial/tests/workload_placement.go
```diff
@@ -1049,19 +1049,7 @@ func allocatableResourceType(nrtInfo nrtv1alpha1.NodeResourceTopology, resName c
...
-		zoneQty, ok := e2enrt.FindResourceAllocatableByName(zone.Resources, resName.String())
-		if !ok {
-			continue
-		}
-		if zoneQty.Cmp(maxResAllocatable) > 0 {
-			maxResAllocatable = zoneQty
-		}
-	}
-
+	maxResAllocatable := e2enrt.GetMaxAllocatableResourceNumaLevel(nrtInfo, resName)
```
fromanirh, line 1052: the intent is good, but when a helper is used only once, we should generally not create it and inline it. The only noteworthy exception is the code extraction makes the code significantly more testable or significantly easier to read.
In this case we have some readability gains and in the followup patch you use it, which passes the code extraction threshold.

Next time please mention in the commit message you plan to use the extracted code and mention the reasons that made you decide to perform the extraction - the fact you extracted the code is evident from the change itself, the commit message should explain why and how, not what.

fromanirh (commented):
/lgtm
/approve

test/e2e/serial/tests/workload_placement.go
shajmakh: I agree, thanks for highlighting this! I'll consider it next time

PR #339, opened 2022-07-18
Must gather e2e test

jlojosnegros:
As must-gather PAO now collects information about NRO ( openshift-kni/performance-addon-operators#919 ) 
we need to add new test suite to check it.

issue: #330

PR #392, opened 2022-08-16
e2e:must-gather: Run must-gather e2e test suite only in OCP Platform

jlojosnegros:
must-gather is only supported on OCP platform so we should skip all the test suite in any other platform

correction over: #339 
issue: #330 
blocks: openshift/release#31102

### DIFF
--- a/test/e2e/must-gather/must_gather_test.go
+++ b/test/e2e/must-gather/must_gather_test.go
@@ -28,10 +28,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/yaml"
 
-	"github.com/onsi/ginkgo"
-	"github.com/onsi/gomega"
 	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects/wait"
+
+	"github.com/onsi/ginkgo"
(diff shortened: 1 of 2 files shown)


test/e2e/must-gather/must_gather_test.go
```diff
@@ -133,8 +136,11 @@ var _ = ginkgo.Describe("[must-gather] NRO data collected", func() {
 					podFolderNames = append(podFolderNames, podFolder.Name())
 				}
 				gomega.Expect(podFolderNames).To(gomega.ContainElement(gomega.MatchRegexp("^numaresources-controller-manager*")))
-				gomega.Expect(podFolderNames).To(gomega.ContainElement(gomega.MatchRegexp("^numaresourcesoperator-mcp-test*")))
 				gomega.Expect(podFolderNames).To(gomega.ContainElement(gomega.MatchRegexp("^secondary-scheduler*")))
+
+				if configuration.Plat == platform.Kubernetes {
```
fromanirh, line 141: that's ok, but thinking about it, must-gather is supported only on OCP platform; so the must-gather e2e tests should just skip (all of them) if running on non-OCP platform.

jlojosnegros: Ok

fromanirh (commented):
/approve
/lgtm

PR #393, opened 2022-08-16
e2e:serial: requesting available resources with replicaset

Tal-or:
This test will schedule a `ReplicaSet` with multiple replicas and check NRT reflects resource consumption correctly.
In addition the test compares between the scheduling time of default and TAS scheduler and make sure we don't have a regression.
IOW, TAS scheduler doesn't take too much time comparing to the default one.

### DIFF
--- a/test/e2e/serial/tests/workload_placement.go
+++ b/test/e2e/serial/tests/workload_placement.go
@@ -19,6 +19,7 @@ package tests
 import (
 	"context"
 	"fmt"
+	"math"
 	"time"
 
 	"github.com/ghodss/yaml"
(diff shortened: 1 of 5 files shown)


test/e2e/serial/tests/workload_placement.go
```diff
@@ -856,6 +857,258 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+					klog.Warningf("inconsistent accounting: no resources consumed by the running pod,\nNRT before test deployment: %s \nNRT after: %s \npod resources: %v", dataBefore, dataAfter, e2ereslist.ToString(rl))
+					return false
+				}
+				return true
+			}).WithTimeout(timeout).WithPolling(10 * time.Second).Should(BeTrue())
+
+			By(fmt.Sprintf("comparing scheduling times between %q and %q", corev1.DefaultSchedulerName, serialconfig.Config.SchedulerName))
+			diff := int64(math.Abs(float64(schedTimeWithTopologyScheduler.Milliseconds() - schedTimeWithDefaultScheduler.Milliseconds())))
+			// 2000 milliseconds diff seems reasonable, but can evaluate later if needed.
+			d := time.Millisecond * 2000
```
Tal-or, line 1104: Maybe we can set a more restrictive diff here

ffromani (commented):
/approve
/lgtm
overall comments are not imporant enough to postpone the merge. Let's improve later.

PR #394, opened 2022-08-16
serial: fix memory requests and free resources on target node

shajmakh:
The memory alignment test was expecting to find a message in the pod events indicating that although the pod resources were available on compute node level, alignment of both containers could not be achieved on a single numa node, one container per numa as per the assumption that the tests run only when TMscope=container.
The test failed to find that expected message because there were no candidate nodes in the first place, the reason was a lack of memory resources. The pod was requesting 9Gi, while the free resources left on the target node were only 8. The reason it was set to 8 is because of the false thought that there will be base memory on that node, which is usually 4.5Gi, which makes sense that 4.5Gi + 8Gi is way more than enough, but that is a mistake. because when noderesourcefit filters the nodes it compares the actual available resources on the node with the pod requested resources, if the available resources can accommodate the pod resources then it will mark the node as a possible candidate, if not it will reject it. The result we get from Fit when the test fails is that there is only 8.3 free memory on the target node, hence it is rejected.

To fix the test, first set minimum free resources on the first numa of the target node: 1Gi, because we know for sure that the baseload will be added to the first numa upon padding the node. the base memory is around 4.5Gi, so we need to consider that both pod containers requests at least 6Gi so none of them could fit into the first numa. To be on the safe side, let both containers request 7Gi.
OTOH the target node should have total free memory that matches the total memory requested by the pod, which is 14Gi, thus let the second numa on the target node have the rest of 13Gi free memory.

### DIFF
--- a/test/e2e/serial/tests/workload_placement_tmpol.go
+++ b/test/e2e/serial/tests/workload_placement_tmpol.go
@@ -1018,7 +1018,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			},
 		),
 	)
-	DescribeTable("[placement][negative] cluster with multiple worker nodes suitable",
+	DescribeTable("[placement][negative] cluster with one worker nodes suitable",
 		func(tmPolicy nrtv1alpha1.TopologyManagerPolicy, setupPadding setupPaddingFunc, errMsg string, podRes podResourcesRequest, unsuitableFreeRes, targetFreeResPerNUMA []corev1.ResourceList) {
 
 			hostsRequired := 2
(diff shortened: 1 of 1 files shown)


fromanirh:
please make sure to fix the PR title ("resuests")

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -1202,11 +1202,16 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 					"hugepages-1Gi":       resource.MustParse("1Gi"),
 				},
 			},
+			// the free resources that should be left on the target node should not depend that there will be some baseload added upon padding the node,
+			// those free resources should match the pod requests in total. The reason behind that is that Noderesourcesfit plugin (the plugin that is
+			// responsible for accepting/rejecting compute nodes as candidates for placing the pod) actually accounts for the baseload, it compares the
+			// actual available resources on node with the pod requested resources, if the available resources can accommodate the pod resources then it
+			// will mark the node as a possible candidate, if not it will reject it.
```
fromanirh, line 1209: note to (my)self: `Fit` does sum all the container resources requests; hence, will account for the baseload when running it's own computation. (note baseload does the same thing/uses the same algo - actually inspired from that).

fromanirh (commented):
/approve
/lgtm
in some cases I would have put different emphasis on different facets, but thinking about it this falls in the matter of personal taste so I can't really think of any objective improvement.
Overall impressive work and very nicely documented. Kudos!

PR #396, opened 2022-08-17
e2e: serial: move code into `internal`, add test predictability helpers

fromanirh:
periodic cleanup to ensure the serial suite is maintenable and keeps being maintenable.
We continue the effort to trim down `test/utils` with (mostly) trivial code movement, and we add a facility to force the selection of a target node for the test suite.


fromanirh:
tested locally. Merging under my responsability. I'll fix any breakage (reverting if needed)

PR #399, opened 2022-08-24
e2e: serial: pod pending when (request + overhead) > zone allocatable

swatisehgal:
Test case where pod overhead is such that even though resources requested
by a pod are enough to be successfully allocated on the target node, the
amount of pod overhead impacts the fate of the pod keeping it in pending
state as the node zone doesn't have enough resources (pod resource request + pod
overhead).

### DIFF
--- a/test/e2e/serial/tests/workload_overhead.go
+++ b/test/e2e/serial/tests/workload_overhead.go
@@ -274,6 +274,176 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
 				}
 
 			})
+
+			It("[test_id:53819][tier2] Pod pending when resources requested + pod overhead don't fit on the target node; NRT objects are not updated", func() {
+				var targetNodeName string
+				var targetNrtInitial *nrtv1alpha1.NodeResourceTopology
+				var targetNrtListInitial nrtv1alpha1.NodeResourceTopologyList
+				var err error
+
+				// even if it is not a hard rule, and even if there are a LOT of edge cases, a good starting point is usually
+				// in the ballpark of 5x the base load. We start like this
+				podResources := corev1.ResourceList{
+					corev1.ResourceCPU:    resource.MustParse("6"),
+					corev1.ResourceMemory: resource.MustParse("6Gi"),
+				}
+
+				// to avoid issues with fractional resources being unaccounted atm, we round up to requests;
+				// for the test proper, as low as cpu=100m and mem=100Mi would have been sufficient.
+				minRes := corev1.ResourceList{
+					corev1.ResourceCPU:    resource.MustParse("1"),
+					corev1.ResourceMemory: resource.MustParse("1Gi"),
+				}
+
+				// need a zone with resources for overhead, pod and a little bit more to avoid zone saturation
+				klog.Infof("kubernetes pod fixed overhead: %s", e2ereslist.ToString(rtClass.Overhead.PodFixed))
+
+				candidateNodeNames := e2enrt.AccumulateNames(nrtTwoZoneCandidates)
+
+				var ok bool
+				targetNodeName, ok = e2efixture.PopNodeName(candidateNodeNames)
+				Expect(ok).To(BeTrue(), "cannot select a node among %#v", candidateNodeNames.List())
+				By(fmt.Sprintf("selecting node to schedule the test pod: %q", targetNodeName))
+
+				err = fxt.Client.List(context.TODO(), &targetNrtListInitial)
+				Expect(err).ToNot(HaveOccurred())
+				targetNrtInitial, err = e2enrt.FindFromList(targetNrtListInitial.Items, targetNodeName)
+				Expect(err).NotTo(HaveOccurred())
+
+				// finding the maximum allocatable resources across all zones and using it as the pod overhead
+				rtClass.Overhead = &nodev1.Overhead{
+					PodFixed: corev1.ResourceList{
+						corev1.ResourceCPU:    e2enrt.GetMaxAllocatableResourceNumaLevel(*targetNrtInitial, corev1.ResourceCPU),
+						corev1.ResourceMemory: e2enrt.GetMaxAllocatableResourceNumaLevel(*targetNrtInitial, corev1.ResourceMemory),
+					},
+				}
+
+				// updating the runtimeClass with the new podoverhead value
+				err = fxt.Client.Update(context.TODO(), rtClass)
+				Expect(err).NotTo(HaveOccurred())
+
+				podFixedOverheadCPU, podFixedOverheadMem := resourcelist.RoundUpCoreResources(*rtClass.Overhead.PodFixed.Cpu(), *rtClass.Overhead.PodFixed.Memory())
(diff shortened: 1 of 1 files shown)


test/e2e/serial/tests/workload_overhead.go
```diff
@@ -274,6 +274,158 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
...
+				targetNodeName, ok = e2efixture.PopNodeName(candidateNodeNames)
+				Expect(ok).To(BeTrue(), "cannot select a node among %#v", candidateNodeNames.List())
+				By(fmt.Sprintf("selecting node to schedule the test pod: %q", targetNodeName))
+
+				err = fxt.Client.List(context.TODO(), &targetNrtListInitial)
+				Expect(err).ToNot(HaveOccurred())
+				targetNrtInitial, err = e2enrt.FindFromList(targetNrtListInitial.Items, targetNodeName)
+				Expect(err).NotTo(HaveOccurred())
+
+				//get maximum available node CPU and Memory
```
Tal-or, line 321: If the pod is asking for maximum available node resources, the target node would be disqualified by the `NodeResourceFit` plugin which is prior to the `TopologyMatch` (our) plugin.
IOW,  we're testing core K8S functionality here.

What we need to do is to make sure that the target node considers a good candidate by the `NodeResourceFit` plugin, but not by the `TopologyMatch`.
We can ask for maximum available resources at the zone level for example.

test/e2e/serial/tests/workload_overhead.go
```diff
@@ -274,6 +274,158 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
...
+				minRes := corev1.ResourceList{
+					corev1.ResourceCPU:    resource.MustParse("1"),
+					corev1.ResourceMemory: resource.MustParse("1Gi"),
+				}
+
+				// need a zone with resources for overhead, pod and a little bit more to avoid zone saturation
+				klog.Infof("kubernetes pod fixed overhead: %s", e2ereslist.ToString(rtClass.Overhead.PodFixed))
+
+				const requiredNUMAZones = 2
+				By(fmt.Sprintf("filtering available nodes with at least %d NUMA zones", requiredNUMAZones))
```
Tal-or, line 302: We are already filtering at the `BeforeEach` (lines 97-107) aren't we?

test/e2e/serial/tests/workload_overhead.go
```diff
@@ -274,6 +274,158 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
...
+				By(fmt.Sprintf("Scheduling the testing deployment with RuntimeClass=%q", rtClass.Name))
+				var deploymentName string = "test-dp"
+				var replicas int32 = 1
+
+				podLabels := map[string]string{
+					"test": "test-dp",
+				}
+				nodeSelector := map[string]string{}
+				deployment := objects.NewTestDeployment(replicas, podLabels, nodeSelector, fxt.Namespace.Name, deploymentName, objects.PauseImage, []string{objects.PauseCommand}, []string{})
+				deployment.Spec.Template.Spec.SchedulerName = serialconfig.Config.SchedulerName
```
Tal-or, line 397: nit:
```
podSpec := &deployment.Spec.Template.Spec
podSpec.SchedulerName = serialconfig.Config.SchedulerName
podSpec.Containers[0].Resources.Limits = podResources
podSpec.RuntimeClassName = &rtClass.Name
```
Or use the ``` NewTestDeploymentWithPodSpec``` func

Tal-or (commented):
Very nice!
A couple of comments from me.

test/e2e/serial/tests/workload_overhead.go
```diff
@@ -274,6 +274,158 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
 				}
 
 			})
+
+			It("[test_id:47582][tier2] Pod pending when resources requested + pod overhead don't fit on the target node; NRT objects are not updated", func() {
```
shajmakh, line 278: let's update the test_id to 53819 as we already have an It with 47582 set.

test/e2e/serial/tests/workload_overhead.go
```diff
@@ -274,6 +274,158 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
...
+				podFixedOverheadCPU, podFixedOverheadMem := resourcelist.RoundUpCoreResources(*rtClass.Overhead.PodFixed.Cpu(), *rtClass.Overhead.PodFixed.Memory())
+				podFixedOverhead := corev1.ResourceList{
+					corev1.ResourceCPU:    podFixedOverheadCPU,
+					corev1.ResourceMemory: podFixedOverheadMem,
+				}
+				klog.Infof("kubernetes pod fixed overhead rounded to: %s", e2ereslist.ToString(podFixedOverhead))
+
+				zoneRequiredResources := podResources.DeepCopy()
+				resourcelist.AddCoreResources(zoneRequiredResources, podFixedOverhead)
+				resourcelist.AddCoreResources(zoneRequiredResources, minRes)
```
shajmakh, line 342: we want to have the total pod + overhead resources allocatable under at least one numa zone of the target node but not available.
having the overhead as the max res of the target node and then rounding it up + the pod resources will end up with the pod pending but no guarantee why, it might be because pod + overhead res are not allocatable (there are no place on the node for the pod + OH because the node might be rejected by Fit as Talor mentioned)

test/e2e/serial/tests/workload_overhead.go
shajmakh: +1 
also it isn't necessary to have two numas, but no need to consider that atm because the ci environments supports at least one worker with 2 numas.

swatisehgal: Good catch, will fix now.

test/e2e/serial/tests/workload_overhead.go
swatisehgal: Sure, will fix. No problem.

test/e2e/serial/tests/workload_overhead.go
swatisehgal: Ack

test/e2e/serial/tests/workload_overhead.go
swatisehgal: Sure, the description in the test plan made me take this approach but what you are suggesting makes sense and I will update shortly.

test/e2e/serial/tests/workload_overhead.go
swatisehgal: Sure, I might have misunderstood the description of this test in test plan. I will update this now.

Tal-or:
Looking great.
Did you test it locally? Unfortunately, the CI  covers nothing here

swatisehgal:
> Looking great. Did you test it locally? Unfortunately, the CI covers nothing here

Yep, I tested this locally and it passed.

shajmakh:
Thanks for all the changes!
/lgtm

PR #415, opened 2022-09-14
e2e:serial:tmpolicy: introduce tmPolicyFuncs

Tal-or:
Specific functions for specific policies start piling up, which makes maintaining harder and adding policy-specific behavior more complex.
In order to solve that, we introduce tmPolicyToolbox which provides a set of functions/variables per a given policy.

This will make future extension much easier and keeps the amount of
test's parameters smaller and cleaner.

The first commit is essentially the addition of another policy-specific function which was the trigger (or the straw that broke the camel's back so to speak)  for the proposed patch.

### DIFF
--- a/test/e2e/serial/tests/workload_placement_tmpol.go
+++ b/test/e2e/serial/tests/workload_placement_tmpol.go
@@ -64,13 +64,22 @@ type podResourcesRequest struct {
 	appCnt  []corev1.ResourceList
 }
 
-type setupPaddingFunc func(fxt *e2efixture.Fixture, nrtList nrtv1alpha1.NodeResourceTopologyList, padInfo paddingInfo) []*corev1.Pod
-type checkConsumedResFunc func(nrtInitial, nrtUpdated nrtv1alpha1.NodeResourceTopology, required corev1.ResourceList) (string, error)
-type filterMatchingResourcesFunc func(nrts []nrtv1alpha1.NodeResourceTopology, requests corev1.ResourceList) []nrtv1alpha1.NodeResourceTopology
+type tmPolicyFuncs struct {
+	name                    func() nrtv1alpha1.TopologyManagerPolicy
+	setupPadding            func(fxt *e2efixture.Fixture, nrtList nrtv1alpha1.NodeResourceTopologyList, padInfo paddingInfo) []*corev1.Pod
+	checkConsumedRes        func(nrtInitial, nrtUpdated nrtv1alpha1.NodeResourceTopology, required corev1.ResourceList) (string, error)
+	filterMatchingResources func(nrts []nrtv1alpha1.NodeResourceTopology, requests corev1.ResourceList) []nrtv1alpha1.NodeResourceTopology
+}
+
+type tmPolicyFuncsHandler map[nrtv1alpha1.TopologyManagerPolicy]tmPolicyFuncs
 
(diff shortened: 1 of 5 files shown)


test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -1568,3 +1540,21 @@ func isHugePageInAvailable(rl corev1.ResourceList) bool {
 	}
 	return false
 }
+
```
fromanirh, line 1543: I think this is the right call.
It's about time we do some factoring in this area, the argument list was a bit wild already.
I think this is the right improvement we need, a code extraction without (yet) a redesign of the flow. We are grouping a bunch of functions together for practicality, there's not yet a separate data type (with state whatsoever).

My only gripe is naming. "tmPolicyToolbox" feels clunky. I'd go with something like
```
- perPolicyHelpers
- tmPolicyFuncs
- perTMPolicyFuncs
```
Probably "tmPolicyFuncs" is the best (= less ugly) I can think atm.
I'd also rename the member funcs like
```
type tmPolicyFuncs struct {
	policyName()            func() string
	setupPadding            func(fxt *e2efixture.Fixture, nrtList nrtv1alpha1.NodeResourceTopologyList, padInfo paddingInfo) []*corev1.Pod
	checkConsumedResources  func(nrtInitial, nrtUpdated nrtv1alpha1.NodeResourceTopology, required corev1.ResourceList) (string, error)
	filterMatchingResources func(nrts []nrtv1alpha1.NodeResourceTopology, requests corev1.ResourceList) []nrtv1alpha1.NodeResourceTopology
}
```

the idea is that every member of `tmPolicyFuncs` should be a `func` (and not a data field). The creation funcs can look like:

```
func newPodScopeTMPolicyFuncs() tmPolicyFuncs {
	return tmPolicyToolbox{
		policyName:              func() string { return  nrtv1alpha1.SingleNUMANodePodLevel },
		setupPadding:            setupPaddingPodLevel,
		checkConsumedRes:        e2enrt.CheckZoneConsumedResourcesAtLeast,
		filterMatchingResources: e2enrt.FilterAnyZoneMatchingResources,
	}
}
```

ffromani (commented):
The fix is needed, the cleanup is welcome. Let's just spend a bit more time sorting out naming and some refinements.

test/e2e/serial/tests/workload_placement_tmpol.go
Tal-or: The naming is always something I'm struggling with.
But the most important thing is that you liked the idea :) 
I'll fix the naming asap.
Thanks for the review.

Tal-or:
@fromanirh Another concern I have is about initializing the new ```tmPolicyFuncs```. This can be a global variable used across all tests.
But instead, we need to recreate it for every ```Describe``` container.
This is redundant overhead and maintainship but I couldn't find a better idea.
We can't move into a separate package since it's using internal functions

fromanirh:
> @fromanirh Another concern I have is about initializing the new `tmPolicyFuncs`. it can be a global variable that is used across all tests, but instead, we need to recreate it for every `Describe` container. This is redundant overhead and maintainship but I couldn't find a better idea. We can't move into a separate package since it's using internal functions

it's a legit concern. Please add a comment nearby `tmPolicyFuncs` to remind us about this. For the time being, let's just recreate it every time.

PR #416, opened 2022-09-14
e2e:serial:tmpolicy: test device NUMA alignment

Tal-or:
Schedule a pod with 2 containers, where the containers` requests are available under two possible nodes, but only one of them has the devices on a single numa in such a way that each device is available on a different numa. The pod is expected to be scheduled on the feasible node with relevant conditions in terms of availability on single numa nodes.

### DIFF
--- a/test/e2e/serial/tests/workload_placement_tmpol.go
+++ b/test/e2e/serial/tests/workload_placement_tmpol.go
@@ -900,6 +900,55 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 				},
 			},
 		),
+		Entry("[test_id:54021][tier1][testtype4][tmscope:container][devices] pod with two gu cnt land on a node with enough resources, containers should be spread on a different zone",
+			tmPolicyFuncsHandler[nrtv1alpha1.SingleNUMANodeContainerLevel],
+			podResourcesRequest{
+				appCnt: []corev1.ResourceList{
+					{
+						corev1.ResourceCPU:    resource.MustParse("4"),
+						corev1.ResourceMemory: resource.MustParse("4Gi"),
+						"example.com/deviceA": resource.MustParse("2"),
+						"example.com/deviceC": resource.MustParse("1"),
+					},
+					{
+						corev1.ResourceCPU:    resource.MustParse("4"),
+						corev1.ResourceMemory: resource.MustParse("4Gi"),
+						"example.com/deviceB": resource.MustParse("2"),
+					},
+				},
+			},
+			// we need keep the gap between Node level fit and NUMA level fit wide enough.
+			// for example if only 2 cpus are separating unsuitable node from becoming suitable,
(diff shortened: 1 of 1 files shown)


Tal-or:
/hold 
Depends on https://github.com/openshift-kni/numaresources-operator/pull/415

fromanirh (approved):
/lgtm

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -908,6 +908,57 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+			nrtv1alpha1.SingleNUMANodeContainerLevel,
+			setupPaddingContainerLevel,
+			e2enrt.CheckNodeConsumedResourcesAtLeast,
+			podResourcesRequest{
+				appCnt: []corev1.ResourceList{
+					{
+						corev1.ResourceCPU:    resource.MustParse("4"),
+						corev1.ResourceMemory: resource.MustParse("4Gi"),
+						"example.com/deviceA": resource.MustParse("2"),
+						"example.com/deviceC": resource.MustParse("1"),
```
shajmakh, line 921: this test will be executed only if those two devices are on the cluster, mentioning these explicitly here will work for current sample devices. we might need to tweak those instances in the future when supporting real devices on the cluster, and have the names of the devices based on environment variables so the test works for any kind of device (real/sample). 
I've started working on this, but still needs some adjustments and testing:
https://github.com/shajmakh/numaresources-operator/commit/64c25286110a19fb03ea35e44fc49d214a807bea

fromanirh: right, I overlooked this part

fromanirh:
/lgtm cancel
https://github.com/openshift-kni/numaresources-operator/pull/416#discussion_r974004808

shajmakh:
thanks for the PR, the test looks good!
was it tested on a live cluster?

Tal-or:
> thanks for the PR, the test looks good! was it tested on a live cluster?

Yes

test/e2e/serial/tests/workload_placement_tmpol.go
Tal-or: Just make sure to update the variables here as well:
https://github.com/openshift-kni/numaresources-operator/tree/main/test/e2e/serial#e2e-serial-testsuite

fromanirh:
/lgtm
let's start with this PR and let's generalize/abstract/handle the device configuration later

fromanirh:
/hold cancel
required PR merged

fromanirh:
/hold
OTOH we can use a rebase to consume the `tmPolicyFuncs`

Tal-or:
> /hold OTOH we can use a rebase to consume the `tmPolicyFuncs`

Yes. This is why I wanted the previous one to get merged first.

fromanirh (approved):
/lgtm

PR #418, opened 2022-09-21
test: fixture: longer and smarter wait for NS del

fromanirh:
Up until now, we had a hardcoded wait of 30s for NS to be deleted. The number was pretty random and made out, and works quite well on lightly loaded (typical CI) cluster.

Sometimes, for more complex tests and/or on shared infra, the teardown can legitimately take longer than 30s to be deleted.

So, we make the wait timeout (much) longer[1] and we add periodic logs to signal the test isn't stuck, but waiting

[1] using another random "long enough" value, but still.


shajmakh:
good usage of InfoS!
thanks for the PR
/lgtm

PR #419, opened 2022-09-21
serial: add tests with device resources

shajmakh:
Add new environment variables for the devices names E2E_NROP_DEVICE_A, E2E_NROP_DEVICE_B, and E2E_NROP_DEVICE_C, that is to enable running the tests with any device type and not limited to specific types.
Add new tests with pods requesting device resources and check the alignment of these pods to the nodes.

### DIFF
--- a/test/e2e/serial/README.md
+++ b/test/e2e/serial/README.md
@@ -38,3 +38,10 @@ they found it before they run.
   for the specified amount of time after each spec, to give time to the cluster to settle up.
 - `E2E_NROP_TEST_TEARDOWN` (accepts string expressing time unit, e.g. `30s`) instructs the suite to wait
   *up* to the specified amount of time while tearing down the resources needed by each spec.
+- `E2E_NROP_DEVICE_A` (accepts string, e.g `example.com/deviceA`) declares name of a device type that exists on
+  the cluster and will be used as a resource of type `device-a` in the tests.
+- `E2E_NROP_DEVICE_B` (accepts string, e.g `example.com/deviceB`) declares name of a device type that exists on
+  the cluster and will be used as a resource of type `device-b` in the tests.
+- `E2E_NROP_DEVICE_C` (accepts string, e.g `example.com/deviceC`) declares name of a device type that exists on
+  the cluster and will be used as a resource of type `device-c` in the tests.
+  
--- a/test/utils/fixture/devices.go
+++ b/test/utils/fixture/devices.go
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2022 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


test/e2e/serial/README.md
```diff
@@ -36,3 +36,9 @@ they found it before they run.
   node matching the given name when a random node is needed.
 - `E2E_NROP_TEST_COOLDOWN` (accepts string expressing time unit, e.g. `30s`) instructs the suite to wait for
   the specified amount of time after each spec, to give tome to the cluster to settle up.
+- `E2E_NROP_DEVICE_A` (accepts string, e.g `example.com/deviceA`) declares name of a device type that exists on
```
fromanirh, line 39: not a fan of this naming, but I can't think of clearly better alternative, so let's move forward this way and improve later

test/utils/devices/devices.go
```diff
@@ -0,0 +1,53 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
fromanirh, line 2: this copyright boilerplate is not really correct. Not wrong either, but we should use the Red Hat copyright for this code. For example, see: https://github.com/openshift-kni/numaresources-operator/blob/v0.4.12-0.2022092201/test/e2e/serial/tests/workload_placement_taint.go#L2

And yes, this means we should fix quite some copyrights boilerplates around the codebase. So let's make sure the new code is correct

test/utils/devices/devices.go
```diff
@@ -0,0 +1,53 @@
+/*
```
fromanirh, line 1: do you mind moving this source file in `test/utils/fixture` ?

fromanirh (commented):
overall LGTM, but there's some inline questions I'd like to discuss

test/e2e/serial/README.md
shajmakh: I agree with you, I'll change it to E2E_NROP_DEVICE_TYPE_1 (sounds more reasonable also when using real devices) for now until we agree on a better naming, thanks for pointing this out.

test/utils/devices/devices.go
shajmakh: Thanks for raising this, do we care about the year?

fromanirh: you're right, I picked a random examples. In this case 2021 seems the correct choice if the code was written in 2021

test/utils/fixture/devices.go
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2022 Red Hat, Inc.
```
fromanirh, line 2: let's review the copyright year per the conversartion in the comments (now I need to review if the correct choice is 2021 or 2022 :) )

test/utils/devices/devices.go
fromanirh: in general: when we do code movement, we keep the older copyright year (obviously); when we write brand new code, we use the current year.

test/utils/fixture/devices.go
fromanirh: ok, this is new code so 2022 is fine

shajmakh: thanks for your input, may you please tell why 2022 and not 2021 (the current year)? I assume this will be available in 4.12 GA, right?

PR #420, opened 2022-09-22
e2e: serial: negative-test: topology alignment of pod requesting devices

swatisehgal:
Schedule a pod with 2 containers each requesting different device types. Enough resources are available on one node but distributed across NUMA nodes such that it is not possibile for the containers to obtain resources from a single numa and therefore the pod stays pending.

### DIFF
--- a/test/e2e/serial/tests/workload_placement_tmpol.go
+++ b/test/e2e/serial/tests/workload_placement_tmpol.go
@@ -1071,8 +1071,8 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			for _, nrt := range nrts {
 				for _, zone := range nrt.Zones {
 					avail := e2enrt.AvailableFromZone(zone)
-					if !isHugePageInAvailable(avail) {
-						Skip(fmt.Sprintf("no hugepages found under node: %q", nrt.Name))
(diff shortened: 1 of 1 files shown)


shajmakh:
Thanks for the PR
/lgtm

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -1590,6 +1642,21 @@ func isHugePageInAvailable(rl corev1.ResourceList) bool {
 	return false
 }
 
+func isRequestingHugepages(podRes podResourcesRequest) bool {
+	for _, appContainer := range podRes.appCnt {
+		if isHugePageInAvailable(appContainer) {
```
fromanirh, line 1647: minor: in a future PR we should rename this function, like `isHugepagePresent` or so. This is a bit misleading in this current form

shajmakh: were you referring to renaming isRequestingHugepages or isHugePageInAvailable? because isHugepagePresent sounds better name for isHugePageInAvailable

fromanirh: I meant that we should move away from `isHugePageInAvailable`, which is no longer fit, and change to something like `isHugepagePresent`, or something else along these lines

PR #422, opened 2022-09-23
e2e: serial: add scheduler cache non-regression

fromanirh:
We add more non-regression tests to cover the future planned addition of the scheduler cache.

We want to cover the cases which are no bother for the default and non-caching NUMA-aware scheduler, but needs to be checked for the caching NUMA-aware scheduler.

All these cases are *NOT* expected to cover the quality of the scheduling decisions - we have plenty of targeted tests for that. Here we want to check the scheduler fundamentals actually works.


fromanirh:
nope, we can't really split from the main functionality :\

PR #424, opened 2022-09-26
e2e: serial: add label to group nonregression test

fromanirh:
With this label we can easily run only the nonregression tests


fromanirh:
adding labels can't hurt anything - and the bulk of tests passed anyway. Let's not waste time for a useless retest.

PR #427, opened 2022-09-28
kustomize: allow the workload type management in ns

Tal-or:
This annotation will allow deployment of the operator inside the numaresources ns when management workload partitioning enabled

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -1,6 +1,8 @@
 apiVersion: v1
 kind: Namespace
 metadata:
+  annotations:
+    workload.openshift.io/allowed: management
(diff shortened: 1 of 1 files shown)


config/manager/manager.yaml
```diff
@@ -1,6 +1,8 @@
 apiVersion: v1
 kind: Namespace
 metadata:
+  annotations:
+    workload.openshift.io/allowed: management
```
fromanirh, line 5: I'm not sure how this will work when deployed by OLM, but adding a annotation won't hurt here

ffromani (commented):
/approve
/lgtm

PR #428, opened 2022-09-28
serial: add negative test with pod scope and devices

shajmakh:
- remove padding functionality for pod scope
- add new negative test to check the alignment of 2 containers on different numa zones when tmscope is pod
- reverting commit: https://github.com/openshift-kni/numaresources-operator/commit/fddf6661bfc2f93b11821a7b9f65a27fd6637dc8

### DIFF
--- a/test/e2e/serial/tests/workload_placement_tmpol.go
+++ b/test/e2e/serial/tests/workload_placement_tmpol.go
@@ -1625,6 +1625,50 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 				},
 			},
 		),
+		Entry("[test_id:54017][tier1][negative][tmscope:pod][devices] pod with two gu cnt keep on pending because cannot align the both containers on single numa",
+			tmPolicyFuncsHandler[nrtv1alpha1.SingleNUMANodePodLevel],
+			"cannot align pod",
+			podResourcesRequest{
+				appCnt: []corev1.ResourceList{
+					{
+						corev1.ResourceCPU:    resource.MustParse("8"),
+						corev1.ResourceMemory: resource.MustParse("4Gi"),
+						corev1.ResourceName(e2efixture.GetDeviceType1Name()): resource.MustParse("3"),
+					},
+					{
+						corev1.ResourceCPU:    resource.MustParse("8"),
+						corev1.ResourceMemory: resource.MustParse("4Gi"),
+						corev1.ResourceName(e2efixture.GetDeviceType2Name()): resource.MustParse("2"),
+					},
+				},
+			},
+			// we need keep the gap between Node level fit and NUMA level fit wide enough.
+			// for example if only 2 cpus are separating unsuitable node from becoming suitable,
+			// it's not good because the baseload should be added as well (which is around 2 cpus)
+			// and then the pod might land on the unsuitable node.
+			[]corev1.ResourceList{
+				{
+					corev1.ResourceCPU:    resource.MustParse("1"),
+					corev1.ResourceMemory: resource.MustParse("2Gi"),
+				},
+				{
+					corev1.ResourceCPU:    resource.MustParse("2"),
+					corev1.ResourceMemory: resource.MustParse("3Gi"),
+				},
+			},
+			[]corev1.ResourceList{
+				{
+					corev1.ResourceCPU:    resource.MustParse("8"),
+					corev1.ResourceMemory: resource.MustParse("4Gi"),
+					corev1.ResourceName(e2efixture.GetDeviceType1Name()): resource.MustParse("3"),
+				},
+				{
+					corev1.ResourceCPU:    resource.MustParse("8"),
+					corev1.ResourceMemory: resource.MustParse("4Gi"),
+					corev1.ResourceName(e2efixture.GetDeviceType2Name()): resource.MustParse("2"),
+				},
+			},
+		),
 	)
 })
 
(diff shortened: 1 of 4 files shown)


shajmakh:
/hold cancel
tests verified

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -1605,62 +1628,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
-	err = fxt.Client.Create(context.TODO(), padPod)
-	ExpectWithOffset(1, err).ToNot(HaveOccurred())
-	paddingPods = append(paddingPods, padPod)
-
-	paddingPodsUnsuitable := setupPaddingForUnsuitableNodes(2, fxt, nrtList, padInfo)
-	return append(paddingPods, paddingPodsUnsuitable...)
-}
-
-func setupPaddingContainerLevel(fxt *e2efixture.Fixture, nrtList nrtv1alpha1.NodeResourceTopologyList, padInfo paddingInfo) []*corev1.Pod {
+func setupPaddingLevel(fxt *e2efixture.Fixture, nrtList nrtv1alpha1.NodeResourceTopologyList, padInfo paddingInfo) []*corev1.Pod {
```
fromanirh, line 1679: this makes sense also because at conceptual level padding for scope=container is a more complex implementation of the padding for scope=pod. IOW, we can reimplement the `setupPaddingPodLevel` using the righe combination of parameters for `setupPaddingContainerLevel` (which you are doing here), while the opposite is not and can't be true.

I like this simplification, but I'd like to see this running for a while. The best way is actually get it running for a while and pay close attention to the test runs. Let's be vigilant for the time being.

While we are at it, let's rename the function `setupPadding` (minor, but still)

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -1778,7 +1746,7 @@ func isRequestingHugepages(podRes podResourcesRequest) bool {
 func newPodScopeTMPolicyFuncs() tmPolicyFuncs {
 	return tmPolicyFuncs{
 		name:                    func() nrtv1alpha1.TopologyManagerPolicy { return nrtv1alpha1.SingleNUMANodePodLevel },
-		setupPadding:            setupPaddingPodLevel,
+		setupPadding:            setupPaddingLevel,
```
fromanirh, line 1797: we should remove this field and call the now-universal `setupPadding` function directly in the call sites

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -482,6 +482,8 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			updatedPod, err := wait.ForPodPhase(fxt.Client, pod.Namespace, pod.Name, corev1.PodRunning, 2*time.Minute)
 			if err != nil {
 				_ = objects.LogEventsForPod(fxt.K8sClient, pod.Namespace, pod.Name)
+				dumpNRTForNode(fxt.Client, targetNodeName, "target")
+				//TODO print scheduler events
```
fromanirh, line 486: do you need events emitted by the scheduler or scheduler logs? The former, the most important among the former, should be emitted already by `LogEventsForPod`

fromanirh (commented):
I was a bit surprised initially, but now I like this direction

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -1778,7 +1822,7 @@ func isRequestingHugepages(podRes podResourcesRequest) bool {
 func newPodScopeTMPolicyFuncs() tmPolicyFuncs {
 	return tmPolicyFuncs{
 		name:                    func() nrtv1alpha1.TopologyManagerPolicy { return nrtv1alpha1.SingleNUMANodePodLevel },
-		setupPadding:            setupPaddingPodLevel,
+		setupPadding:            setupPaddingContainerLevel,
```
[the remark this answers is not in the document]
fromanirh: outdated comment, now I get what you meant

test/e2e/serial/tests/workload_placement_tmpol.go
shajmakh: I actually meant scheduler logs, thanks for raising that, I'll address it in this PR

fromanirh: scheduler logs are potentially very long. We dump events because they are concise. This change would probably need its own discussion and PR

shajmakh: you're right, we'll need some filtering applied. I'll leave this change then.

shajmakh:
> still WIP?

in fact yes :) I am reverting the change of using tmPolicyFuncsHandler, as all the functions there are similar for both scopes and thus we are left with only having the TMscope as the changing field , thus I don't believe it is necessary to still have these commands. let me know what you disagree.
In addition, I wanted to split the negative cases into their own file.

fromanirh:
/hold
until I restore the WIP

fromanirh:
/hold cancel
WIP restored, sorry for the rush :)

shajmakh:
> /hold cancel WIP restored, sorry for the rush :)

Thank you, no worries :)

fromanirh (commented):
/approve
/lgtm

PR #431, opened 2022-10-10
RTE: enable per-node-group tuning

fromanirh:
Introduce per-node-group config (NodeGroupConfig) object, in order to enable fine-tuning of RTE.

This is needed to enable cluster-admins to explore the proper tradeoff between cluster load and scheduling precision.

We will also need to have some slack to explore the best default to set, and we need some scalability knobs for the scale benchmarks (due "soon").

The new Config object means we need to deprecate the `DisablePodsFingerprinting` bool we added in the 4.11 cycle.
This also plays nice with the ongoing deprecation of the bool flags in the k8s APIs.


fromanirh:
/hold
missing controller tests. We don't need e2e tests, but controller tests we definitely need.

fromanirh:
/hold cancel
tests added

bundle/manifests/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -65,6 +65,11 @@ spec:
       - description: Group of Nodes to enable RTE on
         displayName: Group of nodes to enable RTE on
         path: nodeGroups
+      - description: Set the topology info refresh period. Use explicit 0 to disable.
```
swatisehgal, line 68: Why is only the refresh period here and not `infoRefreshMode`? The refresh period is linked to `infoRefreshMode` and depends on whether `Periodic` or `Evented` so was expecting both here.

pkg/objectstate/merge/merge.go
```diff
@@ -96,3 +98,19 @@ func Labels(current, updated client.Object) client.Object {
...
+
+	if updated.PodsFingerprinting != nil {
+		conf.PodsFingerprinting = updated.PodsFingerprinting
+	}
+	if updated.InfoRefreshPeriod != nil {
+		conf.InfoRefreshPeriod = updated.InfoRefreshPeriod
+	}
+	if updated.InfoRefreshMode != nil {
+		conf.InfoRefreshMode = updated.InfoRefreshMode
+	}
```
swatisehgal, line 114: What is the value of `InfoRefreshPeriod` if `Evented` mode is specified? IIUC, we only need `InfoRefreshPeriod` if `Periodic` mode is specified.

config/manifests/bases/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -33,6 +33,11 @@ spec:
       - description: Group of Nodes to enable RTE on
         displayName: Group of nodes to enable RTE on
         path: nodeGroups
+      - description: Set the topology info refresh period. Use explicit 0 to disable.
+        displayName: Topology info refresh period setting
+        path: nodeGroups[0].config.infoRefreshPeriod
```
swatisehgal, line 38: Similar to my query above, would really appreciate some clarification here.

swatisehgal (commented):
Thanks for the PR.
Looking good, just have a few minor queries.

bundle/manifests/numaresources-operator.clusterserviceversion.yaml
fromanirh: good catch, I don't have a answer, need to check.

pkg/objectstate/merge/merge.go
fromanirh: this needs clarification indeed. Thanks for pointing this out.

config/manifests/bases/numaresources-operator.clusterserviceversion.yaml
fromanirh: agreed

pkg/objectstate/merge/merge.go
fromanirh: I think the least surprising behaviour is to propagate fields in the merge code, and make logic about the relationship between fields where we consume the data. I'm gonna change the logic which translate the settings into DS args, but keep the merge logic as simple as possible - hence updating all the fields when != nil.

bundle/manifests/numaresources-operator.clusterserviceversion.yaml
fromanirh: the root cause was a typo in the `operator-sdk` annotation:
```
-       //+operator-sdk:csv:customresourcedefinition:type=spec,displayName="Topology info mechanism setting",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+       //+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Topology info mechanism setting",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
```

swatisehgal:
Thanks for addressing all the review comments.

/lgtm
/approve

PR #432, opened 2022-10-11
serial: fix reboot scenario cases

shajmakh:
The tests that require reboot were failing, this PR fixes them. Please see the commits' messages for more details.

### DIFF
--- a/test/e2e/serial/README.md
+++ b/test/e2e/serial/README.md
@@ -16,7 +16,7 @@ run in isolation with any other workload, and in isolation to each other.
 
 IOW, the suite expects to run against a pristine, unloaded cluster to validate it.
 
-It is responsability of the tests in this suite to clean up properly and to restore the cluster state as
+It is responsibility of the tests in this suite to clean up properly and to restore the cluster state as
 they found it before they run.
 
 ### configuring using the environment variables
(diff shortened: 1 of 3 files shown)


test/e2e/serial/README.md
```diff
@@ -16,7 +16,7 @@ run in isolation with any other workload, and in isolation to each other.
 
 IOW, the suite expects to run against a pristine, unloaded cluster to validate it.
 
-It is responsability of the tests in this suite to clean up properly and to restore the cluster state as
+It is responsibility of the tests in this suite to clean up properly and to restore the cluster state as
```
fromanirh, line 19: this (and following) is a good change and can be split off to its own tiny PR and merged immediately

fromanirh (commented):
The second commit is a no-brainer.
The first commit does things I can make some sense of, but it would be nice to add more details in the commit message to explain why the fixes are fixing, how did we notice and why the code was wrong before; in its current shape, the commit message is reiterating the _what_, not the _why_, which doesn't help much, and is not helpful for our future selves revisiting this code in the future.

Additionally, the first commit _can_ be split in focused commits, e.g. one removing the padding, one fixing the defers and so on.
There's no hard rule here; if all the fixes are needed, it makes sense to join them in the same commit; but there are also arguments for the split: if each commits leaves the codebase compilable, doesn't worsen the quality and is not adding more breakage, focused commits are easier to revert (and to review).

test/e2e/serial/README.md
shajmakh: Thanks for raising that. Done here:
https://github.com/openshift-kni/numaresources-operator/pull/433

shajmakh:
/hold
addressing the review comments

shajmakh:
After merging #433  the reboot tests pass.
closing this PR because I prefer to have it as a future reference with the changes in it. 
a follow-up PR will be posted for more enhancements to the tests.

PR #433, opened 2022-10-11
serial: add the new variable E2E_RTE_CI_IMAGE

shajmakh:
Enable the user to use the RTE image they want in the test by setting this environment variable E2E_RTE_CI_IMAGE. This is especially needed for d/s CI as it is using disconnected environments which cannot pull quay images if they are not mirroredas anticipated.


fromanirh:
/lgtm
/approve
thanks!

PR #434, opened 2022-10-11
e2e: serial: avoid cooldown on test skip

fromanirh:
regardless of the cooldown time, it doesn't make sense to cooldown if the test skipped for unmet prererquisites, which is actually the only reason why we should skip at all in this suite.


fromanirh:
/hold
still testing, looks good so far but not completed yet

PR #435, opened 2022-10-12
serial: enhance test verbosity and reduce execution time

shajmakh:
Enhance tests by:

- Remove redundant padding.
- Split a test case into two gaining two independent tests.
- Add more reports clarifying the test's steps

For more details please find the commit messages.

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -55,14 +55,12 @@ import (
 	e2enrt "github.com/openshift-kni/numaresources-operator/test/utils/noderesourcetopologies"
 	"github.com/openshift-kni/numaresources-operator/test/utils/nrosched"
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
-	e2epadder "github.com/openshift-kni/numaresources-operator/test/utils/padder"
 
 	serialconfig "github.com/openshift-kni/numaresources-operator/test/e2e/serial/config"
 )
 
 var _ = Describe("[serial][disruptive][slow] numaresources configuration management", func() {
 	var fxt *e2efixture.Fixture
-	var padder *e2epadder.Padder
 	var nrtList nrtv1alpha1.NodeResourceTopologyList
 	var nrts []nrtv1alpha1.NodeResourceTopology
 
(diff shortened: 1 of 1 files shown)


test/e2e/serial/tests/configuration.go
```diff
@@ -221,7 +221,7 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
 					}
 				}
 				return true, nil
-			}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(BeTrue())
+			}).WithTimeout(2 * timeout).WithPolling(30 * time.Second).Should(BeTrue())
```
fromanirh, line 224: the problem here is we are introducing a coupling between all the tests which use the timeout variable.
Suppose we need to increase the timeout for a single test for thatever reason. Perhapse the specific test is doing something that takes longer to complete.
This will mean all the tests that use the variable will increase the timeout. Is it correct? is it *more* readable?
Another reason: why it's easier/more readable to reason in terms of block of `timeouts` unit?
IOW, why is `2 * timeout` better than `10 * time.Minute`?

test/e2e/serial/tests/configuration.go
```diff
@@ -118,11 +118,12 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
...
 
 			defer func() {
-				By(fmt.Sprintf("unlabeling node: %q", targetedNode.Name))
+				By(fmt.Sprintf("Restore initial labels of node %q", targetedNode.Name))
+				klog.Infof("unlabeling node: %q with %q", targetedNode.Name, nodes.GetLabelRoleMCPTest())
 				err = unlabelFunc()
 				Expect(err).ToNot(HaveOccurred())
 
-				By(fmt.Sprintf("labeling node: %q", targetedNode.Name))
+				klog.Infof("labeling node: %q with %q", targetedNode.Name, nodes.GetLabelRoleWorker())
```
fromanirh, line 126: let's keep the `By` which is actually more visible than klog. The enhanced log is fine.

fromanirh (commented):
partial review. I'm still unconvinced that using the timeout variable helps.

fromanirh (commented):
/lgtm
/approve
thanks for the updates

fromanirh:
lane failed for infra issue :\

PR #441, opened 2022-10-18
rte: sched: manifest: enable sched plugin tracing

fromanirh:
Create empty directory using tmpfs to hold the NRT cache traces if they are enabled; if not this is not to consume relevant resources, so we can keep it always enabled.

Note this is a NOP at this point in time - the scheduler plugin code is still under development. But the above reasoning still holds: from both operator and scheduler plugin this is a very cheap and backward compatible change.

Enable PFP debugging for  RTEs, consuming the functionality from the latest u/s release. Enable dumping on their PFP status files on tmpfs. Even more than for the scheduler plugin, we can keep this always enabled from storage perspective.


rte/main.go
```diff
@@ -123,6 +123,7 @@ func parseArgs(args ...string) (ProgArgs, error) {
 	flags.BoolVar(&pArgs.Resourcemonitor.PodSetFingerprint, "pods-fingerprint", false, "If enable, compute and report the pod set fingerprint.")
 	flags.BoolVar(&pArgs.Resourcemonitor.ExposeTiming, "expose-timing", false, "If enable, expose expected and actual sleep interval as annotations.")
 	flags.BoolVar(&pArgs.Resourcemonitor.RefreshNodeResources, "refresh-node-resources", false, "If enable, track changes in node's resources")
+	flags.StringVar(&pArgs.Resourcemonitor.PodSetFingerprintStatusFile, "pods-fingerprint-status-file", "", "File to dump the pods fingerprint status. Use \"\" for all namespaces.")
```
fromanirh, line 126: copy/paste mistake

PR #449, opened 2022-10-20
numacell: more kubernetes 1.25 compatibility

fromanirh:
we need more changes to really ensure 1.25 compatibility (and really, better compability in general).
First of all, consume the second fix on device-plugin-manager; then, actually fix the registration process declaring the plugin needs (which are minimal, but let's state that).


shajmakh:
thanks for the PR!
/lgtm

PR #450, opened 2022-10-21
rte: config: fix setting of TM scope

fromanirh:
We used to have the default TM scope "container".
This works - and it is actually the k8s default, but prevents the overriding from every source.

Fix the code allowing override of TM scope.
Note this now requires the value to be always provided in some form. Since we always provide the configmap with the scope and policy settings, this should be ho harm.

### DIFF
--- a/rte/main.go
+++ b/rte/main.go
@@ -15,6 +15,7 @@ limitations under the License.
 package main
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
@@ -56,6 +57,7 @@ type ProgArgs struct {
 	LocalArgs       localArgs
 	Version         bool
 	SysinfoOnly     bool
+	DumpConfig      bool
 }
 
 func main() {
@@ -68,6 +70,15 @@ func main() {
 
 	if parsedArgs.Version {
 		fmt.Printf("%s %s %s %s\n", version.ProgramName(), version.Get(), version.GetGitCommit(), runtime.Version())
+		os.Exit(0)
+	}
+
+	if parsedArgs.DumpConfig {
+		data, err := json.Marshal(parsedArgs)
+		if err != nil {
+			klog.Fatalf("encoding the configuration to JSON")
+		}
+		fmt.Printf("%s\n", string(data))
 		os.Exit(0)
 	}
 
@@ -156,7 +167,7 @@ func parseArgs(args ...string) (ProgArgs, error) {
 	flags.StringVar(&sysReservedMemory, "system-info-reserved-memory", "", "kubelet reserved memory: comma-separated 'numaID=amount' (example: '0=16Gi,1=8192Mi,3=1Gi')")
 	flags.StringVar(&sysResourceMapping, "system-info-resource-mapping", "", "kubelet resource mapping: comma-separated 'vendor:device=resourcename'")
 	flags.BoolVar(&pArgs.SysinfoOnly, "system-info", false, "Output detected system info and exit")
-
+	flags.BoolVar(&pArgs.DumpConfig, "dump-config", false, "Output the configuration settings and exit - the output format is JSON, subjected to change without notice.")
 	flags.BoolVar(&pArgs.Version, "version", false, "Output version and exit")
 
 	err := flags.Parse(args)
--- a/rte/main.go
+++ b/rte/main.go
@@ -37,6 +37,13 @@ import (
 	"github.com/openshift-kni/numaresources-operator/rte/pkg/sysinfo"
 )
 
+const (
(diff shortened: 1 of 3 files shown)


shajmakh:
> Note this now requires the value to be always provided in some form. Since we always provide the configmap with the scope and policy settings, this should be no harm.

what about when the kubeletconfig doesn't set the TMscope/policy? it should be a reasonable configuration, but in this PR the fields are not optional. do I miss something?

fromanirh:
> > Note this now requires the value to be always provided in some form. Since we always provide the configmap with the scope and policy settings, this should be no harm.
> 
> what about when the kubeletconfig doesn't set the TMscope/policy? it should be a reasonable configuration, but in this PR the fields are not optional. do I miss something?

that would cause a RTE crash. The operator needs to ensure something is always set. Let me ensure this.

rte/main.go
```diff
@@ -225,20 +218,41 @@ func defaultHostName() string {
...
 
-func defaultTopologyManagerPolicy() string {
-	if val, ok := os.LookupEnv("TOPOLOGY_MANAGER_POLICY"); ok {
-		return val
+func setupTopologyManagerConfig(pArgs *ProgArgs, conf config.Config) error {
+	policyFrom := "args"
+	// do not overwrite with empty an existing value (e.g. from opts)
+	if pArgs.RTE.TopologyManagerPolicy == "" {
+		policyFrom = "env"
+		pArgs.RTE.TopologyManagerPolicy = os.Getenv("TOPOLOGY_MANAGER_POLICY")
```
shajmakh, line 226: what if the environment variable doesn't align with the kubeletconfig settings? 
also may you tell please where is this env var is set?

fromanirh: env is meant to override arguments. The priority is:
command line arguments overrides env vars overrides conf settings
(this is the standard UNIX way)
noone is setting these envs (anymore), it's for backward compat.

rte/main.go
```diff
@@ -225,20 +229,41 @@ func defaultHostName() string {
...
 	}
-	//https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/#topology-manager-scopes
-	return "container"
+	// still no luck? last try from conf
+	if pArgs.RTE.TopologyManagerScope == "" {
+		scopeFrom = "conf"
+		pArgs.RTE.TopologyManagerScope = conf.TopologyManagerScope
+	}
+	if pArgs.RTE.TopologyManagerScope == "" {
+		return fmt.Errorf("cannot find setting for Topology Manager Scope")
```
shajmakh, line 259: why not return the default?

shajmakh: same for the tmpolicy check above

fromanirh: good point. I'm still a bit torn between fail loudly and being able to fallback to sane defaults. I'll put some more thought in this.

test/e2e/rte/rte_local_test.go
```diff
@@ -0,0 +1,115 @@
...
+	Resourcemonitor resourcemonitor.Args
+	RTE             resourcetopologyexporter.Args
+}
+
+var _ = Describe("[rte][local][config] RTE configuration", func() {
+	Context("with the binary available", func() {
+		It("should have any default for TM params", func() {
+			args := runConfig([]string{}, map[string]string{})
+			Expect(args.RTE.TopologyManagerScope).ToNot(BeEmpty())
+			Expect(args.RTE.TopologyManagerPolicy).ToNot(BeEmpty())
```
shajmakh, line 45: moreover don't we want to make sure they have the default scope/policy (container/none)?

fromanirh: that would make the test and the impl a bit more coupled, because both they will now need external knowledge of what's the correct default. Now only impl has this knowledge, the test does not and this makes it a bit weaker but also a bit more future proof. Do we want to check for the current defaults?

rte/main.go
```diff
@@ -71,6 +73,15 @@ func main() {
 		os.Exit(0)
 	}
 
+	if parsedArgs.DumpConfig {
+		data, err := json.Marshal(parsedArgs)
+		if err != nil {
+			klog.Fatalf("encoding the configuration to JSON")
+		}
+		fmt.Printf("%s\n", string(data))
```
shajmakh, line 81: can we add the values of the environment variables as well?

test/e2e/rte/rte_local_test.go
shajmakh: right makes sense, thanks for clarifying!

rte/main.go
fromanirh: uhm, what's the use case for that?

PR #452, opened 2022-10-24
e2e: fix unsuitable free devices amount

shajmakh:
The test is failing because the test pod was scheduled on a non-target node, hence on a node that was supposed to be unsuitable because it has free resources but are not enough for allocating the test pod.

In order to make the node unsuitable, we must specify the amount of the free resource's type that is being requested by the pod, otherwise, that kind of resource will stay entirely free and will not be padded as expected, making the "unsuitable" nodes as candidates for the test pod, a thing which happened in this test case.

Adjust the unsuitable amount of free devices of Type1 making it impossible to count those nodes as candidates for the test pod.


fromanirh (commented):
/lgtm
/approve
thanks!

swatisehgal (approved):
/lgtm

PR #456, opened 2022-10-26
RTE: proper node group status reporting

fromanirh:
in https://github.com/openshift-kni/numaresources-operator/pull/431 we enabled per-node-group RTE tuning, but the status reporting wasn't there, and instead the operator just reported the not-really helpful defaults.
In this PR we report the nodegroupconfig per-MCP in the status tree, completing the feature.


fromanirh:
/hold
pending controller tests

PR #457, opened 2022-10-26
rte: add exclude list for pod namespaces/name

fromanirh:
Enable the RTE to filter out pods using namespace/name globbing (not regex) when listing through podresouces API.

The main driver for this functionality is to handle lingering terminated pods which would not be detected by the scheduler plugin, and that would make the pod fingerprint computation drift, thus stalling the scheduler.
Filtering is optional and disabled by default.

Enable the handling of the filter list from the operator.


swatisehgal (approved):
/lgtm

PR #459, opened 2022-10-27
serial: update scheduler name properly

shajmakh:
The test fails to set the scheduler name under the test's pod because nroSchedObj.Spec.SchedulerName does not exist, thus it returns an empty value, which later would be translated to the default kube scheduler name instead of the test scheduler name. The scheduler name is defined under Status not under Spec of the numaresourcesscheduler CR.

Modify the scheduler name properly in Status.SchedulerName to check the anticipated test's result.

get it from status instead of spec.


swatisehgal (approved):
/lgtm

fromanirh:
this change looks correct, and now I wonder why the test didn't fail when I run on my cluster

shajmakh:
> this change looks correct, and now I wonder why the test didn't fail when I run on my cluster

same for me. what I thought about is that the test before the kubeletconfig change changed the CR configuration by setting the spec.schedulerName value but for an interruption it was not reverted, a thing that caused the kubeletconfig test to pass because there was a value in spec.schedulerName

PR #460, opened 2022-10-27
RTE: further improvements about node group config status reporting

fromanirh:
This is another refinement of #431 and a followup of #456 
We want to normalize earlier the NodeGroupConfig (see specific commit) to make sure we do meaningful reports of the observed config, and we want to decouple the NodeGroupConfig reporting from the MCP status reporting. Even though both share the same Status object, they should be updated indpendently and in different places in the flow.

Add cleanups and refactorings to improve testability (and some tests) along the way.


fromanirh:
/hold
want to do extra testing locally, let's make sure the third time is the charm

PR #462, opened 2022-10-28
serial: support test for pod TMscope

shajmakh:
The test with its older version is supporting only TMscope=container, because it creates the test's pods with two containers each expected to be placed on different numa zones of the target node. When running this scenario on a cluster with tmscope=pod the test's pod with TAS scheduler fails to run because of insufficient resource availability on a single numa node, while the pod with default scheduler hits TAE.

Support this TC to run on TMscope=pod by padding the target node zones leaving room only for the one pod resource and adjusting the pod.Spec accordingly.

### DIFF
--- a/test/e2e/serial/tests/workload_placement.go
+++ b/test/e2e/serial/tests/workload_placement.go
@@ -499,10 +499,38 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 				ExpectWithOffset(1, failedPodIds).To(BeEmpty(), "padding pod of the target node have failed to run")
 			}
 
+			nrtPostTargetPaddingList, err := e2enrt.GetUpdated(fxt.Client, nrtInitialList, time.Second*10)
+			Expect(err).ToNot(HaveOccurred())
+
+			nrtPostTargetPadding, err := e2enrt.FindFromList(nrtPostTargetPaddingList.Items, targetNodeName)
+			Expect(err).ToNot(HaveOccurred())
+
 			pod.Spec = getPodSpec(tmPolicy, reqResPerNUMA)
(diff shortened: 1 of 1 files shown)


shajmakh:
The test scenario hits a race condition between the TAS-scheduled pod and the default-scheduler-scheduled pod, resulting in a TAE on the default-scheduler-scheduled pod. 
need to validate if this is the expected behavior, if so, then this test isn't actually relevant because the TAE is on the pod that's scheduled by default-scheduler and not something we can control in the scope of this project. another option would be adjusting the scenario so it will verify the fix of: https://github.com/kubernetes-sigs/scheduler-plugins/pull/439

shajmakh:
> The test scenario hits a race condition between the TAS-scheduled pod and the default-scheduler-scheduled pod, resulting in a TAE on the default-scheduler-scheduled pod. need to validate if this is the expected behavior, if so, then this test isn't actually relevant because the TAE is on the pod that's scheduled by default-scheduler and not something we can control in the scope of this project. another option would be adjusting the scenario so it will verify the fix of: [kubernetes-sigs/scheduler-plugins#439](https://github.com/kubernetes-sigs/scheduler-plugins/pull/439)

Turns out this was happening because the testing cluster was configured with tmscope=pod, thus reasonable that the pod which used the default scheduler would hit TAE. 
the test with it's older version supports only container scope, In this PR we are addressing this gap and supporting the test with pod scope.

Tal-or:
I think this test result would be the same for every scheduler-plugin and not TAS specifically.
IOW it looks like we're testing the framework and how it prioritized the different plugins it has.

swatisehgal (approved):
/lgtm
/approve

PR #463, opened 2022-11-02
serial: fix failing test

shajmakh:
Test case 47627 was failing on the NRT consumed resources verification after creating the replica set. The expected resources were negative (CPU: -2) although the actual NRT reflects correct consumption on the target node.
Fix the test by:
- Adding missing checks like waiting for replicaset to be ready & for pods to get deleted
- Checking for consumed resources on node level b/c there are several replicas being scheduled and there is no guarantee of placement of all the pods on one numa zone. 
Additionally to other test code cleanup and enhancements, please refer to the commits for more details.

### DIFF
--- a/internal/wait/replicaset.go
+++ b/internal/wait/replicaset.go
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2022 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package wait
+
+import (
+	"context"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func ForReplicaSetComplete(cli client.Client, rs *appsv1.ReplicaSet, pollInterval, pollTimeout time.Duration) (*appsv1.ReplicaSet, error) {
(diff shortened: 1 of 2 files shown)


shajmakh:
Pending testing on container scope

shajmakh:
> Pending testing on container scope


Test passed on both tmscopes.

internal/wait/replicaset.go
```diff
@@ -0,0 +1,60 @@
...
+			return false, nil
+		}
+
+		klog.Infof("replicaset %s complete", key.String())
+		return true, nil
+	})
+	return updatedRs, err
+}
+
+func AreReplicasAvailable(newStatus *appsv1.ReplicaSetStatus, replicas int32) bool {
```
fromanirh, line 51: nit: if you don't need helper functions, don't export them
I'd even go as far as:
- have `IsReplicasetComplete` unexported -> `isReplicasetComplete`
- inline `AreReplicasAvailable` in the only calling site (being it `isReplicasetComplete`

Also note inconsistent naming: `ForReplicaSetComplete` vs `IsReplicasetComplete`

test/e2e/serial/tests/workload_placement.go
```diff
@@ -925,32 +925,33 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
-					{
-						Name:    "c0",
-						Image:   images.GetPauseImage(),
-						Command: []string{images.PauseCommand},
-						Resources: corev1.ResourceRequirements{
-							Limits: corev1.ResourceList{
-								corev1.ResourceCPU:    resource.MustParse("2"),
-								corev1.ResourceMemory: resource.MustParse("200Mi"),
-							},
+			rsContainers := []corev1.Container{
```
fromanirh, line 928: OTOH this introduces coupling and require the reader to bounce up/down in the code to read the ReplicaSet definition.
So, while it is an improvement, is not obviously better on all fronts.

fromanirh (commented):
/lgtm
/approve

fromanirh:
some nits and minor comments, but nothing that should prevent us to merge this PR.
Good work here!

test/e2e/serial/tests/workload_placement.go
shajmakh: note taken, thank you for highlighting this!

internal/wait/replicaset.go
shajmakh: indeed, i'll address this in a later PR. Thank you for the valuable review!

PR #464, opened 2022-11-03
internal: unexport and rename internal helper functions

shajmakh:
The functions are used only in this internal file and not called in other packages, thus unexport them, and rename ForReplicaSetComplete (to areReplicasAvailable) to keep consistent naming.


fromanirh:
thanks for this cleanup!

PR #468, opened 2022-11-08
serial: pass pod's address properly to the function

shajmakh:
The tests fail when calling corev1qos.GetPodQOS with &pods[0] when pods is a slice of pods resulting in the following error:

```
Test Panicked
In [It] at: /usr/local/go/src/runtime/panic.go:89

runtime error: index out of range [0] with length 0
```

What should be actually passed to the function is the first pod's address in the slice.


fromanirh (commented):
/lgtm
/approve
I was under the impression the old syntax being fixed here should be equivalent to the way you are changing, but I may be wrong. Merging this PR won't hurt anyway.

PR #466, opened 2022-11-04
update k8s libs to 1.25.z and ginkgo to v2

fromanirh:
This PR bundles update to k8s libs 1.25.z and to ginkgo v2
We do these two big changes in one go because:

- kube 1.25 pulls ginkgo v2
- we cannot have ginkgo v1 and v2 in the same binary

so splitting the changes is not really easier than doing everything in one go.

PR #469, opened 2022-11-09
Cleanup 20221109

fromanirh:
Followup found after more field testing of #466 and #468

### DIFF
--- a/test/e2e/sched/install/install_test.go
+++ b/test/e2e/sched/install/install_test.go
@@ -96,6 +96,7 @@ var _ = Describe("[Scheduler] install", func() {
 			By("Check secondary scheduler pod is scheduled on a control-plane node")
 			podList, err := podlist.ByDeployment(e2eclient.Client, *deployment)
 			Expect(err).NotTo(HaveOccurred())
+			Expect(podList).ToNot(BeEmpty(), "cannot find any pods for DP %s/%s", deployment.Namespace, deployment.Name)
 
 			nodeList, err := nodes.GetControlPlane(e2eclient.Client, configuration.Plat)
 			Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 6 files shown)


test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -233,7 +233,8 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 
 			By(fmt.Sprintf("checking deployment pods have been handled by the topology aware scheduler %q but failed to be scheduled on any node", serialconfig.Config.SchedulerName))
 			pods, err := podlist.ByDeployment(fxt.Client, *deployment)
-			Expect(err).NotTo(HaveOccurred(), "Unable to get pods from Deployment %q:  %v", deployment.Name, err)
+			Expect(err).ToNot(HaveOccurred(), "Unable to get pods from Deployment %q:  %v", deployment.Name, err)
```
shajmakh, line 236: I guess this and similar change should be part of the second commit.

test/e2e/sched/install/install_test.go
```diff
@@ -96,6 +96,7 @@ var _ = Describe("[Scheduler] install", func() {
 			By("Check secondary scheduler pod is scheduled on a control-plane node")
 			podList, err := podlist.ByDeployment(e2eclient.Client, *deployment)
 			Expect(err).NotTo(HaveOccurred())
+			Expect(podList).ToNot(BeEmpty(), "cannot find any pods for DS %s/%s", deployment.Namespace, deployment.Name)
```
shajmakh, line 99: here and below , if i'm not mistaken it should  be DP (deployment) instead of DS (daemonset)

fromanirh: right, thanks! will fix

fromanirh:
> thank you for the cleanup! I hope this commit doesn't cause any conflicts in cases of backporting in the future. do we even think to backport the tests changes?

nope, we usually don't backport changes in the serial suite, because we want to keep the suite compatible all across the supported versions. Things can change only when we graduate the API to v1, though.

fromanirh:
I guess I can drop the second commit though, let's narrow down the changes.

PR #470, opened 2022-11-09
serial: update schedulerName via Spec and verify update is successful

shajmakh:
The test failed to update the scheduler name to a new value because it was updating the SchedulerName under Status, while in general, updating a field in k8s object should be done via Spec and the active value of the fields should be fetched from Status.
Fix the test so it modifies the field via Spec and fetches the active value from Status and verify the new value is in place.


swatisehgal (commented):
/lgtm

fromanirh (commented):
/lgtm
/approve

PR #471, opened 2022-11-09
serial tests fixes

shajmakh:
The PR addresses a fix for tests that failed because of checking empty pods that are related to resources like deployment/daemon sets. Please see the different commit messages for more details on each modification.

### DIFF
--- a/internal/wait/daemonset.go
+++ b/internal/wait/daemonset.go
@@ -59,3 +59,24 @@ func AreDaemonSetPodsReady(newStatus *appsv1.DaemonSetStatus) bool {
 	return newStatus.DesiredNumberScheduled > 0 &&
 		newStatus.DesiredNumberScheduled == newStatus.NumberReady
 }
+
+func ForDaemonsetPodsCreation(cli client.Client, ds *appsv1.DaemonSet, expectedPods int, pollInterval, pollTimeout time.Duration) (*appsv1.DaemonSet, error) {
+	key := ObjectKeyFromObject(ds)
+	updatedDs := &appsv1.DaemonSet{}
+	err := wait.PollImmediate(pollInterval, pollTimeout, func() (bool, error) {
+		err := cli.Get(context.TODO(), key.AsKey(), updatedDs)
+		if err != nil {
+			klog.Warningf("failed to get the daemonset %s: %v", key.String(), err)
+			return false, err
+		}
+
+		if int(updatedDs.Status.DesiredNumberScheduled) != expectedPods {
+			klog.Warningf("Waiting for daemonset: %q to have %d pods, current number of created pods: %d", key.String(), expectedPods, updatedDs.Status.DesiredNumberScheduled)
+			return false, nil
+		}
+
+		klog.Infof("pods of daemonset %q are all created", key.String())
+		return true, nil
+	})
+	return updatedDs, err
+}
--- a/test/e2e/serial/tests/workload_unschedulable.go
+++ b/test/e2e/serial/tests/workload_unschedulable.go
@@ -107,13 +107,13 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 	Context("with no suitable node", func() {
 		var requiredRes corev1.ResourceList
 		var nrtListInitial nrtv1alpha1.NodeResourceTopologyList
-
+		var nrtCandidates []nrtv1alpha1.NodeResourceTopology
 		BeforeEach(func() {
 			neededNodes := 1
 
 			requiredNUMAZones := 2
 			By(fmt.Sprintf("filtering available nodes with at least %d NUMA zones", requiredNUMAZones))
-			nrtCandidates := e2enrt.FilterZoneCountEqual(nrts, requiredNUMAZones)
+			nrtCandidates = e2enrt.FilterZoneCountEqual(nrts, requiredNUMAZones)
 			if len(nrtCandidates) < neededNodes {
 				e2efixture.Skipf(fxt, "not enough nodes with 2 NUMA Zones: found %d, needed %d", len(nrtCandidates), neededNodes)
 			}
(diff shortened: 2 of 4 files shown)


test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -676,8 +676,23 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 			err = fxt.Client.Create(context.TODO(), dp)
 			Expect(err).ToNot(HaveOccurred())
 
-			By(fmt.Sprintf("checking deployment pods have been scheduled with the topology aware scheduler %q ", schedulerName))
+			// although the deployment pods will be pending thus the deployment will not be counted as complete,
```
fromanirh, line 679: this is to catch existing flakes or is it a prevention mechanism?

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -676,8 +676,23 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+			// we need to wait until all the replicas are created despite their status before moving forward with the checks
+			dpKey := client.ObjectKeyFromObject(dp)
+			Eventually(func() bool {
+				err = fxt.Client.Get(context.TODO(), dpKey, dp)
+				Expect(err).ToNot(HaveOccurred())
+
+				if dp.Status.Replicas != replicas {
```
fromanirh, line 686: I wonder if we should not just wrap the next checks in a `Eventually` loop, because this check is implicitely expecting the replicas will go up. Basically, we are assuming a positive test. This may be fine here, but I'm literally not sure it holds true in general

fromanirh (commented):
I'm generally OK with the direction but I'd like to think a bit more about the context.

test/e2e/serial/tests/workload_unschedulable.go
shajmakh: its goal is to prevent having flakiness in failures in upcoming checks

fromanirh: ack, good.

test/e2e/serial/tests/workload_unschedulable.go
fromanirh: What I think now is: is the reminder of the test (implicitely) expecting the replicas to be running, or is checking if the replica are actually all running part of the test itself?
IOW, are we:

- making an implicit expectation explicit?
- factoring a common expectation?
- something else?

shajmakh: the test consists of two phases:
1. create a deployment that all its replicas keep on pending (because there are not enough resources)
2. update the requested resources under the deployment so that some of the replicas go running 
so the test is not completely negative, the tag [unsched] might be a bit misleading for this test.
Note though that we do expect the replicas will go up when in the first phase in the test, yet that doesn't mean they are running (running=ready=available), so the deployment pods are there but pending, which is the anticipated behavior.

internal/wait/daemonset.go
```diff
@@ -59,3 +59,24 @@ func AreDaemonSetPodsReady(newStatus *appsv1.DaemonSetStatus) bool {
 	return newStatus.DesiredNumberScheduled > 0 &&
 		newStatus.DesiredNumberScheduled == newStatus.NumberReady
 }
+
+func ForDaemonsetPodsCreation(cli client.Client, ds *appsv1.DaemonSet, expectedPods int, pollInterval, pollTimeout time.Duration) (*appsv1.DaemonSet, error) {
```
fromanirh, line 63: this is a very nice followup to the conversation we had. I like it!

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -289,40 +289,21 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 		})
 
 		It("[test_id:47619][tier3][unsched][default-scheduler] a deployment with a guaranteed pod resources available on one node but not on a single numa; scheduled by default scheduler", func() {
+			By("Scheduling the testing pod")
```
fromanirh, line 292: the code you are adding fixes the issue at hand, but isn't the test trying to exercise exactly this case?
IOW, in this case we do want exactly a Deployment to trigger this scenario (if memory serves).

What the test is supposed to do is to check at any random time that the pods pertaining to this deployment are failing as expected.
This alone should not drag the cluster down though.

Additionally, this test didn't cause any major issues in my own test runs, so it would be nice to understand why is failing/flaking.

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -298,12 +298,14 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 			Expect(err).NotTo(HaveOccurred(), "unable to create pod %q", pod.Name)
 
 			By(fmt.Sprintf("Verify pod failed to be scheduled by default scheduler %q and it has TopologyAffinityError in its events", corev1.DefaultSchedulerName))
-			isFailed, err := nrosched.CheckPODKubeletRejectWithTopologyAffinityError(fxt.K8sClient, pod.Namespace, pod.Name)
-			if err != nil {
-				_ = objects.LogEventsForPod(fxt.K8sClient, pod.Namespace, pod.Name)
-			}
-			Expect(err).ToNot(HaveOccurred())
-			Expect(isFailed).To(BeTrue(), "pod %s/%s with scheduler %s did NOT fail", pod.Namespace, pod.Name, corev1.DefaultSchedulerName)
+			Eventually(func() bool {
```
fromanirh, line 301: this is a fine change

fromanirh (commented):
Some questions inside.

test/e2e/serial/tests/workload_unschedulable.go
shajmakh: The test indeed wanted to test the scenario on deployment/replicaset. 

> What the test is supposed to do is to check at any random time that the pods pertaining to this deployment are failing as expected.
This alone should not drag the cluster down though.

I agree on this except for the term random time. The automation will check all the deployment's pods they hit TAE, but in case the pod was checked in a time that the TAE event wasn't there yet the test would fail. 
this led me to simplify the test to check a single pod and adding the Eventually part. 
However thinking about it again now (and testing it) I could have added the eventually without simplifying the test, so that the event check will be performed on each pod in an Eventually. 
ran the test with the new fix couple of times and it seems stable now.

shajmakh: However, note that we fetch the deployment's pods only once so the test will check only a sample of the deployment's pods, which is enough and makes the test on the safe side than getting stuck in a loop of checking new pods every time.

shajmakh: thanks for pointing this out!
the code is updated now

PR #472, opened 2022-11-10
hack: add ginkgo v2 timeout for serial

shajmakh:
Migrating ginkgo to V2 causes the suites to exit after 1 hour by default: https://onsi.github.io/ginkgo/MIGRATING_TO_V2#timeout-behavior `Finally, the default timeout has been reduced from 24h down to 1h. Users with long-running tests may need to adjust the timeout in their CI scripts.` Serial suite needs more than that, a thing that causes to skip around 18 tests, so increase that timeout to 5 hours for now.

### DIFF
--- a/hack/run-test-serial-e2e.sh
+++ b/hack/run-test-serial-e2e.sh
@@ -104,6 +104,7 @@ function setup() {
 	echo "Running NRO install test suite"
 	runcmd ${BIN_DIR}/e2e-nrop-install.test \
 		--ginkgo.v \
+		--ginkgo.timeout="5h" \
 		--ginkgo.fail-fast \
 		--ginkgo.junit-report=${REPORT_DIR}/e2e-serial-install.xml \
 		--ginkgo.focus='\[Install\] continuousIntegration' \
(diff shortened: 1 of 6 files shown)


hack/run-test-serial-e2e.sh
```diff
@@ -156,7 +156,8 @@ function runtests() {
 		--ginkgo.junit-report=${REPORT_FILE} \
 		${NO_COLOR} \
 		${SKIP} \
-		${FOCUS}
+		${FOCUS} \
+		--ginkgo.timeout="5h"
```
fromanirh, line 160: the fix makes sense. Please move it just after `--ginkgo.v` only because cosmetic reasons (no functional change)

fromanirh (commented):
Please fix the expected output in `https://github.com/openshift-kni/numaresources-operator/blob/main/hack/test-e2e-runner.sh#L29`

fromanirh:
/hold cancel
the fix looks good (and fully makes sense) but it's partial

hack/run-test-serial-e2e.sh
shajmakh: done, thanks!

fromanirh:
/hold
@shajmakh offline pointed out we need the same change for the other suites

shajmakh:
/hold 
some ci failed :)

PR #479, opened 2022-12-05
api: ensure backward compatibility

fromanirh:
during the 4.12 cycle we added APIs which weren't backward compatible. Address them and make them actually backward compatible before 4.12 release. This was made evident running the 4.12 suite against older clusters. Alongside the needed API fixes, enhance the e2e serial suite to tolerate older clusters.


fromanirh:
need to backport down to 4.11

swatisehgal (approved):
/lgtm

PR #489, opened 2022-12-06
[release-4.11][manual] api: ensure backward compatibility

fromanirh:
this is a manual partial cherry-pick of https://github.com/openshift-kni/numaresources-operator/pull/487

during the 4.12 cycle we added APIs which weren't backward compatible. Address them and make them actually backward compatible before 4.12 release. This was made evident running the 4.12 suite against older clusters. Alongside the needed API fixes, enhance the e2e serial suite to tolerate older clusters.


swatisehgal (approved):
/lgtm
/approve

api/numaresourcesoperator/v1alpha1/numaresourcesoperator_types.go
```diff
@@ -104,19 +107,22 @@ type NUMAResourcesOperatorStatus struct {
...
 }
 
 // MachineConfigPool defines the observed state of each MachineConfigPool selected by node groups
 type MachineConfigPool struct {
 	// Name the name of the machine config pool
 	Name string `json:"name"`
-
 	// Conditions represents the latest available observations of MachineConfigPool current state.
 	// +optional
+	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="Optional conditions reported for this NodeGroup"
```
yanirq, line 120: are we tied to operator-sdk here and in future versions ?

fromanirh: could you elaborate please?

yanirq: using operator-sdk frameowrk here ties us to the API it provides (which is ok)  
In case we will want to break the scheme and move to a more native one then it will require refactoring. 
I dont think this will be the case here so we should be fine.

PR #502, opened 2023-01-10
chore: internal: rename import to nropv1alpha1

ffromani:
We use the renamed import `nropv1alpha` to pull in the API package _almost_ everywhere. Fix the few places where we don't, for uniformity's sake and to make the future v1 API transition a little easier.

No intended change in behavior.


swatisehgal (approved):
/lgtm

PR #512, opened 2023-01-18
[release-4.12][manual] chore: make the api package import rename uniform

ffromani:
make sure we rename the api packages always using the same name, which will make transition to newer API easier.

No changes in behaviour


swatisehgal (approved):
/lgtm

PR #513, opened 2023-01-18
api: layout: api/numaresourcesoperator/v* -> api/v*

ffromani:
trivial path rename to comply with layout standards; no changes in content or behaviour, trivial
update change.

We don't have yet any consumer of the operator API (nor we probably should), so this is still a good
time to do this invasive (albeit simple) change.

But why did we had the `api/numaresources/v1alpha` layout in the first place?
The exact reason is lost in the beginning of the repo, but as far as the original initiator (@ffromani) memory goes, this was never a conscious design decision, rather an artifact of how `operator-sdk` was used (possibly even with some unusual but legal parameter combination).

Further hints that the current layout is historical artifact can be seen in the layout of the spiritual ancestor of
numaresources-operator: 
https://github.com/openshift-kni/rte-operator/commit/3570aba082154d3dbf99f6947f484365dd58c58a


ffromani:
/hold
needs design review

swatisehgal (approved):
/lgtm

No objections with the changes proposed here.

/hold
to allow time for additional feedback from other team members.

ffromani:
we had a braingstorming session and while this makes the code nicer, we are trying to avoid changes which can hurt backportability. So we are postponing this work for now

PR #519, opened 2023-01-26
[manual][release-4.12] add support for scheduler cache using the reserve plugin

ffromani:
Enable the scheduler-side cache using the reserve plugin. The functionality is controlled solely by a new optional scheduler spec parameter, `CacheResyncSeconds`. If present and > 0, it will enable the functionality.
Default is disabled for backward compatibility.

Manual backport of https://github.com/openshift-kni/numaresources-operator/pull/443


ffromani:
/hold
let's check the backporting timeline before

ffromani:
/hold cancel
if the change is not backward compatible as it should, the sooner we know the better

ffromani:
/retest
failed for a different reason (installplan failed)

ffromani:
So:

1. we will need a manual, albeit trivial, rebase because we backported PRs in a different order than the one was used for main branch. This is partly my (ffromani's) fault, partly a byproduct of circumstances. We will revisit the backport process and improve where possible
2. the RTE errors we are experiencing are IMO red herrings. The core issue is the scheduler plugin was NOT backported yet. But the operator always sends the `cacheResyncPeriod` param even when 0 (to allow the enabled->disabled transition in the easiest way). So the scheduler plugin is likely to fail to startup (not demonstrated but all the hints point there) which cause cascading failures later on
3. it would be nice to investigate further why the `exporter` binary is missing and if we are silently skipping tests, but this, albeit important, is a separate task.

ffromani:
reopen because the reserve plugin/cache support was backported to scheduler-plugin 4.12, and new container image have been built by automation

ffromani:
/hold cancel
all concerns addressed (there could be lower-priority followups later)

swatisehgal (approved):
/lgtm
/approve

NIT: I would add reference to the original [PR](https://github.com/openshift-kni/numaresources-operator/pull/443) being backported in the PR description.

ffromani:
> /lgtm /approve
> 
> NIT: I would add reference to the original [PR](https://github.com/openshift-kni/numaresources-operator/pull/443) being backported in the PR description.

good point, fixed

PR #522, opened 2023-01-31
Update prow configuration 20230131

ffromani:
update aliases set and perform long due update to the names of the groups


swatisehgal (approved):
/lgtm
/approve

PR #525, opened 2023-02-01
[manual][release-4.12] API: add v1 

ffromani:
Add v1 API. The v1 API is v1alpha1 with the deprecated `DisablePodsFingerprinting` removed, and` NodeGroupConfig` cleaned up accordingly.

We add preparation for webhook, including conversion functions, but this PR does NOT wire them, intentionally.
Deploying conversion webhooks would require changes in the CSV, which would in turn break the upgrade flow (upgrades will still be possible, but they will require manual fixes), and would make the backport even harder.

The consequence of no conversion hooks is the conversion is unidirectional v1alpha1 -> v1.
The operator will **accept** both v1alpha1 and v1, but it will only **serve** v1. Deploying manifests won't be affected, but code interacting with the operator needs to be prepared to handle v1 objects as response.

This means we need to update all our testsuite, and that the serial suite is no longer backward compatible down to `release-4.10` branch.

All the above is very very similar to what we did in https://github.com/openshift-kni/performance-addon-operators/pull/305

manual backport of #505


swatisehgal (approved):
/lgtm
/approve

PR #549, opened 2023-04-26
e2e: serial: fix expected event on error

ffromani:
Catch up with u/s https://github.com/kubernetes-sigs/scheduler-plugins/pull/547

We fix the expected message within events when a container/pod is unschedulable. Previously, we had to use custom strings, but now we can leverage the constants declared in the `nrosched` utility package, on which we depend already anyway.


swatisehgal (approved):
/lgtm

shajmakh:
Thanks for the fix, we need it in 4.13 test version too.
/cherry-pick release-4.13

PR #551, opened 2023-05-01
serial(v1beta1): set expected scheduler to empty for default sched

shajmakh:
A test that verifies that a pod scheduled with default scheduler failed because the expected Event.ReportingController is not "default-scheduler". Looking into the issue, it turned out that Event.ReportingController is set only if the scheduler is non-default starting v1beta1, if it is default the field is left empty.

Adopt this update in the test and pass an empty value as the expected scheduler.

### DIFF
--- a/test/e2e/serial/tests/workload_placement.go
+++ b/test/e2e/serial/tests/workload_placement.go
@@ -988,7 +988,9 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 
 			By(fmt.Sprintf("checking the pods were scheduled with scheduler %q", corev1.DefaultSchedulerName))
 			for _, pod := range pods {
-				schedOK, err := nrosched.CheckPODWasScheduledWith(fxt.K8sClient, pod.Namespace, pod.Name, corev1.DefaultSchedulerName)
+				// starting v1beta1, if a pod is scheduled by the default scheduler, the ReportingController is left empty (unset)
+				// thus we pass an empty string as the expected scheduler that took charge of scheduling the workload.
+				schedOK, err := nrosched.CheckPODWasScheduledWith(fxt.K8sClient, pod.Namespace, pod.Name, "")
 				Expect(err).ToNot(HaveOccurred())
 				Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s", pod.Namespace, pod.Name, corev1.DefaultSchedulerName)
 


shajmakh:
/hold
requires testing. No available cluster currently.

test/e2e/serial/tests/workload_placement.go
```diff
@@ -988,7 +988,9 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 
 			By(fmt.Sprintf("checking the pods were scheduled with scheduler %q", corev1.DefaultSchedulerName))
 			for _, pod := range pods {
-				schedOK, err := nrosched.CheckPODWasScheduledWith(fxt.K8sClient, pod.Namespace, pod.Name, corev1.DefaultSchedulerName)
+				// starting v1beta1, if a pod is scheduled by the default scheduler, the ReportingController is left empty (unset)
+				// thus we pass an empty string as the expected scheduler that took charge of scheduling the workload.
+				schedOK, err := nrosched.CheckPODWasScheduledWith(fxt.K8sClient, pod.Namespace, pod.Name, "")
```
ffromani, line 993: I don't think `v1beta1` is relevant here. This is a behavior of the default scheduler, which is completely unaware of all the TAS machinery - so the latter is hardly relevant.
While I'm actually kinda OK with the fix proper, I'd like to have a look into the k8s codebase to learn a bit more, I find weird the default scheduler is not setting `ReportingController` in the event.

ffromani:
@shajmakh on which version we see this error?

test/e2e/serial/tests/workload_placement.go
shajmakh: I'll research more the root cause of not setting the def-sched and update here.

shajmakh:
> @shajmakh on which version we see this error?

I've seen this on the available 4.13 results. Haven't checked yet 4.12. but I recall before my leave that was not the case in 4.12. I'll double-verify that as well.

ffromani:
> > @shajmakh on which version we see this error?
> 
> I've seen this on the available 4.13 results. Haven't checked yet 4.12. but I recall before my leave that was not the case in 4.12. I'll double-verify that as well.

interesting, because is not happening here on

```
$ oc version
Client Version: 4.13.0-XXXXX
Kustomize Version: v4.5.7
Server Version: 4.13.0-0.nightly-2023-04-01-062001
Kubernetes Version: v1.26.2+7195e44
```

this means something fishy is going on and we need to spend more time on this fix.
I'll send a PR shortly aiming to improve the debuggability.

shajmakh:
Closing this PR after getting to the conclusion that this was a bug in one of the OCP builds and is fixed in the latest.

PR #552, opened 2023-05-02
e2e: nrosched: add extra logs and filters

ffromani:
We are getting a flake lately on which we see default scheduler events missing the `ReportingController` field populated which is unexpected and, well, makes the test flake.
Add extra logs and more precise filtering to narrow down the issue.
xref: https://github.com/openshift-kni/numaresources-operator/pull/551


shajmakh:
/lgtm
Thanks for the PR.

PR #558, opened 2023-05-18
validator: add validation for scheduler cache

ffromani:
Add helper packages to check the state of the scheduler cache. The first and most obvious troubleshooting step is to check if the scheduler cache is clean (vs dirty, thus not sending pods to a node because the overallocation).

To do so, we rely on the tracing functionality introduced in podfingerprint package >= 0.1.0. Components computing the pod fingerprint can now expose the debug data in JSON format, which this code reads, consumes and crosschecks.

The output can tell the client (human or machine) if the scheduler cache is clean for each node, and if not, why the computation drifted.

Drifted is caused by different perspective on which pods are present on which node. There's only one key scenario on which this happen and it is not eventually self-healed: lingering terminated pods, which stay around and are terminated *before* the scheduler plugin starts.

Arguably, the scheduler plugin (aka the go k8s client) can make steps to make sure it gets these pods, but until (if) this happens, better to detect this scenario.


ffromani:
I'm thinking about splitting into multiple commits (no real reason for not having done that).

ffromani:
/cherry-pick release-4.13

(too hasty before)

PR #562, opened 2023-05-22
rte: enable PFP computation considering only pods with exclusive resources assigned

ffromani:
Add option (new enum value in `NodeConfig`) to compute the PFP using only pods which have exclusive resources allocated, which are also the only kind of resources which can be accounted per-NUMA.

IOW considering all pods actually create unnecessary noise.
Accounting only exclusive resources is the most correct thing to do.

Note this is meant to (greatly) improve the jobs situation, but not completely solve it.
The kubelet is supposed to return the correct set of pods straight on, so it's likely we have a bug in the kubelet side.

### DIFF
--- a/pkg/objectupdate/rte/rte.go
+++ b/pkg/objectupdate/rte/rte.go
@@ -207,7 +207,8 @@ func isPodFingerprintEnabled(conf *nropv1.NodeGroupConfig) bool {
 		// not specified -> use defaults
 		conf = &cfg
 	}
-	return *conf.PodsFingerprinting == nropv1.PodsFingerprintingEnabled
+	return (*conf.PodsFingerprinting == nropv1.PodsFingerprintingEnabled ||
+		*conf.PodsFingerprinting == nropv1.PodsFingerprintingEnabledExclusiveResources)
 }
 
 func isNotifyFileEnabled(conf *nropv1.NodeGroupConfig) bool {
--- a/pkg/objectupdate/rte/rte.go
+++ b/pkg/objectupdate/rte/rte.go
@@ -19,6 +19,7 @@ package rte
 import (
 	"fmt"
 	"path/filepath"
+	"strconv"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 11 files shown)


pkg/objectupdate/rte/rte.go
```diff
@@ -146,11 +147,12 @@ func DaemonSetArgs(ds *appsv1.DaemonSet, conf nropv1.NodeGroupConfig) error {
 		flags.SetOption("--sleep-interval", refreshPeriod)
 	}
 
-	pfpEnabled := isPodFingerprintEnabled(&conf)
+	pfpEnabled, pfpRestricted := isPodFingerprintEnabled(&conf)
 	klog.V(2).InfoS("DaemonSet update: pod fingerprinting status", "daemonset", ds.Name, "enabled", pfpEnabled)
 	if pfpEnabled {
 		flags.SetToggle("--pods-fingerprint")
 		flags.SetOption("--pods-fingerprint-status-file", filepath.Join(pfpStatusDir, "dump.json"))
+		flags.SetOption("--pods-fingerprint-unrestricted", strconv.FormatBool(!pfpRestricted)) // note the "not"!
```
swatisehgal, line 155: Perhaps it is better to have a variable name that is more expressive here.  It is not very descriptive and it is not very evident from here what is the restriction about?  How about `--pods-fingerprint-exclusive-resource-only` ?

pkg/objectupdate/rte/rte_test.go
```diff
@@ -74,7 +74,7 @@ func TestUpdateDaemonSetArgs(t *testing.T) {
 			name: "defaults",
 			conf: nropv1.DefaultNodeGroupConfig(),
 			expectedArgs: []string{
-				"--pods-fingerprint", "--refresh-node-resources", "--sleep-interval=10s", "--notify-file=/run/rte/notify",
+				"--pods-fingerprint", "--pods-fingerprint-unrestricted=false", "--refresh-node-resources", "--sleep-interval=10s", "--notify-file=/run/rte/notify",
```
swatisehgal, line 77: Another reason I am not a fan of this name is because it inherently has a negative in the name which makes this confusing. I would prefer it it was `--pods-fingerprint-exclusive-resource-only=true` making the code more readable and clear on what is expected here.

pkg/objectupdate/rte/rte.go
ffromani: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/issues/177

pkg/objectupdate/rte/rte_test.go
ffromani: point taken!

swatisehgal (approved):
/lgtm

Thanks for filing the issue. I don't want this PR to be blocked because of naming issue.

/hold
to allow time in case others would like to review. Feel free to unhold when you think it is suitable.

ffromani:
> /lgtm
> 
> Thanks for filing the issue. I don't want this PR to be blocked because of naming issue.
> 
> /hold to allow time in case others would like to review. Feel free to unhold when you think it is suitable.

it seems more and more we can't backport the option rename in v0.11 - but we can and we will fix in 0.12. However 0.12 is nowehere ready yet and we should not wait for 0.12.
We handle this rename internally, is not user-visible. We can move forward.
So, I'm un-holding this PR.

PR #563, opened 2023-05-22
serial: fix reboot testcase and properly wait on mcp updates

shajmakh:
Running test 47674 was leaving crashed mcp settings on one of the nodes causing the mcp to stuck in updating status forever, hence incompatible env for the next test to run.
    The reason behind this is that the test was not waiting properly for the updates on relevant mcps to start, and then finish updating.
    When updating the NROP label to the new mcp, this triggers an update on the old mcp, which takes time for it to finish. Before this PR, the test was only waiting on the new mcp to get updated while ignoring the old mcp status. The test continues to run the later checks and even tearsdown while the old mcp was still in updating status. One part of the teardown is restoring initial labels on the nodes and at last, delete the new mcp that was created for the test purpose. Added all up: initial mcp is being updated, nodes labels are restored and new mcp is deleted results in a situation of a node searching for not-found mcp causing the initial mcp to stuck in updating because one node is in degraded state. For example:
    
    Node worker-1 is reporting: "machineconfig.machineconfiguration.openshift.io \"rendered-mcp-test-36f40ba3c1038c7ce5ce54f8a840a58f\" not found"

    
  To resolve this, add waiting loops on each mcp like the following:
  once the operator object is updated with the new mcp:
  1. wait on the new mcp to get updated.
  2. wait on the initial mcp to start updating.
  3. wait on the initial mcp to complete updating.
  
  And by the end of the test when restoring initial configuration:
  1. wait on the initial mcp to start updating.
  2. wait on the initial mcp to complete updating.

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -147,7 +147,25 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
 				Expect(err).ToNot(HaveOccurred())
 
 				var wg sync.WaitGroup
+				By("waiting for mcp to start updating")
 				for _, mcp := range mcps {
+					wg.Add(1)
+					klog.Infof("wait for mcp %q to start updating", mcp.Name)
+					go func(mcpool *machineconfigv1.MachineConfigPool) {
+						defer GinkgoRecover()
+						defer wg.Done()
+						err = wait.With(fxt.Client).
+							Interval(configuration.MachineConfigPoolUpdateInterval).
+							Timeout(configuration.MachineConfigPoolUpdateTimeout).
+							ForMachineConfigPoolCondition(context.TODO(), mcpool, machineconfigv1.MachineConfigPoolUpdating)
+						Expect(err).ToNot(HaveOccurred())
+					}(mcp)
+				}
+				wg.Wait()
+
+				By("wait for mcp to get updated")
+				for _, mcp := range mcps {
+					klog.Infof("wait for mcp %q to get updated", mcp.Name)
 					wg.Add(1)
 					go func(mcpool *machineconfigv1.MachineConfigPool) {
 						defer GinkgoRecover()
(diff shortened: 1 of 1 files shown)


test/e2e/serial/tests/configuration.go
```diff
@@ -216,13 +216,22 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
...
-			for i := range nroOperObj.Spec.NodeGroups {
-				nroOperObj.Spec.NodeGroups[i].MachineConfigPoolSelector.MatchLabels = mcp.Labels
-			}
-
-			// TODO: this shoould be retried
-			err = fxt.Client.Update(context.TODO(), nroOperObj)
-			Expect(err).ToNot(HaveOccurred())
+			// for i := range nroOperObj.Spec.NodeGroups {
+			// 	nroOperObj.Spec.NodeGroups[i].MachineConfigPoolSelector.MatchLabels = mcp.Labels
+			// }
```
ffromani, line 221: please remove commented code

ffromani (commented):
looks nice. Good work.
I wonder how often the bug fixed here triggers - can't pinpoint failures I've seen lately to this bug, but I do routinely skip some tests on my env, that could be relevant.

Please remove commented out code and we're good to merge.

ffromani:
> > looks nice. Good work. I wonder how often the bug fixed here triggers - can't pinpoint failures I've seen lately to this bug, but I do routinely skip some tests on my env, that could be relevant.
> > Please remove commented out code and we're good to merge.
> 
> The bug was happening all the time (at least I observed it in all 4.13 runs). In CI we run reboot tests separately from other tests and we trigger their run after the rest of the tests. We have 2 tests requiring reboot. and this one always runs first leaving a misconfigured mcp for the next test, causing it to skip because there are "not enough nodes" with tmp policy.

ok, this makes sense indeed (and reboot requiring tests are among the ones I seldom run). Thanks for catching this.

test/e2e/serial/tests/configuration.go
shajmakh: thanks for pointing this out. done.

PR #587, opened 2023-06-01
api: scheduler: add tunable for dedicated informer

ffromani:
api: scheduler: add tunable for dedicated informer

Expose and enable by default a scheduler spec option to make the KNI-specific scheduler plugin to use a separate scheduler, not the framework-provided.
This will ensure the resync algo eventually converges.


ffromani:
/hold

let's sort out the test situation

ffromani:
still missing: sched controller tests. e2e tests will be added on a separate PR.

ffromani:
no longer WIP: e2e tests will be added in a separate PR

PR #595, opened 2023-06-14
manifests: drop obsolete node-role labels

ffromani:
We can now use the modern node-role labels to identify control plane nodes. No need to use anymore obsolete terms.


shajmakh:
@ffromani thanks for the PR. Is this needed following a new change or was this available from the beginning?

swatisehgal (approved):
/lgtm

ffromani:
> @ffromani thanks for the PR. Is this needed following a new change or was this available from the beginning?

new recommended role names are available from a while. The old role names will be removed "soon" so let's not lag too much behind, we use old names only because obsolete defaults.

However the CI failures look legit, I'll have a look later

ffromani:
/test ci-e2e

just 1 (one) retry until I fix my local env

ffromani:
/retest

looks like a glitch

PR #598, opened 2023-06-17
e2e: serial: resacct improve and generalize expectations

ffromani:
We should in general very, VERY careful when factorizing code in e2e tests.
We totally DO NOT want to add coupling to tests, and when we factor code we need to make really sure we're factoring code which does so general that a failure or a bug _always_ means a critical failure in the context of tests.

Easy candidates are setup helpers and expectations.

Turns out we check quite some expectations about proper NRT resource accounting, so let's reuse expectations for the
most obvious candidate sites.
Previously, we just checked for _A_ change in the NRT list, which was very crude and caused frequent flakes.
Now we wait for a targeted change in the very NRT object we care about, then we check that change is what we expected.
This is more precise, more correct and, at least in the initial runs, less flaky.

### DIFF
--- a/test/e2e/serial/tests/workload_placement.go
+++ b/test/e2e/serial/tests/workload_placement.go
@@ -1002,7 +1002,6 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 
 			By(fmt.Sprintf("verifying resources allocation correctness for NRT target: %q", targetNodeName))
 			var nrtAfterRSCreation nrtv1alpha2.NodeResourceTopologyList
-			podQoS := corev1qos.GetPodQOS(&pods[0])
 			Eventually(func() bool {
 				nrtAfterRSCreation, err := e2enrt.GetUpdated(fxt.Client, nrtInitial, timeout)
 				Expect(err).ToNot(HaveOccurred())
@@ -1017,18 +1016,8 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 
 				rl := e2ereslist.FromReplicaSet(*rs)
 
-				dataBefore, err := yaml.Marshal(nrtInitialTarget)
-				Expect(err).ToNot(HaveOccurred())
-				dataAfter, err := yaml.Marshal(updatedTargetNrt)
-				Expect(err).ToNot(HaveOccurred())
-				match, err := e2enrt.CheckNodeConsumedResourcesAtLeast(*nrtInitialTarget, *updatedTargetNrt, rl, podQoS)
-				Expect(err).ToNot(HaveOccurred())
-
-				if match == "" {
-					klog.Warningf("inconsistent accounting: no resources consumed by the running pod,\nNRT before test replicaset: %s \nNRT after: %s \npod resources: %v", dataBefore, dataAfter, e2ereslist.ToString(rl))
-					return false
-				}
-				return true
+				_, match := checkNRTConsumedResources(fxt, *nrtInitialTarget, rl, &pods[0])
+				return match != ""
 			}).WithTimeout(timeout).WithPolling(10 * time.Second).Should(BeTrue())
 
 			By(fmt.Sprintf("deleting replicaset %s/%s", fxt.Namespace.Name, rsName))
@@ -1106,7 +1095,6 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			}
 
 			By(fmt.Sprintf("verifying resources allocation correctness for NRT target: %q", targetNodeName))
-			podQoS = corev1qos.GetPodQOS(&pods[0])
 			Eventually(func() bool {
 				nrtAfterDPCreation, err := e2enrt.GetUpdated(fxt.Client, nrtInitial, timeout)
 				Expect(err).ToNot(HaveOccurred())
@@ -1121,18 +1109,8 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 
 				rl := e2ereslist.FromReplicaSet(*rs)
 
-				dataBefore, err := yaml.Marshal(nrtInitialTarget)
-				Expect(err).ToNot(HaveOccurred())
-				dataAfter, err := yaml.Marshal(updatedTargetNrt)
-				Expect(err).ToNot(HaveOccurred())
-				match, err := e2enrt.CheckNodeConsumedResourcesAtLeast(*nrtInitialTarget, *updatedTargetNrt, rl, podQoS)
-				Expect(err).ToNot(HaveOccurred())
-
-				if match == "" {
-					klog.Warningf("inconsistent accounting: no resources consumed by the running pod,\nNRT before test replicaset: %s \nNRT after: %s \npod resources: %v", dataBefore, dataAfter, e2ereslist.ToString(rl))
-					return false
-				}
-				return true
+				_, match := checkNRTConsumedResources(fxt, *nrtInitialTarget, rl, &pods[0])
+				return match != ""
 			}).WithTimeout(timeout).WithPolling(10 * time.Second).Should(BeTrue())
 
 			By(fmt.Sprintf("comparing scheduling times between %q and %q", corev1.DefaultSchedulerName, serialconfig.Config.SchedulerName))
--- a/test/e2e/serial/tests/workload_placement_tmpol.go
+++ b/test/e2e/serial/tests/workload_placement_tmpol.go
@@ -27,7 +27,6 @@ import (
 	corev1qos "k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"github.com/ghodss/yaml"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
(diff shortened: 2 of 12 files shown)


test/e2e/serial/tests/workload_placement.go
```diff
@@ -1017,18 +1016,8 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
-				Expect(err).ToNot(HaveOccurred())
-				match, err := e2enrt.CheckNodeConsumedResourcesAtLeast(*nrtInitialTarget, *updatedTargetNrt, rl, podQoS)
-				Expect(err).ToNot(HaveOccurred())
-
-				if match == "" {
-					klog.Warningf("inconsistent accounting: no resources consumed by the running pod,\nNRT before test replicaset: %s \nNRT after: %s \npod resources: %v", dataBefore, dataAfter, e2ereslist.ToString(rl))
-					return false
-				}
-				return true
+				_, match := checkNRTConsumedResources(fxt, *nrtInitialTarget, rl, &pods[0])
```
shajmakh, line 1019: checkNRTConsumedResources (now) checks resources consumption on numa zones. need to be cautious when using this function to check the NRT getting updated especially if we are checking pod scope. This test for example is anticipated to run on both scopes. Assuming it run on pod scope, it runs 10 replicas (each with 2 containers), using checkNRTConsumedResources to check the NRT updates will check if the total requested resources are consumed from one numa, which is not accrate because we are testing pod scope, and the scheduler may place different containers on different numas.

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -853,31 +852,39 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
...
+				wg.Add(1)
+
+				go func(ref *nrtv1alpha2.NodeResourceTopology) {
+					defer GinkgoRecover()
+					defer wg.Done()
+
+					Consistently(func(g Gomega) {
+						var updatedNrt nrtv1alpha2.NodeResourceTopology
+						err := fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(ref), &updatedNrt)
+						g.Expect(err).ToNot(HaveOccurred(), "cannot get updated NRT for %q", ref.Name)
```
shajmakh: This means we'll be fitching the each node NRT in different times. how about that we get the list of the updated NRTs before entering the for loop, and when comparing the objects to the initial we fetch the relevant one by the idx? 
fetching the current NRT object within the loop may find unexpected changes in NRT but I cannot think how can this be helpful in troubleshooting, I even believe it'll make it even hard to spot the root cause because we are checking different versions of NRT overall.

test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -108,9 +112,19 @@ func CheckEqualAvailableResources(nrtInitial, nrtUpdated nrtv1alpha2.NodeResourc
...
 		}
 	}
+	klog.Infof("=> NRT %d zones equal", len(nrtInitial.Zones))
 	return true, nil
 }
 
+func logPFP(nrt nrtv1alpha2.NodeResourceTopology, tag string) {
+	attr, ok := nrtv1alpha2attr.Get(nrt.Attributes, podfingerprint.Attribute)
+	if !ok {
+		klog.Warningf("=> NRT %s had no PFP attribute")
```
shajmakh, line 122: missing the the nrt.Name that should replace the string format

test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -194,7 +194,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 			}
 		})
 
-		It("[test_id:47617][tier2][unsched] workload requests guaranteed pod resources available on one node but not on a single numa", func() {
+		It("[test_id:47617][tier2][unsched][failalign] workload requests guaranteed pod resources available on one node but not on a single numa", Label("unsched", "failalign"), func() {
```
shajmakh, line 181: why do we want to add the labels in the test description too?

test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -66,6 +67,29 @@ func GetUpdated(cli client.Client, ref nrtv1alpha2.NodeResourceTopologyList, tim
...
+			return false, err
+		}
+		// very cheap test to rule out false negatives
+		if updatedNrt.ObjectMeta.ResourceVersion == ref.ObjectMeta.ResourceVersion {
+			klog.Warningf("NRT for %q resource version didn't change", ref.Name)
+			return false, nil
+		}
+		equalZones = apiequality.Semantic.DeepEqual(ref.Zones, updatedNrt.Zones)
+		klog.Infof("NRT change: updated to %s", e2enrt.ToString(updatedNrt))
+		return !equalZones, nil
```
shajmakh, line 89: IIUC the function will return with the first NRT it spots. Yet, it won't be enough because releasing/consuming resources cannot be controlled, it periodically getting updated. What do you think about using the ForNodeResourceTopologiesSettled before taking the final decision and return with the updated NRT?

shajmakh (approved):
thanks for the PR, great work on this! 
I've added few comments, let me know what do you think.

test/e2e/serial/tests/workload_placement.go
shajmakh: I'll anyway run this test from on my env to better understand the behavior.

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -175,10 +175,9 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 					e2efixture.Skipf(fxt, "not enough nodes with policy %q - found %d", string(tmPolicy), len(nrts))
 				}
 
-				targetNrtListInitial, err := e2enrt.GetUpdated(fxt.Client, nrtList, 1*time.Minute)
+				var targetNrtInitial nrtv1alpha2.NodeResourceTopology
+				err := fxt.Client.Get(context.TODO(), client.ObjectKey{Name: targetNodeName}, &targetNrtInitial)
 				Expect(err).ToNot(HaveOccurred())
-				targetNrtInitial, err := e2enrt.FindFromList(targetNrtListInitial.Items, targetNodeName)
-				Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
ffromani: Likely bad habits I contributed to establish. This is indeed better and only slightly more verbose.

test/e2e/serial/tests/workload_placement.go
ffromani: This is a good point. To save time I just run container scope, so this is likely to be still broken (or more broken!) on pod scope.

test/e2e/serial/tests/workload_unschedulable.go
ffromani: I understand what you mean, but I feel depending on the NRT list rather than on specific NRT values is deceptive. NRT objects are updated independently and I'm now very skeptical that checking the list is a good and safe simplification (I may be wrong though). I think that at this stage we should acknowledge the system complexity, so check each NRT specifically, and tame it rather than hide it.

test/e2e/serial/tests/workload_unschedulable.go
ffromani: we probably should not anymore (I don't see us ever going back to ginkgo v1), I'll stop adding these and just use ginkgov2 `Label` from now on.

test/utils/noderesourcetopologies/noderesourcetopologies.go
ffromani: yes, it will return once it detects a change (any change) in NRT counters. We need indeed to have stronger guarantees, for example waiting for NRT to settle before look for the value. We probably need to iterate more in this area

test/e2e/serial/tests/workload_placement.go
shajmakh: I'll verify on pod scope then

test/e2e/serial/tests/workload_unschedulable.go
shajmakh: no problem, we can keep it like that. 

> NRT objects are updated independently

why? do you mean in code or in actual? because in code I see the list is simply a list of NodeResourceTopology so i'd assume that fetching the NRT "items" from the cluster will get us updated NRT single objects.

ffromani: NRT objects are updated independently by RTE/NFDs, in the live cluster. The intent is to iterate over a list just to get the set of NRT object we should watch to learn when we reach the state we need.

test/e2e/serial/tests/workload_unschedulable.go
ffromani: We need to evaluate this indeed

test/e2e/serial/tests/workload_unschedulable.go
ffromani: (clarified offline) @shajmakh thinking about it, you are probably right about the list. We can implement the same check just checking the list and it's also probably simpler codewise.
Not sure we save time performing individual checks, so we should probably use list and move on.

ffromani: Implemented the change. I'm inclined towards re-evaluating (gradually, over time) changes we made switching to `GetUpdatesForNode`. Most if not all of them seem to be nice (albeit sometimes minor) improvements anyway, so I don't think we need any revert anytime soon. Further fixes and tunings are likely needed and expected.

PR #606, opened 2023-06-20
e2e: serial: Fail fast if policies not aligned

jlojosnegros:
Enhance the serial suite to fail inmmediately if the topology policies in the NodeResourceTopology objects are not aligned with the TAS prerequisites:
- policy: single-numa-node
- scope: pod/container

### DIFF
--- a/pkg/validator/validator.go
+++ b/pkg/validator/validator.go
@@ -230,6 +230,22 @@ func Validate(data ValidatorData) (Report, error) {
 	}, nil
 }
 
+func GetKubeletConfigurationsFromTASEnabledNodes(ctx context.Context, cli client.Client) (map[string]*kubeletconfigv1beta1.KubeletConfiguration, error) {
+	tasEnabledNodeNames, err := GetNodesByNRO(ctx, cli)
+	if err != nil {
+		return nil, fmt.Errorf("error while trying to get TAS enabled nodes. error: %w", err)
+	}
+
+	validatorData := ValidatorData{
+		tasEnabledNodeNames: tasEnabledNodeNames,
+	}
+	if err := CollectKubeletConfig(ctx, cli, &validatorData); err != nil {
+		return nil, fmt.Errorf("error while collecting kubeletconfigs from TAS enabled nodes. error: %w", err)
+	}
+
+	return validatorData.kConfigs, nil
+}
+
 func getClusterVersionInfo() (*version.Info, error) {
 	cli, err := clientutil.NewDiscoveryClient()
 	if err != nil {
--- a/test/e2e/serial/config/config.go
+++ b/test/e2e/serial/config/config.go
@@ -17,15 +17,27 @@
 package config
 
 import (
+	"context"
+	"encoding/json"
+	"fmt"
 	"os"
 
+	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
+
 	. "github.com/onsi/gomega"
+
+	nrtv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
+
+	"github.com/openshift-kni/numaresources-operator/pkg/validator"
+	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
+	e2enrt "github.com/openshift-kni/numaresources-operator/test/utils/noderesourcetopologies"
 )
 
 const (
-	MultiNUMALabel    = "numa.hardware.openshift-kni.io/cell-count"
-	nropTestCIImage   = "quay.io/openshift-kni/resource-topology-exporter:test-ci"
-	SchedulerTestName = "test-topology-scheduler"
+	MultiNUMALabel                   = "numa.hardware.openshift-kni.io/cell-count"
+	nropTestCIImage                  = "quay.io/openshift-kni/resource-topology-exporter:test-ci"
+	SchedulerTestName                = "test-topology-scheduler"
+	minNumberOfNodesWithSameTopology = 2
 )
 
 // This suite holds the e2e tests which span across components,
@@ -66,4 +78,96 @@ func GetRteCiImage() string {
 		return pullSpec
 	}
 	return nropTestCIImage
+}
+
+func contains(items []string, st string) bool {
+	for _, item := range items {
+		if item == st {
+			return true
+		}
+	}
+	return false
+}
+
+var policyScopeToTmPolicy = map[string]nrtv1alpha2.TopologyManagerPolicy{
+	"restricted:pod":             nrtv1alpha2.RestrictedPodLevel,
+	"restricted:container":       nrtv1alpha2.RestrictedContainerLevel,
+	"best-effort:pod":            nrtv1alpha2.BestEffortPodLevel,
+	"best-effort:container":      nrtv1alpha2.BestEffortContainerLevel,
+	"none:":                      nrtv1alpha2.None,
+	"single-numa-node:pod":       nrtv1alpha2.SingleNUMANodePodLevel,
+	"single-numa-node:container": nrtv1alpha2.SingleNUMANodeContainerLevel,
+}
+
+func toTopologyPolicy(topologyManagerPolicy, topologyManagerScope string) (nrtv1alpha2.TopologyManagerPolicy, error) {
+	chain := topologyManagerPolicy + ":" + topologyManagerScope
+	nrtPolicy, ok := policyScopeToTmPolicy[chain]
+	if !ok {
+		return "", fmt.Errorf("Unable to convert %q to TmPolicy", chain)
+	}
+	return nrtPolicy, nil
+}
+
+func getTopologyConsistencyErrors(kconfigs map[string]*kubeletconfigv1beta1.KubeletConfiguration, nrts []nrtv1alpha2.NodeResourceTopology) map[string]error {
+	ret := make(map[string]error)
+	for nodeName, kconfig := range kconfigs {
+		nrt, err := e2enrt.FindFromList(nrts, nodeName)
+		if err != nil {
+			ret[nodeName] = fmt.Errorf("Unable to find NRT for node %q", nrt.Name)
+			continue
+		}
+
+		tmPolicy, err := toTopologyPolicy(kconfig.TopologyManagerPolicy, kconfig.TopologyManagerScope)
+		if err != nil {
+			ret[nodeName] = fmt.Errorf("Unable to convert kc.policy/kc.scope to TopologyManagerPolicy. node:%q, err: %w", nodeName, err)
+			continue
+		}
+
+		if !contains(nrt.TopologyPolicies, string(tmPolicy)) {
+			ret[nodeName] = fmt.Errorf("Incoherent KubeletConfig/NRT for node %q", nodeName)
+		}
+
+	}
+
+	for _, nrt := range nrts {
+		if _, ok := kconfigs[nrt.Name]; !ok {
+			ret[nrt.Name] = fmt.Errorf("Unable to find KubeletConfig for node %q", nrt.Name)
+		}
+	}
+
+	return ret
+}
+
+func CheckNodesTopology(ctx context.Context) error {
+	kconfigs, err := validator.GetKubeletConfigurationsFromTASEnabledNodes(ctx, e2eclient.Client)
+	if err != nil {
+		return fmt.Errorf("error while trying to get KubeletConfigurations. error: %w", err)
+	}
+
+	var nrtList nrtv1alpha2.NodeResourceTopologyList
+	err = e2eclient.Client.List(context.TODO(), &nrtList)
+	if err != nil {
+		return fmt.Errorf("error while trying to get NodeResourceTopology objects. error: %w", err)
+	}
+
+	errorMap := getTopologyConsistencyErrors(kconfigs, nrtList.Items)
+	if len(errorMap) != 0 {
+		prettyMap, err := json.MarshalIndent(errorMap, "", "  ")
+		if err != nil {
+			return fmt.Errorf("Found some nodes with incoherent info in KubeletConfig/NRT data")
+		}
+		return fmt.Errorf("Following nodes have incoherent info in KubeletConfig/NRT data:\n%s\n", string(prettyMap))
+	}
+
+	snnPodLevelNRTs := e2enrt.FilterTopologyManagerPolicy(nrtList.Items, nrtv1alpha2.SingleNUMANodePodLevel)
+	numSnnPodLevelNRTs := len(snnPodLevelNRTs)
+
+	snnContainerLevelNRTs := e2enrt.FilterTopologyManagerPolicy(nrtList.Items, nrtv1alpha2.SingleNUMANodeContainerLevel)
+	numSnnContainerLevelNRTs := len(snnContainerLevelNRTs)
+
+	if numSnnPodLevelNRTs < minNumberOfNodesWithSameTopology && numSnnContainerLevelNRTs < minNumberOfNodesWithSameTopology {
+		return fmt.Errorf("Not enough nodes with either %q topology (found:%d) nor %q topology (found: %d). Need at least %d", nrtv1alpha2.SingleNUMANodePodLevel, numSnnPodLevelNRTs, nrtv1alpha2.SingleNUMANodeContainerLevel, numSnnContainerLevelNRTs, minNumberOfNodesWithSameTopology)
+	}
+
+	return nil
 }
--- a/test/e2e/serial/e2e_serial_test.go
+++ b/test/e2e/serial/e2e_serial_test.go
@@ -17,6 +17,7 @@
 package serial
 
 import (
+	"context"
 	"math/rand"
 	"testing"
 	"time"
@@ -34,6 +35,7 @@ import (
 )
 
 var afterSuiteReporters = []Reporter{}
+var setupExecuted = false
 
 var randomSeed int64
 
@@ -53,10 +55,17 @@ var _ = BeforeSuite(func() {
 
 	klog.Infof("using random seed %v", randomSeed)
 
+	Expect(serialconfig.CheckNodesTopology(context.TODO())).Should(Succeed())
 	serialconfig.Setup()
+	setupExecuted = true
 })
 
-var _ = AfterSuite(serialconfig.Teardown)
+var _ = AfterSuite(func() {
+	if setupExecuted {
+		serialconfig.Teardown()
+	}
+
+})
 
 var _ = ReportAfterSuite("TestTests", func(report Report) {
 	for _, reporter := range afterSuiteReporters {


test/e2e/serial/e2e_serial_test.go
```diff
@@ -45,10 +54,34 @@ func TestSerial(t *testing.T) {
 var _ = BeforeSuite(func() {
 	// this must be the very first thing
 	rand.Seed(time.Now().UnixNano())
+
+	var nrtList nrtv1alpha2.NodeResourceTopologyList
+
+	err := e2eclient.Client.List(context.TODO(), &nrtList)
+	if err != nil {
+		suiteSkipped = true
+		Skip(fmt.Sprintf("error while trying to get NodeResourceTopology objects. error: %q", err.Error()))
```
ffromani, line 63: just use `%v` for errors

test/e2e/serial/e2e_serial_test.go
```diff
@@ -45,10 +54,34 @@ func TestSerial(t *testing.T) {
 var _ = BeforeSuite(func() {
 	// this must be the very first thing
 	rand.Seed(time.Now().UnixNano())
+
+	var nrtList nrtv1alpha2.NodeResourceTopologyList
```
ffromani, line 58: the code looks correct, but we have plans to re-add these tests in `cnf-tests` (someday...) so this should be wrapped in a helper. Moving this code into `serialconfig` is an option, or we can add another abstraction if needed.

ffromani (commented):
We likely need to extend the suite anyway, so the direction seems OK.

test/e2e/serial/e2e_serial_test.go
jlojosnegros: thanks.

test/e2e/serial/e2e_serial_test.go
jlojosnegros: I moved the code to a new function in `serialConfig`. Not really proud of the name though

test/e2e/serial/config/config.go
```diff
@@ -67,3 +75,23 @@ func GetRteCiImage() string {
...
+		return fmt.Errorf("error while trying to get NodeResourceTopology objects. error: %w", err)
+	}
+	// we're ok with any TM policy as long as the updater can handle it,
+	// we use this as proxy for "there is valid NRT data for at least X nodes
+	policies := []nrtv1alpha2.TopologyManagerPolicy{
+		nrtv1alpha2.SingleNUMANodeContainerLevel,
+		nrtv1alpha2.SingleNUMANodePodLevel,
+	}
+	nrts := e2enrt.FilterByPolicies(nrtList.Items, policies)
+	if len(nrts) < minNumberOfNodesWithRightTopology {
```
shajmakh, line 93: this is exactly the case we want to avoid. If there are 2 nodes with different policies there must be something wrong. NRT data are expected to align with kubeletconfig settings, if it's not then there is no point of running the tests thus skipping the suite. 
what I'd do is to fetch the kubelet configuration of numa topology and scope policy and compare it to what's there in NRT, so the check is correct but the condition on the policies is not.

shajmakh:
Thanks for the PR! 
added one comment.

test/e2e/serial/config/config.go
shajmakh: right. 
the reason we need this change is that in case the cluster was corrupted for some reason (bad configuration, mcp misconfiguration, tests leftovers.. ) the tests continued being executed and then getting skipped later because when they checked for the minimum nodes with matching tmpolicy there weren't enough nodes. so via this modification, we aim to fail fast and skip the suite even earlier.

jlojosnegros:
Change the way of checking to follow [this](https://github.com/openshift-kni/numaresources-operator/pull/606#discussion_r1236991130) .

Code could be improved but would like to check if this is the proper approach.

test/e2e/serial/config/config.go
```diff
@@ -67,3 +76,62 @@ func GetRteCiImage() string {
...
+	}
+
+	if podLevelNumber >= minNumberOfNodesWithRightTopology {
+		for _, nrt := range nrtsSingleNumaNodePodLevel {
+			kconfig, ok := kconfigs[nrt.Name]
+			if !ok {
+				return fmt.Errorf("Unable to find KubeletConfig for node %q", nrt.Name)
+			}
+			if kconfig.TopologyManagerPolicy != "single-numa-node" || kconfig.TopologyManagerScope != "pod" {
+				return fmt.Errorf("Uncoheren KubeletConfig/NRT for node %q", nrt.Name)
```
jlojosnegros, line 118: typo: uncoheren -> uncoherent

test/e2e/serial/config/config.go
```diff
@@ -67,3 +76,62 @@ func GetRteCiImage() string {
...
+
+	if containerLevelNumber >= minNumberOfNodesWithRightTopology {
+		nrtsSingleNumaNodeContainerLevel := e2enrt.FilterTopologyManagerPolicy(nrts, nrtv1alpha2.SingleNUMANodeContainerLevel)
+		for _, nrt := range nrtsSingleNumaNodeContainerLevel {
+			kconfig, ok := kconfigs[nrt.Name]
+			if !ok {
+				return fmt.Errorf("Unable to find KubeletConfig for node %q", nrt.Name)
+			}
+			if kconfig.TopologyManagerPolicy != "single-numa-node" || kconfig.TopologyManagerScope != "container" {
+				return fmt.Errorf("Uncoheren KubeletConfig/NRT for node %q", nrt.Name)
```
jlojosnegros, line 131: typo: uncoheren -> uncoherent

pkg/validator/validator.go
```diff
@@ -230,6 +230,22 @@ func Validate(data ValidatorData) (Report, error) {
 	}, nil
 }
 
+func GetKubeletConfigurationsFromTASEnabledNodes(ctx context.Context, cli client.Client) (map[string]*kubeletconfigv1beta1.KubeletConfiguration, error) {
```
ffromani, line 233: this is calling for some refactoring, but can be postponed

test/e2e/serial/config/config.go
```diff
@@ -17,15 +17,24 @@
...
 )
 
 const (
-	MultiNUMALabel    = "numa.hardware.openshift-kni.io/cell-count"
-	nropTestCIImage   = "quay.io/openshift-kni/resource-topology-exporter:test-ci"
-	SchedulerTestName = "test-topology-scheduler"
+	MultiNUMALabel                    = "numa.hardware.openshift-kni.io/cell-count"
+	nropTestCIImage                   = "quay.io/openshift-kni/resource-topology-exporter:test-ci"
+	SchedulerTestName                 = "test-topology-scheduler"
+	minNumberOfNodesWithRightTopology = 2
```
ffromani, line 37: with _Right_ Topology or with the _Same_ topology?

test/e2e/serial/config/config.go
```diff
@@ -67,3 +76,62 @@ func GetRteCiImage() string {
 	}
 	return nropTestCIImage
 }
+
+func CheckNodesTopology() error {
+	kconfigs, err := validator.GetKubeletConfigurationsFromTASEnabledNodes(context.TODO(), e2eclient.Client)
```
ffromani, line 81: let's create the context from the top level, so this function should accept a context and pass it through

test/e2e/serial/config/config.go
```diff
@@ -67,3 +76,62 @@ func GetRteCiImage() string {
...
+	}
+	nrts := e2enrt.FilterByPolicies(nrtList.Items, policies)
+	singleNumaNodeNumber := len(nrts)
+	if singleNumaNodeNumber < minNumberOfNodesWithRightTopology {
+		return fmt.Errorf("not enough nodes with valid policy - found %d, need at least %d", len(nrts), minNumberOfNodesWithRightTopology)
+	}
+
+	nrtsSingleNumaNodePodLevel := e2enrt.FilterTopologyManagerPolicy(nrts, nrtv1alpha2.SingleNUMANodePodLevel)
+	podLevelNumber := len(nrtsSingleNumaNodePodLevel)
+	containerLevelNumber := singleNumaNodeNumber - podLevelNumber
```
ffromani, line 105: I'd not just add this assumption here. Let's count explicitely, perhaps also raising a warning (= log)/error if the sum of containerLevel and podLevel is different than singleNumaNodeNumber

test/e2e/serial/config/config.go
```diff
@@ -67,3 +76,62 @@ func GetRteCiImage() string {
...
+	}
+
+	if podLevelNumber >= minNumberOfNodesWithRightTopology {
+		for _, nrt := range nrtsSingleNumaNodePodLevel {
+			kconfig, ok := kconfigs[nrt.Name]
+			if !ok {
+				return fmt.Errorf("Unable to find KubeletConfig for node %q", nrt.Name)
+			}
+			if kconfig.TopologyManagerPolicy != "single-numa-node" || kconfig.TopologyManagerScope != "pod" {
+				return fmt.Errorf("Uncoherent KubeletConfig/NRT for node %q", nrt.Name)
```
ffromani, line 118: typo: Incoherent

other than that this check is probably better moved in the RTE testsuite, or in an explicit e2e test in general. We can and should assume _here_ we can trust the NRT data, and we can trust it correctly represents the kubelet config

test/e2e/serial/config/config.go
```diff
@@ -67,3 +76,62 @@ func GetRteCiImage() string {
...
+
+	if containerLevelNumber >= minNumberOfNodesWithRightTopology {
+		nrtsSingleNumaNodeContainerLevel := e2enrt.FilterTopologyManagerPolicy(nrts, nrtv1alpha2.SingleNUMANodeContainerLevel)
+		for _, nrt := range nrtsSingleNumaNodeContainerLevel {
+			kconfig, ok := kconfigs[nrt.Name]
+			if !ok {
+				return fmt.Errorf("Unable to find KubeletConfig for node %q", nrt.Name)
+			}
+			if kconfig.TopologyManagerPolicy != "single-numa-node" || kconfig.TopologyManagerScope != "container" {
+				return fmt.Errorf("Uncoherent KubeletConfig/NRT for node %q", nrt.Name)
```
ffromani, line 131: ditto

test/e2e/serial/e2e_serial_test.go
```diff
@@ -45,10 +47,20 @@ func TestSerial(t *testing.T) {
...
+		Skip(fmt.Sprintf("error on Nodes Topology validation. error: %v", err))
+	}
 	serialconfig.Setup()
 })
 
-var _ = AfterSuite(serialconfig.Teardown)
+var _ = AfterSuite(func() {
+	if !suiteSkipped {
+		serialconfig.Teardown()
+	}
```
ffromani, line 62: let's align the happy path on the left whenever possible:
```
if suiteSkipped {
    return
}
serialconfig.Teardown()
```

ffromani (commented):
/approve

I'm fine with the general direction

test/e2e/serial/config/config.go
jlojosnegros: Ok, lets sync here. 
I added all the kubeletconfig stuff because I understood from [this comment](https://github.com/openshift-kni/numaresources-operator/pull/606#discussion_r1236755683) of @shajmakh that we should check both KubeletConfiguration and NRTs where coherent in this fields.
Did I understand correctly? 
Is there a possible case where the KubeletConfiguration and the NRTs had different values in these fields?

pkg/validator/validator.go
jlojosnegros: If we do not need to check KubeletConfigurations in the end we could avoid this function completely

test/e2e/serial/config/config.go
```diff
@@ -67,3 +76,64 @@ func GetRteCiImage() string {
...
+	}
+
+	policies := []nrtv1alpha2.TopologyManagerPolicy{
+		nrtv1alpha2.SingleNUMANodeContainerLevel,
+		nrtv1alpha2.SingleNUMANodePodLevel,
+	}
+	nrts := e2enrt.FilterByPolicies(nrtList.Items, policies)
+	singleNumaNodeNumber := len(nrts)
+	if singleNumaNodeNumber < minNumberOfNodesWithSameTopology {
+		return fmt.Errorf("not enough nodes with valid policy - found %d, need at least %d", len(nrts), minNumberOfNodesWithSameTopology)
```
shajmakh, line 100: this can be removed now that the accurate check is done in L109

test/e2e/serial/config/config.go
```diff
@@ -67,3 +76,64 @@ func GetRteCiImage() string {
...
+	}
+
+	nrtsSingleNumaNodePodLevel := e2enrt.FilterTopologyManagerPolicy(nrts, nrtv1alpha2.SingleNUMANodePodLevel)
+	podLevelNumber := len(nrtsSingleNumaNodePodLevel)
+
+	nrtsSingleNumaNodeContainerLevel := e2enrt.FilterTopologyManagerPolicy(nrts, nrtv1alpha2.SingleNUMANodeContainerLevel)
+	containerLevelNumber := len(nrtsSingleNumaNodePodLevel)
+
+	if podLevelNumber < minNumberOfNodesWithSameTopology && containerLevelNumber < minNumberOfNodesWithSameTopology {
```
shajmakh, line 109: this will 100% be false in our CI envs because either we have 2 nodes with container scope OR two with pod scope. What I would think of is to take one thing as the source of truth, say kubeletconfig config, and based the checks on it. so for example, we should expect numberOfNRONodes = (number Of Nodes With Policy Level as configured in kubeletconfig).

shajmakh:
Thanks for the enhancements. inserted a comment following the latest updates. 
@jlojosnegros the failing lane needs the fix in https://github.com/openshift-kni/numaresources-operator/commit/cec6850179495819dc5fcfb35f85a91eecfbd102 I think rebase would resolve this.

jlojosnegros:
> Thanks for the enhancements. inserted a comment following the latest updates. @jlojosnegros the failing lane needs the fix in [cec6850](https://github.com/openshift-kni/numaresources-operator/commit/cec6850179495819dc5fcfb35f85a91eecfbd102) I think rebase would resolve this.

Thanks for the comments and for the heads up. 
Rebasing over last main content.

test/e2e/serial/config/config.go
jlojosnegros: Let's do this on another PR to refactor/cleanup all validation code, ok?

ffromani: if this check we're adding here is superseded by another test done later in this very function being added by this PR, it seems simpler to just remove this redundant check, though.

test/e2e/serial/config/config.go
ffromani: since the whole point of this PR is adding a initial check verifying NRT is consistent with kubeletconfig, the source of truth should probably be kubeletconfig indeed

jlojosnegros: Sure but, the whole point is to check that both sources are coherent ... and even if the NRT data seems "ok-ish" here and this check does not fail, it would be checked around line 113 that each of the NRT node data is consistent with its Kubeletconfig

test/e2e/serial/config/config.go
jlojosnegros: Sorry my bad, I readed it in a hurry and thought we were talking about the checks in the tests cases .... 

yes, this is deleted in the last version.

test/e2e/serial/config/config.go
jlojosnegros: For the sake of completeness let me summarize the agreement we have reached here
 1. keep the existing code, but fail (not skip) if the misalignment is detected
 2. file a bug/issue/jira card to track refactoring/enhancements in this area in the validation code 
 3. file a bug/issue/jira card to track adding a NROP/RTE e2e test to check this

test/e2e/serial/config/config.go
jlojosnegros: I have changed the approach in the last commit.

Thanks for the comments .

test/e2e/serial/e2e_serial_test.go
```diff
@@ -45,10 +48,21 @@ func TestSerial(t *testing.T) {
 var _ = BeforeSuite(func() {
 	// this must be the very first thing
 	rand.Seed(time.Now().UnixNano())
+
+	err := serialconfig.CheckNodesTopology(context.TODO())
+	if err != nil {
+		suiteSkipped = true
+		Skip(fmt.Sprintf("error on Nodes Topology validation. error: %v", err))
```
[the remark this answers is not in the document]
jlojosnegros: done

jlojosnegros:
rebasing over last main to resolve conflicts.

jlojosnegros:
/hold cancel 
As we already agree on the way forward

test/e2e/serial/config/config.go
```diff
@@ -67,3 +79,77 @@ func GetRteCiImage() string {
...
+func toTopolyPolicy(topologyManagerPolicy, topologyManagerScope string) (nrtv1alpha2.TopologyManagerPolicy, error) {
+	chain := topologyManagerPolicy + ":" + topologyManagerScope
+	tmPolicy, ok := policyScopeToTmPolicy[chain]
+	if !ok {
+		return "", fmt.Errorf("Unable to convert %q to TmPolicy", chain)
+	}
+	return tmPolicy, nil
+}
+
+func getTopologyConsistencyErrors(kconfigs map[string]*kubeletconfigv1beta1.KubeletConfiguration, nrts []nrtv1alpha2.NodeResourceTopology) map[string]error {
```
ffromani, line 111: what if `kconfigs` is a subset of `nrts`?

test/e2e/serial/config/config.go
```diff
@@ -67,3 +79,77 @@ func GetRteCiImage() string {
...
+	ret := make(map[string]error)
+	for nodeName, kconfig := range kconfigs {
+		nrt, err := e2enrt.FindFromList(nrts, nodeName)
+		if err != nil {
+			ret[nodeName] = fmt.Errorf("Unable to find KubeletConfig for node %q", nrt.Name)
```
shajmakh, line 116: "unable to find NRT for node.."

test/e2e/serial/config/config.go
```diff
@@ -67,3 +79,77 @@ func GetRteCiImage() string {
...
+	"best-effort:pod":            nrtv1alpha2.BestEffortPodLevel,
+	"best-effort:container":      nrtv1alpha2.BestEffortContainerLevel,
+	"none:":                      nrtv1alpha2.None,
+	"single-numa-node:pod":       nrtv1alpha2.SingleNUMANodePodLevel,
+	"single-numa-node:container": nrtv1alpha2.SingleNUMANodeContainerLevel,
+}
+
+func toTopolyPolicy(topologyManagerPolicy, topologyManagerScope string) (nrtv1alpha2.TopologyManagerPolicy, error) {
+	chain := topologyManagerPolicy + ":" + topologyManagerScope
+	tmPolicy, ok := policyScopeToTmPolicy[chain]
```
shajmakh, line 104: I think I'd call it nrtPolicy to avoid confusion between the known topology manager policy

test/e2e/serial/config/config.go
```diff
@@ -67,3 +79,77 @@ func GetRteCiImage() string {
...
+	"restricted:pod":             nrtv1alpha2.RestrictedPodLevel,
+	"restricted:container":       nrtv1alpha2.RestrictedContainerLevel,
+	"best-effort:pod":            nrtv1alpha2.BestEffortPodLevel,
+	"best-effort:container":      nrtv1alpha2.BestEffortContainerLevel,
+	"none:":                      nrtv1alpha2.None,
+	"single-numa-node:pod":       nrtv1alpha2.SingleNUMANodePodLevel,
+	"single-numa-node:container": nrtv1alpha2.SingleNUMANodeContainerLevel,
+}
+
+func toTopolyPolicy(topologyManagerPolicy, topologyManagerScope string) (nrtv1alpha2.TopologyManagerPolicy, error) {
```
shajmakh, line 102: function name typo: toTopologyPolicy(..)

test/e2e/serial/config/config.go
```diff
@@ -67,3 +79,77 @@ func GetRteCiImage() string {
...
+	for nodeName, kconfig := range kconfigs {
+		nrt, err := e2enrt.FindFromList(nrts, nodeName)
+		if err != nil {
+			ret[nodeName] = fmt.Errorf("Unable to find KubeletConfig for node %q", nrt.Name)
+			continue
+		}
+
+		tmPolicy, err := toTopolyPolicy(kconfig.TopologyManagerPolicy, kconfig.TopologyManagerScope)
+		if err != nil {
+			ret[nodeName] = fmt.Errorf("Unable to convert kc.policy/kc.scope to TopologyMangerPolicy. node:%q, err: %w", nodeName, err)
```
shajmakh, line 122: same comment here regarding the naming of the output: TopologyMangerPolicy

test/e2e/serial/config/config.go
```diff
@@ -67,3 +79,77 @@ func GetRteCiImage() string {
...
+
+		if !contains(nrt.TopologyPolicies, string(tmPolicy)) {
+			ret[nodeName] = fmt.Errorf("Incoherent KubeletConfig/NRT for node %q", nodeName)
+		}
+
+	}
+	return ret
+}
+
+func CheckNodesTopology(ctx context.Context) error {
```
shajmakh, line 134: I believe we can avoid passing cxt to the function, and we can simply use context.TODO()

test/e2e/serial/config/config.go
```diff
@@ -17,15 +17,27 @@
...
 )
 
 const (
-	MultiNUMALabel    = "numa.hardware.openshift-kni.io/cell-count"
-	nropTestCIImage   = "quay.io/openshift-kni/resource-topology-exporter:test-ci"
-	SchedulerTestName = "test-topology-scheduler"
+	MultiNUMALabel                   = "numa.hardware.openshift-kni.io/cell-count"
+	nropTestCIImage                  = "quay.io/openshift-kni/resource-topology-exporter:test-ci"
+	SchedulerTestName                = "test-topology-scheduler"
+	minNumberOfNodesWithSameTopology = 2
```
shajmakh, line 40: we missed using this. If not mistaken, currently this code checks that the nrts are aligned with the kubeletconfig which is good and is needed. the suite fails if these configurations are not aligned, and that is the expected behavior. yet it is not enough to address the main problem. I expect the suite to still be executed with different topologies on different policies circling back to the main problem. 
I believe the only left thing to do is to check that there are at least 2 ( minNumberOfNodesWithSameTopology ) nodes with similar nrt.TopologyPolicy && it is equal to (either nrtv1alpha2.SingleNUMANodeContainerLevel or nrtv1alpha2.SingleNUMANodePodLevel). Let me know what you think.

test/e2e/serial/config/config.go
ffromani: We can, but we should not. We should not hardcode `context.TODO()` in the low-level function. The context should be created by the top-most function and passed through accordingly. TL;DR: I'd keep the context here and in all the utility functions.

shajmakh: ah good to learn about that, thanks for sharing it.

test/e2e/serial/config/config.go
jlojosnegros: I have added an additional check if there are some NRTs with no kubeletconfiguration.

test/e2e/serial/config/config.go
shajmakh: Thanks for all the modifications!
/lgtm
/hold 
I guess more improvements can be done in a later PR but hold for other reviewers' approval.

PR #608, opened 2023-06-21
e2e: serial: resacct improve and generalize expectations, take 2

ffromani:
Fixes to utilities to be more robust and to reduce flakiness, emerged during the review of #598

### DIFF
--- a/internal/wait/noderesourcetopologies.go
+++ b/internal/wait/noderesourcetopologies.go
@@ -20,9 +20,12 @@ import (
 	"context"
 	"fmt"
 
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/util/sets"
 	k8swait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
+
+	intnrt "github.com/openshift-kni/numaresources-operator/internal/noderesourcetopology"
 
 	nrtv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
 	nrtv1alpha2attr "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2/helper/attribute"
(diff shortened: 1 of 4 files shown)


test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -79,19 +92,29 @@ func GetUpdatedForNode(cli client.Client, ctx context.Context, ref nrtv1alpha2.N
...
-		if updatedNrt.ObjectMeta.ResourceVersion == ref.ObjectMeta.ResourceVersion {
-			klog.Warningf("NRT for %q resource version didn't change", ref.Name)
+		equalZones = isEqualNRT(ref, updatedNrt)
+		if equalZones {
 			return false, nil
 		}
-		equalZones = apiequality.Semantic.DeepEqual(ref.Zones, updatedNrt.Zones)
-		klog.Infof("NRT change: updated to %s", e2enrt.ToString(updatedNrt))
-		return !equalZones, nil
+		return true, nil
```
shajmakh, line 99: I believe the condition here can be avoided and simply return with !isEqualNRT.., also because there are no logs specific to each case.

ffromani:
reviewers note the concepts here are stable but I have a cleanup pending. I'll post later after a validation run

test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -79,19 +92,29 @@ func GetUpdatedForNode(cli client.Client, ctx context.Context, ref nrtv1alpha2.N
...
 	return updatedNrt, err
 }
 
+func isEqualNRT(initialNrt, updatedNrt nrtv1alpha2.NodeResourceTopology) bool {
+	// very cheap test to rule out false negatives
+	if updatedNrt.ObjectMeta.ResourceVersion == initialNrt.ObjectMeta.ResourceVersion {
+		klog.Warningf("NRT %q resource version didn't change", initialNrt.Name)
+		return true
+	}
+	equalZones := apiequality.Semantic.DeepEqual(initialNrt.Zones, updatedNrt.Zones)
```
shajmakh, line 156: I wonder if it takes the same time to update the different resources on NRT. for example if the consumed devices are updated by the same time that the consumed cpus are. 
this PR is definitely needed. I'll have the tests triggered on this once merged to make sure it's stable for all resources checked.

shajmakh (commented):
Thank you for this. 
minor comments below.
/lgtm

test/utils/noderesourcetopologies/noderesourcetopologies.go
ffromani: right, this is a relic of past work. I'll simplify the code

test/utils/noderesourcetopologies/noderesourcetopologies.go
ffromani: it's a good point to explore. I think the code is robust enough atm to deal with that, but this is something we should totally look into for the future

test/utils/noderesourcetopologies/noderesourcetopologies.go
ffromani: ...and add logs. Note we need to bubble up and return `equalZones`, this is why we have the explicit assignment

test/utils/noderesourcetopologies/noderesourcetopologies.go
Tal-or: What about `AttributeList` https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/blob/master/pkg/apis/topology/v1alpha2/types.go#L54?

internal/wait/noderesourcetopologies.go
```diff
@@ -110,3 +113,65 @@ func (wt Waiter) ForNodeResourceTopologiesSettled(ctx context.Context, threshold
...
+func findNRTByName(nrts []nrtv1alpha2.NodeResourceTopology, name string) (*nrtv1alpha2.NodeResourceTopology, error) {
+	for idx := 0; idx < len(nrts); idx++ {
+		if nrts[idx].Name == name {
+			return &nrts[idx], nil
+		}
+	}
+	return nil, fmt.Errorf("failed to find NRT for %q", name)
+}
+
+func isNRTEqual(initialNrt, updatedNrt nrtv1alpha2.NodeResourceTopology) bool {
```
Tal-or, line 166: We already have it here https://github.com/openshift-kni/numaresources-operator/pull/608/commits/ee9fa855df1a0ab9d6a91dd4f0a38ea271c7086b#diff-98853065ac58ab9f346c413d9e8a25d4268fc59f11d0012dc087d28b09798e81R107 don't we?

internal/wait/noderesourcetopologies.go
```diff
@@ -110,3 +113,65 @@ func (wt Waiter) ForNodeResourceTopologiesSettled(ctx context.Context, threshold
...
+			}
+			klog.Info("NRT for %s matches reference", referenceNrt.Name)
+		}
+		klog.Infof("Matched reference for all %d NRT objects", len(nrtListReference.Items))
+		return true, nil
+	})
+	return updatedNrtList, err
+}
+
+func findNRTByName(nrts []nrtv1alpha2.NodeResourceTopology, name string) (*nrtv1alpha2.NodeResourceTopology, error) {
```
Tal-or, line 157: I think we also have this function too in `utils`

internal/wait/noderesourcetopologies.go
```diff
@@ -110,3 +113,65 @@ func (wt Waiter) ForNodeResourceTopologiesSettled(ctx context.Context, threshold
...
+	err := k8swait.Poll(wt.PollInterval, wt.PollTimeout, func() (bool, error) {
+		err := wt.Cli.List(ctx, &updatedNrtList)
+		if err != nil {
+			klog.Errorf("cannot get the NRT List: %v", err)
+			return false, err
+		}
+		for idx := range nrtListReference.Items {
+			referenceNrt := &nrtListReference.Items[idx]
+
+			if !nrtFilter(referenceNrt) {
```
Tal-or, line 135: nit: Readability-wise I would expect to read it as 
```
if nrtFilter(referenceNrt) {
    klog.Warningf("skipping NRT %q because of filter", referenceNrt.Name)
    continue
}
without the `!` mark 
```

Tal-or (commented):
Thank you for the PR. I left couple of small comments

internal/wait/noderesourcetopologies.go
ffromani: we do, but this package can't depend on `test/utlls`, while the other way around is allowed

internal/wait/noderesourcetopologies.go
ffromani: yep, we do, but again wrong dependency direction

internal/wait/noderesourcetopologies.go
ffromani: I'm not sure I'm getting you right because if I do that's a bug, not a nit :)

test/utils/noderesourcetopologies/noderesourcetopologies.go
ffromani: Good point. This also should be addressed in the followup. The saving grace is the only mutating field (PFP) is eventually a function of the content of the `Zones`.

ffromani:
/test ci-e2e

infra issue

test/utils/noderesourcetopologies/noderesourcetopologies.go
ffromani: filed https://github.com/openshift-kni/numaresources-operator/issues/609 for tracking

internal/wait/noderesourcetopologies.go
Tal-or: So why not having it the other way around?

internal/wait/noderesourcetopologies.go
ffromani: that's actually a good point. Let me see.

internal/wait/noderesourcetopologies.go
ffromani: because `IsEqualNRT` and `FindNRTByName` don't really fit in the `wait` package. They belong more to a package dealing with NRT objects rather than a package dealing with wait for conditions.
Lacking (time for) a refactoring to repurpose and reshuffle the code, [a little duplication seems better than a little dependency](https://www.efekarakus.com/2021/09/23/a-little-copying-is-better-than-a-little-dependency.html)

Tal-or: I see, let's move on, it's not a big deal anyway

ffromani:
backport of this PR depends on #605

shajmakh:
#605 is merged, let's backport this to 4.13:
/cherry-pick release-4.13

PR #610, opened 2023-06-22
e2e: verify nrt data after each test

shajmakh:
Following https://github.com/openshift-kni/numaresources-operator/pull/598, https://github.com/openshift-kni/numaresources-operator/pull/608 stabilize the e2e tests even more by verifying after each test that the NRT data is settled to the initial. This will help in troubleshooting failures related to to unclean NRT data and in a meaninful cooldown.

PS: we assume that the cluster is clean before running the suite, as in it doesn't have workloads apart from the system and setup workloads that are needed for executing serial. Also, we assume no sudden interrupts, resources consumptions/releases during the run, which was also the case before this update.

### DIFF
--- a/test/e2e/serial/config/fixture.go
+++ b/test/e2e/serial/config/fixture.go
@@ -68,7 +68,7 @@ func NewFixtureWithOptions(nsName string, options e2efixture.Options) (*E2EConfi
 		NROSchedObj: &nropv1.NUMAResourcesScheduler{},
 	}
 
-	cfg.Fixture, err = e2efixture.SetupWithOptions(nsName, options)
+	cfg.Fixture, err = e2efixture.SetupWithOptions(nsName, nrtv1alpha2.NodeResourceTopologyList{}, options)
 	if err != nil {
 		return nil, err
 	}
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -71,7 +71,7 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
 		Expect(serialconfig.Config.Ready()).To(BeTrue(), "NUMA fixture initialization failed")
 
 		var err error
-		fxt, err = e2efixture.Setup("e2e-test-configuration")
+		fxt, err = e2efixture.Setup("e2e-test-configuration", serialconfig.Config.NRTList)
 		Expect(err).ToNot(HaveOccurred(), "unable to setup test fixture")
 
 		err = fxt.Client.List(context.TODO(), &nrtList)
--- a/test/e2e/serial/tests/non_regression_fundamentals.go
+++ b/test/e2e/serial/tests/non_regression_fundamentals.go
@@ -51,7 +51,7 @@ var _ = Describe("[serial][fundamentals][scheduler][nonreg] numaresources fundam
 		Expect(serialconfig.Config.Ready()).To(BeTrue(), "NUMA fixture initialization failed")
 
 		var err error
-		fxt, err = e2efixture.Setup("e2e-test-non-regression-fundamentals")
+		fxt, err = e2efixture.Setup("e2e-test-non-regression-fundamentals", serialconfig.Config.NRTList)
 		Expect(err).ToNot(HaveOccurred(), "unable to setup test fixture")
 
 		err = fxt.Client.List(context.TODO(), &nrtList)
--- a/test/e2e/serial/tests/resource_accounting.go
+++ b/test/e2e/serial/tests/resource_accounting.go
@@ -65,7 +65,7 @@ var _ = Describe("[serial][disruptive][scheduler][resacct] numaresources workloa
 		Expect(serialconfig.Config.Ready()).To(BeTrue(), "NUMA fixture initialization failed")
 
 		var err error
-		fxt, err = e2efixture.Setup("e2e-test-resource-accounting")
+		fxt, err = e2efixture.Setup("e2e-test-resource-accounting", serialconfig.Config.NRTList)
 		Expect(err).ToNot(HaveOccurred(), "unable to setup test fixture")
 
 		padder, err = e2epadder.New(fxt.Client, fxt.Namespace.Name)
--- a/test/e2e/serial/tests/resource_accounting.go
+++ b/test/e2e/serial/tests/resource_accounting.go
@@ -287,8 +287,7 @@ var _ = Describe("[serial][disruptive][scheduler][resacct] numaresources workloa
 			Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s", updatedPod.Namespace, updatedPod.Name, serialconfig.Config.SchedulerName)
 
 			By("Waiting for the NRT data to stabilize")
-			wait.With(fxt.Client).Interval(11*time.Second).Timeout(1*time.Minute).ForNodeResourceTopologiesSettled(context.TODO(), 3)
-
+			e2efixture.WaitForNRTSettle(fxt)
 			// Check that NRT of the target node reflect correct consumed resources
 			By("Verifying NRT is updated properly when running the test's pod")
 			expectNRTConsumedResources(fxt, *targetNrtBefore, requiredRes, updatedPod)
--- a/test/utils/fixture/fixture.go
+++ b/test/utils/fixture/fixture.go
@@ -51,8 +51,9 @@ type Fixture struct {
 }
 
 const (
-	defaultTeardownTime = 180 * time.Second
-	defaultCooldownTime = 30 * time.Second
+	defaultTeardownTime      = 180 * time.Second
+	defaultCooldownTime      = 30 * time.Second
+	defaultCooldownThreshold = 5
 )
 
 type Options uint
@@ -63,13 +64,15 @@ const (
 )
 
 var (
-	teardownTime time.Duration
-	cooldownTime time.Duration
+	teardownTime      time.Duration
+	cooldownTime      time.Duration
+	cooldownThreshold int
 )
 
 func init() {
 	cooldownTime = getCooldownTimeFromEnvVar()
 	teardownTime = getTeardownTimeFromEnvVar()
+	cooldownThreshold = getCooldownThresholdFromEnvVar()
 }
 
 func SetupWithOptions(name string, nrtList nrtv1alpha2.NodeResourceTopologyList, options Options) (*Fixture, error) {
@@ -132,6 +135,12 @@ func Cooldown(ft *Fixture) {
 	klog.Warningf("cooling down for %v", cooldownTime)
 	time.Sleep(cooldownTime)
 }
+
+func WaitForNRTSettle(fxt *Fixture) (*nrtv1alpha2.NodeResourceTopologyList, error) {
+	klog.Info("cooldown by verifying NRTs data is settled")
+	return intwait.With(fxt.Client).Interval(11*time.Second).Timeout(1*time.Minute).ForNodeResourceTopologiesSettled(context.TODO(), cooldownThreshold)
+}
+
 func setupNamespace(cli client.Client, baseName string, randomize bool) (corev1.Namespace, error) {
 	name := baseName
 	if randomize {
@@ -208,6 +217,19 @@ func getTeardownTimeFromEnvVar() time.Duration {
 	if err != nil {
 		klog.Errorf("cannot parse the provided test teardown time (fallback to default: %v): %v", defaultTeardownTime, err)
 		return defaultTeardownTime
+	}
+	return val
+}
+
+func getCooldownThresholdFromEnvVar() int {
+	raw, ok := os.LookupEnv("E2E_NROP_COOLDOWN_THRESHOLD")
+	if !ok {
+		return defaultCooldownThreshold
+	}
+	val, err := strconv.Atoi(raw)
+	if err != nil {
+		klog.Errorf("cannot parse the provided test cooldown threshold for resources (fallback to default: %v): %v", defaultCooldownThreshold, err)
+		return defaultCooldownThreshold
 	}
 	return val
 }
--- a/test/utils/fixture/fixture.go
+++ b/test/utils/fixture/fixture.go
@@ -32,7 +32,10 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 
+	nrtv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
+
 	"github.com/openshift-kni/numaresources-operator/internal/objects"
+	intwait "github.com/openshift-kni/numaresources-operator/internal/wait"
 	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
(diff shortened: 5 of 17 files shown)


shajmakh:
WIP:
need to add checks by end of each the test

test/e2e/serial/config/fixture.go
```diff
@@ -91,5 +94,14 @@ func NewFixtureWithOptions(nsName string, options e2efixture.Options) (*E2EConfi
 	cfg.SchedulerName = cfg.NROSchedObj.Status.SchedulerName
 	klog.Infof("detected scheduler name: %q", cfg.SchedulerName)
 
+	initialPFPState, err = intnrt.GetCurrentPFPState(cfg.Fixture.Client, context.TODO())
```
ffromani, line 97: does this belong to the Fixture or the E2Econfig?
if it belongs to the Fixture, it is recorded multiple times during the testsuite lifecycle, each per group of test (belonging to the same fixture!)
If it belongs to the E2EConfig, it is recorded once when all the testsuite runs, before any Fixture runs

internal/noderesourcetopology/noderesourcestopology.go
```diff
@@ -0,0 +1,87 @@
...
+		return 1
+	}
+	count.Count++
+	ps[nodeName] = count
+	return count.Count
+}
+
+func (ps PFPState) String() string {
+	return "TODO"
+}
```
ffromani, line 65: please delete

internal/noderesourcetopology/noderesourcestopology.go
```diff
@@ -0,0 +1,87 @@
...
+		nrt := &nrtList.Items[idx]
+		attr, ok := nrtv1alpha2attr.Get(nrt.Attributes, podfingerprint.Attribute)
+		if !ok {
+			return nil, fmt.Errorf("NRT %s missing PFP attribute", nrt.Name)
+		}
+		pfpState.Observe(nrt.Name, attr.Value)
+	}
+
+	return pfpState, nil
+}
```
ffromani, line 87: I'd actually extract more code from `ForNodeResourceTopologiesSettled`, all the code which gets the PFP state (including the optional `NodeName` filter) and reimplement `ForNodeResourceTopologiesSettled` on top of this helper

test/e2e/serial/config/fixture.go
shajmakh: It actually belongs to the e2econfig, that's the idea to record the pfp in before suite. Thanks for catching this, will fix it.

internal/noderesourcetopology/noderesourcestopology.go
```diff
@@ -0,0 +1,87 @@
+package noderesourcetopology
```
ffromani, line 1: please add the copyright boilerplate (check other files in `internal/`)

test/utils/fixture/fixture.go
```diff
@@ -95,10 +102,43 @@ func Teardown(ft *Fixture) error {
...
+		equal, err := isPFPEqual(ft.Client, initialPFP)
+		if err != nil {
+			return err
+		}
+		if !equal {
+			return errors.New("NRT data is not settled to the initial, PFP is different")
+		}
+	}
+
 	Cooldown()
```
shajmakh, line 116: not sure we still need this cooldown time. before removing it need to run the suite few times with 0 cooldown and see. note that cooldown is also used within the tests and not only here

ffromani: I think the logic you added should be put inside `Cooldown`.
If no initialPFP state available, just sleep like today. If available, don't sleep and wait for PFP to be restored (or fail)

ffromani:
After some more thinking I think we can pivot a bit.
in `test/e2e/serial/config/fixture.go` we have the **Suite fixture** (which is the context per-suite implemented in `E2EConfig` - naming is indeed inconsistent here). This is meant to be called once per suite run, using `SetupFixture` and `TeardownFixture` in turn called by ginkgo's `Setup` and `Teardown`. We should hook here.

When the suite starts, we do grab a snapshot of `nrtList` (check `NewFixtureWithOptions`).
Note, for extra confusion, we DO reuse the test-level fixture code to create the `e2e-test-infra` pseudo fixture which has the same lifecycle of the suite, outspanning the test-group-fixtures.

But the takeaway is that (we need a cleanup and rename and) `E2EConfig` holds a `nrtList` grabbed right at the startup when the cluster is assumed idle and stable, so we can use THAT `nrtList` in` Cooldown` to wait for the cluster to get back to that state. For that we can probably just reuse `internal/wait/ForNodeResourceTopologiesEqualTo`

test/utils/fixture/fixture.go
shajmakh: good point! thanks for raising it. will handle it.

internal/noderesourcetopology/noderesourcestopology.go
shajmakh: Thanks for your note. no PFP updates are done thus no longer relevant.

test/e2e/serial/tests/non_regression_fundamentals.go
```diff
@@ -66,11 +66,11 @@ var _ = Describe("[serial][fundamentals][scheduler][nonreg] numaresources fundam
 	Context("using the NUMA-aware scheduler without NRT data", func() {
 		var testPod *corev1.Pod
 
-		// BeforeEach(func() {
```
shajmakh: likely to be addressed in a later PR

shajmakh:
Following offline conversation, the points highlighted by Francesco were taken into consideration and the PR now verifies that the NRTs data after each test is matching to the initial data that was recorded by the suite's setup. 
Thanks @ffromani for the valuable review and insights!

ffromani:
please rebase on top of https://github.com/openshift-kni/numaresources-operator/pull/614

test/e2e/serial/config/fixture.go
```diff
@@ -77,7 +77,7 @@ func NewFixtureWithOptions(nsName string, options e2efixture.Options) (*E2EConfi
 	if err != nil {
 		return nil, err
 	}
-
+	cfg.Fixture.InitialNRTList = cfg.NRTList
```
ffromani, line 80: Is this needed? the test fixture we use here is for infra (numacell ds) so it seems this can be avoided?

test/e2e/serial/tests/configuration.go
```diff
@@ -71,7 +71,7 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
 		Expect(serialconfig.Config.Ready()).To(BeTrue(), "NUMA fixture initialization failed")
 
 		var err error
-		fxt, err = e2efixture.Setup("e2e-test-configuration")
+		fxt, err = e2efixture.Setup("e2e-test-configuration", serialconfig.Config.NRTList)
```
ffromani, line 74: I wonder if we can just pass `serialconfig.Config`. But this generalization should probably wait for a future PR

test/utils/fixture/fixture.go
```diff
@@ -62,6 +62,16 @@ const (
...
+var (
+	teardownTime time.Duration
+	cooldownTime time.Duration
+)
+
+func init() {
+	cooldownTime = getCooldownTimeFromEnvVar()
+	teardownTime = getTeardownTimeFromEnvVar()
+}
+
```
ffromani: this is a good change because these 2 values are expected to be stable across the serial test suite run, so it makes sense (it's actually more correct) to fetch the value once at suite start

test/utils/fixture/fixture.go
```diff
@@ -203,3 +211,11 @@ func getTeardownTime() time.Duration {
...
 	return val
 }
+
+func GetCoolDownTime() time.Duration {
+	return cooldownTime
+}
+
+func GetTeardownTime() time.Duration {
+	return teardownTime
+}
```
ffromani, line 237: I'm not sure we need these getters.

test/utils/fixture/fixture.go
```diff
@@ -63,13 +64,15 @@ const (
 )
 
 var (
-	teardownTime time.Duration
-	cooldownTime time.Duration
+	teardownTime         time.Duration
+	cooldownTime         time.Duration
+	resCooldownThreshold int
```
ffromani, line 69: let's call this `cooldownThreshold`

test/utils/fixture/fixture.go
```diff
@@ -212,10 +215,27 @@ func getTeardownTimeFromEnvVar() time.Duration {
...
 	return cooldownTime
 }
 
 func GetTeardownTime() time.Duration {
 	return teardownTime
 }
+
+func GetResourcesCooldownThreshold() int {
+	return resCooldownThreshold
+}
```
ffromani, line 241: why do we need this getter?

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -287,7 +287,7 @@ var _ = Describe("[serial][disruptive][scheduler][resacct] numaresources workloa
 			Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s", updatedPod.Namespace, updatedPod.Name, serialconfig.Config.SchedulerName)
 
 			By("Waiting for the NRT data to stabilize")
-			wait.With(fxt.Client).Interval(11*time.Second).Timeout(1*time.Minute).ForNodeResourceTopologiesSettled(context.TODO(), 3)
+			wait.With(fxt.Client).Interval(11*time.Second).Timeout(1*time.Minute).ForNodeResourceTopologiesSettled(context.TODO(), e2efixture.GetResourcesCooldownThreshold())
```
ffromani, line 290: let's just call `Cooldown` here (and in all the other occorrences below)
Or, equally or perhaps better:

- extract a helper from `Cooldown` which waits for the NRTs to Settle (`Settle(ft *e2efixture.Fixture)`?) or does nothing if the env var has the right value/no initial NRT data
- call the new helper in `Cooldown` (trivial code movement)
- call the new helper here (and in the following)

ffromani (commented):
/approve

thanks for the PR, the direction looks good, commenting about improvements

test/e2e/serial/tests/configuration.go
shajmakh: we certainly can, and I think since we are already on it let's have it at once, unless you see a strong reason not to. I'll have it updated to `serialconfig.Config` let me know if it works for you.

test/e2e/serial/config/fixture.go
shajmakh: right, I might have mixed between the setups. Thanks for raising that.

test/e2e/serial/tests/configuration.go
shajmakh: certainly we can, but this will require a change in the structure of the suite. the minimal change would be moving E2EConfig struct to somewhere else than under serial/ (for example utils/configuration/configuration.go or some other file with known types across the project - not necessarily e2e). on one hand this enables us passing the whole suite config to anywhere, but OTOH its only instance would be initialized in serial. I don't see it suites anywhere else other than serial though. 

we cannot simply pass the serialconfig.Config to utils/fixture because this will create an import loop. and if otherwise we could have just used serialconfig.Config.NRTList directly in utils/fixture.

test/utils/fixture/fixture.go
shajmakh: right, we don't need these two. thanks for pointing this out.

test/utils/fixture/fixture.go
shajmakh: because we want to protect this from manipulations by other parts of the code, and to retain its original value we got from the env var.

test/e2e/serial/tests/configuration.go
ffromani: right, there could be an import loop. I think there's no need to embark in a large refactoring now and we can proceed with the direction you already took.

test/utils/fixture/fixture.go
```diff
@@ -51,8 +50,9 @@ type Fixture struct {
 }
 
 const (
-	defaultTeardownTime = 180 * time.Second
-	defaultCooldownTime = 30 * time.Second
+	defaultTeardownTime         = 180 * time.Second
+	defaultCooldownTime         = 30 * time.Second
+	defaultResCooldownThreshold = 5
```
ffromani, line 55: `defaultCooldownThreshold`?

test/e2e/serial/tests/resource_accounting.go
ffromani: I think we do the above. The reason why I still think this is the best approach is that, doing like this, all the consumers of the variable are within the `fixture` package. This means we don't need the getter anymore and that we keep more variables private to the package, reducing the coupling between packages (`fixture` and test code)

ffromani (commented):
thanks for your updates and your work. I think we're almost there, just the last round of questions/suggestions

test/utils/fixture/fixture.go
ffromani: Adding a getter is indeed a way (pretty much the only way in golang) to enforce the fact the client code should access the value in read-only and not change it. The change is, of course, done intentionally (there are no accidents in code) but is missed during the review or the testing or both and causes an unwanted mutation.

However adding a getter fixes the problem, but increases the API surface (functions AND exposed variables are part of the API! the former is better than the latter). A even better fix is not increasing the API surface, and incresae cohesion while reducing coupling.

We can call "cohesion" how much the related code is logically close together. Accessing a global variable, for example, has likely low cohesion because references to the variable can be spread all across the codebase.
Accessing a local private member has higher cohesion, because all the code has to be confined in the same package.

We can call "coupling" how much two pieces of code depend on each other. A package coupled to another, perhaps because of bad API, need to change when the other package change, because for example it depends on internal behavior and not on the interface, or because the interface is too big.

Having a getter can _potentially_ decrease cohesion and (by definition, because it is adding to the API surface) always increase coupling.
There's a possible way, I believe, to increse cohesion _AND_ reduce coupling if we change `Cooldown` a bit, and I'm suggesting that in another comment.

test/e2e/serial/tests/resource_accounting.go
shajmakh: I agree with the idea. But confused that fixture cooldown wait for the NRTs to settle to initial while in the other places where we use the cooldown threshold it's not necessarily that we wait for the NRTs to settle to initial, but actually to simply settle. so we cannot call cooldown as is in these places. but they can be combined. will add a commit to addressing this.

ffromani: sure, calling newly-added `Settle` wherever needed is fine

shajmakh:
/hold 
commits are not aligned to the changes, need to fix that.

test/e2e/serial/tests/resource_accounting.go
shajmakh: apologies for the late updates, the commits are ready for another round of review when possible. 
so eventually i didn't like the combination of the two part (settle-for-initial and just-settle-for-the-cooldown-threshold because each is used in different places, so either this or that. Yet, the cooldown threshold is called only in the fixture. let me know what you think.

shajmakh:
/hold cancel
Thanks @ffromani for the beneficial review and comments!

PR #613, opened 2023-06-23
e2e: remove unwanted skip

shajmakh:
Seems like 3 tests were accedentally skipped in both scopes because of this check. The removed check is pointeless because the nrtList is never empty because it is collected in upper level BeforeEach. And even if it is, the tests should not be skipped. I believe the intention was to announce that there tests doesn't actually care about NRT data, as all they care about is verifying that pods of different QoSs can be scheduled with SS.

### DIFF
--- a/test/e2e/serial/tests/non_regression_fundamentals.go
+++ b/test/e2e/serial/tests/non_regression_fundamentals.go
@@ -63,14 +63,8 @@ var _ = Describe("[serial][fundamentals][scheduler][nonreg] numaresources fundam
 		Expect(err).NotTo(HaveOccurred())
 	})
 
-	Context("using the NUMA-aware scheduler without NRT data", func() {
+	Context("using the NUMA-aware scheduler without verifying NRT data", func() {
 		var testPod *corev1.Pod
-
-		BeforeEach(func() {
-			if len(nrtList.Items) > 0 {
-				e2efixture.Skip(fxt, "this test require empty NRT data")
-			}
-		})
 
 		AfterEach(func() {
 			if testPod == nil {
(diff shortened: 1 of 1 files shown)


ffromani:
the intent of these tests is to do non-regression tests and check the numa-aware scheduler tolerate (surely for best-effort and burstable) lack of NRT data. So the skip somehow conveys the intent, even thoug it's possible there's a bug which always make the tests skip.
Let's fix the code, but just removing the skip doesn't seem the correct approach considering the intent

shajmakh:
> check the numa-aware scheduler tolerate (surely for best-effort and burstable) lack of NRT data. 

Thanks for your comment. If I get you correctly, then what could be done to verify that ^ is checking that the initial NRT is equivalent to the one after scheduling the workload. OTOH, IIRC there are several distributed tests along serial suite that don't check NRT updates. If we want to address this area in all other tests we can possibly create a separate PR to tackle that but I think it'll overload the logs and the code. what do you think?

shajmakh:
/hold 
needs https://github.com/openshift-kni/numaresources-operator/pull/610/commits/cec6850179495819dc5fcfb35f85a91eecfbd102

shajmakh:
For these tests, an NRT check post workload scheduling was added.

shajmakh:
/retest
looks like a connection issue:

`Caching blobs under "/var/cache/blobs".
Trying to pull registry.build05.ci.openshift.org/ci/managed-clonerefs@sha256:c59da5f1725324d953791ca9b80221a653a338f710449cf8efbd62830e66fb89...
Getting image source signatures
error: error creating buildah builder: reading signatures: downloading signatures for sha256:c59da5f1725324d953791ca9b80221a653a338f710449cf8efbd62830e66fb89 in registry.build05.ci.openshift.org/ci/managed-clonerefs: received unexpected HTTP status: 504 Gateway Time-out
INFO[2023-06-27T21:02:17Z] Build src previously failed from an infrastructure error (FetchImageContentFailed), retrying... 
INFO[2023-06-27T21:03:17Z] Created build "src"                          `

ffromani:
> > check the numa-aware scheduler tolerate (surely for best-effort and burstable) lack of NRT data.
> 
> Thanks for your comment. If I get you correctly, then what could be done to verify that ^ is checking that the initial NRT is equivalent to the one after scheduling the workload. OTOH, IIRC there are several distributed tests along serial suite that don't check NRT updates. If we want to address this area in all other tests we can possibly create a separate PR to tackle that but I think it'll overload the logs and the code. what do you think?

The main thing is we should have some tests which run WITHOUT NRT data at all (no object!). This means we also need a way to pause/stop RTEs populating them, which we don't have atm

ffromani:
filed https://github.com/openshift-kni/numaresources-operator/issues/631 for tracking

shajmakh:
Relevant modifications will be handled in #700.

PR #614, opened 2023-06-23
ghactions: kind: update image

ffromani:
The runners are consuming kind 0.20 now, and we're stuck on kube 1.24 for no good reason. Bump to 1.26.


swatisehgal (approved):
/lgtm

PR #617, opened 2023-06-26
project: layout: move helpers in `cmd/`

ffromani:
move utilities in the `cmd` tree; also add a note to explain why the operator entry point ad the RTE entry point are NOT here.


ffromani:
Reviewers please note this PR should **not** be backported and is relevant only for 4.14 and onwards.

PR #623, opened 2023-06-27
e2e: print random seed being used

ffromani:
We reset the random seed, so let's make sure we print what we use. Ideally we should integrate with ginkgo and/or make the seed configurable. Let's at least make it explicit the seed we use as starting point.

### DIFF
--- a/test/e2e/rte/rte_e2e_suite_test.go
+++ b/test/e2e/rte/rte_e2e_suite_test.go
@@ -37,6 +37,12 @@ import (
 	_ "github.com/k8stopologyawareschedwg/resource-topology-exporter/test/e2e/topology_updater"
 )
 
+var (
+	BinaryPath string
+
+	randomSeed int64
+)
+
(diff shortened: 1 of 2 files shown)


test/e2e/rte/rte_e2e_suite_test.go
```diff
@@ -37,6 +37,12 @@ import (
 	_ "github.com/k8stopologyawareschedwg/resource-topology-exporter/test/e2e/topology_updater"
 )
 
+var (
+	BinaryPath string
+
```
shajmakh, line 42: purposed empty line?

shajmakh:
/lgtm
Thanks for the PR.

test/e2e/rte/rte_e2e_suite_test.go
ffromani: yes!

swatisehgal (approved):
/lgtm

PR #628, opened 2023-06-28
e2e cleanup 20230628

ffromani:
further improvements and cleanups following the implementation of the NRT settlement support

### DIFF
--- a/internal/noderesourcetopology/equality_test.go
+++ b/internal/noderesourcetopology/equality_test.go
@@ -1,0 +1,208 @@
+/*
+ * Copyright 2023 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package noderesourcetopology
+
+import (
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+
+	nrtv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
+)
+
+func TestEqualZones(t *testing.T) {
+	testCases := []struct {
+		name     string
+		data1    nrtv1alpha2.ZoneList
+		data2    nrtv1alpha2.ZoneList
+		expected bool
+	}{
+		{
+			name: "sorted",
+			data1: nrtv1alpha2.ZoneList{
+				{
+					Name: "Zone-000",
+					Resources: nrtv1alpha2.ResourceInfoList{
+						{
+							Name:        "foo",
+							Capacity:    resource.MustParse("2"),
+							Allocatable: resource.MustParse("2"),
+							Available:   resource.MustParse("2"),
+						},
+						{
+							Name:        "bar",
+							Capacity:    resource.MustParse("4"),
+							Allocatable: resource.MustParse("4"),
+							Available:   resource.MustParse("4"),
+						},
+					},
+				},
+				{
+					Name: "Zone-001",
+					Resources: nrtv1alpha2.ResourceInfoList{
+						{
+							Name:        "bar",
+							Capacity:    resource.MustParse("4"),
+							Allocatable: resource.MustParse("4"),
+							Available:   resource.MustParse("4"),
+						},
+						{
+							Name:        "foo",
+							Capacity:    resource.MustParse("2"),
+							Allocatable: resource.MustParse("2"),
+							Available:   resource.MustParse("2"),
+						},
+					},
+				},
+			},
+			data2: nrtv1alpha2.ZoneList{
+				{
+					Name: "Zone-001",
+					Resources: nrtv1alpha2.ResourceInfoList{
+						{
+							Name:        "foo",
+							Capacity:    resource.MustParse("2"),
+							Allocatable: resource.MustParse("2"),
+							Available:   resource.MustParse("2"),
+						},
+						{
+							Name:        "bar",
+							Capacity:    resource.MustParse("4"),
+							Allocatable: resource.MustParse("4"),
+							Available:   resource.MustParse("4"),
+						},
+					},
+				},
+				{
+					Name: "Zone-000",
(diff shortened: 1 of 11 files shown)


swatisehgal (approved):
/lgtm

Changes look reasonable to me. I will defer whether we want to backport this or not to @shajmakh.

/hold
to avoid inadvertent merge.

shajmakh:
Thank you for the cleanup.
/lgtm too.
up to you @ffromani to backport following your comment [here](https://github.com/openshift-kni/numaresources-operator/pull/626#issuecomment-1610816523)

shajmakh:
Thank you for the cleanup.
/lgtm
up to you @ffromani to backport following your comment https://github.com/openshift-kni/numaresources-operator/pull/626#issuecomment-1610816523

test/utils/fixture/fixture.go
```diff
@@ -128,17 +134,19 @@ func Skipf(ft *Fixture, format string, args ...interface{}) {
 
 func Cooldown(ft *Fixture) {
 	if len(ft.InitialNRTList.Items) > 0 {
-		ginkgo.By("cooldown by verifying NRTs data is settled to the initial state")
-		intwait.With(ft.Client).Interval(5*time.Second).Timeout(1*time.Minute).ForNodeResourceTopologiesEqualTo(context.TODO(), &ft.InitialNRTList, intwait.NRTIgnoreNothing)
+		interval := 5 * time.Second
+		ginkgo.By(fmt.Sprintf("cooldown by verifying NRTs data is settled to the initial state (interval=%v timeout=%v)", interval, settleTimeout))
+		intwait.With(ft.Client).Interval(interval).Timeout(settleTimeout).ForNodeResourceTopologiesEqualTo(context.TODO(), &ft.InitialNRTList, intwait.NRTIgnoreNothing)
+		// TODO: shall we wait for the data to settle?
```
shajmakh, line 149: I don't think we need this TODO because we're already waiting enough time for the NRT to equal to initial.

test/e2e/serial/config/infra.go
```diff
@@ -97,8 +98,20 @@ func setupNUMACell(fxt *e2efixture.Fixture, nodeGroups []nropv1.NodeGroup, timeo
...
+	klog.Infof("daemonsets ready (%d)", len(dss))
+
+	klog.Infof("e2e infra setup completed")
+}
+
+func waitAllDSReady(fxt *e2efixture.Fixture, dss []*appsv1.DaemonSet, timeout time.Duration) {
 	var wg sync.WaitGroup
-	for _, ds := range dss {
+	for idx := range dss {
+		ds := dss[idx]
```
shajmakh, line 113: why? we're not using idx

test/utils/fixture/fixture.go
```diff
@@ -136,8 +137,12 @@ func Cooldown(ft *Fixture) {
 	if len(ft.InitialNRTList.Items) > 0 {
 		interval := 5 * time.Second
 		ginkgo.By(fmt.Sprintf("cooldown by verifying NRTs data is settled to the initial state (interval=%v timeout=%v)", interval, settleTimeout))
-		intwait.With(ft.Client).Interval(interval).Timeout(settleTimeout).ForNodeResourceTopologiesEqualTo(context.TODO(), &ft.InitialNRTList, intwait.NRTIgnoreNothing)
 		// TODO: shall we wait for the data to settle?
+		currentNrtList, err := intwait.With(ft.Client).Interval(interval).Timeout(settleTimeout).ForNodeResourceTopologiesEqualTo(context.TODO(), &ft.InitialNRTList, intwait.NRTIgnoreNothing)
+		if err != nil {
+			diff := cmp.Diff(ft.InitialNRTList.Items, currentNrtList.Items)
+			klog.Warningf("NRT MISMATCH:\n----\n%s\n---\n", diff)
+		}
```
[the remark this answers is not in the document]
ffromani: Not sure still. Probably yes.

test/utils/fixture/fixture.go
ffromani: sounds good. Removing.

ffromani:
the PR is now complete and ready for review

test/utils/fixture/fixture.go
```diff
@@ -136,8 +137,12 @@ func Cooldown(ft *Fixture) {
 	if len(ft.InitialNRTList.Items) > 0 {
 		interval := 5 * time.Second
 		ginkgo.By(fmt.Sprintf("cooldown by verifying NRTs data is settled to the initial state (interval=%v timeout=%v)", interval, settleTimeout))
-		intwait.With(ft.Client).Interval(interval).Timeout(settleTimeout).ForNodeResourceTopologiesEqualTo(context.TODO(), &ft.InitialNRTList, intwait.NRTIgnoreNothing)
 		// TODO: shall we wait for the data to settle?
+		currentNrtList, err := intwait.With(ft.Client).Interval(interval).Timeout(settleTimeout).ForNodeResourceTopologiesEqualTo(context.TODO(), &ft.InitialNRTList, intwait.NRTIgnoreNothing)
+		if err != nil {
+			diff := cmp.Diff(ft.InitialNRTList.Items, currentNrtList.Items)
```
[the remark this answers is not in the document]
ffromani: The variable is initialized outside the `k8swait.PollImmediate` and captured by its closure, so it should always contain the most recent Get value, isn't it?

ffromani:
Once approved and merged, I plan to backport this PR down to 4.12, but not immediately. Let's see few runs of it against 4.14 before doing all the backport dance.

test/utils/fixture/fixture.go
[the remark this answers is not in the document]
shajmakh: I see the variable is declared but not initialized so it's zero-value will be nil. you are right, it will always contain the most recent value of Get/List but in case List kept failing the value will stay nil. Unless List do initialize an empty nrt-list in case of failure. I'm not sure about that, I'm interested to verify this.
Anyhow,its possibility to fail is close to zero, so we can keep it as is.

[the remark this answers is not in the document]
ffromani: true, but if `currentNrtList` is zero-value, then `err` will be != nil, so we clearly catch this case

internal/noderesourcetopology/equality_test.go
```diff
@@ -0,0 +1,132 @@
...
+						{
+							Name:        "bar",
+							Capacity:    resource.MustParse("4"),
+							Allocatable: resource.MustParse("4"),
+							Available:   resource.MustParse("4"),
+						},
+					},
+				},
+			},
+			expected: true,
```
shajmakh, line 110: I think it's nearly enough to test this case. 
just to clarify, don't we want usually to add negative case chack, like the expected to be false, just to verify the equality check also works for the inner items as expected?

ffromani: good point. Yes we should add at least a negative case

shajmakh:
Thanks for the beneficial enhancements!
/lgtm

shajmakh:
Thanks for adding the test
/lgtm

PR #629, opened 2023-06-28
e2e: add missing wait for NRT to settle

shajmakh:
Some tests fail often in CI on inconsistent accounting in NRT data, and pass after rerun. To assure the NRT data is relaible at the checking point, need to wait enough for the data to be settled and only then to compare the outputs.


shajmakh:
will add more cleanups later, but this one helps stabilize the tests, so it can be alone in the PR.

ffromani (commented):
/lgtm
/approve

using settle as done here seems like a good idea in general

ffromani:
@shajmakh thanks for the PR. Please backport as you see fit.

PR #635, opened 2023-07-05
e2e: serial: support kubelet defaults in validation

ffromani:
The kubelet sets defaults for Topology Manager policy and scope when a value is not explicitely set in the configuration.
Let's support this scenario, copying the kubelet defaults. 
We should get them programmatically and from the right kubelet version, but the saving grace is the values don't change
frequently (if at all).

In addition, starting NRT API v1alpha2, the "topologyPolicies" field is deprecated in favor of Attributes.
Albeit not  needed for the enhancement,  we take the chance to make this code more forward-looking switching to comparing attributes.


shajmakh:
/lgtm
Thanks for the fix and the enhancments.

PR #638, opened 2023-07-06
Makefile: build: use shared libraries

ffromani:
In some cases we don't really need the fully static compilation and we can depend on system libraries.


swatisehgal (approved):
/lgtm

Please unhold when you see fit.

jlojosnegros (commented):
/lgtm

ffromani:
/hold cancel

let's not backport this yet (if ever)

PR #639, opened 2023-07-06
operator: rte: default inforefreshmode=periodic

ffromani:
We documented that the default InfoRefreshMode value should be "Periodic", not "PeriodicAndEvents", so let's fix the code accordingly.


swatisehgal (approved):
/lgtm

ffromani:
infra issues so far

shajmakh:
@ffromani do we need this cherry-picked to release-4.13?

ffromani:
yes and also to 4.12. But this will require full validation (= full run of the serial suite) so let's plan accordingly.

PR #556, opened 2023-05-11
scheduler: enable reserve plugin and cache by default

ffromani:
Time to enable by default the cacheResyncPeriod.
We will handle the compatibility just requiring in the release notes a minimum version of the scheduler plugin.

PR #640, opened 2023-07-07
serial: fetch CacheResyncPeriod from Status

shajmakh:
When running the tests on a cluster with default configuration, the default value of scheduler's CacheResyncPeriod is reflected in the object's status and not in spec. The tests do not change this field so they run on default configuration and when fetching it from spec the value is not found hence skipping tests.

Fetch the scheduler's CacheResyncPeriod from its status instead to enable running the tests. This is also true if the value is not default.


ffromani (approved):
good catch. Since we enabled by default in #556 and its backports, this is the right thing to do.

ffromani:
/lgtm

@shajmakh please backpor t as you see fit

ffromani:
infra issues, let's retry later because it seems we're running low on CI capacity

PR #653, opened 2023-07-14
e2e: fix RTE lane

ffromani:
We previously failed to bail out early if the RTE lane was failing.
The runner script ultimately returns the exit code of the last executed command. Problem is: also the latest `if` block
is a command, and the evaluation of the command was successful, swallowing the RTE lane failure (if any).

The test per se seems to be failed because we incorrectly always check for the exporter binary availability. But the exporter binary is not meant to be here, in this lane: it is tested explicitely in the separate RTE local lane.

### DIFF
--- a/hack/run-test-e2e.sh
+++ b/hack/run-test-e2e.sh
@@ -40,6 +40,10 @@ trap '{ if [ "${NO_TEARDOWN}" = false ]; then
              echo "Running NROScheduler uninstall test suite";
              ${BIN_DIR}/e2e-nrop-sched-uninstall.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.junit-report=${REPORT_DIR}/sched-uninstall.xml
           fi
+
+          echo "Undeploying sample devices for RTE tests"
+          rte/hack/undeploy-devices.sh
+
           echo "Running NRO uninstall test suite";
 	  ${BIN_DIR}/e2e-nrop-uninstall.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.junit-report=${REPORT_DIR}/uninstall.xml
         fi }' EXIT SIGINT SIGTERM SIGSTOP
@@ -56,6 +60,21 @@ fi
 
 
 echo "Running Functional Tests: ${GINKGO_SUITS}"
+
+echo "Deploying sample devices for RTE tests"
+rte/hack/deploy-devices.sh
+if [ $? -ne 0 ]; then
+    echo "Failed to deploy sample device plugin for RTE tests"
+    exit 2
+fi
+
+rte/hack/check-ds.sh oc sampledevices device-plugin-a-ds
+if [ $? -ne 0 ]; then
+    echo "Failed to verify sample device plugin for RTE tests"
+    exit 4
+fi
+
+echo "Running RTE tests"
 export E2E_TOPOLOGY_MANAGER_POLICY="${E2E_TOPOLOGY_MANAGER_POLICY:-SingleNUMANodePodLevel}"
 # -v: print out the text and location for each spec before running it and flush output to stdout in realtime
 # -timeout: exit the suite after the specified time
--- a/hack/run-test-e2e.sh
+++ b/hack/run-test-e2e.sh
@@ -75,13 +75,15 @@ if [ $? -ne 0 ]; then
 fi
 
 echo "Running RTE tests"
-export E2E_TOPOLOGY_MANAGER_POLICY="${E2E_TOPOLOGY_MANAGER_POLICY:-SingleNUMANodePodLevel}"
+export E2E_TOPOLOGY_MANAGER_POLICY="${E2E_TOPOLOGY_MANAGER_POLICY}"
+export E2E_TOPOLOGY_MANAGER_SCOPE="${E2E_TOPOLOGY_MANAGER_SCOPE}"
 # -v: print out the text and location for each spec before running it and flush output to stdout in realtime
 # -timeout: exit the suite after the specified time
 # -r: run suites recursively
 # --fail-fast: ginkgo will stop the suite right after the first spec failure
 # --flake-attempts: rerun the test if it fails
 # -requireSuite: fail if tests are not executed because of missing suite
+echo "TM config policy=[${E2E_TOPOLOGY_MANAGER_POLICY}] scope=[${E2E_TOPOLOGY_MANAGER_SCOPE}]"
 ${BIN_DIR}/e2e-nrop-rte.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.fail-fast --ginkgo.flake-attempts=2 --ginkgo.junit-report=${REPORT_DIR}/e2e.xml --ginkgo.skip='\[Disruptive\]|\[StateDirectories\]|\[NodeRefresh\]|\[local\]' --ginkgo.focus='\[release\]'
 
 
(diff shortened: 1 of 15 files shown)


swatisehgal (approved):
/lgtm

ffromani:
/retest

it seems infra issue?

ffromani:
actual error:
```
End Captured GinkgoWriter Output[0m

  [38;5;9merr=unable to upgrade connection: container not found ("resource-topology-exporter") stderr=
  Unexpected error:
      <*errors.errorString | 0xc0001145a0>: {
          s: "unable to upgrade connection: container not found (\"resource-topology-exporter\")",
      }
      unable to upgrade connection: container not found ("resource-topology-exporter")
  occurred[0m
  [38;5;9mIn [1m[It][0m[38;5;9m at: [1m/go/src/github.com/openshift-kni/numaresources-operator/test/e2e/rte/rte_test.go:315
```

hack/run-test-e2e.sh
```diff
@@ -64,7 +64,10 @@ export E2E_TOPOLOGY_MANAGER_POLICY="${E2E_TOPOLOGY_MANAGER_POLICY:-SingleNUMANod
 # --flake-attempts: rerun the test if it fails
 # -requireSuite: fail if tests are not executed because of missing suite
 ${BIN_DIR}/e2e-nrop-rte.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.fail-fast --ginkgo.flake-attempts=2 --ginkgo.junit-report=${REPORT_DIR}/e2e.xml --ginkgo.skip='\[Disruptive\]|\[StateDirectories\]|\[NodeRefresh\]|\[local\]'
```
Tal-or, line 66: We can annotate the bash script with `set -e` if we expect nothing to fail, instead of checking the exit code for each test

ffromani: Yes, probably better to switch to `-e`. I was looking for a targeted fix but it seems we need a deep rework here.

ffromani: yes

ffromani:
/hold

surely broken until RTE v0.12.0 os released and consumed

ffromani:
depends on #663

ffromani:
requires: https://github.com/openshift/release/pull/41614

ffromani:
ended up rewriting a good part of the lane, and there are a lot of potential followups, but fixed at last and reviewable.

hack/run-test-e2e.sh
```diff
@@ -56,6 +60,21 @@ fi
 
 
 echo "Running Functional Tests: ${GINKGO_SUITS}"
+
+echo "Deploying sample devices for RTE tests"
+rte/hack/deploy-devices.sh
+if [ $? -ne 0 ]; then
+    echo "Failed to deploy sample device plugin for RTE tests"
+    exit 2
```
Tal-or: Why different exit codes? we have the log above for indicating where the flow breaks

rte/hack/deploy-devices.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -ex
```
Tal-or, line 3: Do we want to failed hard here in case one of the created objects already exists?

rte/hack/manifests/sample-devices/daemonset-dpa.yaml
```diff
@@ -0,0 +1,36 @@
...
+    metadata:
+      labels:
+        name: dpa
+    spec:
+      serviceAccountName: sampledevices-sa
+      containers:
+        - name: dpa-container
+          securityContext:
+            privileged: true
+          image: quay.io/k8stopologyawareschedwg/sample-device-plugin:v0.1.1
```
Tal-or, line 20: This image is fine for past and future releases?
If not we should make it configurable

rte/hack/manifests/sample-devices/daemonset-dpa.yaml
```diff
@@ -0,0 +1,36 @@
...
+            - name: DEVICE_RESOURCE_NAME
+              value: "example.com/deviceA"
+          volumeMounts:
+            - name: kubeletsockets
+              mountPath: /var/lib/kubelet/device-plugins
+            - name: device-a-config-vol
+              mountPath: /etc/devices
+      volumes:
+        - name: kubeletsockets
+          hostPath:
```
Tal-or, line 32: I would add 
`type: Socket`
```
 hostPath:
   path: /var/lib/kubelet/device-plugins
   type: Socket
```

tools/catkubeletconfmap/catkubeletconfmap.go
```diff
@@ -0,0 +1,55 @@
+package main
```
Tal-or, line 1: license statement?

hack/run-test-e2e.sh
```diff
@@ -86,11 +37,26 @@ export E2E_TOPOLOGY_MANAGER_SCOPE="${E2E_TOPOLOGY_MANAGER_SCOPE}"
...
+  echo "Running Functional Tests: ${GINKGO_SUITS}"
+  # -v: print out the text and location for each spec before running it and flush output to stdout in realtime
+  # -timeout: exit the suite after the specified time
+  # -r: run suites recursively
+  # --fail-fast: ginkgo will stop the suite right after the first spec failure
+  # --flake-attempts: rerun the test if it fails
+  # -requireSuite: fail if tests are not executed because of missing suite
+  ${BIN_DIR}/e2e-nrop-sched.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.fail-fast --ginkgo.flake-attempts=2 --ginkgo.junit-report=${REPORT_DIR}/e2e-sched.xml
+
+  echo "Running NROScheduler uninstall test suite";
```
Tal-or, line 54: If it used only for CI purposes then the uninstall part just consumes time, because the cluster would get deleted anyway.

hack/run-test-e2e.sh
ffromani: because this way we have a relatively easy way to handle different failures programmatically. In this case however it is going to ben overridden by another change, so I guess it won't matter much

rte/hack/manifests/sample-devices/daemonset-dpa.yaml
ffromani: good point, I'll add.

tools/catkubeletconfmap/catkubeletconfmap.go
ffromani: good point, adding the license boilerplate. Silly me just forgot.

hack/run-test-e2e.sh
ffromani: you're right. I'm reluctant to remove it just yet because if memory serves we do some test in the deinstall path. I can soublecheck and remove if I'm remember wrongly.

rte/hack/manifests/sample-devices/daemonset-dpa.yaml
ffromani: adding `type`, but we want to mount a `Directory` so changing accordingly.

hack/run-test-e2e.sh
ffromani: No tests on uninstall path, but let's discuss the removal in a separate PR

tools/catkubeletconfmap/catkubeletconfmap.go
ffromani: let's add them all in one go: https://github.com/openshift-kni/numaresources-operator/pull/673

hack/run-test-e2e.sh
Tal-or: ack

Tal-or:
/lgtm
/approve
Thanks for the changes

PR #654, opened 2023-07-15
e2e: rte lane cleanups

ffromani:
improve the e2e rte lane to prepare to for future fixes (see #653 ).


ffromani:
on second thought we probably don't need this split anymore. Focusing on #653

PR #658, opened 2023-07-21
controllers: add related-objects support

jlojosnegros:
Add relatedObjects support to better integrate with must-gather for both numaresourcesopeator and numaresourcesscheduler objects


controllers/numaresourcesoperator_controller.go
```diff
@@ -590,6 +592,57 @@ func (r *NUMAResourcesOperatorReconciler) mcpToNUMAResourceOperator(mcpObj clien
...
+	ret := []configv1.ObjectReference{
+		{
+			Resource: "namespaces",
+			Name:     r.Namespace,
+		},
+		{
+			Group:     "machineconfiguration.openshift.io",
+			Resource:  "kubeletconfigs",
+			Namespace: "",
+			Name:      "",
```
ffromani, line 606: Do we need to set namespace/name explicitely here? If so, let's add this data; if not, no need to explicit initialization with zero-value strings (= let's just omit these here and below)

controllers/numaresourcesscheduler_controller.go
```diff
@@ -153,6 +156,24 @@ func (r *NUMAResourcesSchedulerReconciler) reconcileResource(ctx context.Context
 
 }
 
+func (r *NUMAResourcesSchedulerReconciler) getRelatedObjects(dpStatus nropv1.NamespacedName) []configv1.ObjectReference {
+	// 'Resource' should be in lowercase and plural
+	return []configv1.ObjectReference{
+		{
+			Group:     "",
```
ffromani, line 163: same comment for zero-value objects

ffromani (commented):
Looks nice (but few inline comments). Let's have test coverage at least with `controllers/` integration tests

ffromani (commented):
looks good. Let's add the tests and it should be good to go. We will evaluate backports later.

controllers/controllers_suite_test.go
```diff
@@ -53,6 +53,7 @@ var _ = BeforeSuite(func() {
 	testEnv = &envtest.Environment{
 		CRDDirectoryPaths:     []string{filepath.Join("..", "config", "crd", "bases")},
 		ErrorIfCRDPathMissing: true,
+		BinaryAssetsDirectory: "/home/jojosneg/.local/share/kubebuilder-envtest/k8s/1.24.2-linux-amd64",
```
ffromani, line 56: please remove

controllers/numaresourcesoperator_controller.go
```diff
@@ -590,6 +592,47 @@ func (r *NUMAResourcesOperatorReconciler) mcpToNUMAResourceOperator(mcpObj clien
 	return requests
 }
 
+func (r *NUMAResourcesOperatorReconciler) getRelatedObjects(dsStatuses []nropv1.NamespacedName) []configv1.ObjectReference {
+	// 'Resource' should be in lowercase and plural
```
ffromani, line 596: could we have a quote (link to doc, source code) for this?

controllers/numaresourcesoperator_controller_test.go
```diff
@@ -468,7 +470,66 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
 						}
 						Expect(reconciler.Client.Get(context.TODO(), mcp2DSKey, ds)).ToNot(HaveOccurred())
 					})
-
+					When(" daemonsets are ready", func() {
+						BeforeEach(func() {
+							key := client.ObjectKeyFromObject(nro)
+
+							reconciler.RTEManifests.DaemonSet.Status.DesiredNumberScheduled = int32(len(nro.Spec.NodeGroups))
+							reconciler.RTEManifests.DaemonSet.Status.NumberReady = reconciler.RTEManifests.DaemonSet.Status.DesiredNumberScheduled
```
ffromani, line 478: this works (but is not great) but pollutes the global state - we should at least restore the initial values

controllers/numaresourcesscheduler_controller.go
```diff
@@ -153,6 +156,22 @@ func (r *NUMAResourcesSchedulerReconciler) reconcileResource(ctx context.Context
 
 }
 
+func (r *NUMAResourcesSchedulerReconciler) getRelatedObjects(dpStatus nropv1.NamespacedName) []configv1.ObjectReference {
+	// 'Resource' should be in lowercase and plural
```
ffromani, line 160: could we have a quote (link to doc, source code) for this?

ffromani (commented):
thanks for adding the tests! few comments inline

controllers/numaresourcesoperator_controller_test.go
jlojosnegros: yeah I am not a fan of this neither but I did not found another way to do it. 
if there is a better way please let me know and I will change it with pleasure.

About restore initial values, you're totally right ... on it.

controllers/numaresourcesscheduler_controller.go
jlojosnegros: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

> Most resource types require a name that can be used as a DNS subdomain name as defined in [RFC 1123](https://tools.ietf.org/html/rfc1123). This means the name must:
> 
> - contain no more than 253 characters
> - contain only lowercase alphanumeric characters, '-' or '.'
> - start with an alphanumeric character
> - end with an alphanumeric character

ffromani: sure, let's add this info to the comment. Thanks!

PR #659, opened 2023-07-21
serial: fix nil pointer in log

shajmakh:
In case NRTs don't exist that will return an error and the node name should be logged based on what's there in kconfig map, because nrt list will be empty, and addressing an item in an empty list will cause the program to panic.


ffromani (commented):
/lgtm
/approve

Issue #660, opened 2023-07-21
rebase images on top of ubi9


ffromani:
we should rebase our container images on top of ubi9

ffromani:
fixed by #661

PR #661, opened 2023-07-21
base image: ubi8 to ubi9

shajmakh:
For RHAOS 4.14 we should rebase on ubi9 for binary artifacts


ffromani:
I'm not sure the change is properly exercised because we discovered that failures in the internal lanes are silently skipped (#653)
Nevertheless, this change is correct and

fixes #660

PR #662, opened 2023-07-21
golang: move to 1.20

shajmakh:
RHAOS 4.14 is rhel9 and golang 1.20


ffromani:
/approve
/lgtm

I expect some lanes to fail because of the usual gotcha related to `Dockerfile.tools` - if not proably we have some obsolete bits in our build infra we should clean up later.

PR #665, opened 2023-07-24
doc: reformat README.md

ffromani:
reformat for better readability


swatisehgal (approved):
/lgtm

ffromani:
no longer necessary. The issue is the prow CI config was not moved to ubi9, so the binaries failed to start for linking error, so the subscription never completes, so the lane fails.

PR #668, opened 2023-07-24
e2e: build: use fully static build for serial

ffromani:
The serial suite is NOT meant to be distribute outside CI and NOT supported anyway, so we can use fully static build for this suite executable.


swatisehgal (approved):
/lgtm

ffromani:
requires: https://github.com/openshift/release/pull/41614

PR #669, opened 2023-07-25
e2e: rte: testing infra improvements

ffromani:
generic improvements and ground work for #653

### DIFF
--- a/test/e2e/rte/rte_test.go
+++ b/test/e2e/rte/rte_test.go
@@ -200,14 +200,8 @@ var _ = ginkgo.Describe("with a running cluster with all the components", func()
 				klog.Infof("RTE config: %#v", rc)
 				gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
-				// we intentionally don't check the values themselves - atm this would
-				// be to complex, effectively rewriting most of the controller logic
-				if len(kc.ReservedSystemCPUs) > 0 {
-					gomega.Expect(rc.Resources.ReservedCPUs).ToNot(gomega.BeEmpty())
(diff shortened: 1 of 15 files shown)


Tal-or:
I like the changes in general but how that helps to improve the infra? those changes are all part of the test suite

ffromani:
> I like the changes in general but how that helps to improve the infra? those changes are all part of the test suite

rephrased. I meant the testing libraries/utilities.

test/e2e/rte/rte_test.go
```diff
@@ -200,14 +200,8 @@ var _ = ginkgo.Describe("with a running cluster with all the components", func()
...
 
-				// we intentionally don't check the values themselves - atm this would
-				// be to complex, effectively rewriting most of the controller logic
-				if len(kc.ReservedSystemCPUs) > 0 {
-					gomega.Expect(rc.Resources.ReservedCPUs).ToNot(gomega.BeEmpty())
-				}
-				if len(kc.ReservedMemory) > 0 {
-					gomega.Expect(rc.Resources.ReservedMemory).ToNot(gomega.BeEmpty())
-				}
+				gomega.Expect(rc.TopologyManagerPolicy).To(gomega.Equal(kc.TopologyManagerPolicy), "TopologyManager Policy mismatch")
```
Tal-or, line 203: I would also add what we go was what expected.

ffromani: the annotation is adding context to the expected annotation, so the value got vs expected should already be there in the failed expectation error message (+ the extra annotations)

PR #674, opened 2023-07-27
sysinfo and podrescompat removal: superseded by getallocatableResources

swatisehgal:
`sysinfo` and `podrescompat` packages were introduced as a way to obtain resource information from the system and serve the endpoints provided by podresources in the absence of getAllotableResource endpoint in podresource API.

Support for getallocatableResource endpoint graduated to GA recently and it is safe to assume that this endpoint is enabled by default and provides all the necesary information for determining node resource allocatable.

We therefore no longer need the aforementioned packages and remove them.


ffromani (commented):
nice work, thanks. Some comments inline

ffromani:
I think we should also do a pass of `go mod tidy && go mod vendor` in the context of this PR (if you resubmit AND CI doesn't complain - it seems it doesn't)

rte/main.go
```diff
@@ -86,38 +82,18 @@ func main() {
...
-	klog.Infof(`
-=== System information ===
-%s
-==========================`, sysInfo)
-
-	if parsedArgs.SysinfoOnly {
-		os.Exit(0)
-	}
-
 	k8sCli, err := podres.GetClient(parsedArgs.RTE.PodResourcesSocketPath)
```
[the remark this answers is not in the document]
swatisehgal: Ack.

rte/main.go
```diff
@@ -86,30 +82,13 @@ func main() {
...
 		klog.Fatalf("failed to start prometheus server: %v", err)
 	}
 
-	sysCli := k8sCli
-	if !parsedArgs.LocalArgs.SysConf.IsEmpty() {
-		sysCli = podrescompat.NewSysinfoClientFromLister(k8sCli, parsedArgs.LocalArgs.SysConf)
-	}
-
-	cli := sharedcpuspool.NewFromLister(sysCli, parsedArgs.RTE.Debug, parsedArgs.RTE.ReferenceContainer)
+	sysCli := cli
```
ffromani, line 90: no need, we can just use
```
cli = sharedcpuspool.NewFromLister(sysCli, parsedArgs.RTE.Debug, parsedArgs.RTE.ReferenceContainer)
```

ffromani (approved):
LGTM

minor nit, please don't bother unless you need to resubmit for other reasons

rte/main.go
swatisehgal: Sure, can fix this as well.

PR #678, opened 2023-08-01
e2e: untie k8s e2e framework from rte test suite

Tal-or:
A continuation of the work done at https://github.com/k8stopologyawareschedwg/resource-topology-exporter
in order to remove the dependencies in the k8s e2e framework.

The framework client replaced with the local clients provided by the `clients` package.

PR #679, opened 2023-08-01
main: use resmonitor `Handle`

Tal-or:
Due to the recent change on RTE side we need to fix the function's parameters in main.


Tal-or:
prerequisite for #678

PR #680, opened 2023-08-02
e2e:rte: run local test separately

Tal-or:
The local rte tests are cluster-independent.

Although we're using ginkgo focus for running only the local tests,
the suite still tries to initialize the code in the other tests.

The test passed so far only becuase of the fact that
k8s e2e framework knows how to deal with the absence of kubeconfig file.

Therefore, we should have a sperate suite for the local tests
which doesn't import other tests that are depending on the
presence of a cluster.

### DIFF
--- a/test/e2e/rte/local/local_suite_test.go
+++ b/test/e2e/rte/local/local_suite_test.go
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2023 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package local
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestLocal(t *testing.T) {
+	RegisterFailHandler(Fail)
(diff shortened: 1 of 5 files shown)


test/e2e/rte/local/local_suite_test.go
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright 2021 Red Hat, Inc.
```
ffromani, line 2: is this a rename of existing code? If so it's ok, if not let's use 2023

test/e2e/rte/local/local_suite_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var (
+	randomSeed int64
+)
+
+func TestRTE(t *testing.T) {
```
ffromani, line 31: `TestRTELocal` or `TestLocal`?

test/e2e/rte/rte_e2e_suite_test.go
```diff
@@ -35,6 +35,7 @@ import (
 
 	_ "github.com/k8stopologyawareschedwg/resource-topology-exporter/test/e2e/rte"
 	_ "github.com/k8stopologyawareschedwg/resource-topology-exporter/test/e2e/topology_updater"
+	_ "github.com/openshift-kni/numaresources-operator/test/e2e/rte/local"
```
ffromani, line 38: Uhm, but if we want to isolate the local tests in their own suite (good), why are we importing it here again?

ffromani (commented):
looks generally OK with a question inside

test/e2e/rte/local/local_suite_test.go
Tal-or: Again copy paste mistake

test/e2e/rte/rte_e2e_suite_test.go
Tal-or: Because I thought there's nothing bad at running the tests again as part of the general suite

ffromani: Ok, I see no problem with that

ffromani (commented):
/lgtm

PR #682, opened 2023-08-10
e2e: fix flaky test 47591

shajmakh:
The test fails in CI and passes locally. The test modifies the pod resources under the deployment and do some checks on the new pod. The part that is causing the test to fail is that it doesn't wait "long enough" for the new pod to get created, causing the automation to pick an old pod, hence failing in later node name check.
 Investigating it deeper, even with same oc client the test kept passing locally. Yet managed eventually to reproduce it by running it using an ssh connection to the cluster. Not clear about the root cause at the moment, but it seems that when the automation is connected to the cluster actually matters in terms of how quick the check can be, as via exporting the KC the test had no chance of failing, but via ssh the test continuously failed, possible indicator that ssh connection has a faster reaction than exporting KC.

To fix the flakiness, we add a waiting loop to guarantee we do not move further until the new pod's replicaset gets created.

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -585,7 +585,7 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
 			nroSchedKey := objects.NROSchedObjectKey()
 			err = fxt.Client.Get(context.TODO(), nroSchedKey, nroSchedObj)
 			Expect(err).ToNot(HaveOccurred(), "cannot get %q in the cluster", nroSchedKey.String())
-			schedulerName := nroSchedObj.Spec.SchedulerName
+			schedulerName := nroSchedObj.Status.SchedulerName
 
 			nrtPreCreatePodList, err := e2enrt.GetUpdated(fxt.Client, initialNrtList, timeout)
 			Expect(err).ToNot(HaveOccurred())
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -510,6 +510,39 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
 			}
 			wg.Wait()
 
+			defer func() {
+				By("reverting kubeletconfig changes")
+				Eventually(func(g Gomega) {
+					err = fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(targetedKC), targetedKC)
+					Expect(err).ToNot(HaveOccurred())
+
+					kcObj, err = kubeletconfig.MCOKubeletConfToKubeletConf(targetedKC)
+					Expect(err).ToNot(HaveOccurred())
+
+					kcObj.TopologyManagerScope = initialTopologyManagerScope
+					err = kubeletconfig.KubeletConfToMCKubeletConf(kcObj, targetedKC)
+					Expect(err).ToNot(HaveOccurred())
+
+					err = fxt.Client.Update(context.TODO(), targetedKC)
+					g.Expect(err).ToNot(HaveOccurred())
+				}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(Succeed())
+
+				By("waiting for MachineConfigPools to get updated")
+				for _, mcp := range mcps {
+					wg.Add(1)
+					go func(mcpool *machineconfigv1.MachineConfigPool) {
+						defer GinkgoRecover()
+						defer wg.Done()
+						err = wait.With(fxt.Client).
+							Interval(configuration.MachineConfigPoolUpdateInterval).
+							Timeout(configuration.MachineConfigPoolUpdateTimeout).
+							ForMachineConfigPoolCondition(context.TODO(), mcpool, machineconfigv1.MachineConfigPoolUpdated)
+						Expect(err).ToNot(HaveOccurred())
+					}(mcp)
+				}
+				wg.Wait()
+			}()
+
 			By("checking that NUMAResourcesOperator's ConfigMap has changed")
 			cmList := &corev1.ConfigMapList{}
 			err = fxt.Client.List(context.TODO(), cmList)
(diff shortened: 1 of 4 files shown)


internal/replicasets/replicasets.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"context"
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func GetReplicasetByDeployment(cli client.Client, ctx context.Context, deployment appsv1.Deployment) ([]appsv1.ReplicaSet, error) {
```
ffromani, line 27: I'd merge this function into `internal/podlist` mimcing what the `DeploymentByOwnerReference` is doing (bar the necessary adjustments, so perhaps `ReplicaSetByDeployment`?)

admittedly `podlist` is not the best location, but a new `replicasets` package is not obviously good so I'd keep everything packed so we can fix all the mess in one go in the future

test/e2e/serial/tests/workload_placement.go
```diff
@@ -276,10 +280,21 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+			// wait for the replicasets under the deployment are increased by 1
+			namespacedDpName := fmt.Sprintf("%s/%s", updatedDp.Namespace, updatedDp.Name)
+			Eventually(func() int {
+				r, err := replicasets.GetReplicasetByDeployment(fxt.Client, context.TODO(), *updatedDp)
+				Expect(err).ToNot(HaveOccurred())
+				klog.Infof("number of replicasets under deployment %q is %d", namespacedDpName, len(r))
+				return len(r)
+			}).WithTimeout(10*time.Second).WithPolling(1*time.Second).Should(Equal(len(dpReplicas)+1), "deployment %q replicasets were not increased by 1", namespacedDpName)
+
+			dpReplicas, err = replicasets.GetReplicasetByDeployment(fxt.Client, context.TODO(), *updatedDp)
```
ffromani, line 292: Can we simplify this check just checking the related Deployment is ready? If not fine, but please document why.

test/e2e/serial/tests/workload_placement.go
```diff
@@ -371,6 +386,13 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 				return fxt.Client.Update(context.TODO(), updatedDp)
 			}).WithTimeout(2 * time.Minute).WithPolling(10 * time.Second).ShouldNot(HaveOccurred())
 
+			Eventually(func() int {
+				r, err := replicasets.GetReplicasetByDeployment(fxt.Client, context.TODO(), *updatedDp)
+				Expect(err).ToNot(HaveOccurred())
+				klog.Infof("number of replicasets under deployment %q is %d", namespacedDpName, len(r))
+				return len(r)
+			}).WithTimeout(10*time.Second).WithPolling(1*time.Second).Should(Equal(len(dpReplicas)+1), "deployment %q replicasets were not increased by 1", namespacedDpName)
```
ffromani, line 394: likewise, I'm missing what this check adds that the previous check right below for the deployment to complete is neglecting. It seems we missed to check a key detail, but which one?

ffromani (commented):
thanks for your fix! I just would like to understand more the context

internal/replicasets/replicasets.go
shajmakh: I was thinking similarly but I didn't see ReplicaSetByDeployment fits under podList just because of the naming of this folder. It conveys that the functionality presented there lists pod under other controllers. so I created a new folder for replicaset (although there is a high chance we'll not gonna use it for other methods in the near future as we concentrate on pods mainly).
Anyhow, no big deal of moving it under podList, so will do so.

test/e2e/serial/tests/workload_placement.go
shajmakh: once I have a cluster available I'll add some debugging log like replica number under the deployment to validate that the test failed because missing the new replica

internal/replicasets/replicasets.go
shajmakh: let me know what do you think if we need to rename podList to something else perhaps findby

ffromani: I concur `podlist` is not the best fit for this function and considering only this change moving the new code here is actually a pessimization. At project level there's no clear direction here so I prefer to keep all the related code packed instead of experimenting with independent solutions. I'd rather refactor everything in one go

test/e2e/serial/tests/workload_placement.go
ffromani: I don't mind adding extra utility functions (if their use is warranted), but in the test code I'd rather spend more time understanding what's going on. Either we are checking the wrong condition, or there's more going on and we're are kinda sidestepping the root cause.

ffromani (commented):
thanks for the updates, I'll need to have another pass later

test/e2e/serial/tests/workload_placement.go
shajmakh: so after adding more logs to examine the replicas under the deployment, my guess is justified because we see this:

```
I0821 09:03:37.771974  247702 deployment.go:42] deployment e2e-test-workload-placement-6409/testdp47591 complete
STEP: replicasets number is 2 should be 3 08/21/23 09:03:37.782
```

and the debugging code was added after verifying the dp was complete. 
this means we check the dp is complete "too early" -> earlier than having the new replicaset created.

ffromani: thanks, this indeed proves you found something which is worth a fix. I think we should double down and understand why the existing check is not good enough, and fix that. If this proves to be impossible, or too hard, then we should add and additional check. But in general I'd expect the existing check to cover this case. It doesn't and this means the existing code is buggy, or incomplete.

shajmakh: ForDeploymentComplete only checks if the pod under the deployment is ready. In the context of this test, we change the resources of the container resulting in creating new replicaset -> recreating pod. 
I tried verifying that the DP is indeed updated (fetching the current dp and comparing it with the new expected spec) but that didn't help. 
From what I see manually, and using the automation, ForDeploymentComplete will always report success in this context because the deployment is not new, but updated. Simplest example to justify that is verifying the case where the new pod is pending, and it turns that in this case the deployment will still be considered as complete because the old pod is still there and running.

```
[root@registry ~]# oc get pod,deployment,replicaset
NAME                               READY   STATUS    RESTARTS   AGE
pod/padder-2588                    1/1     Running   0          59s
pod/padder-6902                    1/1     Running   0          59s
pod/testdp47591-586c9f5494-cngvq   0/2     Pending   0          12s
pod/testdp47591-5b9d4d78fd-qxspf   2/2     Running   0          2m16s
NAME                          READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/testdp47591   1/1     1            1           3m15s
NAME                                     DESIRED   CURRENT   READY   AGE
replicaset.apps/testdp47591-586c9f5494   1         1         0       12s
replicaset.apps/testdp47591-5b9d4d78fd   1         1         1       2m16s
replicaset.apps/testdp47591-64894fcd94   0         0         0       3m15s
```

Because of this, the deployment complete check is not enough for this test and we need to check what happens in between the checks boiling down to verify the replicaset number is actually increased. From there we guarantee either a success or a clear failure that will be caught in later stages (e.g having 2 pods ,one in pending/failing state). Thus I believe we need to have this intermediate check. I have added clarification regarding how deployment readiness is decided so it will be taken into consideration once used.

shajmakh: The ForDeploymentComplete function is used in many places, but as of now, there is no obvious flakiness caused in other tests, and if there will be I believe it best be handled by adding additional checks (before/after the function call) as the context of the test requires.

ffromani: ok, thanks for the extra investigation. This justifies the extra code specific to this test.
At the same time, I can't help but notice there's a disconnect in what we expect (and how we use) deployments and the way they actually behave. We will need to handle this somehow in the future.

test/e2e/serial/tests/configuration.go
```diff
@@ -510,6 +510,39 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
 			}
 			wg.Wait()
 
+			defer func() {
```
ffromani, line 513: you are doing a good change, but I think we can move this further above on line 496, don't we?

ffromani (commented):
/lgtm
/approve

a minor(ish) comment is not IMO worth further delays.

test/e2e/serial/tests/workload_placement.go
shajmakh: I agree there is room for improvement in this regards. I have checked other uses of this function, and I can truly tell that if there wasn't the wait-for-NRT-settle and if the test had a conflicting expectation to the old deployment configuration, then there would be a good chance the test may fail due to this same disconnection.

test/e2e/serial/tests/configuration.go
shajmakh: thought about this, I think there is catch 22 here: what if the mcp got stuck in updating, don't we want to see the actual current configuration of the kubeletconfig when debugging? OTOH what if MCP never got updated -> the test will fail and defer will not be called -> kubeletconfig may reflect misleading data. And after all, only if mcp got updated the new configuration of the object is actually applied..

shajmakh: what do you think less evil here?

ffromani: I would prefer to explicitly log the config we should see and make all efforts to revert. But this is already an improvement over the old code, so not blocking.

PR #685, opened 2023-08-17
rte manifests: update sample-devices image

shajmakh:
Consume the enhanced sample device plugin image that allows reregistration of the devices in case of kubelet restart.

P.S: there is no failure noted before this change but seems more right to consume the enhanced version.

### DIFF
--- a/rte/hack/manifests/sample-devices/daemonset-dpa.yaml
+++ b/rte/hack/manifests/sample-devices/daemonset-dpa.yaml
@@ -17,11 +17,16 @@ spec:
         - name: dpa-container
           securityContext:
             privileged: true
-          image: quay.io/k8stopologyawareschedwg/sample-device-plugin:v0.1.1
+          image: quay.io/k8stopologyawareschedwg/sample-device-plugin:v0.2.2
           imagePullPolicy: IfNotPresent
           env:
             - name: DEVICE_RESOURCE_NAME
               value: "example.com/deviceA"
+          command:
+            - /bin/deviceplugin
+          args:
+            - --alsologtostderr
+            - --config-dir=/etc/devices
           volumeMounts:
             - name: kubeletsockets
               mountPath: /var/lib/kubelet/device-plugins


ffromani:
let's not retest: let's actually check the setup is correct with the new sample device plugin. Perhaps some other updates are due.

shajmakh:
> let's not retest: let's actually check the setup is correct with the new sample device plugin. Perhaps some other updates are due.

d/s change is also failing, I'll check the reason later today

ffromani:
we need to update the daemonset manifests (like https://github.com/openshift-kni/numaresources-operator/blob/main/rte/hack/manifests/sample-devices/daemonset-dpa.yaml)
adding
```
        command:
          - /bin/deviceplugin
        args:
          - --alsologtostderr
          - --config-dir=/etc/devices
```
into the container template so it reads like
```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: device-plugin-a-ds
  namespace: sample-dev-ns
spec:
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      name: device-plugin-a
  template:
    metadata:
      labels:
        name: device-plugin-a
    spec:
      containers:
      - args:
        - --alsologtostderr
        - --config-dir=/etc/devices
        command:
        - /bin/deviceplugin
        env:
        - name: DEVICE_RESOURCE_NAME
          value: example.com/deviceA
        image: XXX
        imagePullPolicy: IfNotPresent
        name: device-plugin-a-container
        securityContext:
          privileged: true
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
        volumeMounts:
        - mountPath: /var/lib/kubelet/device-plugins
          name: kubeletsockets
        - mountPath: /etc/devices
          name: device-a-config-vol
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      serviceAccount: sampledevices-sa
      serviceAccountName: sampledevices-sa
      terminationGracePeriodSeconds: 30
      volumes:
      - hostPath:
          path: /var/lib/kubelet/device-plugins
          type: Directory
        name: kubeletsockets
      - configMap:
          defaultMode: 420
          name: device-a-configmap
        name: device-a-config-vol
  updateStrategy:
    rollingUpdate:
      maxSurge: 0
      maxUnavailable: 1
    type: RollingUpdate

```

ffromani (commented):
/lgtm

PR #686, opened 2023-09-12
modified render to add must-gather tests

SargunNarula:
This PR contains the modifications required to enable running the must gather tests in the NROP pipeline.

### DIFF
--- a/Dockerfile.tests
+++ b/Dockerfile.tests
@@ -10,6 +10,6 @@ FROM registry.access.redhat.com/ubi9/ubi-minimal
 WORKDIR /usr/local/bin
 COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/e2e-nrop*.test .
 COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/run-e2e-nrop*.sh .
-COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/numacell .
-COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/pause .
+COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/numacell /bin/numacell
+COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/pause /pause
 CMD ["./run-e2e-nrop-serial.sh"]
--- a/Dockerfile.tests
+++ b/Dockerfile.tests
@@ -7,8 +7,9 @@ RUN make build-e2e-all
 RUN make build-numacell
 
 FROM registry.access.redhat.com/ubi9/ubi-minimal
-COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/e2e-nrop*.test /usr/local/bin
-COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/run-e2e-nrop-serial.sh /usr/local/bin
-COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/numacell /bin
-COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/pause /
-ENTRYPOINT ["/usr/local/bin/run-e2e-nrop-serial.sh"]
+WORKDIR /usr/local/bin
+COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/e2e-nrop*.test .
+COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/run-e2e-nrop*.sh .
+COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/numacell .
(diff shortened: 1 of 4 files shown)


ffromani:
I understand this is still WIP, but still I'm more inclined towards adding different specific entrypoints rather than a generic entrypoint which can be configured. We can discuss this approach and weight pros and cons

ffromani:
if ready to review, please un-draft.

question: how do you folks plan to consume the tests image after this change?

Dockerfile.tests
```diff
@@ -11,4 +11,4 @@ COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/
 COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/run-e2e-nrop-serial.sh /usr/local/bin
 COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/numacell /bin
 COPY --from=builder /go/src/github.com/openshift-kni/numaresources-operator/bin/pause /
-ENTRYPOINT ["/usr/local/bin/run-e2e-nrop-serial.sh"]
+ENTRYPOINT ["./entrypoint.sh"]
```
ffromani, line 15: this should be turned into `CMD`, to read
```
CMD ["/usr/local/bin/run-e2e-nrop-serial.sh"]
```

this is not strictly needed, but it help in the context of this change to clarify the intent, and make the flow more extensible.
Then, rather than add this wrapper, just create your script the way you do below, but just add it into `/usr/local/bin`

Next, the container image will still have the existing run-e2e-nrop-serial.sh as entry point, but you will be able to override it and use the must-gather script you added.

You will need to pass parameters anyway, but this way is IMO cleaner and surely simpler: less wrappers, less code around:
```
podman run -ti quay.io/openshift-kni/numaresources-tests /usr/local/bin/run-e2e-must-gather.sh
```

hack/render-e2e-runner.sh
```diff
@@ -5,16 +5,29 @@ set -eu
...
-MARKER=$(grep -Fn '### serial suite runner' hack/run-test-serial-e2e.sh | cut -d: -f1)
+if [[ "$1" == 'must-gather' ]]; then
+    MARKER=$(grep -Fn '### must-gather suite runner' hack/run-test-must-gather-e2e.sh | cut -d: -f1)
+    TEST_FILE='hack/run-test-must-gather-e2e.sh'
+    FILE='run-e2e-nrop-must-gather.sh'
+    
+elif [[ $1 == 'serial' ]]; then
+    MARKER=$(grep -Fn '### serial suite runner' hack/run-test-serial-e2e.sh | cut -d: -f1)
+    TEST_FILE='hack/run-test-serial-e2e.sh'
+    FILE='run-e2e-nrop-serial.sh'
```
shajmakh, line 16: this is good, and I wonder if we can untie this for only MG or serial tests. I agree we need to have a default script, and serial it is, but what if we want to execute other tests later on that are containerized in the image, but because of this limitation we can't actually reach and run them freely.

hack/run-test-must-gather-e2e.sh
```diff
@@ -1,15 +1,176 @@
...
+done
+
+if [ -z "${REPORT_FILE}" ]; then
+	REPORT_FILE="${REPORT_DIR}/e2e-must-gather-run.xml"
+fi
+
+function setup() {
+	if [[ "${SETUP}" != "true" ]]; then
+		return 0
+	fi
```
shajmakh, line 102: I understand this is how we've been doing it for serial suite, but why do we need it? don't we limit our selves here? what if we see the need for other new ginkgo flags? 
If we want to execute other commands that aren't supported by this, we'll be banned
https://github.com/openshift-kni/numaresources-operator/pull/686/files#diff-63231dd0167658c2ecd90e27f927a8312a61ac0849cec95b8fe72b1263a06360R88
do we really need it that strict? 
plus I see the tests binaries are copied under the image:
https://github.com/openshift-kni/numaresources-operator/pull/686/files#diff-1d0fed26e7daaae4aa24a5ed9bab7c25d82cac3a51199b90cf9c73e01da4b5b8R10 so why not enable using them directly on new container?

shajmakh (commented):
Thanks for the proposed changes, I left some comments but they are not blocking.

hack/run-test-must-gather-e2e.sh
shajmakh: just to clarify my point, what I think - and might be mistaken- is that if we had another entrypoint (not command) like bin/sh and we do not limit the command to be within the ginkgo flags specified above, then I think we wouldn't be doing or needing this change now since the tests binaries are in fact shipped into the image (an till now we use only the serial suite .test).

shajmakh:
/lgtm
/hold
leaving more review time and the final say to @ffromani

ffromani (commented):
thanks, this looks promising. But the paths doesn't look fine yet.

ffromani (commented):
/approve
/lgtm

PR #688, opened 2023-09-15
pkg: images: reuse client

ffromani:
no need to create a specific k8s client to learn the image details, we can safely reuse the manager builtin client.

### DIFF
--- a/main.go
+++ b/main.go
@@ -196,7 +196,7 @@ func main() {
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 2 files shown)


ffromani:
CI failure may be legit, need to investigate

ffromani:
/retest

we had a bunch of infra-related flakes lately

ffromani:
let's rethink the client management entirely

Issue #495, opened 2022-12-16
add spell-checker helper, and possibly CI test


ffromani:
add script helper which checks the spelling in the source code, like other projects (k8s comes to mind) do.
Evaluate if this check should be added to CI blocking lanes.

jlojosnegros:
Would https://github.com/openshift/cluster-node-tuning-operator/pull/680 be a good way to solve this issue? 
It would add spellcheck among other checks.

ffromani:
yes, that could be a good way to address this issue

danielmellado:
@jlojosnegros happy to have a chat with you about that if you'd like to.

PR #689, opened 2023-09-15
Add golangci-lint

jlojosnegros:
This PR adds a new golangci-lint target to the makefile and sets up a common config along the lines of other openshift operators.

- Address misspell violations
- Address staticheck violations

Signed-off-by: Daniel Mellado [dmellado@redhat.com](mailto:dmellado@redhat.com)
Signed-off-by: Jose Luis Ojosnegros [jojosneg@redhat.com](mailto:jojosneg@redhat.com)

issue: #495

### DIFF
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -1,0 +1,26 @@
+name: CI Linter with golangci-lint
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v3
+
+    - name: Set up golang
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.20.6
+
+    - name: Run Linter
+      run: make golangci-lint
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,36 @@
+run:
+  concurrency: 4
+  timeout: 5m
+  tests: true
+  skip-dirs:
+    - vendor
+    - test
+  modules-download-mode: vendor
+linters:
+  disable-all: true
+  enable:
+    - errcheck
+    - bodyclose
+    - exportloopref
+    - gosimple
+    - govet
+    - ineffassign
+    - misspell
+    - nolintlint
+    - nosprintfhostport
+    - staticcheck
+    - tenv
+    - typecheck
+    - unconvert
+    - unused
+    - wastedassign
+    - whitespace
+linters-settings:
+  misspell:
+    locale: US
+    ignore-words:
+      - NRO
+      - nro
+      - numaresource
+      - numa
+      - NUMA
--- a/Makefile
+++ b/Makefile
@@ -80,6 +80,13 @@ SHELL = /usr/bin/env bash -o pipefail
 
 KUSTOMIZE_DEPLOY_DIR ?= config/default
 
+# golangci-lint variables
+GOLANGCI_LINT_VERSION=1.54.2
+GOLANGCI_LINT_NAME=golangci-lint-$(GOLANGCI_LINT_VERSION)-$(GOOS)-$(GOARCH)
+GOLANGCI_LINT_ARTIFACT_FILE=$(GOLANGCI_LINT_NAME).tar.gz
+GOLANGCI_LINT_EXEC_NAME=golangci-lint
+GOLANGCI_LINT=$(BIN_DIR)/$(GOLANGCI_LINT_EXEC_NAME)
+
 all: build
 
 ##@ General
@@ -386,3 +393,21 @@ verify-generated: bundle generate
 
 install-git-hooks:
 	git config core.hooksPath .githooks	
+
+
+GOLANGCI_LINT_LOCAL_VERSION := $(shell command ${GOLANGCI_LINT} --version 2> /dev/null | awk '{print $$4}')
+.PHONY: golangci-lint
+golangci-lint:
+	@if [ ! -x "$(GOLANGCI_LINT)" ]; then\
+		echo "Downloading golangci-lint from https://github.com/golangci/golangci-lint/releases/download/v$(GOLANGCI_LINT_VERSION)/$(GOLANGCI_LINT_ARTIFACT_FILE)";\
+		mkdir -p $(BIN_DIR);\
+		curl -JL https://github.com/golangci/golangci-lint/releases/download/v$(GOLANGCI_LINT_VERSION)/$(GOLANGCI_LINT_ARTIFACT_FILE) -o $(BIN_DIR)/$(GOLANGCI_LINT_ARTIFACT_FILE);\
+		pushd $(BIN_DIR);\
+		tar --no-same-owner -xzf $(GOLANGCI_LINT_ARTIFACT_FILE)  --strip-components=1 $(GOLANGCI_LINT_NAME)/$(GOLANGCI_LINT_EXEC_NAME);\
+		chmod +x $(GOLANGCI_LINT_EXEC_NAME);\
+		rm -f $(GOLANGCI_LINT_ARTIFACT_FILE);\
+		popd;\
+	else\
+		echo "Using golangci-lint cached at $(GOLANGCI_LINT), current version $(GOLANGCI_LINT_LOCAL_VERSION) expected version: $(GOLANGCI_LINT_VERSION)";\
+	fi
+	$(GOLANGCI_LINT) run --verbose --print-resources-usage -c .golangci.yaml
--- a/api/numaresourcesoperator/v1/numaresourcesoperator_types.go
+++ b/api/numaresourcesoperator/v1/numaresourcesoperator_types.go
@@ -94,7 +94,7 @@ type NodeGroup struct {
 	// MachineConfigPoolSelector defines label selector for the machine config pool
 	// +optional
 	MachineConfigPoolSelector *metav1.LabelSelector `json:"machineConfigPoolSelector,omitempty"`
-	// Config defines the RTE behaviour for this NodeGroup
+	// Config defines the RTE behavior for this NodeGroup
 	// +optional
 	Config *NodeGroupConfig `json:"config,omitempty"`
 }
--- a/cmd/nrtcacheck/main.go
+++ b/cmd/nrtcacheck/main.go
@@ -66,7 +66,7 @@ func init() {
 type ProgArgs struct {
 	Version   bool
 	DumpNodes bool
-	NodeNames sets.String
+	NodeNames sets.Set[string]
 }
 
 func main() {
@@ -116,7 +116,7 @@ func main() {
 
 	var nodeNames []string
 	if parsedArgs.NodeNames.Len() > 0 {
-		nodeNames = parsedArgs.NodeNames.List()
+		nodeNames = sets.List(parsedArgs.NodeNames)
 		klog.V(2).Infof("using provided node list with %d items", len(nodeNames))
 	} else {
 		workers, err := nodes.GetWorkerNodes(cli, ctx)
@@ -174,19 +174,19 @@ func main() {
 			continue
 		}
 
-		podsByRTE := sets.String{}
+		podsByRTE := sets.New[string]()
 		for _, nn := range st.Pods {
 			podsByRTE.Insert(nn.String())
 		}
 
 		fmt.Printf("%s: pods on sched, not on RTE: [\n", nodeName)
-		for _, name := range podsBySched.Difference(podsByRTE).List() {
+		for _, name := range sets.List(podsBySched.Difference(podsByRTE)) {
 			fmt.Printf(" - %s\n", name)
 		}
 		fmt.Printf("]\n")
 
 		fmt.Printf("%s: pods on RTE, not on sched: [\n", nodeName)
-		for _, name := range podsByRTE.Difference(podsBySched).List() {
+		for _, name := range sets.List(podsByRTE.Difference(podsBySched)) {
 			fmt.Printf(" - %s\n", name)
 		}
 		fmt.Printf("]\n")
@@ -241,7 +241,7 @@ func parseArgs(args ...string) (ProgArgs, error) {
 
 	klog.InitFlags(flags)
 	err := flags.Parse(args)
-	pArgs.NodeNames = sets.NewString(flags.Args()...)
+	pArgs.NodeNames = sets.New[string](flags.Args()...)
 	return pArgs, err
 }
 
--- a/controllers/numaresourcesoperator_controller_test.go
+++ b/controllers/numaresourcesoperator_controller_test.go
@@ -18,7 +18,6 @@ package controllers
 
 import (
 	"context"
-	"encoding/json"
 	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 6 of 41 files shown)


Makefile
```diff
@@ -386,3 +390,22 @@ verify-generated: bundle generate
...
+GOLANGCI_LINT := $(shell command -v golangci-lint 2> /dev/null)
+GOLANGCI_LINT_LOCAL_VERSION := $(shell command golangci-lint --version | awk '{print $$4}')
+golangci-lint:
+ifdef GOLANGCI_LINT
+	@echo "Found golangci-lint"
+ifeq ($(GOLANGCI_LINT_LOCAL_VERSION),$(GOLANGCI_LINT_VERSION))
+		@echo golangcilint using same version
+else
+		@echo "Mismatch version, installing golangci-lint $(GOLANGCI_LINT_VERSION)"
+		curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(BIN_DIR) $(GOLANGCI_LINT_VERSION)
```
ffromani, line 403: downloading binaries is ok-ish, but `curl | sh` is not unless 1. it's a critical piece AND 2. there are no other practical ways. Can we just curl the binary?
In addition, I'd download a tag/fixed release, not vs master

Makefile
```diff
@@ -386,3 +390,22 @@ verify-generated: bundle generate
 
 install-git-hooks:
 	git config core.hooksPath .githooks	
+
+GOLANGCI_LINT := $(shell command -v golangci-lint 2> /dev/null)
+GOLANGCI_LINT_LOCAL_VERSION := $(shell command golangci-lint --version | awk '{print $$4}')
+golangci-lint:
+ifdef GOLANGCI_LINT
```
ffromani, line 397: can we have the check like we do for `operator-sdk` and `opm` ? I'd also omit the version check/postpone to later

Makefile
jlojosnegros: Using `install.sh` is the recommended way to install golangci-lint https://golangci-lint.run/usage/install/
`install.sh` script just checks OS, platform and arch and download the proper "tar.gz" with the binary.

It is not downloading the `master` but the version specified in `GOLANGCI_LINT_VERSION`. 
It is using the `master` version of the `install.sh` script but that can be change if we end using it and it is an issue

ffromani: Right about only fetching `install.sh` from master (even though it's still a moving target, is a bit less evil).
In out case we know exactly the OS, platform and arch, so I'd still unconvinced a `curl | shell` is worth the challenges.

Makefile
jlojosnegros: done

jlojosnegros:
> failures not relevant for this PR. Rather than adding a prow lane/job, I'd use golangci-lint in a github action, at least for starters.

I thought about that, but using it this way make it easier to launch it locally and solve problems before submitting which could improve development cycle.

Makefile
```diff
@@ -386,3 +395,23 @@ verify-generated: bundle generate
 
 install-git-hooks:
 	git config core.hooksPath .githooks	
+
+
+GOLANGCI_LINT_LOCAL_VERSION := $(shell command ${GOLANGCI_LINT} --version 2> /dev/null | awk '{print $$4}')
+.PHONY: golangci-lint
+golangci-lint:
```
Tal-or, line 402: Can we add this target as prerequisite  for the `build` target?

Tal-or:
> > failures not relevant for this PR. Rather than adding a prow lane/job, I'd use golangci-lint in a github action, at least for starters.
> 
> I thought about that, but using it this way make it easier to launch it locally and solve problems before submitting which could improve development cycle.

I don't think the two proposal are contradict. we can ran it manually, but as well having a github action to run on it against any PR

jlojosnegros:
> > > failures not relevant for this PR. Rather than adding a prow lane/job, I'd use golangci-lint in a github action, at least for starters.
> > 
> > 
> > I thought about that, but using it this way make it easier to launch it locally and solve problems before submitting which could improve development cycle.
> 
> I don't think the two proposal are contradict. we can ran it manually, but as well having a github action to run on it against any PR

To run it against any PR I was thinking on doing as for NTO, create a new prow job to run them using the same makefile target. 
Something like https://github.com/openshift/release/pull/43550

Makefile
jlojosnegros: Tie both targets in Makefile will not allow us to run a build without running this check and I think we could wanna do that on some environments.

The idea is to run them automatically using a prow job once this PR is merge, but keep that job as `optional` (non-blocking for merge) in the beginning until we feel confident enough to change it.

If running this in prow is "too late" and the problem is to avoid forget running this before pushing changes, I think a git hook could do the tick.

WDYT?

Tal-or: > If running this in prow is "too late" and the problem is to avoid forget running this before pushing changes, I think a git hook could do the tick.

yes that could work indeed. Thanks.

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,26 @@
...
+  pull_request:
+    branches: [ main ]
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
```
Tal-or, line 15: nit: let's try to run with `ubuntu-22.04` which is the most updated host version provided by GH actions nowdays

internal/resourcelist/resourcelist_test.go
```diff
@@ -168,6 +168,7 @@ func TestSubCoreResources(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(ToString(tc.expected), func(t *testing.T) {
 			res := tc.res.DeepCopy()
+			//nolint:errcheck
```
swatisehgal, line 171: We are we choosing to make the linter happy by doing this rather than returning error and ensuring that the expected error is nil?

controllers/numaresourcesoperator_controller_test.go
```diff
@@ -933,6 +933,7 @@ func reconcileObjects(nro *nropv1.NUMAResourcesOperator, mcp *machineconfigv1.Ma
 	return reconciler
 }
 
+//nolint:unused
 func nodeGroupConfigToString(conf nropv1.NodeGroupConfig) string {
```
swatisehgal, line 937: Why are we choosing to keep this rather than removing it? I would incline towards removing unused functions. We can always add them if needed in the future.

cmd/nrtcacheck/main.go
```diff
@@ -116,7 +116,7 @@ func main() {
 
 	var nodeNames []string
 	if parsedArgs.NodeNames.Len() > 0 {
-		nodeNames = parsedArgs.NodeNames.List()
+		nodeNames = parsedArgs.NodeNames.UnsortedList()
```
swatisehgal, line 119: Use of `sets.List(parsedArgs.NodeNames)` seems more appropriate here. It has been documented here: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#List. Is there any particular reason we are going with [UnsortedList](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Byte.UnsortedList) instead of List method provided in the sets package?

Also applies to similar updates below.

controllers/numaresourcesoperator_controller_test.go
jlojosnegros: you're right

internal/resourcelist/resourcelist_test.go
jlojosnegros: you're right 
thanks for the comment

cmd/nrtcacheck/main.go
jlojosnegros: Just thought order was not important in those cases so I used the UnsortedList function but your option seems better. 
Thanks for the comment.

.github/workflows/golangci-lint.yaml
jlojosnegros: thanks for the comment. 
changed.

cmd/nrtcacheck/main.go
ffromani: From the algorhythm perspective the order is not really relevant, actually I introduced room for bugs using `List` (vs `UnsortedList`. In this case it's better to preserve the old (and questionable) semantic than to sneak the change (I'd see fix) here.

.golangci.yaml
```diff
@@ -0,0 +1,38 @@
...
+  misspell:
+    locale: US
+    ignore-words:
+      - NRO
+      - nro
+      - numaresource
+      - numa
+      - NUMA
+      - behaviour
+      - flavour
```
ffromani, line 38: this is questionable, we seem to prefer the american spelling (vs the english one) but it can be changed later

Makefile
ffromani: I don't mind this approach but I'd start with an informing (non-blocking) CI lane and "enforce" the fixes during the review.

Makefile
```diff
@@ -386,3 +395,23 @@ verify-generated: bundle generate
...
+	@if [ ! -x "$(GOLANGCI_LINT)" ]; then\
+		echo "Downloading golangci-lint $(GOLANGCI_LINT_VERSION)";\
+		echo "Downloading golangci-lint from https://github.com/golangci/golangci-lint/releases/download/$(GOLANGCI_LINT_VERSION_TAG)/$(GOLANGCI_LINT_ARTIFACT_FILE)";\
+		mkdir -p $(BIN_DIR);\
+		curl -JL https://github.com/golangci/golangci-lint/releases/download/$(GOLANGCI_LINT_VERSION_TAG)/$(GOLANGCI_LINT_ARTIFACT_FILE) -o $(BIN_DIR)/$(GOLANGCI_LINT_ARTIFACT_FILE);\
+		pushd $(BIN_DIR);\
+		echo "we are in `pwd`";\
```
ffromani, line 409: please remove this

Makefile
```diff
@@ -386,3 +395,23 @@ verify-generated: bundle generate
 
 install-git-hooks:
 	git config core.hooksPath .githooks	
+
+
+GOLANGCI_LINT_LOCAL_VERSION := $(shell command ${GOLANGCI_LINT} --version 2> /dev/null | awk '{print $$4}')
+.PHONY: golangci-lint
+golangci-lint:
+	@if [ ! -x "$(GOLANGCI_LINT)" ]; then\
+		echo "Downloading golangci-lint $(GOLANGCI_LINT_VERSION)";\
```
ffromani, line 404: this seems to be redundant given the line right below

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -85,7 +85,7 @@ type NodeGroupConfig struct {
 	// InfoRefreshPeriod sets the topology info refresh period. Use explicit 0 to disable.
 	// +optional
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Topology info refresh period setting",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
-	InfoRefreshPeriod *metav1.Duration `json:"infoRefreshPeriod,omitEmpty"`
+	InfoRefreshPeriod *metav1.Duration `json:"infoRefreshPeriod,omitempty"`
```
ffromani, line 88: this is a fix which should be backport ASAP, should be extracted from this PR, merged and backported ASAP

main.go
```diff
@@ -341,6 +341,7 @@ func renderRTEManifests(rteManifests rtemanifests.Manifests, namespace string, i
 		return mf, err
 	}
 	_ = rteupdate.DaemonSetUserImageSettings(mf.DaemonSet, "", imageSpec, images.NullPolicy)
+	//nolint:errcheck
 	rteupdate.DaemonSetPauseContainerSettings(mf.DaemonSet)
```
ffromani, line 345: we should actually bubble up the error here (both calls)

internal/resourcelist/resourcelist_test.go
```diff
@@ -168,7 +168,9 @@ func TestSubCoreResources(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(ToString(tc.expected), func(t *testing.T) {
 			res := tc.res.DeepCopy()
-			SubCoreResources(res, tc.resToSub)
+			if err := SubCoreResources(res, tc.resToSub); err != nil {
+				t.Errorf("Error while calculating resources: %s", err.Error())
+			}
```
ffromani, line 173: we prefer to NOT inline the call:
```
err := SubCoreResources(res, tc.resToSub)
if err != nil {
    ...
```

internal/remoteexec/command.go
```diff
@@ -62,7 +63,7 @@ func CommandOnPodByNames(c kubernetes.Interface, podNamespace, podName, cntName
 		return nil, nil, err
 	}
 
-	err = exec.Stream(remotecommand.StreamOptions{
+	err = exec.StreamWithContext(context.Background(), remotecommand.StreamOptions{
```
ffromani, line 66: this should be exposed at function level (add context at first argument) and the context should be instantiated at top-level rather than created here

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - name: Check out code
+      uses: actions/checkout@v3
+
+    - name: Set up golang
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.20.6
+
+    - name: Run Linter
+      run: make golangci-lint
```
ffromani, line 26: let's start like this, but I'm eyeing the official action for lesser resource usage, saving us CI time.

ffromani (commented):
thanks, some comments inline and we should be ready

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
jlojosnegros: PR created https://github.com/openshift-kni/numaresources-operator/pull/696

Will rebase over main as soon as merged.

Makefile
jlojosnegros: thanks for the comment.
Done

Makefile
jlojosnegros: deleted.
Thanks for the comment

main.go
jlojosnegros: done.

.golangci.yaml
jlojosnegros: deleted and changed the typos.

jlojosnegros:
https://github.com/openshift-kni/numaresources-operator/pull/689#discussion_r1342403101

I was pretending to do that with a prow ci-lane but with the gh action I think I do not have enough privileges to mark the action as "non-required"

Makefile
jlojosnegros: I was pretending to do that with a prow ci-lane but with the gh action I think I do not have enough privileges to mark the action as "non-required"

ffromani: In our case prow considers all GH actions lane as non-binding. Admittedly this is incidental configuration, but good enough for now.

ffromani:
please rebase on top of the current main branch

main.go
```diff
@@ -341,7 +341,11 @@ func renderRTEManifests(rteManifests rtemanifests.Manifests, namespace string, i
 		return mf, err
 	}
 	_ = rteupdate.DaemonSetUserImageSettings(mf.DaemonSet, "", imageSpec, images.NullPolicy)
```
ffromani, line 343: this one could have fixed as well, but we can clean up in a follow PR

ffromani (approved):
/approve
/lgtm

thanks for the updates and all your work!

jlojosnegros:
> /approve /lgtm
> 
> thanks for the updates and all your work!

thank you for your comments :)

PR #694, opened 2023-09-26
Relax condition on reboot and a cleanup

shajmakh:
The cooldown is performed at the end of the test and its main goal is to validate that the NRT is reset to the initial value.
On reboot scenarios, this check always fails because it turns out that the memory distribution across the same numa zones of the same node is not similar after the reboot, but is always found the same on the node level.After investigation, it turns out that new capacities after reboot are reported as so by the kernel. Since this is a behavior we cannot control, we need to adapt it, and thus here we suggest relaxing the equality check of memory resources. We allow a deviation of 52Mi which should cover these differences.

### DIFF
--- a/internal/noderesourcetopology/equality.go
+++ b/internal/noderesourcetopology/equality.go
@@ -18,11 +18,15 @@ package noderesourcetopology
 
 import (
 	"fmt"
+	"strings"
 
 	nrtv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 )
 
-func EqualZones(zonesA, zonesB nrtv1alpha2.ZoneList) (bool, error) {
+func EqualZones(zonesA, zonesB nrtv1alpha2.ZoneList, isRebootTest bool) (bool, error) {
 	if len(zonesA) != len(zonesB) {
 		return false, fmt.Errorf("unequal zone count")
 	}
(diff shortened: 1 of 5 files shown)


internal/noderesourcetopology/equality.go
```diff
@@ -80,3 +88,35 @@ func EqualResourceInfo(resInfoA, resInfoB nrtv1alpha2.ResourceInfo) (bool, error
...
+
+func EqualResourceInfoWithDeviation(resInfoA, resInfoB nrtv1alpha2.ResourceInfo) (bool, error) {
+	/*
+		tests that involve reboot show a difference in the memory capacity on numa nodes, e.g memory capacity
+		 is reallocated across numa nodes upon startup. This is a a kernel behavior which we have no control
+		 on thus we want to work it around. The tests failed when comapring initial (pre-reboot) NRT with the
+		 final (post reboot) NRT, and finds out a difference of 53760000 of memory (=52500Ki ~=52Mi).
+		 For those case, we allow a deviation of 52Mi when comparing the memory topologies.
+	*/
+	dev, _ := resource.ParseQuantity("54525952") //52 Mi
```
swatisehgal, line 100: How did we come up with this value? The commit message and the PR says that we allow a deviation of 1Mi but we seem to be allowing a deviation of roughly 52Mi here.

shajmakh: good catch, I need to update the commit message. After testing it turned out that 1Mi is not enough to cover the difference, and usually the difference we get is 53760000 which equal to ~51 Mi units of memory, I rounded that diff up and made it 52Mi. I'll fix the PR and the commit msg. sorry for the confusion.

shajmakh: fixed, thanks

swatisehgal (approved):
/lgtm
/approve

shajmakh:
> /lgtm /approve

shajmakh:
> /lgtm /approve

The bot didn't seem to recognize the lgtm label for some unclear reason, but only the approval

shajmakh:
(I closed the PR by mistake so I reopened it and that triggered the tests again, although all tests were passing before the close)

swatisehgal (commented):
/lgtm

ffromani:
/cherry-pick release-4.14

backporting e2e test improvements is generally fine

PR #695, opened 2023-09-27
serial: fix flaky test

shajmakh:
TBD


test/e2e/serial/tests/workload_placement.go
```diff
@@ -476,6 +476,17 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+				This test reveals a race between the 2 schedulers, since we do not wait for the pod handled by TAS
+				scehduler to get created, we cannot guarantee that it will pick up the resources before the pod that
+				is handled by the default scheduler. because of this the test is fragile and fails frequently.
+				to overcome this we have the following options:
+				1. omit the test as it adds no value as it is now and only introduces a noise
+				2. wait for the pods to get created before checking and mark the test as non-reg
+				3. change the check to verify that 1 out of the 2 pods is running, with a warning of the status of the
+				 second pod (due to the race, we could hit TAE)
+				2. update the test goal to compare another aspect of both schedulers: scheduling time (already covered by another test); node pickup and in case not the same node keep that as a warning - (not much value in this eaither);
+			*/
```
ffromani, line 489: we should drop the test. Keeping it will worse the signal, being too flaky with no hope to make it robust, and changing it delivers litte to no value.
let's just keep the comment explaining we HAD such a test (including test summary), we removed it and why

shajmakh: thanks for your feedback, true indeed. I'll do the necessary.

ffromani (commented):
/approve
/lgtm

PR #696, opened 2023-10-02
json options should be lower case

jlojosnegros:
Typo found in some json options


ffromani:
/approve
/lgtm

thanks!

PR #700, opened 2023-10-05
controller: support suppressing NRT updates

shajmakh:
The main update in this PR is the addition of a new flag under the numaresourcesoperator controller to control
publishing RTE updates into the NRTs. When this mode is enabled,
--no-publish will be passed to the RTE ds indicating that the NRT
updates should be paused unless disabled in the numaresop object for the
corresponding mcp.
When the updates are paused or NRTs are not found; the
numaresourcesscheduler will not be functional and all workloads that are
using this scheduler will keep pending until the updates are re-enabled.

This is followed with some couple of tests (e2e & unit tests) fixes. see commits for more details.


shajmakh:
note for myself:
the commit message of this commit https://github.com/openshift-kni/numaresources-operator/pull/700/commits/17ef3cbea04fc2fbbbbe538f20382dfb928c5a77 is weird although was phrased differently (!). The commit is for running "make generate bundle manifests"

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -42,6 +42,12 @@ type NUMAResourcesOperatorSpec struct {
 	// +optional
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Optional ignore pod namespace/name glob patterns"
 	PodExcludes []NamespacedName `json:"podExcludes,omitempty"`
+	// Valid values are: "enabled", "disabled".
+	// Defaults to "enabled".
+	// +optional
+	// +kubebuilder:default=enabled
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="RTE Sync" //UPDATE ME IN CRD
+	RTESync string `json:"rteSync,omitempty"`
```
ffromani: this should be per-nodeGroup, not global

controllers/numaresourcesoperator_controller.go
```diff
@@ -222,6 +222,10 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
 		}
 	}
 
+	if instance.Spec.RTESync == "disabled" {
```
ffromani: this needs a log messages, perhaps at V >= 4

api/numaresourcesoperator/v1/numaresourcesoperator_defaults.go
```diff
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
```
ffromani, line 23: I prefer to change all the instances in one go (to be done in a follow up PR) rather than add just for this change

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -92,6 +86,12 @@ type NodeGroupConfig struct {
 	// +optional
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Topology info refresh period setting",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
 	InfoRefreshPeriod *metav1.Duration `json:"infoRefreshPeriod,omitempty"`
+	// Valid values are: true, false.
+	// Defaults to "false".
+	// +optional
+	// +kubebuilder:default=false
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="RTE Pause"
+	RTEPause *bool `json:"rtepause,omitempty"`
```
ffromani, line 94: you moved it into the right place, but I tend to prefer not using bools. Will add the k8s API design suggestions when I find it again

controllers/numaresourcesoperator_controller.go
```diff
@@ -222,7 +222,7 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
 		}
 	}
 
-	if !instance.Spec.RTESync {
+	if !*instance.Spec.NodeGroups[0].Config.RTEPause {
```
ffromani: make sure to: 1. find the right `NodeGroup` and 2. handle the `nil` value (API layer should guarantee non-nil)

api/numaresourcesoperator/v1/numaresourcesoperator_defaults.go
shajmakh: not sure I understand, what instances are you referring to? 
in another comment I see you do not like boolean so will make that a string with 2 options: enable & disable. that way we won't need this lib

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
shajmakh: will change it for the meanwhile to string (e.g enabled/disabled). just to understand better, what's not so good in bool? more complicated to understand compared to enabled/disabled maybe? bound us from adding a third option without large maintainance? or is there another reason?

controllers/numaresourcesoperator_controller.go
shajmakh: very good point. thanks for highlighting, will handle it

controllers/numaresourcesoperator_controller.go
shajmakh: done, thanks

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
ffromani: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types - all the document is a great read which I keep always handy

ffromani: also https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md

api/numaresourcesoperator/v1/numaresourcesoperator_defaults.go
ffromani: there are other places in the codebase which can benefit to move to `k8s.io/utils/pointer`. I actually welcome this change. But it should be done all in once: the only thing I'm not fine with is having both the old clunky way I started (reinventing `k8s.io/utils/pointer` locally) and using `k8s.io/utils/pointer`.  Either you submit a PR to convert to `k8s.io/utils/pointer` everywhere and get that PR merged before to move on with this one (fine) or you use the old way in this new PR and convert to `k8s.io/utils/pointer` with a later PR

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
ffromani: note that using bool is not per se automatically wrong. But is not a no-brainer either. We should just think about using them (vs. enums) and evaluate the consequences vs just slapping a bool here.

api/numaresourcesoperator/v1/numaresourcesoperator_defaults_test.go
```diff
@@ -98,3 +114,8 @@ func ptrToDuration(d time.Duration) *metav1.Duration {
 	}
 	return &v
 }
+
+func ptrToRTEMode(m RTEPauseMode) *RTEPauseMode {
```
ffromani, line 118: you can just `return &m`

api/numaresourcesoperator/v1/numaresourcesoperator_defaults.go
```diff
@@ -56,3 +59,8 @@ func defaultInfoRefreshPeriod() *metav1.Duration {
 	}
 	return &period
 }
+
+func defaultRTEPause() *RTEPauseMode {
+	rtePause := RTEPauseDisabled
```
ffromani, line 64: `RTEPauseDisabled` is a constant, here's the reason we use an intermediate local variable

config/crd/bases/nodetopology.openshift.io_numaresourcesoperators.yaml
```diff
@@ -300,6 +305,11 @@ spec:
                           - Enabled
                           - EnabledExclusiveResources
                           type: string
+                        rtepause:
+                          default: false
+                          description: 'Valid values are: true, false. Defaults to
+                            "false".'
+                          type: boolean
```
ffromani, line 312: this will need an update (not urgent)

config/manifests/bases/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -51,6 +51,9 @@ spec:
         path: nodeGroups[0].config.podsFingerprinting
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:text
+      - description: 'Valid values are: true, false. Defaults to "false".'
+        displayName: RTE Pause
+        path: nodeGroups[0].config.rtepause
```
ffromani, line 56: ditto

controllers/numaresourcesoperator_controller.go
```diff
@@ -222,6 +222,11 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
 		}
 	}
 
+	// TODO modify the trees to include only rtepause disabled (nil/empty/disabled)
+	// log "RTEsync is disabled for nodegroups ..."
+	// check length of the new trees, if >0 continue with the sync, otherwise return:
+	//		return ctrl.Result{}, status.ConditionAvailable, nil
+
```
ffromani, line 229: the code here is a bit more convoluted than I like it to be. You may likely want to look about how we create per-nodeGroup daemonsets when we iterate over `existing.State` around line 410

ffromani: *MAYBE* just change the `daemonsetUpdater` function

shajmakh: thanks for your review and comments, I haven't gone deeply into this yet, I'll surely verify you suggestion

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
shajmakh: thanks for the references, I'll look into them

api/numaresourcesoperator/v1/numaresourcesoperator_defaults.go
shajmakh: okay makes sense, thanks for clarifying. I will handle this in a later PR.

shajmakh:
still not final nor ready for review: pending more unit tests; ds args alignment with the new flag (if needed here at all?);

pkg/objectupdate/rte/rte.go
```diff
@@ -136,7 +136,12 @@ func DaemonSetArgs(ds *appsv1.DaemonSet, conf nropv1.NodeGroupConfig) error {
 		return fmt.Errorf("cannot modify the arguments for container %s", cnt.Name)
 	}
 
-	flags.SetToggle("--refresh-node-resources")
+	rtePauseEnabled := isRTEPauseEnabled(&conf)
+	klog.V(2).InfoS("DaemonSet update: RTEPause status", "daemonset", ds.Name, "enabled", rtePauseEnabled)
+	if !rtePauseEnabled {
+		klog.V(4).InfoS("DaemonSet update: enable node resources refresh")
+		flags.SetToggle("--refresh-node-resources")
```
ffromani, line 143: this flag doesn't what you probably think: it asks the RTE to update the capacity (not just the current allocation): https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/main/pkg/resourcemonitor/resourcemonitor.go#L404

it would be ok to add a new flag to RTE, but we will need a RTE PR

rte/main.go
```diff
@@ -83,6 +83,13 @@ func main() {
 		os.Exit(0)
 	}
 
+	if !parsedArgs.Resourcemonitor.RefreshNodeResources {
```
ffromani, line 86: this is an interesting idea and avoid us a RTE PR. But we still need a new option (can be specific of rte/main.go, we have already some custom options), because repurposing existing option is not fine.

ffromani: or just use `--no-publish`?

controllers/numaresourcesoperator_controller_test.go
```diff
@@ -787,6 +791,41 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
...
+			reconciler := reconcileObjects(nro, mcp)
+
+			key := client.ObjectKeyFromObject(nro)
+
+			nroCurrent := &nropv1.NUMAResourcesOperator{}
+			Expect(reconciler.Client.Get(context.TODO(), key, nroCurrent)).NotTo(HaveOccurred())
+			Expect(nroCurrent.Spec.NodeGroups[0].Config.RTEPause).To(Equal(conf.RTEPause))
+
+			//we do not check the ds here because we don't want to update ds and cause a restart of the pods,
+			// we only pause the pods sync with NRTs
```
ffromani, line 808: we should check that the DS spec command line does NOT include the `--no-publish` flag

ffromani: we should have tests which check that if when the setting is given with pause disabled no changes are done

controllers/numaresourcesoperator_controller_test.go
```diff
@@ -787,6 +791,41 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
...
+			reconciler := reconcileObjects(nro, mcp)
+
+			key := client.ObjectKeyFromObject(nro)
+
+			nroCurrent := &nropv1.NUMAResourcesOperator{}
+			Expect(reconciler.Client.Get(context.TODO(), key, nroCurrent)).NotTo(HaveOccurred())
+			Expect(nroCurrent.Spec.NodeGroups[0].Config.RTEPause).To(Equal(&rteMode))
+
+			// we do not check the ds here because we don't want to update ds and cause a restart of the pods,
+			// we only pause the pods sync with NRTs
```
ffromani, line 827: in this case we DO WANT to update the DSs and cause a pod restart. And the commandline should include `--no-publish`

and we should have a third test which checks that --no-publish is added when pause is enabled AND THEN removed when pause is disabled

pkg/objectupdate/rte/rte.go
```diff
@@ -136,6 +136,13 @@ func DaemonSetArgs(ds *appsv1.DaemonSet, conf nropv1.NodeGroupConfig) error {
 		return fmt.Errorf("cannot modify the arguments for container %s", cnt.Name)
 	}
 
+	rtePauseEnabled := isRTEPauseEnabled(&conf)
+	klog.V(2).InfoS("DaemonSet update: RTEPause status", "daemonset", ds.Name, "enabled", rtePauseEnabled)
+	if rtePauseEnabled {
+		klog.V(4).InfoS("DaemonSet update: disable publishing NRTs updates")
```
ffromani, line 142: this should be rare enough that it deserves to be logged with `V(2)` level

pkg/objectupdate/rte/rte.go
```diff
@@ -136,6 +136,13 @@ func DaemonSetArgs(ds *appsv1.DaemonSet, conf nropv1.NodeGroupConfig) error {
...
+		flags.SetToggle("--no-publish")
+	}
```
ffromani, line 144: if `rtePauseEnabled` is false, we need to remove the toggle if present. Otherwise, when added, it will be always kept here

ffromani (commented):
need to figure out a better name than `RTEPause`. Other than that the logic looks fine. We need more tests though.

pkg/objectupdate/rte/rte.go
ffromani: and let's put the log outside the if (for which we will need an `else` anyway, see below)

shajmakh:
status: pending validating e2e tests + renaming the new flag.

pkg/objectupdate/rte/rte_test.go
```diff
@@ -179,6 +179,16 @@ func TestUpdateDaemonSetArgs(t *testing.T) {
...
 
 func expectCommandLine(t *testing.T, ds, origDs *appsv1.DaemonSet, testName string, expectedArgs []string) {
+	uncontrolledArgs := []string{
+		//	"--notify-file=/run/rte/notify",
+		"--podresources-socket=unix:///host-var/lib/kubelet/pod-resources/kubelet.sock",
+		"--sysfs=/host-sys",
+		"--topology-manager-policy=restricted",
+		"--topology-manager-scope=pod",
+	}
+	if len(expectedArgs) != len(ds.Spec.Template.Spec.Containers[0].Args)-len(uncontrolledArgs) {
```
ffromani, line 189: this is  good idea, but then let's go full throttle.
1. we can rename 'uncontrollerdArgs' to `commonArgs`, or any better name which conveys the fact this is a subset of arguments we want to be there
2. turn the two slices into sets, adding `commonArgs` to any set to ensure the two sets contains them. The sets semantics ensure we don't have duplicates
3. turn the two sets into **sorted** lists
4. compare the two resulting sorted lists. If they aren't equal, fail the test

PS: please remove the commented out code :)

api/numaresourcesoperator/v1/numaresourcesoperator_defaults.go
```diff
@@ -38,6 +38,9 @@ func (ngc *NodeGroupConfig) Default() {
 	if ngc.InfoRefreshMode == nil {
 		ngc.InfoRefreshMode = defaultInfoRefreshMode()
 	}
+	if ngc.RTEPause == nil {
```
ffromani, line 41: (here and everywhere) I'm still contemplating if we should rename `RTEPause` to `InfoRefreshPause` or not. What do you think?

pkg/objectupdate/rte/rte.go
```diff
@@ -136,6 +136,14 @@ func DaemonSetArgs(ds *appsv1.DaemonSet, conf nropv1.NodeGroupConfig) error {
 		return fmt.Errorf("cannot modify the arguments for container %s", cnt.Name)
 	}
 
+	rtePauseEnabled := isRTEPauseEnabled(&conf)
+	klog.V(2).InfoS("DaemonSet update: RTEPause status", "daemonset", ds.Name, "enabled", rtePauseEnabled)
+	if rtePauseEnabled {
+		flags.SetToggle("--no-publish")
+	} else if _, ok := flags.GetFlag("--no-publish"); ok {
```
ffromani, line 143: `flagcodec` should allow you to remove the flag directly, without check+delete. Please check

test/e2e/serial/tests/non_regression_fundamentals.go
```diff
@@ -19,16 +19,20 @@ package tests
 import (
 	"context"
 	"fmt"
+	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
```
ffromani, line 22: minor: please add import to the proper group (sdlib imports, k8s imports, 3rd party imports, local imports...)

test/e2e/serial/tests/non_regression_fundamentals.go
```diff
@@ -63,91 +67,157 @@ var _ = Describe("[serial][fundamentals][scheduler][nonreg] numaresources fundam
...
+			dsKey := wait.ObjectKey{
+				Namespace: nropObjInitial.Status.DaemonSets[0].Namespace,
+				Name:      nropObjInitial.Status.DaemonSets[0].Name,
+			}
+
+			Eventually(func() []string {
+				dsObj := appsv1.DaemonSet{}
+				err = fxt.Client.Get(context.TODO(), client.ObjectKey(dsKey), &dsObj)
+				Expect(err).ToNot(HaveOccurred())
+				cnt := k8swgobjupdate.FindContainerByName(dsObj.Spec.Template.Spec.Containers, "resource-topology-exporter")
```
ffromani, line 112: we should have a constant for the RTE name, please check

pkg/objectupdate/rte/rte.go
```diff
@@ -148,7 +148,9 @@ func DaemonSetArgs(ds *appsv1.DaemonSet, conf nropv1.NodeGroupConfig) error {
 
 	notifEnabled := isNotifyFileEnabled(&conf)
 	klog.V(2).InfoS("DaemonSet update: event notification", "daemonset", ds.Name, "enabled", notifEnabled)
-	if !notifEnabled {
+	if notifEnabled {
```
ffromani, line 151: this is just bug prevention or did you find an actual bug here?

ffromani (commented):
partial review; still need to have a proper look at tests and to document the expected scheduler behavior. But this PR is shaping up real nice nevertheless.

api/numaresourcesoperator/v1/numaresourcesoperator_defaults.go
shajmakh: yep, agree. I had several options in mind ( nrtDataPublish/ nrtUpdatingMode/ rteFunctionalityMode/ NRTInfoSuppress) but InfoRefreshPause sounds better as it goes along with the rest of the other fields under the ngc.

pkg/objectupdate/rte/rte.go
shajmakh: indeed, it already checks that internally. 
https://github.com/k8stopologyawareschedwg/deployer/blob/main/pkg/flagcodec/flagcodec.go#L94C3-L94C3
thanks for pointing this out.

pkg/objectupdate/rte/rte.go
shajmakh: The source of the RTE pods spec is derived from the ds args which is in turn derived from the the NROP.  Unless I missed it, the option was not set in this place. It is set by default but in case of any modification and there is no way back to enable it.

shajmakh: sure, I can handle that

ffromani: thanks a lot. Let's start checking we have controller tests to exercise this flow. e2e tests here looks like overkill. If we have test coverage and we do the right thing, then it turns into a code readability issue, which is important but less urgent.

shajmakh: sure, I'll  revisit and let you know

pkg/objectupdate/rte/rte_test.go
shajmakh: thanks for your detailed comment. This is indeed a good idea if duplicates are not important. If sets are used and there happen to be a duplicate the conversation should fail not succeed and continue. one example is if an arg is duplicated then something is fishy and could turn out that in some place an arg was set while the intention is to set some other arg. in this case we want to fail.
However I do like adopting appending the slices. I've updated the code let me know what do you think. As for sorting slices I don't think that is helping a lot we anyway do a deepEqual and want enough debugging logs (like what arg was missing) in case this failed.

ffromani:
```
 hack/verify-generated.sh
uncommitted generated files. run 'make generate bundle manifests' and commit results.
```

pkg/objectupdate/rte/rte_test.go
```diff
@@ -179,14 +179,8 @@ func TestUpdateDaemonSetArgs(t *testing.T) {
...
 func expectCommandLine(t *testing.T, ds, origDs *appsv1.DaemonSet, testName string, expectedArgs []string) {
-	uncontrolledArgs := []string{
-		//	"--notify-file=/run/rte/notify",
-		"--podresources-socket=unix:///host-var/lib/kubelet/pod-resources/kubelet.sock",
-		"--sysfs=/host-sys",
-		"--topology-manager-policy=restricted",
-		"--topology-manager-scope=pod",
-	}
-	if len(expectedArgs) != len(ds.Spec.Template.Spec.Containers[0].Args)-len(uncontrolledArgs) {
+	expectedArgs=append(expectedArgs,commonArgs...)
```
ffromani: I need to check if this idiom can leak the appended args in the original slice. I think in the common case it will not, but in some unlikely but possible corner cases it can.
the safest (and clunkiest) way is to clone `expectedArgs`:
```
expectedArgs = append([]string{}, expecredArgs...)
expectedArgs = append(expectedArgs, commonArgs...)
```

(can be done without variable shadowing in the first line).

Likely not worth resubmitting for this, though.

ffromani:
I think this looks good for 4.15 once CI is green and changes are squashed as planned. I'll have a final pass then but looks good to go.

pkg/objectupdate/rte/rte_test.go
shajmakh: interesting, I'll validate. thanks for rasing that

pkg/objectupdate/rte/rte_test.go
shajmakh: sounds like a good easy check to exclude duplicates, I'll add it. thanks for pointing that out.

ffromani:
If you split the last two commits into a new PR, we can merge them right away (and likely backport at least to 4.15)

test/e2e/serial/tests/non_regression_fundamentals.go
```diff
@@ -25,12 +25,11 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	nrtv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
```
ffromani, line 28: nit: misplaced import: should go around line 33 below

test/e2e/serial/tests/operator_profile_update.go
```diff
@@ -0,0 +1,282 @@
+/*
+ * Copyright 2022 Red Hat, Inc.
```
ffromani, line 2: 2022?

test/e2e/serial/tests/operator_profile_update.go
```diff
@@ -0,0 +1,282 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package tests
+
+import (
+	"context"
+	"fmt"
+	"k8s.io/klog/v2"
```
ffromani, line 22: nit: mixed import, should go around line 33 below

test/e2e/serial/tests/operator_profile_update.go
```diff
@@ -0,0 +1,282 @@
...
+			Expect(err).ToNot(HaveOccurred(), "cannot get %q from the cluster", nroKey.String())
+
+			updatedNROPObj := nropObjInitial.DeepCopy()
+			infoRefreshPauseMode := nropv1.InfoRefreshPauseEnabled
+			//TODO adapt the existence of multiple node groups,https://issues.redhat.com/browse/CNF-10552
+			updatedNROPObj.Spec.NodeGroups[0].Config = &nropv1.NodeGroupConfig{
+				InfoRefreshPause: &infoRefreshPauseMode,
+			}
+
+			err = fxt.Client.Update(context.TODO(), updatedNROPObj)
```
ffromani, line 94: updates should always be done in a loop (like `Eventually`). Updating the NROP objects is unlikely to fail, but still

test/e2e/serial/tests/operator_profile_update.go
```diff
@@ -0,0 +1,282 @@
...
+			}
+			Eventually(func() []string {
+				dsObj := appsv1.DaemonSet{}
+				err = fxt.Client.Get(context.TODO(), client.ObjectKey(dsKey), &dsObj)
+				Expect(err).ToNot(HaveOccurred())
+				cnt := k8swgobjupdate.FindContainerByName(dsObj.Spec.Template.Spec.Containers, manifests.ContainerNameRTE)
+				Expect(cnt).NotTo(BeNil(), "cannot find container data for %q", manifests.ContainerNameRTE)
+				return cnt.Args
+
+			}).WithTimeout(time.Minute).WithPolling(9*time.Second).Should(ContainElement(ContainSubstring("--no-publish")), "ds was not updated as expected: \"--no-pubish\" arg is missing.")
```
ffromani, line 122: this is misplaced. Another test (perhaps controller test) must test that if we set the pause field in the API object THEN the `--no-publish` command line argument pops in the DS command line args. We should not retest this here, but depend on the fact this specific bit works.

test/e2e/serial/tests/operator_profile_update.go
```diff
@@ -0,0 +1,282 @@
...
+			klog.Info("waiting for DaemonSet to be ready")
+			_, err = wait.With(e2eclient.Client).Interval(10*time.Second).Timeout(3*time.Minute).ForDaemonSetReadyByKey(context.TODO(), dsKey)
+			Expect(err).ToNot(HaveOccurred(), "failed to get the daemonset %s: %v", dsKey.String(), err)
+		})
+
+		AfterEach(func() {
+			restoreNROP(fxt, *nropObjInitial)
```
ffromani, line 129: if a function is called only in a single place, please inline it. Let's consider extracting functions only when we have at least two call sites.

test/e2e/serial/tests/operator_profile_update.go
```diff
@@ -0,0 +1,282 @@
...
+			By(fmt.Sprintf("verify pod events contain the correct message %q", eventMsgInvalidTopology))
+			ok, err := verifyPodEvents(fxt, testPod, eventReasonFailedScheduling, eventMsgInvalidTopology)
+			Expect(err).ToNot(HaveOccurred())
+			if !ok {
+				_ = objects.LogEventsForPod(fxt.K8sClient, testPod.Namespace, testPod.Name)
+			}
+			Expect(ok).To(BeTrue(), "failed to find the expected event with Reason=%q and Message contains %q", eventReasonFailedScheduling, eventMsgInvalidTopology)
+
+			By("restore NROP configuration and verify the pod goes running")
+			restoreNROP(fxt, *nropObjInitial)
```
ffromani, line 171: so we can try to restore twice?

test/e2e/serial/tests/operator_profile_update.go
```diff
@@ -0,0 +1,282 @@
...
+			if !ok {
+				_ = objects.LogEventsForPod(fxt.K8sClient, testPod.Namespace, testPod.Name)
+			}
+			Expect(ok).To(BeTrue(), "failed to find the expected event with Reason=%q and Message contains %q", eventReasonFailedScheduling, eventMsgInvalidTopology)
+
+			By("verify NRTs has no changes")
+			//
+
+			By("restore NROP configuration and verify the pod goes running")
+			//restoreNROP(fxt, *nropObjInitial)
```
ffromani, line 209: why commented out?

test/e2e/serial/tests/operator_profile_update.go
```diff
@@ -0,0 +1,282 @@
...
+	err := fxt.Client.Get(context.TODO(), nroKey, currentObject)
+	Expect(err).ToNot(HaveOccurred())
+	//TODO adapt the existence of multiple node groups,https://issues.redhat.com/browse/CNF-10552
+	if reflect.DeepEqual(currentObject.Status.MachineConfigPools[0].Config, initialObj.Status.MachineConfigPools[0].Config) {
+		return
+	}
+
+	newNROP := currentObject.DeepCopy()
+	newNROP.Spec = initialObj.Spec
+	err = fxt.Client.Update(context.TODO(), newNROP)
```
ffromani, line 236: ditto about updating in a loop

test/e2e/serial/tests/operator_profile_update.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	Eventually(func() []string {
+		dsObj := appsv1.DaemonSet{}
+		err := fxt.Client.Get(context.TODO(), client.ObjectKey(dsKey), &dsObj)
+		Expect(err).ToNot(HaveOccurred())
+		cnt := k8swgobjupdate.FindContainerByName(dsObj.Spec.Template.Spec.Containers, manifests.ContainerNameRTE)
+		Expect(cnt).NotTo(BeNil(), "cannot find container data for %q", manifests.ContainerNameRTE)
+		return cnt.Args
+	}).WithTimeout(time.Minute).WithPolling(9*time.Second).ShouldNot(ContainElement(ContainSubstring("--no-publish")), "ds was not updated as expected: \"--no-pubish\" arg is missing.")
```
ffromani, line 265: ditto about (not) checking command line args

test/e2e/serial/tests/operator_profile_update.go
```diff
@@ -0,0 +1,282 @@
+/*
```
ffromani, line 1: let's think a bit more about the file name. The purpose of the tests is to check when we don't have any NRT data. The fact we leverate the newly added option and we update a profile is an implementation detail, not so important it leaks on the file name

shajmakh:
> If you split the last two commits into a new PR, we can merge them right away (and likely backport at least to 4.15)

thank you for the review and new comments. I completely agree, except that I suspect there is a problem in the way the logs are reported and I'm still digging to understand further if it is a bug or a feature. I was testing this and this is the reason you see some code parts are commented out so this shouldn't be the final shape of this PR yet

test/e2e/serial/tests/operator_profile_update.go
shajmakh: agree. this will be used in later multi-staged tests. This is why I extracted it to a method.

test/e2e/serial/tests/operator_profile_update.go
shajmakh: right, but we do check if a restore is needed before actually applying the initial config (again).

test/e2e/serial/tests/operator_profile_update.go
shajmakh: for testing and to make it faster to debug :) 
will fix it eventually.

test/e2e/serial/tests/operator_profile_update.go
shajmakh: I actualy recall now why I added this, it is because if the test proceed to the next step of checking if ds is ready, and finds I as so, it is likely that it still has the old config. so there were 2 ways: 1. wait for ds pods to get recreated=wait for ds to become not ready Or 2. wait for the new config to be reflected and only then wait for the ds to get ready. 
I chose to go with the second option because if the check failed this options gives us the bottom of the problem which is whether new config is reflected or not.

shajmakh:
Update: The tests hit a bug in the CR status when running one after the other but pass w/o issues when running one at a time (!!). The bug doesn't seem to be in the tests but rather in the controller somewhere (further debugging is needed) because it reproduces manually. Yet, there are no clear steps for reproduction. I need to finalize the last commits/open new prereq PRs as fit, thus I don't recommend you review the patch with the current shape.

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -173,3 +175,11 @@ type NUMAResourcesOperatorList struct {
 func init() {
 	SchemeBuilder.Register(&NUMAResourcesOperator{}, &NUMAResourcesOperatorList{})
 }
+
+func (ngc *NodeGroupConfig) ToString() string {
```
ffromani, line 179: `ToString` is the variant we (mostly me) invented when we want a string serialization but we don't feel ready, or we prefer to, not implement the [Stringer interface](https://go.dev/tour/methods/17)

So in general let's be conscious if our stringification functions are ready or not to be promoted to full Stringer interface. In this case, do you want or prefer to do so?

api/numaresourcesoperator/v1/numaresourcesoperator_defaults.go
```diff
@@ -24,11 +24,11 @@ import (
 
 func DefaultNodeGroupConfig() NodeGroupConfig {
 	ngc := NodeGroupConfig{}
-	ngc.Default()
+	ngc.FillNilFeilds()
```
ffromani, line 27: Typo: should be Fields.

But arguably we ARE filling empty resource objects with default values, aren't we? after all, "nil" is the default zero value for pointers, and we are setting it to default values - pointing to constants with default values.

In general, this field is similar to String vs ToString: wants to be like the default initialization functions into kube apis, but we're not there yet or there are small/subtle differences we want to keep, hence we use a API as similar as possible but not the real API.

pkg/loglevel/loglevel.go
```diff
@@ -61,7 +61,7 @@ func UpdatePodSpec(podSpec *corev1.PodSpec, cntName string, level operatorv1.Log
 		return fmt.Errorf("cannot modify the arguments for container %s", cnt.Name)
 	}
 	flags.SetOption("--v", kLog.String())
-	klog.InfoS("container klog level", "container", cnt.Name, "-v", kLog.String())
+	klog.InfoS("container klog level", "container", cnt.Name, "--v", kLog.String())
```
ffromani, line 64: I prefer the other way around: single-letter identifiers should use a single dash
```
-v
--long-option
```

api/numaresourcesoperator/v1/numaresourcesoperator_defaults.go
shajmakh: yes we can consider the nil as empty thus filling it with default. but the reason I changed this is because Default() sounds like it returns the object with default values. This is different from filling Empty/nil fields with the default because if the object was partially initialized, the non-empty fields will not be set to default. 
in this specific line it doesn't make any difference because the object is empty (was just created)

ffromani: Fair enough, so let's use `SetDefaults` which is closer (identical?) the real API does

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
shajmakh: Thanks for this.
I saw the need for this mainly because https://github.com/openshift-kni/numaresources-operator/blob/main/controllers/numaresourcesoperator_controller.go#L359 doesn't print readable value (address).
 I don't see any harm in making this the official stringer implementation for NodeGroupConfig as long we agree on the content presentation:
1. completely nil/empty struct is printed as empty  string ""
2. otherwise, nil fields are filled with the default and all are printed out.
The only concern would be maintenance, e.g adjusting the method to reflect all fields, but this should be part of every update done on the config struct.wdyt?

pkg/loglevel/loglevel.go
shajmakh: addressing this here: https://github.com/openshift-kni/numaresources-operator/pull/848

shajmakh:
Note: the tests after the rebase on top of the scheduler's latest updates are behaving in a way that the pods will go running, the only change is that for gu pod the updates are suppressed (as wanted by this change) and reflected only upon restoring the setting.

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
ffromani: I think maintaining is fine. We can add tests to enforce this, perhaps later:
1. add test(s) which uses String() with non-defaults values
2. add test(s) which do a round trip: initialize object, call String() on it, pass to a function which create an object from its string representation, check it is equal to the initial object (roundtrip test)

pkg/objectupdate/rte/rte_test.go
```diff
@@ -195,3 +206,15 @@ func sliceIndex(sl []string, s string) int {
 	}
 	return -1
 }
+func getSetFromStringList(args []string) map[string]string {
```
ffromani, line 209: nit: this is not a set, it's more a map. But not worth resubmitting because of this

internal/wait/daemonset.go
```diff
@@ -36,11 +36,12 @@ func (wt Waiter) ForDaemonSetReadyByKey(ctx context.Context, key ObjectKey) (*ap
 		}
 
 		if !AreDaemonSetPodsReady(&updatedDs.Status) {
-			klog.Warningf("daemonset %s desired %d scheduled %d ready %d",
+			klog.Warningf("daemonset %s desired %d scheduled %d ready %d up-to-date %d",
```
ffromani, line 39: this is an important fix and should probably be moved in its own PR. But let's keep it like this for now

internal/wait/daemonset.go
```diff
@@ -81,3 +81,30 @@ func (wt Waiter) ForDaemonsetPodsCreation(ctx context.Context, ds *appsv1.Daemon
 	})
 	return updatedDs, err
 }
+
+func (wt Waiter) ForDaemonSetUpdateByKey(ctx context.Context, key ObjectKey) (*appsv1.DaemonSet, error) {
+	updatedDs := &appsv1.DaemonSet{}
+
+	immediate := true
+	err := k8swait.PollUntilContextTimeout(ctx, wt.PollInterval, wt.PollTimeout, immediate, func(aContext context.Context) (bool, error) {
```
ffromani, line 89: nit: rather than `aContext`, let's prefer `ctx` all other factors being equal

ffromani (commented):
/approve
/lgtm

Impressive work and impressive persistence and discipline in following this through over the months. Kudos!

internal/wait/daemonset.go
shajmakh: done: https://github.com/openshift-kni/numaresources-operator/pull/851

pkg/objectupdate/rte/rte_test.go
shajmakh: addressed here: https://github.com/openshift-kni/numaresources-operator/pull/852

shajmakh:
Thanks @ffromani, for your invaluable guidance through this, your patience, and your helpful reviews!

ffromani:
> Thanks @ffromani, for your invaluable guidance through this, your patience, and your helpful reviews!

Thank you for the great contribution!
I'm deeply sorry for the revert (see #851 and #852), it's on me.
I noticed the incorrect commits only post-merge, when doing cherry-picks. Unfortunately the issue was evident only after close inspection using `git` and not from the `github` UI, but still I should have noticed before the merge. The silver lining is we added an automated check (#853) to prevent these occurrences.

PR #708, opened 2023-10-11
enable skipped tests & stabilize flaky tests

shajmakh:
Some tests were skipped for unexpected reasons, fix that and fix the flakiness by waiting enough time for the NRTs to settle. See commits for more details on each change.

### DIFF
--- a/test/e2e/serial/tests/scheduler_cache.go
+++ b/test/e2e/serial/tests/scheduler_cache.go
@@ -337,7 +337,7 @@ var _ = Describe("[serial][scheduler][cache][tier1] scheduler cache", Label("sch
 				klog.Infof("hosts required %d desired pods %d expected pending %d", hostsRequired, desiredPods, expectedPending)
 
 				// so we can't support ATM zones > 2. HW with zones > 2 is rare anyway, so not to big of a deal now.
-				// TOOD: when we support NUMA zones > 2, switch to FilterZoneCountAtLeast
+				// TODO: when we support NUMA zones > 2, switch to FilterZoneCountAtLeast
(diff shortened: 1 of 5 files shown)


test/e2e/serial/tests/scheduler_cache.go
```diff
@@ -326,7 +326,7 @@ var _ = Describe("[serial][scheduler][cache][tier1] scheduler cache", Label("sch
 				// In this testcase, having running pods is not good enough: we want to ALSO have pods which keep being
 				// pending "forever". We can't really check "forever", so we just check "long enough".
 
-				deviceName := e2efixture.GetDeviceType1Name()
+				deviceName := e2efixture.GetDeviceType3Name()
```
ffromani, line 329: in hindsight, we probably need to either use more descriptive names or the abstract these devices a bit further.
Stuff for a new PR anyway

ffromani (commented):
/approve
/lgtm

real nice work here

test/e2e/serial/tests/scheduler_cache.go
shajmakh: Thanks for the review and the comment. I agree, we set up the devices' settings using a separate CI job and tried as much as possible to include the needed combinations. I'll enhance the naming in a later PR.

ffromani:
do we want a backport to 4.14?
probably not worth to backport further than that

shajmakh:
> do we want a backport to 4.14? probably not worth to backport further than that


I don't think so, we started using the latest 4.15 tests image, I think it also contains fixes of other tests (reboot cases)

PR #714, opened 2023-10-12
serial: wait for NRT to settle part-2

shajmakh:
Wait for NRT to settle before verifying resources are updated in missing places.


shajmakh:
/hold 
for testing and more fixes if needed

test/e2e/serial/tests/non_regression.go
```diff
@@ -366,6 +366,9 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			failedPodIds := e2efixture.WaitForPaddingPodsRunning(fxt, paddingPods)
 			Expect(failedPodIds).To(BeEmpty(), "some padding pods have failed to run")
 
+			By("waiting for the NRT data to settle")
+			e2efixture.WaitForNRTSettle(fxt)
```
ffromani, line 370: let's use brand new `MustSettleNRT` (please rebase on top of current head of main)

shajmakh: done, thanks.

shajmakh:
The tests ran several times with full pass rate

shajmakh:
/hold 
for confirmation if this is the correct handling of the case.

ffromani (commented):
/approve
/lgtm

thanks for the fixes. We are now able to get rid of the silly cooldown with Sleep (ugh) and waiting for the settle seems better than checking the conditions in Eventually blocks.

ffromani:
> > > The tests ran several times with full pass rate
> > 
> > 
> > I think we should have this PR. It makes the code more correct and more robust. But wondering: how much slower on average does this PR make the test runs go? Negligible? (like 1-2 mins) Noticeable? (like 10m) Do we have data?
> 
> for the affected running tests it is ~20 secs per test on average, so for the total 13 tests it's was ~ 4:20 mins. this is based on the runs I have in hand which I executed only for the affected tests.

seems no bother. We should totally move forward.

PR #722, opened 2023-10-16
[release-4.10][manual] deps: bump 20231016

ffromani:
bump deps


shajmakh:
4.10 is out of maintenance, do we still need this?

ffromani:
> 4.10 is out of maintenance, do we still need this?

if it is so indeed, we're golden and we can close this PR

PR #723, opened 2023-10-19
operator: metrics: hooks: add disabling options

ffromani:
add command line flags and options to disable
the functionality, and disable it by default.

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -110,6 +114,8 @@ func main() {
 	flag.BoolVar(&showVersion, "version", false, "outputs the version and exit")
 	flag.BoolVar(&enableScheduler, "enable-scheduler", false, "enable support for the NUMAResourcesScheduler object")
 	flag.BoolVar(&enableWebhooks, "enable-webhooks", false, "enable conversion webhooks")
+	flag.BoolVar(&enableWebhooks, "enable-metrics", false, "enable metrics server")
```
fedepaol, line 117: you already have a flag (metricsAddr), wouldn't it be better to change the default value? Otherwise the cli ux is misleading, you set the address but nothing happens

ffromani: Yes, I'm still evaluating the best approach here

PR #737, opened 2023-10-25
e2e: Check Attributes when checking NRT mutation

### LINKED ISSUE #609: better handling of NRT Attributes in the e2e suite
The e2e suite currently:

- [doesn't check top-level Attributes when checking for NRT mutations](https://github.com/openshift-kni/numaresources-operator/pull/608#discussion_r1236804230)
- uses the deprecated `topologyPolicies` field as deciding factor, while it should fetch the same information from the `Attributes`

we need to improve how we handle Attributes in the test suite to fix both cases

jlojosnegros:
Attributes should be used when checking for NRT mutation.

This PR fixes #609


ffromani (commented):
/approve
/lgtm

swatisehgal (commented):
/lgtm

PR #744, opened 2023-10-26
[release-4.11] metrics: add option to toggle on/off

ffromani:
and disable by default because served unsecurely.
4.11 is approaching EOL; 4.12 will restore correct support.


swatisehgal (approved):
/lgtm
/approve

PR #745, opened 2023-10-26
[release-4.11] nrop: webhook: serve on localhost

ffromani:
webhooks are unneeded and expected to be disabled by default; for extra safety, force them on loclahost.


swatisehgal (approved):
/lgtm

PR #746, opened 2023-10-26
k8s: taints: import package

ffromani:
import the taints package, used just once in the e2e serial suite, to remove the last dependency from k8s.io/kubernetes
this PR won't actually remove that dep, but other PRs in the pipeline will.

no intended changes in behavior


shajmakh:
thanks for the PR. 
/lgtm 
/approve

PR #764, opened 2023-11-07
deployer: upgrade to v0.15.z

ffromani:
Update deployer deps from ancient v0.12.z to much more recent v0.15.z in steps, to enable (some form of) bisecting and to review API changes.
Still, deployer API changed significantly (and for the better, but still) so we will need quite some testing to move forward.


jlojosnegros (commented):
/lgtm

PR #765, opened 2023-11-07
chore: lint: fixes 20231107

ffromani:
partial address of lint warnings after we enabled in the test subtree in #758


swatisehgal (commented):
We have a few more lint failures in the last two commits because of which CI linter is failing but other than that the PR looks straightforward.

ffromani:
> We have a few more lint failures in the last two commits because of which CI linter is failing but other than that the PR looks straightforward.

yes, the missing bits are nontrivial and will be addressed in separate PRs.

jlojosnegros (commented):
/lgtm

ffromani:
> > We have a few more lint failures in the last two commits because of which CI linter is failing but other than that the PR looks straightforward.
> 
> yes, the missing bits are nontrivial and will be addressed in separate PRs.

like: https://github.com/openshift-kni/numaresources-operator/pull/766

PR #766, opened 2023-11-08
e2e: serial: rewrite deployment wait

ffromani:
the linter complains about code in this tests; rather than add a trivial patch, we take the chance to rewrite the wait loop using facilities gained by libs since the test was first introduced.
No intended change in behavior.


swatisehgal (approved):
/lgtm
/approve

PR #767, opened 2023-11-08
e2e: serial: remove half-baked randomSeed

ffromani:
We tried to make serial suite runs reproducible,
but the effort was half baked:
- we emit the seed WE configure, but we don't allow to set it
- there's no way to ensure any other package resets the global seed

turns out this code is more harmful than useful, so we remove it and we reset to a clean slate.


swatisehgal (approved):
/lgtm

PR #773, opened 2023-11-10
serial: remove ReporterAfterSuite to fix linter

shajmakh:
Linter is warning about using depricated ginkgo reporter code to create polarion report. Looking deeper the reporter is not used while publishing the tests report, as we use the junit file which has its own autogenerated ReporterAfterSuite.Thus removing the related code which cuase the noise in linter check.


shajmakh:
/hold
for testing

shajmakh:
the test passed. results are reported as expected on all venues.
/hold cancel

PR #652, opened 2023-07-13
Deprecate topology policies

jlojosnegros:
`topologyPolicies` field is deprecated and should not be used. 
The new recommended approach is to use the attributes. 
Update our codebase to use an attribute instead of that field.

Fixes #630

PR #774, opened 2023-11-13
[release-4.13][manual] Deprecate TopologyPolicies field

ffromani:
topologyPolicies field is deprecated and should not be used. The new recommended approach is to use the attributes.

Added two new attributes to handle policy and scope.

manual backport of #652


shajmakh:
Thanks for clarifying and for handling this.
/lgtm

PR #775, opened 2023-11-13
e2e: serial: more precise skip location

ffromani:
fix our helper to report which e2e code is skipping
- not where the actual ginkgo.Skip is ultimately invoked.


ffromani:
/cherry-pick release-4.14

want this down to 4.12

PR #776, opened 2023-11-13
vendor: remove unused deps

shajmakh:
cleanup after #773 and remove kubevirt deps.


test/e2e/serial/tests/scheduler_cache.go
```diff
@@ -1,650 +0,0 @@
-/*
```
ffromani, line 1: I don't think this should be deleted

test/e2e/serial/tests/scheduler_cache_stall.go
```diff
@@ -1,372 +0,0 @@
-/*
```
ffromani, line 1: I don't think this should be deleted

ffromani (commented):
it seems a bad commit was pushed? also `vendor/modules.txt` should be fixed and pushed

ffromani:
/approve
/lgtm

thanks for the cleanup!

PR #790, opened 2023-11-23
controller: test: relax condition

ffromani:
since e8432043 we have e2e test coverage, so we can relax this condition and tolerate the case on which we miss daemonsets.


swatisehgal (approved):
/lgtm

PR #801, opened 2023-11-28
improve support to coexist with standard scheduler 

ffromani:

1. react to any foreign pod (= pod not scheduled by this scheduler)
2. consider all the pods when computing the expected node state


ffromani:
missing e2e tests
missing docs/commit messages

ffromani:
the podresources socket failure is very suspicious, one last retest to rule out CI flakes (the fact the bot takes minutes to react warrants the test)

ffromani:
the PR is now ready for review

test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -508,3 +508,45 @@ func ResourceInfoListToResourceList(ri nrtv1alpha2.ResourceInfoList) corev1.Reso
...
+}
+
+func ResourceInfoProvidingAtMost(resources []nrtv1alpha2.ResourceInfo, resName string, resQty resource.Quantity) bool {
+	zeroQty := resource.MustParse("0")
+	zoneQty, ok := FindResourceAvailableByName(resources, resName)
+	klog.Infof("  +--> checking if resources include %q in (0, %s] (zoneQty=%s found=%v)", resName, resQty.String(), zoneQty.String(), ok)
+	if !ok {
+		return false
+	}
+	if zoneQty.Cmp(zeroQty) <= 0 {
```
Tal-or, line 545: You can use `zoneQty.IsZero`

test/e2e/serial/tests/scheduler_cache.go
```diff
@@ -612,12 +597,25 @@ var _ = Describe("[serial][scheduler][cache][tier1] scheduler cache", Label("sch
 				// NRT updater, resync loop, scheduler retry loop.
 				failedPods, updatedPods = wait.With(fxt.Client).Timeout(5*time.Minute).ForPodListAllRunning(context.TODO(), expectedRunningPods)
 				Expect(len(updatedPods)).To(Equal(hostsRequired))
+				dumpFailedPodInfo(fxt, failedPods)
```
Tal-or, line 600: In case of failed assertion in the line above, we won't get this information.

Tal-or:
Thanks for the PR, small comments from my side, but the general direction LGTM

test/utils/noderesourcetopologies/noderesourcetopologies.go
ffromani: thanks, fixed

test/e2e/serial/tests/scheduler_cache.go
ffromani: good catch, let me address this

ffromani:
> don't we need to backport this commit: [e658290](https://github.com/openshift-kni/numaresources-operator/commit/e6582903af1d5552d9d56961d481eee5d6d5aae6)?

That commit changes the default behavior, so it should backported very carefully. We don't for now, because this feature is meant for newer NUMA-scheduler stacks

PR #802, opened 2023-12-06
e2e: serial: resource accounting: take cluster snapshot post padding

ffromani:
Previously, the test fixture took its NRT snapshot before padding.
But padding is an essential component of testing, and requires guaranteed class pods to actually consume numa-local resources.

If we later check that conditions match, we will never ever converge because padding pods will be (correctly) counted to assess the node resource consumption.

The correct approach is then to take the expected state AFTER padding, considering padding pods as part of the cluster state, and right before any test-specific pod is scheduled.

Or, alternatively, to check the state excluding the padding pods which will make the test more complex (and thus fragile), and/or of questionable value.

The only remaining question is why this failure was gone unnoticed up until now.

### DIFF
--- a/test/e2e/serial/tests/resource_accounting.go
+++ b/test/e2e/serial/tests/resource_accounting.go
@@ -327,10 +327,14 @@ var _ = Describe("[serial][disruptive][scheduler][resacct] numaresources workloa
 			Expect(ok).To(BeTrue(), "cannot select a node among %#v", e2efixture.ListNodeNames(nrtCandidateNames))
 			By(fmt.Sprintf("selecting node to schedule the test pod: %q", targetNodeName))
 
-			targetNrtListInitial, err = e2enrt.GetUpdated(fxt.Client, nrtList, 1*time.Minute)
+			// TODO: just use nrtList?
+			err = fxt.Client.List(context.TODO(), &targetNrtListInitial)
 			Expect(err).ToNot(HaveOccurred())
+			klog.Infof("initial NRT List: %s", intnrt.ListToString(targetNrtListInitial.Items, " initial list"))
+
 			targetNrtInitial, err = e2enrt.FindFromList(targetNrtListInitial.Items, targetNodeName)
 			Expect(err).NotTo(HaveOccurred())
+			klog.Infof("initial NRT target: %s", intnrt.ToString(*targetNrtInitial))
 
 			//calculate base load on the target node
 			baseload, err := nodes.GetLoad(fxt.K8sClient, context.TODO(), targetNodeName)
--- a/test/e2e/serial/tests/resource_accounting.go
+++ b/test/e2e/serial/tests/resource_accounting.go
@@ -200,6 +200,9 @@ var _ = Describe("[serial][disruptive][scheduler][resacct] numaresources workloa
 			failedPodIds = e2efixture.WaitForPaddingPodsRunning(fxt, targetPaddingPods)
 			Expect(failedPodIds).To(BeEmpty(), "some padding pods have failed to run")
 
+			By("waiting for the NRT data to settle")
+			e2efixture.MustSettleNRT(fxt)
+
 			By("saturating nodes we want to be unsuitable")
 			for idx, unsuitableNodeName := range unsuitableNodeNames {
 				nrtInfo, err := e2enrt.FindFromList(nrtList.Items, unsuitableNodeName)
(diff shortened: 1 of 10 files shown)


ffromani:
the (nonblocking) security lane failure seems not worrying and will be handled with a later PR

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -90,6 +90,10 @@ var _ = Describe("[serial][disruptive][scheduler][resacct] numaresources workloa
 	AfterEach(func() {
 		err := padder.Clean()
 		Expect(err).NotTo(HaveOccurred())
+
+		By("waiting for the NRT data to settle")
```
shajmakh, line 94: is this (AfterEach()) the intended place? 
note that teardown is waiting for the NRT settlement to the initial data.

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -331,10 +331,14 @@ var _ = Describe("[serial][disruptive][scheduler][resacct] numaresources workloa
...
 
-			targetNrtListInitial, err = e2enrt.GetUpdated(fxt.Client, nrtList, 1*time.Minute)
+			// TODO: just use nrtList?
+			err = fxt.Client.List(context.TODO(), &targetNrtListInitial)
 			Expect(err).ToNot(HaveOccurred())
+			klog.Infof("initial NRT List: %s", intnrt.ListToString(targetNrtListInitial.Items, " initial list"))
+
 			targetNrtInitial, err = e2enrt.FindFromList(targetNrtListInitial.Items, targetNodeName)
 			Expect(err).NotTo(HaveOccurred())
+			klog.Infof("initial NRT target: %s", intnrt.ToString(*targetNrtInitial))
```
shajmakh, line 343: I think this is still not the right place to get the NRT data because padding is starting at line 359

shajmakh (commented):
Thank you for catching this. I really am curious why this was not seen before, I can only guess it is because we didn't wait enough after the padding so it moved quickly to the test and managed to have the same data for several attempts and it falsly passed. I'll verify in older runs the log of the test.

test/e2e/serial/tests/resource_accounting.go
shajmakh: I see you added that later after the padding. if so can we remove the var definition completely and start it only after the padding?

test/e2e/serial/tests/resource_accounting.go
ffromani: right, that was the intent and I somehow fumbled. Fixing. Thanks for catching!

however I'm now also contemplating the actual utility about this change; we would have two waits back to back, whose value is questionable.

shajmakh:
> /hold
> 
> let's try to bisect the latest working version to understand why it broke

it looks like this was broken indeed but never shown (no error handling), but thanks to #765 this was caught :
https://github.com/openshift-kni/numaresources-operator/commit/9e70a9842c9eea9cf922c90675c41fc0caf333cd#diff-b45c0de4295a0e0f027a7f5717bf64c8d648ee724c76c5d80f41f43f7aa3da35R413

shajmakh:
> Thank you for catching this. I really am curious why this was not seen before, I can only guess it is because we didn't wait enough after the padding so it moved quickly to the test and managed to have the same data for several attempts and it falsly passed. I'll verify in older runs the log of the test.

I was verifying how the test behaved in previous runs, and because it passed, there was not enough debug info. but even before that, if we check the code history, we see that there was no proper error handling, which is why this passed.

ffromani:
> > Thank you for catching this. I really am curious why this was not seen before, I can only guess it is because we didn't wait enough after the padding so it moved quickly to the test and managed to have the same data for several attempts and it falsly passed. I'll verify in older runs the log of the test.
> 
> I was verifying how the test behaved in previous runs, and because it passed, there was not enough debug info. but even before that, if we check the code history, we see that there was no proper error handling, which is why this passed.

this makes sense, but is a bit scary. Let's do better (yours truly first and foremost) from now on.

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -85,11 +85,15 @@ var _ = Describe("[serial][disruptive][scheduler][resacct] numaresources workloa
 
 		// Note that this test, being part of "serial", expects NO OTHER POD being scheduled
 		// in between, so we consider this information current and valid when the It()s run.
+
+		By("waiting for the NRT data to settle")
```
shajmakh, line 89: I'm still unsure we need this here, because 1. BeforeEach() do not do any padding. 2. teardown from prev test will check this.

ffromani: Very fair. Let me tinker with it, and likely remove.

test/e2e/serial/tests/resource_accounting.go
shajmakh: from what I saw all the tests I have found using this var are either using it to get the updated list (e2enrt.GetUpdated()) or compare post padding. because the context holding these tests has it's main goal to create non-gu pods while expecting no NRT changes. we can still keep it yet I just hope it be not misused cyrcling back to the situation the PR is addressing, which is the eng responsibility to verify but still.

shajmakh:
Thank you again for this. 
left non-blocking comments
/lgtm
/hold 
please cancel when you see fit

PR #803, opened 2023-12-06
 flagcodec: always use flag normalization

ffromani:
this enables us to remove a longstanding ambiguity about `--v` vs `-v` and to fix a silent bug on which we BOTH set `--v` and `-v`, with the latter winning, opening the door for surprising bugs.


ffromani:
need to wait for `deployer` to cherry-pick the `flagcodec` changes before we can consider a backport.

ffromani:
gentle ping

Tal-or:
/lgtm
/approve 
Thank you

PR #804, opened 2023-12-12
pause: add C-based implementation

ffromani:
Also add Makefile helpers, defaulting to the C-based (not the existing shell-based) implementation.


shajmakh:
I'm just curious, why this is needed, and how does it help to have a C-based image?

ffromani:
> I'm just curious, why this is needed, and how does it help to have a C-based image?

It's a combination of factors:

- using a static binary allow us to have  a `FROM scratch` image, minimizing the deps (we only need pause here)
- we don't NEED C, but in this case the C implementation is 4 lines of code (including spaces) and the final static binary is less than 1 megabyte

PR #806, opened 2023-12-25
OCPBUGS-25763: Set base CSV replaces field

rbaturov:
If we release our newer operator version without the correct metadata, it will release that new version just fine but at the same time it will unpublish all earlier versions in the channel, making the new version effectively the only entry in that channel.

This happens when a new operator bundle ships with a "skipRange" annotation but without the "replaces" property in the operators OLM metadata (CSV)

This PR adds the "replaces" property in addition to the "skipRange" property.

### DIFF
--- a/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
@@ -599,4 +599,5 @@ spec:
   minKubeVersion: 1.23.0
   provider:
     name: Red Hat
+  replaces: numaresources-operator.v4.14.999-snapshot
   version: 4.15.999-snapshot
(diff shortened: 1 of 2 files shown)


ffromani:
please copy paste the (good) PR description in the commit message, so the rationale for the change is recorded in the git history

bundle/manifests/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -599,4 +599,5 @@ spec:
   minKubeVersion: 1.23.0
   provider:
     name: Red Hat
+  replaces: numaresources-operator.v4.14.999-snapshot
```
ffromani, line 602: this is correct, but since we don't publish (yet?) on operatorhub.io, this is going to help us little. Let's have it nevertheless

ffromani (commented):
/approve

rbaturov:
> please copy paste the (good) PR description in the commit message, so the rationale for the change is recorded in the git history

Done

ffromani (approved):
/lgtm

Issue #809, opened 2024-01-05
deprecate or remove the generated k8sclientset


ffromani:
Did we ever used it somehow in the last 3-4 releases?
If not, time to get rid of it

PR #810, opened 2024-01-05
tree: clientset: remove unused and unsupported code

ffromani:
remove autogenerated clientset. No record of it being used since 4.12 at least, perhaps earlier. It was introduced and never really got traction, so high time to get rid of it

Fixes: #809


ffromani:
/hold

before to move forward, let's check the backward compatibility concerns, because this is a published package

ffromani:
/hold cancel

we figured out the backward compatibility concerns and we can move on with this change, let's just never backport it for any reason

PR #812, opened 2024-01-10
pkg: k8sclient: deprecate package

ffromani:
deprecate the autogenerated k8sclientset package;
use controller-runtime generic client instead.


ffromani:
/retest

adding comments can't cause image building failure if compilation passes (and it does)

ffromani:
the fact unit test lanes are passing is sufficient verification - we're only adding comments here. CI is acting again, so let's avoid the churn: https://github.com/openshift-kni/numaresources-operator/pull/811

Issue #815, opened 2024-01-10
add hierarchical test tiers


ffromani:
currently running tier2 tests means we run ONLY tier2 tests, not tier1 *and* tier2 tests. This is strictly speaking correct but is inconvenient. We need a way to run a tier of test including the higher tiers. E.g. run "up to" tier2 tests mean run tier0+tier1+tier2 tests

ffromani:
this can probably solved or at least greatly mitigated in the test runner

PR #816, opened 2024-01-10
hack: e2e: support hierarchical tiers in the runner

ffromani:
Enhance the runner script to support hierarchical tiers, abusing --focus for now, labels in the future.

What is "hierarchical tiers"?
Currently running tier2 tests means we run ONLY tier2 tests, not tier1 and tier2 tests. This is strictly speaking correct but is inconvenient. Introducing hierarchy means that if we select a tier (and enable hierarchy modes) then the tests which will run are the ones belonging to that tier and to all higher tier)

IOW the selected tier will mean "run all tiers up to this one, included".

Fixes: #815


ffromani:
failures not relevant, (unfortunately) the change is not exercised by CI, tested offline

Issue #814, opened 2024-01-10
introduce tier0 tests


ffromani:
running tier1 tests with the current codebase yields
```
Will run 47 of 93 specs
```

which is starting to defy the point to have tiers and most notably to have tier1. We should isolate the really critical tests in tier0 and keep the number at bay

PR #832, opened 2024-01-23
serial: tests: introduce tier 0

ffromani:
Now that we have a simple way to run all tests up to a given tier, included (see: f69ac68a6375fec6ec355c44bd1c6d02a36a7593) we can review our tiers.

The ter1 tests had 47 tests, roughly, which is great on a hand but bad on the other hand, because the very purpose of tiering was to have a strict and well delimited set of tests.

We promote the most critical tier1 tests to tier0 tests. Tiering is not set in stone - tests will be moved among tiers.

But tier size should always kept at bay in the coming changes. As rule of thumb, no tier should contain more than 20-25% of total tests; and this is especially true for high (tier0/tier1) tests.

Fixes: #814

### DIFF
--- a/test/e2e/install/install_test.go
+++ b/test/e2e/install/install_test.go
@@ -63,7 +63,7 @@ var _ = Describe("[Install] continuousIntegration", func() {
 	})
 
 	Context("with a running cluster with all the components", func() {
-		It("[test_id:47574][tier1] should perform overall deployment and verify the condition is reported as available", func() {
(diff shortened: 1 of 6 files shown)


shajmakh:
Thanks for this.
/lgtm

shajmakh:
/hold
@ffromani feel free to unhold when you see fit

ffromani:
/unhold

additional test done locally are good

PR #835, opened 2024-01-31
consume support for scheduler plugin informerMode from upstream

rbaturov:
Previously, the NUMA-aware scheduler plugin had d/s-specific code to enable the separate informer, which was in turn needed to get the full pod list vs the filtered pod list the scheduler framework provides. Thus, we had d/s specific enablement in the operator. Since commit [93c518b](https://github.com/kubernetes-sigs/scheduler-plugins/pull/599) the upstream added support for the separate informer, consumed by the KNI scheduler since commit [780ea84](https://github.com/openshift-kni/scheduler-plugins/pull/161). Remove the d/s specific enablement and set the proper configuration value.

We expect no changes from UX perspective. This is an internal change only.

### DIFF
--- a/controllers/numaresourcesscheduler_controller.go
+++ b/controllers/numaresourcesscheduler_controller.go
@@ -252,14 +252,20 @@ func configParamsFromSchedSpec(schedSpec nropv1.NUMAResourcesSchedulerSpec, cach
 
 	var foreignPodsDetect string
 	var resyncMethod string = k8swgmanifests.CacheResyncAutodetect
+	var informerMode string
 	if *schedSpec.CacheResyncDetection == nropv1.CacheResyncDetectionRelaxed {
 		foreignPodsDetect = k8swgmanifests.ForeignPodsDetectOnlyExclusiveResources
 	} else {
 		foreignPodsDetect = k8swgmanifests.ForeignPodsDetectAll
 	}
+	if *schedSpec.SchedulerInformer == k8swgmanifests.CacheInformerDedicated {
+		informerMode = k8swgmanifests.CacheInformerDedicated
+	} else {
+		informerMode = k8swgmanifests.CacheInformerShared
+	}
 	params.Cache.ResyncMethod = &resyncMethod
 	params.Cache.ForeignPodsDetectMode = &foreignPodsDetect
-	
+	params.Cache.InformerMode = &informerMode
 	klog.InfoS("setting cache parameters", dumpConfigCacheParams(params.Cache)...)
 
 	return params
@@ -270,6 +276,7 @@ func dumpConfigCacheParams(ccp *k8swgmanifests.ConfigCacheParams) []interface{} 
 		"resyncPeriod", strInt64Ptr(ccp.ResyncPeriodSeconds),
 		"resyncMethod", strStringPtr(ccp.ResyncMethod),
 		"foreignPodsDetectMode", strStringPtr(ccp.ForeignPodsDetectMode),
+		"informerMode", strStringPtr(ccp.InformerMode),
 	}
 }
 
--- a/controllers/numaresourcesscheduler_controller_test.go
+++ b/controllers/numaresourcesscheduler_controller_test.go
@@ -342,10 +342,6 @@ var _ = ginkgo.Describe("Test NUMAResourcesScheduler Reconcile", func() {
 					Name:  schedupdate.PFPStatusDumpEnvVar,
 					Value: schedupdate.PFPStatusDir,
 				},
-				{
-					Name:  schedupdate.NRTInformerEnvVar,
-					Value: schedupdate.NRTInformerVal,
-				},
 			} {
 				gotEv := schedupdate.FindEnvVarByName(cnt.Env, ev.Name)
 				gomega.Expect(gotEv).ToNot(gomega.BeNil(), "missing environment variable %q in %q", ev.Name, cnt.Name)
@@ -388,9 +384,6 @@ var _ = ginkgo.Describe("Test NUMAResourcesScheduler Reconcile", func() {
 				{
 					Name: schedupdate.PFPStatusDumpEnvVar,
 				},
-				{
-					Name: schedupdate.NRTInformerEnvVar,
-				},
 			} {
 				gotEv := schedupdate.FindEnvVarByName(cnt.Env, ev.Name)
 				gomega.Expect(gotEv).To(gomega.BeNil(), "unexpected environment variable %q in %q", ev.Name, cnt.Name)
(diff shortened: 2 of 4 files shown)


Tal-or:
@rbaturov retest won't help here
You haver some linter issues:
```
pkg/objectupdate/sched/sched.go:73: unnecessary trailing newline (whitespace)
```

ffromani:
> This PR introduces changes required to fix the numaresources operator to consume the upstreamed separate informer support. Now, the informerMode field is passed through the topology-aware scheduler config map as expected from the scheduler to consume. The previous implementation of passing the informerMode field through the environment variable is removed.

This text is good, but still heavily skewed about the "who". IOW, it is summarizing the changes in plain english. This is not the content we should aim for. We should aim for text which describes the "why", giving context when needed to understand "why", why now? why is important? why is better than before (for a given definition of "better")?

We will (or we should...) go back and read git log and/or PR description to learn the motive and remember the context, not to get a quick grasp of the changes.

So what about something like
"Previously, the NUMA-aware scheduler plugin had d/s-specific code to enable the separate informer, which was in turn needed to get the full pod list vs the filtered pod list the scheduler framework provides. Thus, we had d/s specific enablement in the operator. Since commit XXX (TODO: figure it out) the upstream added support for the separate informer, consumed by the KNI scheduler since commit YYY (TODO: figure out). Remove the d/s specific enablement and set the proper configuration value.

We expect no changes from UX perspective. This is an internal change only."

controllers/numaresourcesscheduler_controller.go
```diff
@@ -259,7 +259,7 @@ func configParamsFromSchedSpec(schedSpec nropv1.NUMAResourcesSchedulerSpec, cach
 	}
 	params.Cache.ResyncMethod = &resyncMethod
 	params.Cache.ForeignPodsDetectMode = &foreignPodsDetect
-
+
```
ffromani: unnecessary change, please remove

controllers/numaresourcesscheduler_controller.go
```diff
@@ -259,7 +259,7 @@ func configParamsFromSchedSpec(schedSpec nropv1.NUMAResourcesSchedulerSpec, cach
 	}
```
ffromani, line 260: the commit message is empty here, but this is one of the rare cases on which is OK.
For extra point you can hint (1-2 lines) which feature you want to consume which prompted the bump, but the reason why is already pretty obvious from the PR context, so minimalistic commit message is fine.

controllers/numaresourcesscheduler_controller.go
```diff
@@ -252,14 +252,20 @@ func configParamsFromSchedSpec(schedSpec nropv1.NUMAResourcesSchedulerSpec, cach
 
 	var foreignPodsDetect string
```
ffromani, line 253: please expand the commit message to explain succintly the "why" of this change. Feel free to take parts of the PR description and paste them here (or, equivalently, the PR description should build and expand on the content of the most important commit message(s) of the series, which in this case is this one).

controllers/numaresourcesscheduler_controller_test.go
```diff
@@ -342,10 +342,6 @@ var _ = ginkgo.Describe("Test NUMAResourcesScheduler Reconcile", func() {
 					Name:  schedupdate.PFPStatusDumpEnvVar,
```
ffromani, line 342: this commit should be squashed in the previous

pkg/objectupdate/sched/sched_test.go
```diff
@@ -73,10 +73,6 @@ var dpAllOptions = &appsv1.Deployment{
 								Name:  "PFP_STATUS_DUMP",
 								Value: "/run/pfpstatus",
 							},
-							{
```
ffromani, line 76: this change is good. I think we do want to preserve the intent of the tests, so check that we can effectively change the setting, using the new way (ConfigParams), don't we?

ffromani:
Thanks for the PR! I left a few comments to improve over this good start. I'll wait for further review from both @Tal-or and @shajmakh !

ffromani:
no point in further retests until the quick and basic lanes (e.g. CI Linter) are happy

ffromani:
/retitle consume support for scheduler plugin informerMode from upstream

ffromani:
bad rebase it seems?

ffromani:
needs to be backported down to 4.14

Tal-or:
Please add the CNF ticket in the PR header in order to sync it with Jira

ffromani:
> Please add the CNF ticket in the PR header in order to sync it with Jira

I removed it. I'm unconvinced we should have this synchronization.

Tal-or:
> > Please add the CNF ticket in the PR header in order to sync it with Jira
> 
> I removed it. I'm unconvinced we should have this synchronization.

This is mostly for stakeholders which are monitoring mainly with Jira so they'll have an easy way to associate the ticket with the PR

ffromani:
> > > Please add the CNF ticket in the PR header in order to sync it with Jira
> > 
> > 
> > I removed it. I'm unconvinced we should have this synchronization.
> 
> This is mostly for stakeholders which are monitoring mainly with Jira so they'll have an easy way to associate the ticket with the PR

The question is if this project accepts these kinds of tags pointing to trackers external to the project.

controllers/numaresourcesscheduler_controller_test.go
```diff
@@ -417,7 +410,55 @@ var _ = ginkgo.Describe("Test NUMAResourcesScheduler Reconcile", func() {
...
+
+			key := client.ObjectKeyFromObject(nrs)
+			_, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: key})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+			expectCacheParams(reconciler.Client, depmanifests.CacheResyncAutodetect, depmanifests.CacheResyncOnlyExclusiveResources, depmanifests.CacheInformerShared)
+
+		})
+
+		ginkgo.It("should allow to change the informerMode to Dedicated", func() {
```
ffromani, line 444: this test makes sense only if we
1. start with informerMode dedicated
2. we can correctly change to shared
3. we can then correctly change it back again to dedicated

if you followed another test which doesn't do that, the other test is wrong and needs to be fixed in another PR

pkg/objectupdate/sched/sched_test.go
```diff
@@ -328,9 +324,6 @@ func TestDeploymentEnvVarSettings(t *testing.T) {
 	cacheResyncDebugEnabled := nropv1.CacheResyncDebugDumpJSONFile
 	cacheResyncDebugDisabled := nropv1.CacheResyncDebugDisabled
 
-	schedInformerShared := nropv1.SchedulerInformerShared
-	schedInformerDedicated := nropv1.SchedulerInformerDedicated
-
```
ffromani, line 333: why was this removed?

ffromani (commented):
/lgtm

minor comments can be addressed in a later PR

ffromani:
/approve

we need to do this work to catch up with scheduler-plugins code upstream

ffromani:
let's start the backports once this code merges

ffromani:
sweet

PR #836, opened 2024-02-01
WIP: api: numaresourcesopereator: MCP status reflection is optional

ffromani:
The NUMAResourcesOperator object initially reflected the status of the selected MCP in its own status. This helped the initial (very initial, bootstrapping) stage, but quickly becomes large and it is essentially unnecessaty and wasteful.

We start declaring the values optional.
Future changes, backward compatibility allowing, will stop reflecting the status to save bandwidth, and unblock future compatibility.


ffromani:
no, bad idea. The biggest offender is `status.MachineConfigPool[*].Conditions` which is marked optional anyway. We can't retroactively mark this optional, and actually doing so could harm our extensions plan.

We can and probably should reuse this field for the aforementioned plans (in general), deferring the cleanup to the api v2, if we will ever get to that.

PR #837, opened 2024-02-02
controller: stop reflecting MCP conditions

ffromani:
Reflecting MCP conditions in the NRO object conditions is heavyweight and gives little extra value. Let's stop do that - the nested conditions are optional anyway, but let's add a command line option to re-enable it if really needed.

Note: "heavyweight" has to be put in context


ffromani:
/hold cancel

additional test locally

ffromani:
ci-install-e2e failure not critical: failed in must-gather (infra issue), our tests passed cleanly

PR #843, opened 2024-02-06
snyk: add exclude list for 4.13

rbaturov:
Added this file to be consumed by the snyk static analysis check.
Thereby, we can exclude packages that raise vulnerability warnings to silence them.


ffromani:
failures are legit. We need to look into them.

ffromani:
please rebase on top of the head of the branch

PR #846, opened 2024-02-08
[release-4.14][manual] consume support for scheduler plugin informerMode from upstream

ffromani:
Previously, the NUMA-aware scheduler plugin had d/s-specific code to enable the separate informer, which was in turn needed to get the full pod list vs the filtered pod list the scheduler framework provides. Thus, we had d/s specific enablement in the operator. Since commit https://github.com/kubernetes-sigs/scheduler-plugins/pull/599 the upstream added support for the separate informer, consumed by the KNI scheduler since commit https://github.com/openshift-kni/scheduler-plugins/pull/161. Remove the d/s specific enablement and set the proper configuration value.

We expect no changes from UX perspective. This is an internal change only.

backport notice: add targeted fixes to make this PR independent from (the backport of) #801 
backport notice: manual backport of #845

### DIFF
--- a/controllers/numaresourcesscheduler_controller.go
+++ b/controllers/numaresourcesscheduler_controller.go
@@ -176,16 +176,9 @@ func (r *NUMAResourcesSchedulerReconciler) syncNUMASchedulerResources(ctx contex
 	defer klog.V(4).Info("SchedulerSync stop")
 
 	schedSpec := instance.Spec.Normalize()
-
 	cacheResyncPeriod := unpackAPIResyncPeriod(schedSpec.CacheResyncPeriod)
 
-	resyncPeriod := int64(cacheResyncPeriod.Seconds())
-	params := k8swgmanifests.ConfigParams{
-		ProfileName: schedSpec.SchedulerName,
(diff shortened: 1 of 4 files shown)


ffromani:
/test ci-e2e

is this a flake with a terrible timing?

ffromani:
ok, not a flake. Investigating.

ffromani:
/hold

checking the scheduler side. Probably unnecessary bordering harmful

ffromani:
we don't want this change now: https://github.com/openshift-kni/scheduler-plugins/pull/165

ffromani:
/reopen
we will merge once https://github.com/openshift-kni/scheduler-plugins/pull/223 merges

ffromani:
the failure is expected because https://github.com/openshift-kni/scheduler-plugins/pull/223 merged but the lane restarted before automation pushed an updated image

ffromani:
re-closing with better explanation of the reason. Would depend on the config AP v1beta3. the branch 4.14 was using v1beta2, and we can't update mid-Z-stream, especially so late in the stream. xref: https://github.com/openshift-kni/scheduler-plugins/pull/224

PR #848, opened 2024-02-11
flags: identify -v with single hyphen

shajmakh:
Follow the convention by referring to single-letter identifier, -v here, with single hyphen.

A scheduler and deployer with similar PR will follow as well as bumping the versions.

### DIFF
--- a/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
@@ -511,7 +511,7 @@ spec:
                         operator: Exists
               containers:
               - args:
-                - --v=4
+                - -v=4
                 - --leader-elect
                 - --enable-scheduler
                 command:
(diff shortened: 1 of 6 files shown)


shajmakh:
/hold 
(to include the version bump)

shajmakh:
/cc @ffromani 
not sure we are in need of this, it is only meant to follow the rules by the book, but considering the [comment](https://github.com/k8stopologyawareschedwg/deployer/blob/main/pkg/flagcodec/flagcodec.go#L100C1-L109C29), I think it is not bad of keeping it as is. also this seems to be known in other common [repos](https://github.com/onsi/ginkgo/blob/master/ginkgo/command/program.go#L123).

ffromani (commented):
When we consume deployer >= 0.18 this is no longer necessary, but it's still better to be consistent. Thanks for cleaning this up!

ffromani (commented):
/approve
/lgtm

ffromani:
please rebase on top of main because we merged https://github.com/openshift-kni/numaresources-operator/pull/853

PR #849, opened 2024-02-12
Revert "controller: support suppressing NRT updates"

ffromani:
Reverts openshift-kni/numaresources-operator#700


ffromani:
reverted and force-pushed (see: #850)

PR #851, opened 2024-02-12
internal: verify ds pods are updated

shajmakh:
If a ds is updated and the current method is used to wait for the ds to be ready, the check will pass but in reality the ds pods will still need to be updated. this is because the check doesn't ensure the pods are updated.

Ensure that by adding the condition that the number of up to date pods under the daemoonset equals to the expected and ready number pods.


shajmakh:
/assign @ffromani 
as per comment: https://github.com/openshift-kni/numaresources-operator/pull/700#discussion_r1485816105

ffromani:
please rebase on top of main because we merged #853

ffromani:
/approve
/lgtm

thanks for splitting this out and fixing very quickly

PR #863, opened 2024-03-05
OCPBUGS-30236: controller: kubelet: skip config with empty payload

ffromani:
MCO's KubeletConfig can be sent with empty payload, for example when configuring autosizing:

```
apiVersion: machineconfiguration.openshift.io/v1
kind: KubeletConfig
metadata:
  name: autosizing-master
spec:
  autoSizingReserved: true
  machineConfigPoolSelector:
    matchLabels:
      pools.operator.machineconfiguration.openshift.io/master: ""
```

In general there's not much we can do when obeserving these kubeletconfig objects anyway, so we skip them.

Please note: the control loop still depends on the fact there
1. is just one kubeletconfig object *with payload*
2. there's no merge logic

as it stands today both still should be core assumptions regarding how MCO works.

### DIFF
--- a/controllers/kubeletconfig_controller_test.go
+++ b/controllers/kubeletconfig_controller_test.go
@@ -129,6 +129,22 @@ var _ = Describe("Test KubeletConfig Reconcile", func() {
 				event := <-fakeRecorder.Events
 				Expect(event).To(ContainSubstring("ProcessFailed"))
 			})
+
+			It("should skip invalid kubeletconfig", func() {
+				invalidMcoKc := testobjs.NewKubeletConfigWithoutData("test1", label1, mcp1.Spec.MachineConfigSelector)
+				reconciler, err := NewFakeKubeletConfigReconciler(nro, mcp1, invalidMcoKc)
+				Expect(err).ToNot(HaveOccurred())
+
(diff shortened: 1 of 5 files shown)


controllers/kubeletconfig_controller_test.go
```diff
@@ -129,6 +129,22 @@ var _ = Describe("Test KubeletConfig Reconcile", func() {
...
+
+				key := client.ObjectKeyFromObject(mcoKc1)
+				_, err = reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: key})
+				Expect(err).ToNot(HaveOccurred())
+
+				// verify creation event
+				fakeRecorder, ok := reconciler.Recorder.(*record.FakeRecorder)
+				Expect(ok).To(BeTrue())
+				event := <-fakeRecorder.Events
+				Expect(event).To(ContainSubstring("ProcessSkip"))
```
yanirq, line 146: "ProcessSkip"  is a constant ?  should it be ?   (I'm trying to think how to link the desired test string here to the code if at all)

ffromani: we don't have a formal API contract about these values (so yes, the API is leaking)

Issue #873, opened 2024-03-07
ignore kubeletconfig not matching the configured nodegroups


ffromani:
in #863 we added code to safely ignore kubeletconfigs without payload. The fix in itself is correct, but it's too late in the flow.
We should first and foremost check we are looking at `kubeletconfig` objects matching the `nodeGroups` we care about, discarding everything else.

PR #874, opened 2024-03-07
kubeletconfig: ignore nonmatching kubeletconfig objects

ffromani:
In #863 (https://github.com/openshift-kni/numaresources-operator/commit/3caa8294757cf4b98d554f0af813b7c77e5ed454) we added support to handle gracefully payload-less kubeletconfig objects. There's however another issue lurking: we can have multiple kubeletconfig objects.

Previously, the code was incorrectly assuming there is at most one kubeletconfig object in the cluster, targeting the MCP
worker pool.

But we can actually have more kubeletconfig objects as long as they don't target overlapping MCPs (e.g. with 2 MCPs we can have 2 kubeletconfig objects, say, or perhaps even 4 considering payloadless objects).

Thus, we need to fix our matching logic to handle this occurrence.
The fix is simple: just look for kubeletconfig which we target NodeGroups we are interested into, silently skipping everything else.

Fixes: #873

### DIFF
--- a/controllers/kubeletconfig_controller.go
+++ b/controllers/kubeletconfig_controller.go
@@ -123,10 +123,15 @@ func (r *KubeletConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 type InvalidKubeletConfig struct {
 	ObjectName string
+	Err        error
 }
 
 func (e *InvalidKubeletConfig) Error() string {
 	return "invalid KubeletConfig object: " + e.ObjectName
+}
+
+func (e *InvalidKubeletConfig) Unwrap() error {
+	return e.Err
 }
 
 func (r *KubeletConfigReconciler) reconcileConfigMap(ctx context.Context, instance *nropv1.NUMAResourcesOperator, kcKey client.ObjectKey) (*corev1.ConfigMap, error) {
@@ -134,6 +139,26 @@ func (r *KubeletConfigReconciler) reconcileConfigMap(ctx context.Context, instan
 	if err := r.Client.Get(ctx, kcKey, mcoKc); err != nil {
 		return nil, err
 	}
+
+	mcps, err := machineconfigpools.GetListByNodeGroupsV1(ctx, r.Client, instance.Spec.NodeGroups)
+	if err != nil {
+		return nil, err
+	}
+
+	mcp, err := machineconfigpools.FindBySelector(mcps, mcoKc.Spec.MachineConfigPoolSelector)
+	if err != nil {
+		klog.ErrorS(err, "cannot find a matching mcp for MCO KubeletConfig", "name", kcKey.Name)
+		var notFound *machineconfigpools.NotFound
+		if errors.As(err, &notFound) {
+			return nil, &InvalidKubeletConfig{
+				ObjectName: kcKey.Name,
+				Err:        notFound,
+			}
+		}
+		return nil, err
+	}
+
+	klog.V(3).InfoS("matched MCP to MCO KubeletConfig", "kubeletconfig name", kcKey.Name, "MCP name", mcp.Name)
 
 	// nothing we care about, and we can't do much anyway
 	if mcoKc.Spec.KubeletConfig == nil {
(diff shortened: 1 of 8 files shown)


ffromani:
/hold cancel

the e2e test looks good (enough)

internal/machineconfigpools/machineconfigpools_test.go
```diff
@@ -0,0 +1,197 @@
+/*
```
shajmakh, line 1: likely in this file, but not completely sure where this should be, shouldn't we have tests to cover multiple kubeletconfigs that the first one in the list is not matching the nodegroups configured in nrop resource, and the second one does?

test/e2e/serial/tests/configuration.go
```diff
@@ -736,6 +739,41 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
...
+
+			By("recording the current kubeletconfig and configmap status")
+			kcCmsPre, err := getKubeletConfigMapsSoftOwnedBy(ctx, fxt.Client, nroOperObj.Name)
+			Expect(err).ToNot(HaveOccurred(), "cannot list KubeletConfig ConfigMaps in the cluster (PRE)")
+			kcCmNamesPre := sets.List[string](accumulateKubeletConfigNames(kcCmsPre))
+			klog.Infof("initial set of configmaps from kubeletconfigs: %v", strings.Join(kcCmNamesPre, ","))
+
+			By("creating extra ctrplane kubeletconfig")
+			ctrlPlaneKc := intobjs.NewKubeletConfigAutoresizeControlPlane()
+			err = fxt.Client.Create(ctx, ctrlPlaneKc)
```
shajmakh, line 758: does this trigger an update on the controlplane nodes? if so I believe we better wait for that update to finish here and in defer

shajmakh (commented):
Thank you for working on this.
left a few comment below.

controllers/kubeletconfig_controller.go
```diff
@@ -123,30 +123,23 @@ func (r *KubeletConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 type InvalidKubeletConfig struct {
 	ObjectName string
+	Err        error
 }
 
 func (e *InvalidKubeletConfig) Error() string {
 	return "invalid KubeletConfig object: " + e.ObjectName
 }
 
+func (e *InvalidKubeletConfig) Unwrap() error {
```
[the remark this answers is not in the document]
ffromani: This is called by `errors.As`: https://pkg.go.dev/errors#As

controllers/kubeletconfig_controller.go
```diff
@@ -123,30 +123,23 @@ func (r *KubeletConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	return e.Err
+}
+
 func (r *KubeletConfigReconciler) reconcileConfigMap(ctx context.Context, instance *nropv1.NUMAResourcesOperator, kcKey client.ObjectKey) (*corev1.ConfigMap, error) {
 	mcoKc := &mcov1.KubeletConfig{}
 	if err := r.Client.Get(ctx, kcKey, mcoKc); err != nil {
```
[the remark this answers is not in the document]
ffromani: the namespace is present but may be "" (empty), which is the standard for cluster-scoped resources;
likewise, cluster-scoped resources must have unique names, so we are reasonnably sure to pick the right object

internal/machineconfigpools/machineconfigpools_test.go
ffromani: good idea, I'll work on that

controllers/kubeletconfig_controller.go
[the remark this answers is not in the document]
shajmakh: good to learn, thanks for sharing that!

shajmakh:
thank you for adding the test.
/lgtm
/hold 
leaving room for others to review, please ungate as you see better.

controllers/kubeletconfig_controller.go
```diff
@@ -123,18 +123,43 @@ func (r *KubeletConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		if errors.As(err, &notFound) {
+			return nil, &InvalidKubeletConfig{
+				ObjectName: kcKey.Name,
+				Err:        notFound,
+			}
+		}
+		return nil, err
+	}
+
+	klog.InfoS("matched MCP to MCO KubeletConfig", "kubeletconfig name", kcKey.Name, "MCP name", mcp.Name)
```
yanirq, line 161: what would be the frequency of this log message ?   (just asking to see if the log can get bloated from this info msg)

internal/machineconfigpools/machineconfigpools_test.go
```diff
@@ -0,0 +1,197 @@
...
+		expectedErr error
+	}
+
+	testCases := []testCase{
+		{
+			name:        "all empty",
+			expectedErr: &NotFound{},
+		},
+		{
+			name: "mcps but missing sel",
```
yanirq, line 43: sel = selector ?

controllers/kubeletconfig_controller.go
ffromani: It will be emitted each time a reconciliation is attempted. I can probably move to V(3).

internal/machineconfigpools/machineconfigpools_test.go
ffromani: yes, since I resubmit I'll expand

ffromani:
/hold cancel

LGTM is gone anyway

PR #887, opened 2024-03-26
e2e: api: add and use support to sort tolerations

ffromani:
Make sure we compare sorted tolerations in the e2e tests, so we minimize flakes.
No intended changes in behavior.

PR #888, opened 2024-03-26
build: golang: remove 'latest' floating tag refs

ffromani:
never use floating tags in the build system, not
even for CI tests. Always pin explicit versions.


ffromani:
required by #887

PR #900, opened 2024-03-27
e2e: add tests for user-provided toleration scenarios

shajmakh:
TBD


shajmakh:
need a rebase + testing is on going offline

test/e2e/serial/tests/tolerations.go
```diff
@@ -350,6 +351,83 @@ var _ = Describe("[serial][disruptive][slow][rtetols] numaresources RTE tolerati
...
+						return nil
+					}
+
+					klog.Infof("adding taint: %q to node: %q", tnt.String(), updatedNode.Name)
+					err = fxt.Client.Update(ctx, updatedNode)
+					if err != nil {
+						return err
+					}
+					return nil
+				}).WithPolling(1 * time.Second).WithTimeout(1 * time.Minute).ShouldNot(HaveOccurred())
```
ffromani, line 397: please move to `applyTaintToNode` (or let's talk why is not suitable)

test/e2e/serial/tests/tolerations.go
```diff
@@ -350,6 +351,83 @@ var _ = Describe("[serial][disruptive][slow][rtetols] numaresources RTE tolerati
...
+				Expect(len(pods)).To(Equal(len(workers)), "pods number is not as expected for RTE daemonset: expected %d found %d", len(workers), ds.Status.DesiredNumberScheduled)
+
+				var podToDelete corev1.Pod
+				for _, pod := range pods {
+					if pod.Spec.NodeName == taintedNode.Name {
+						podToDelete = pod
+						break
+					}
+				}
+				Expect(podToDelete).NotTo(BeNil())
```
ffromani, line 418: this is an object (not a pointer) so can it ever be nil?

test/e2e/serial/tests/tolerations.go
```diff
@@ -350,6 +351,83 @@ var _ = Describe("[serial][disruptive][slow][rtetols] numaresources RTE tolerati
...
+				Expect(err).ToNot(HaveOccurred(), "failed to get the daemonset ready: %v", err)
+
+				klog.Info("verify there is no RTE pod running on the tainted node")
+				pods, err = podlist.With(fxt.Client).ByDaemonset(ctx, *updatedDs)
+				Expect(err).NotTo(HaveOccurred(), "Unable to get pods from Daemonset %q: %v", ds.Name, err)
+				for _, pod := range pods {
+					Expect(pod.Spec.NodeName).NotTo(Equal(taintedNode.Name), "found RTE pod running on tainted node without toleration on NROP obj")
+				}
+			})
+
```
ffromani, line 439: This works. Maybe we can do better: get all the pods running on the tainted node, verify none of them is a RTE pod (there's a label for that, and we can also check the command line and/or the image)

test/e2e/serial/tests/tolerations.go
```diff
@@ -433,6 +438,181 @@ var _ = Describe("[serial][disruptive][slow][rtetols] numaresources RTE tolerati
...
+							return nil
+						}
+
+						klog.Infof("adding taint: %q to node: %q", tnt.String(), updatedNode.Name)
+						err = fxt.Client.Update(ctx, updatedNode)
+						if err != nil {
+							return err
+						}
+						return nil
+					}).WithPolling(1 * time.Second).WithTimeout(1 * time.Minute).ShouldNot(HaveOccurred())
```
ffromani, line 489: again let's use `applyTaintToNode` (or let's discuss why not)

test/e2e/serial/tests/tolerations.go
```diff
@@ -433,6 +438,181 @@ var _ = Describe("[serial][disruptive][slow][rtetols] numaresources RTE tolerati
...
+					Expect(err).NotTo(HaveOccurred(), "Unable to get pods from Daemonset %q: %v", ds.Name, err)
+					Expect(len(pods)).To(Equal(len(workers)), "pods number is not as expected for daemonset %s/%s: expected %d found %d", ds.Namespace, ds.Name, len(workers)-1, len(pods))
+					found := false
+					for _, pod := range pods {
+						if pod.Spec.NodeName == taintedNode.Name {
+							found = true
+							break
+						}
+					}
+					Expect(found).To(BeTrue())
```
ffromani, line 573: if we add a helper like suggested previously to find a RTE pod among all the pods running on a node (and I think there's something somewhere already, perhaps as a starting point? maybe u/s in the deployer or RTE repo), we can use the same code here to make sure such pod exists

test/e2e/serial/tests/tolerations.go
```diff
@@ -433,6 +438,181 @@ var _ = Describe("[serial][disruptive][slow][rtetols] numaresources RTE tolerati
...
```
ffromani, line 612: again this seems to be a nice candidate for the helper function to find a RTE pod among all pods running on a node

test/e2e/serial/tests/tolerations.go
```diff
@@ -614,6 +614,91 @@ var _ = Describe("[serial][disruptive][slow][rtetols] numaresources RTE tolerati
...
+						return nil
+					}
+
+					klog.Infof("adding taint: %q to node: %q", tnt.String(), updatedNode.Name)
+					err = fxt.Client.Update(ctx, updatedNode)
+					if err != nil {
+						return err
+					}
+					return nil
+				}).WithPolling(1 * time.Second).WithTimeout(1 * time.Minute).ShouldNot(HaveOccurred())
```
ffromani, line 654: `applyTaintToNode` again

test/e2e/serial/tests/tolerations.go
```diff
@@ -614,6 +614,91 @@ var _ = Describe("[serial][disruptive][slow][rtetols] numaresources RTE tolerati
...
+				pods, err = podlist.With(fxt.Client).ByDaemonset(ctx, *updatedDs)
+				Expect(err).NotTo(HaveOccurred(), "Unable to get pods from Daemonset %s: %v", dsKey.String(), err)
+				found := false
+				for _, pod := range pods {
+					if pod.Spec.NodeName == taintedNode.Name {
+						found = true
+						break
+					}
+				}
+				Expect(found).To(BeTrue())
```
ffromani, line 697: ditto

ffromani (commented):
very nice work here. Let's discuss (minorish) improvements

test/e2e/serial/tests/tolerations.go
ffromani: this is perhaps the biggest/only actual defect I found so far in this PR

test/e2e/serial/tests/tolerations.go
shajmakh: that is also another way to do this check. could you please tell why getting all node's pods and check the label on each pod is better than checking the assigned node of the all the RTE pods? 
one thing I see better in the current way is that it will need to go over less pods (3 vs 30+) considering our CI supports at best 3 workers.

ffromani: the main reason is I'd like to encourage variety in the test flow. I'd like to crosscheck results coming from different angles, so the final check could (and IMO should) take a different route than keep depending more and more on daemonset reports.

Checking all pods on a selected node is not something concerning performance wise, if the test code or even worse our platform can't cope with the load caused by listing all pods on a node we have way worse issues in our hands.

shajmakh: okay I see your point now. That is indeed good idea. I'll improve this part as you suggested. Thanks for calrifying.

test/e2e/serial/tests/tolerations.go
shajmakh: thanks for pointing out, I missed that.

test/e2e/serial/tests/workload_placement_taint.go
```diff
@@ -131,31 +131,8 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
-					klog.Infof("adding taint: %q to node: %q", tnt.String(), updatedNode.Name)
-					err = fxt.Client.Update(context.TODO(), updatedNode)
-					if err != nil {
-						return err
-					}
-					updatedNodeNames = append(updatedNodeNames, updatedNode.Name)
-					return nil
-				}).WithPolling(1 * time.Second).WithTimeout(1 * time.Minute).ShouldNot(HaveOccurred())
+				applyTaintToNode(context.TODO(), fxt.Client, refNode, tnt)
+				updatedNodeNames = append(updatedNodeNames, refNode.Name)
```
ffromani, line 135: good catch, thanks

test/e2e/serial/tests/tolerations.go
shajmakh: indeed the object won't be nil. changed the check to verify the pod name isn't empty. Thank you for catching this.

ffromani:
/approve
/lgtm

thanks for addressing the review comments

PR #895, opened 2024-03-27
CNF-11848: [release-4.15][manual] RTE: add support to add extra tolerations

ffromani:
add support to inject (and override) the RTE tolerations.
The user-provided tolerations will completely override the manifests tolerations.
If the user removes the tolerations, then the operator will enforce again the manifest ones (possibly, no tolerations).

The recommended flow to add/change the default tolerations (if any) is to
1. find the owned daemonsets: nrop.Status.DaemonSets
2. lookup the owned daemonsets using the namespace/name pair found in
   the previous step
3. inspect the DS manifests to learn about the current tolerations
4. set new tolerations in nrop.Spec.NodeGroup[*].Config.Tolerations
   taking into account the existing ones

cumulative backport of #808 #887 #894 #901

PR #901, opened 2024-03-27
controller: fix daemonUpdater

ffromani:
some of the changes in the daemonUpdater callback
are platform agnostic and should be performed anyway.


ffromani:
backport merged in #895

PR #910, opened 2024-04-03
e2e: wait for ds creation before checking it is ready

shajmakh:
The test deletes ds and immediately checks for it to be ready using its object key. When there is no ds yet the test would fail because the object it is trying to search is not found on the cluster.

To prevent this from happening, add wait after the ds is deleted to make sure it is recreated with the expected pod count (not ready yet though).

This was also driving another failure in the next test (72861) where when it checks the number of pods owned by ds it retrieves a double count because of the terminating pods that still exist on the cluster:

```
shajmakh@shajmakh-thinkpadt590 ~ $ oc get ds,pod
NAME                                          DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR                     AGE
daemonset.apps/numaresourcesoperator-worker   2         2         2       2            2           node-role.kubernetes.io/worker=   26s

NAME                                                    READY   STATUS        RESTARTS   AGE
pod/numaresources-controller-manager-65b9bf8d6f-f5bx6   1/1     Running       0          29h
pod/numaresourcesoperator-worker-2fv87                  2/2     Terminating   0          33s
pod/numaresourcesoperator-worker-2xc2q                  2/2     Terminating   0          66s
pod/numaresourcesoperator-worker-ppdm5                  2/2     Running       0          24s
pod/numaresourcesoperator-worker-psx85                  2/2     Running       0          24s
pod/secondary-scheduler-75777d89cc-vqgcf                1/1     Running       0          22h
shajmakh@shajmakh-thinkpadt590 ~ $
```

### DIFF
--- a/test/e2e/serial/tests/tolerations.go
+++ b/test/e2e/serial/tests/tolerations.go
@@ -266,7 +266,9 @@ var _ = Describe("[serial][disruptive][slow][rtetols] numaresources RTE tolerati
 				By("deleting the DS to force the system recreate the pod")
 				// hack! remove the DS object to make the operator recreate the DS and thus restart all the pods
 				Expect(fxt.Client.Delete(ctx, updatedDs)).Should(Succeed())
-				By(fmt.Sprintf("checking that the DaemonSet is still ready - should match worker nodes count %d", len(workers)))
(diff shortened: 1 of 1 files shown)


shajmakh:
rebase on top of #900

PR #916, opened 2024-04-04
CVE-2023-45288: deps: bump x/net and related

ffromani:
Update x/net to pick up golang/net@ba87210
To fully fix we need to build with go >= 1.22.2.

xref: golang/go#65051


ffromani:
/hold

still evaluating the impact and the perf consequences

ffromani:
/hold cancel
we move ahead for main branch but we still on hold for the other branches

PR #918, opened 2024-04-06
e2e: tolerations: test CR and taint updates

shajmakh:
Add two tests to check that when updating taints or tolerations in a way that shakes the matching criteria the RTE pod  will get evicted.


ffromani (approved):
/lgtm

PR #930, opened 2024-04-18
images: streamline overhaul the image management

ffromani:
We bundle operator and operand (RTE) in the same image. The operator learn which is its own image by inspecting its own pod, and the user can override the value setting the corresponding field in the NodeConfig struct.

Since the beginning we wanted also to make the images configurable for development purposes. This feature never materialized.

In addition, the detection and image replacement code got messier over time, with no real cleanup.

To address both these needs, overhaul and simplify the image management. All changes are expected to
be transparent for regular users, outside development environments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,29 +138,31 @@ test-must-gather-e2e: build-must-gather-e2e
 ##@ Build
 
 binary: build-tools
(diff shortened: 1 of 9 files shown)


swatisehgal (commented):
/lgtm

PR #933, opened 2024-05-08
sched: api: generate config v1

ffromani:
The v1beta3 scheduler configuration is being phased out from the scheduler plugin repo (at least), so let's jump to v1. There are non functional changes anyway.


shajmakh:
/lgtm
/hold 
feel free to unhold when you see fit

ffromani:
/hold cancel
the sooner we merge the more time we have to verify

PR #934, opened 2024-05-20
e2e serial: initiate NRT list in each test independently

shajmakh:
Before this change, each test used the data of NRT list that was
gathered at the beginning of the suite and compared it with the NRT data gathered after each test to ensure the NRTs are settled. This is fine as long as the resources topology is guaranteed not to be updated in a way that messes up the teardown comparison.

Since we cannot specify the tools used to set up the resources in the
testing environment, hence no guarantee the topology would stay the same from the beginning of the suite till the end; we record the NRT data on each test fetching it directly from the cluster, independent from the data gathered at the start of the suite.

### DIFF
--- a/test/e2e/serial/config/fixture.go
+++ b/test/e2e/serial/config/fixture.go
@@ -66,7 +66,7 @@ var Config *E2EConfig
 
 func SetupFixture() error {
 	var err error
-	Config, err = NewFixtureWithOptions("e2e-test-infra", e2efixture.OptionRandomizeName|e2efixture.OptionAvoidCooldown)
+	Config, err = NewFixtureWithOptions("e2e-test-infra", e2efixture.OptionRandomizeName|e2efixture.OptionAvoidCooldown|e2efixture.OptionStaticClusterData)
 	return err
 }
 
--- a/test/e2e/serial/tests/resource_hostlevel.go
+++ b/test/e2e/serial/tests/resource_hostlevel.go
@@ -56,8 +56,6 @@ var _ = Describe("[serial] numaresources host-level resources", Serial, func() {
 
 	AfterEach(func() {
 		Expect(e2efixture.Teardown(fxt)).To(Succeed())
-		By("waiting for the NRT data to settle")
-		e2efixture.MustSettleNRT(fxt)
 	})
 
 	Context("with at least two nodes suitable", func() {
--- a/test/utils/fixture/fixture.go
+++ b/test/utils/fixture/fixture.go
@@ -41,6 +41,7 @@ import (
 	"github.com/openshift-kni/numaresources-operator/internal/objects"
 	intwait "github.com/openshift-kni/numaresources-operator/internal/wait"
 	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
+	"github.com/openshift-kni/numaresources-operator/test/utils/noderesourcetopologies"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -67,9 +68,10 @@ const (
 type Options uint
 
 const (
-	OptionNone          = 0
-	OptionRandomizeName = 1 << iota
-	OptionAvoidCooldown = 2 << iota
+	OptionNone              = 0
+	OptionRandomizeName     = 1 << iota
+	OptionAvoidCooldown     = 2 << iota
+	OptionStaticClusterData = 4 << iota
 )
 
 var (
@@ -89,18 +91,42 @@ func init() {
 }
 
 func SetupWithOptions(name string, nrtList nrtv1alpha2.NodeResourceTopologyList, options Options) (*Fixture, error) {
+	ctx := context.Background()
 	if !e2eclient.ClientsEnabled {
 		return nil, fmt.Errorf("clients not enabled")
 	}
 	randomizeName := (options & OptionRandomizeName) == OptionRandomizeName
 	avoidCooldown := (options & OptionAvoidCooldown) == OptionAvoidCooldown
+	staticClusterData := (options & OptionStaticClusterData) == OptionStaticClusterData
+	ginkgo.By("set up the test namespace")
 	ns, err := setupNamespace(e2eclient.Client, name, randomizeName)
 	if err != nil {
 		klog.Errorf("cannot setup namespace %q: %v", name, err)
 		return nil, err
 	}
-	ginkgo.By(fmt.Sprintf("set up the test namespace %q", ns.Name))
+	klog.Infof("test namespace %q was set up successfully", ns.Name)
+	if !staticClusterData {
+		ginkgo.By("pull NRT data before test starts")
+		var nrtAtTestSetup nrtv1alpha2.NodeResourceTopologyList
+		immediate := true
+		err := wait.PollUntilContextTimeout(ctx, 10*time.Second, 30*time.Second, immediate, func(ctx context.Context) (bool, error) {
+			err := e2eclient.Client.List(ctx, &nrtAtTestSetup)
+			return err == nil, nil
+		})
+		if err != nil {
+			klog.Errorf("failed to pull NRT items: %v", err)
+			return nil, err
+		}
 
+		ginkgo.By("verify the collected NRT data is similar to the data gathered in the beginning of the suite")
+		ok, _ := noderesourcetopologies.EqualNRTListsItems(nrtAtTestSetup, nrtList)
+		if !ok {
+			klog.Warning("WARNING! NRT MISMATCH:\n")
+			klog.Infof(intnrt.ListToString(nrtList.Items, "-----NRT at Suite Setup"))
+			klog.Infof(intnrt.ListToString(nrtAtTestSetup.Items, "-----NRT at Test Setup"))
+		}
+		nrtList = nrtAtTestSetup
+	}
 	klog.Infof("set up the fixture reference NRT List: %s", intnrt.ListToString(nrtList.Items, " fixture initial"))
 
 	return &Fixture{
--- a/test/utils/noderesourcetopologies/noderesourcetopologies.go
+++ b/test/utils/noderesourcetopologies/noderesourcetopologies.go
@@ -561,3 +561,22 @@ func ResourceInfoProvidingAtMost(resources []nrtv1alpha2.ResourceInfo, resName s
 	}
 	return true
 }
+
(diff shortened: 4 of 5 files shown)


shajmakh:
/hold 
for local tests to complete

test/e2e/serial/tests/resource_hostlevel.go
```diff
@@ -56,8 +56,6 @@ var _ = Describe("[serial] numaresources host-level resources", Serial, func() {
 
 	AfterEach(func() {
 		Expect(e2efixture.Teardown(fxt)).To(Succeed())
-		By("waiting for the NRT data to settle")
-		e2efixture.MustSettleNRT(fxt)
```
ffromani, line 60: thanks for this optimization

test/utils/fixture/fixture.go
```diff
@@ -101,6 +101,12 @@ func SetupWithOptions(name string, nrtList nrtv1alpha2.NodeResourceTopologyList,
 	}
 	ginkgo.By(fmt.Sprintf("set up the test namespace %q", ns.Name))
 
+	if len(nrtList.Items) == 0 {
+		gomega.Eventually(func(g gomega.Gomega) {
+			err = e2eclient.Client.List(context.TODO(), &nrtList)
+			g.Expect(err).ToNot(gomega.HaveOccurred())
+		}).WithTimeout(30*time.Second).WithPolling(10*time.Second).Should(gomega.Succeed(), "failed to pull NRT items")
+	}
```
ffromani, line 109: unfortunately this is not OK because the bug was to have the `ginkgo.By` in this function 2 lines above. We should not have assertions/expectations in this function. Actually, only functions prefixed with `Must` should have them. So let's convert this loop in a wait loop (example: `wait.PollUntilContextTimeout()`) and let's return error on failure

ffromani (commented):
changes here LGTM but I'd like more context about the root cause before to move forward.

test/utils/fixture/fixture.go
ffromani: If we fetch the nrtList before every suite, which I'd not be against, we should perhaps check the nrtList we just collected against the one we collected when everything started. If they differ, we should emit a fat big warning because this is not expected to happen. WDYT?

shajmakh: Thanks for your comments. I'll reposition ginkgo.by few lines above to align with the ns setup step.
And thanks for sharing the info on functions starting with Must. I'll update this part and adopt your suggestion.

test/utils/fixture/fixture.go
```diff
@@ -70,6 +71,7 @@ const (
 	OptionNone          = 0
 	OptionRandomizeName = 1 << iota
 	OptionAvoidCooldown = 2 << iota
+	OptionIsSuite       = 4 << iota
```
shajmakh, line 74: I was reluctant if these options are meant only for single tests but found it handy instead of passing another parameter to the setupWithOption to indicate a suite setup, @ffromani WDYT about keeping it?

test/utils/fixture/fixture.go
shajmakh: addressed in last commit

test/utils/fixture/fixture.go
ffromani: we need to review how use `iota` here (mostly my fault)

test/e2e/serial/config/fixture.go
```diff
@@ -66,7 +66,7 @@ var Config *E2EConfig
 
 func SetupFixture() error {
 	var err error
-	Config, err = NewFixtureWithOptions("e2e-test-infra", e2efixture.OptionRandomizeName|e2efixture.OptionAvoidCooldown)
+	Config, err = NewFixtureWithOptions("e2e-test-infra", e2efixture.OptionRandomizeName|e2efixture.OptionAvoidCooldown|e2efixture.OptionIsSuite)
```
ffromani, line 69: I'm ok with the concept but let's brainstorm a better name :)

test/utils/fixture/fixture.go
```diff
@@ -94,13 +96,63 @@ func SetupWithOptions(name string, nrtList nrtv1alpha2.NodeResourceTopologyList,
...
 	avoidCooldown := (options & OptionAvoidCooldown) == OptionAvoidCooldown
+	isSuiteSetup := (options & OptionIsSuite) == OptionIsSuite
+	ginkgo.By("set up the test namespace")
 	ns, err := setupNamespace(e2eclient.Client, name, randomizeName)
 	if err != nil {
 		klog.Errorf("cannot setup namespace %q: %v", name, err)
 		return nil, err
 	}
-	ginkgo.By(fmt.Sprintf("set up the test namespace %q", ns.Name))
+	klog.Infof("test namespace %q was set up successfully", ns.Name)
```
ffromani, line 106: this should be logged the same way as "set up the test namespace", so this should be a `ginkgo.By`

test/utils/fixture/fixture.go
```diff
@@ -94,13 +96,63 @@ func SetupWithOptions(name string, nrtList nrtv1alpha2.NodeResourceTopologyList,
...
+	klog.Info("Going Once .... \n\n\n")
```
ffromani, line 107: please remove debug logs eventually

test/utils/fixture/fixture.go
```diff
@@ -94,13 +96,63 @@ func SetupWithOptions(name string, nrtList nrtv1alpha2.NodeResourceTopologyList,
...
+		ginkgo.By("pull NRT data before test starts")
+		var nrtAtTestSetup nrtv1alpha2.NodeResourceTopologyList
+		immediate := true
+		err := wait.PollUntilContextTimeout(context.Background(), 10*time.Second, 30*time.Second, immediate, func(ctx context.Context) (bool, error) {
+			err := e2eclient.Client.List(context.TODO(), &nrtAtTestSetup)
+			if err != nil {
+				return false, nil
+			}
+			return true, nil
+		})
```
ffromani, line 118: good, better: 
```
err := wait.PollUntilContextTimeout(context.Background(), 10*time.Second, 30*time.Second, immediate, func(ctx context.Context) (bool, error) {
	err := e2eclient.Client.List(context.Background(), &nrtAtTestSetup)
        return err == nil, nil
})
```

even better: create a single `context` at the beginning of this function and use it consistenly

test/utils/fixture/fixture.go
```diff
@@ -94,13 +96,63 @@ func SetupWithOptions(name string, nrtList nrtv1alpha2.NodeResourceTopologyList,
...
+			klog.Info("Going loop .... \n")
+
+		}
+		if nrtMismatch {
+			klog.Warning("NRT MISMATCH:\n")
+			a, _ := yaml.Marshal(nrtList.Items)
+			klog.Infof("-----NRT at Suite Setup: \n%s\n", a)
+			b, _ := yaml.Marshal(nrtAtTestSetup.Items)
+			klog.Infof("-----NRT at Test Setup: \n%s\n", b)
+		}
```
ffromani, line 151: this should probably be put in a separate function. Also, please check functions in `internal/noderesourcetopology/tostring.go`

ffromani (commented):
thnaks for the changes. The direction is better, let's keep it

test/utils/fixture/fixture.go
ffromani, line 151: When we add this kind of logic, let's always evaluate if we can move it in a (new) function inside a package, like `internal/noderesourcetopologies`. This way we can add unit tests to the new logic, even if we use it initially only in tests.

test/e2e/serial/config/fixture.go
ffromani: Perhaps `OptionStaticClusterData` ? `OptionInitialClusterInfo`?

test/utils/fixture/fixture.go
shajmakh: thanks for catching. removed

test/utils/fixture/fixture.go
shajmakh: why? I understand ginkgo.By usually used to highlight the next step to be executed. Since I moved By to line 100 to declare that the test is about to set up the ns, when the setup is done, I log the info pointing out the ns name. I did the same for NRT data.
I usually like to use ginkgo.By to highlight milestones/main steps of the test/function and klog for additional details while running the step's code

ffromani: for consistency. We notify the begin of an action using `ginkgo.By`, which is fine and I agree with your way of using it, but we notify the end of the same action using `klog.Infof`.
Not a big deal though

test/utils/fixture/fixture.go
shajmakh: you're right. updated in latest updates, let me know if the place of the function is fine, details found in the commit msg.

ffromani (commented):
/lgtm

thanks for the updates! nice work!

test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -561,3 +561,22 @@ func ResourceInfoProvidingAtMost(resources []nrtv1alpha2.ResourceInfo, resName s
 	}
 	return true
 }
+
+func EqualNRTListsItems(nrtListA, nrtListB nrtv1alpha2.NodeResourceTopologyList) (bool, error) {
```
ffromani, line 565: optional followup: can this be placed or reimplemented in terms of code in `internal/noderesourcetopology/equality.go`?

test/utils/fixture/fixture.go
```diff
@@ -89,18 +91,42 @@ func init() {
 }
 
 func SetupWithOptions(name string, nrtList nrtv1alpha2.NodeResourceTopologyList, options Options) (*Fixture, error) {
+	ctx := context.Background()
 	if !e2eclient.ClientsEnabled {
 		return nil, fmt.Errorf("clients not enabled")
 	}
 	randomizeName := (options & OptionRandomizeName) == OptionRandomizeName
 	avoidCooldown := (options & OptionAvoidCooldown) == OptionAvoidCooldown
+	StaticClusterData := (options & OptionStaticClusterData) == OptionStaticClusterData
```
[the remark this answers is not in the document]
shajmakh: thanks for catching, fixed

test/utils/noderesourcetopologies/noderesourcetopologies.go
ffromani: completely agree. This is non-urgent and can wait a followup PR later on.

PR #936, opened 2024-05-21
e2e: WithOffset -> GinkgoHelper

ffromani:
rather than use WithOffset variants, which composes badly, let's leverage ginkgo.GinkgoHelper() to achieve the same result in a more composable and easier way.


test/e2e/serial/tests/workload_placement_taint.go
```diff
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/onsi/ginkgo/v2"
```
shajmakh, line 24: imported twice

ffromani: fixed, thanks

PR #937, opened 2024-05-23
e2e: hostlevel: cover all pod QoSes

ffromani:
extend the existing non-regression test coverage to all possible QoSes besides Guaranteed.

### DIFF
--- a/test/e2e/serial/tests/resource_hostlevel.go
+++ b/test/e2e/serial/tests/resource_hostlevel.go
@@ -143,6 +143,88 @@ var _ = Describe("[serial][hostlevel] numaresources host-level resources", Seria
 				},
 				corev1.PodQOSBestEffort,
 			),
+			Entry("[qos:gu] with ephemeral storage, multi-container",
+				intnrt.SingleNUMANode,
+				// required resources for the test pod
+				[]corev1.ResourceList{
+					{
+						corev1.ResourceCPU:              resource.MustParse("2"),
+						corev1.ResourceMemory:           resource.MustParse("256Mi"),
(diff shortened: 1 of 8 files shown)


test/e2e/serial/tests/resource_hostlevel.go
```diff
@@ -113,6 +151,16 @@ var _ = Describe("[serial] numaresources host-level resources", Serial, func() {
...
+				corev1.PodQOSBurstable,
+			),
+			Entry("[tmscope:cnt][qos:be] with ephemeral storage",
+				intnrt.SingleNUMANode,
+				intnrt.Container,
+				// required resources for the test pod
+				corev1.ResourceList{
+					corev1.ResourceEphemeralStorage: resource.MustParse("256Mi"),
+				},
+				corev1.PodQOSBestEffort,
```
shajmakh, line 163: since the tests are dealing with pod with one container, we can omit passing the tmscope. this way the test would be executed on all available configured tmscopes whenever the test is triggered.

shajmakh (commented):
Thanks for extending the coverage.
LGTM.
Small comment below. and did you get a chance to run the new tests?

ffromani:
keeping the hold because I need to run the tests locally

test/e2e/serial/tests/resource_hostlevel.go
ffromani: Ok, _really_ fixed now. Took me 3 reads to really get your point, sorry!

ffromani:
also verified locally, hold removed

shajmakh:
Thanks for the additional coverage.
/lgtm

PR #938, opened 2024-05-30
CNF-11823: build manager and exporter with strictfipsruntime flag

rbaturov:
As part of the for effort of achieving FIPS compliancy for NROP operator, this step is needed - both binaries must be built with the strictfipsruntime flag


ffromani:
/hold

let's consider this change after the branch-out of 4.16

PR #939, opened 2024-05-31
e2e: host-level: verify NRT updates

shajmakh:
Add a check to verify NRT is updated accordingly for gu pods or not
updated for non-gu pods.

### DIFF
--- a/internal/resourcelist/resourcelist.go
+++ b/internal/resourcelist/resourcelist.go
@@ -107,16 +107,23 @@ func SubCoreResources(res, resToSub corev1.ResourceList) error {
 	return nil
 }
 
-func Accumulate(ress []corev1.ResourceList) corev1.ResourceList {
+func Accumulate(ress []corev1.ResourceList, filterFunc func(corev1.ResourceName, resource.Quantity) bool) corev1.ResourceList {
 	ret := corev1.ResourceList{}
 	for _, rr := range ress {
 		for resName, resQty := range rr {
+			if !filterFunc(resName, resQty) {
+				continue
+			}
 			qty := ret[resName]
 			qty.Add(resQty)
 			ret[resName] = qty
 		}
 	}
 	return ret
+}
+
+func AllowAll(_ corev1.ResourceName, _ resource.Quantity) bool {
+	return true
 }
 
 func Equal(ra, rb corev1.ResourceList) bool {
--- a/internal/resourcelist/resourcelist.go
+++ b/internal/resourcelist/resourcelist.go
@@ -126,6 +126,19 @@ func AllowAll(_ corev1.ResourceName, _ resource.Quantity) bool {
 	return true
 }
 
+func FilterExclusive(resName corev1.ResourceName, resQty resource.Quantity) bool {
+	if resName == corev1.ResourceEphemeralStorage || resName == corev1.ResourceStorage {
+		return false
+	}
+
+	if resName == corev1.ResourceCPU {
+		if resQty.MilliValue()%1000 > 0 {
+			return false
+		}
+	}
+	return true
+}
+
 func Equal(ra, rb corev1.ResourceList) bool {
 	if len(ra) != len(rb) {
 		return false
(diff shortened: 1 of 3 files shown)


shajmakh:
/hold
pending updates to [unsched] cases

shajmakh:
/cc @ffromani 
just fyi for now, this is still pending more verification

internal/resourcelist/resourcelist.go
```diff
@@ -119,6 +120,28 @@ func Accumulate(ress []corev1.ResourceList) corev1.ResourceList {
...
+				}
+			}
+
+			qty := ret[resName]
+			qty.Add(resQty)
+			ret[resName] = qty
+		}
+	}
+	return ret
+}
```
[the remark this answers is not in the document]
shajmakh: This is a really nice idea, as it enables flexibility in resources accounting, yet I'd be in favor of it more when the filter function implementation is not too diverse and conditional. when we adopt it here, it seems to me that we are carrying half of the implementation in the filter function. I've added it anyway (in next update), let me know what you think.

test/e2e/serial/tests/resource_hostlevel.go
```diff
@@ -108,6 +113,31 @@ var _ = Describe("[serial][hostlevel] numaresources host-level resources", Seria
...
+				e2efixture.MustSettleNRT(fxt)
+
+				targetNrtInitial, err := e2enrt.FindFromList(nrtList.Items, updatedPod.Spec.NodeName)
+				Expect(err).NotTo(HaveOccurred())
+
+				if expectedQOS == corev1.PodQOSGuaranteed {
+					By("verify resources consumption is reflected in the NRTs")
+					exclusiveRequired := intreslist.AccumulateExclusiveForNRT(requiredRes)
+					klog.Infof("exclusive required resources: %+v", exclusiveRequired)
+					expectNRTConsumedResources(fxt, *targetNrtInitial, exclusiveRequired, updatedPod)
```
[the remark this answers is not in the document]
shajmakh: I tried to prefer clarity over brevity here, I'll fix it

shajmakh:
testing on both TM scopes is in progress

shajmakh:
passed on both scopes:

`
Ran 7 of 104 Specs in 423.545 seconds
SUCCESS! -- 7 Passed | 0 Failed | 0 Pending | 97 Skipped
`

test/e2e/serial/tests/resource_hostlevel.go
```diff
@@ -108,6 +108,32 @@ var _ = Describe("[serial][hostlevel] numaresources host-level resources", Seria
...
+						hostLevel := resName == corev1.ResourceEphemeralStorage
+						excCpus := true
+						if resName == corev1.ResourceCPU {
+							_, err := strconv.Atoi(resQty.String())
+							if err != nil && resQty.Value()%1000 > 0 {
+								excCpus = true
+							}
+						}
+						return !(hostLevel || excCpus)
+
```
ffromani, line 130: IIUC the intent:
```
if resName == corev1.ResourceEphemeralStorage {
    return false
}
if resName == corev1.ResourceCPU {
    cpuAmount, ok := cpuQty.AsInt64()
    if !ok {
        return false
    }
    if cpuAmount % 1000 > 0 {
        return false
    }
}
return true
```

ffromani (commented):
not completely sure about the intent of a code snippet, otherwise LGTM

test/e2e/serial/tests/resource_hostlevel.go
ffromani: Feel free to borrow code (and point out mistakes, if any) from https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/noderesourcetopology/resourcerequests/exclusive.go

shajmakh: yes that's the intention, I'll fix the confusion

PR #940, opened 2024-06-05
e2e: drop host-level resources when padding numa cells

shajmakh:
A problem was caught during workload placement tests requiring pre-padding numa zones as test setup. The problem was that the cluster had leftover workloads that consumed host-level resources and that was counted in the base load of the node. Later on, the test expected to pad Numa zones consider host-level resources to be part of the padder resources. The setup failed because host-level resources are not numa scoped.

Adjust the relevant methods that pad numa zones to filter specified resources such as ephemeral storage.

### DIFF
--- a/test/utils/noderesourcetopologies/noderesourcetopologies.go
+++ b/test/utils/noderesourcetopologies/noderesourcetopologies.go
@@ -225,8 +225,11 @@ func accumulateNodeAvailableResources(nrt nrtv1alpha2.NodeResourceTopology, reas
 	klog.Infof("resInfoList available %s: %s", reason, e2enrt.ResourceInfoListToString(resInfoList))
 	return resInfoList, nil
 }
-func SaturateZoneUntilLeft(zone nrtv1alpha2.Zone, requiredRes corev1.ResourceList) (corev1.ResourceList, error) {
+func SaturateZoneUntilLeft(zone nrtv1alpha2.Zone, requiredRes corev1.ResourceList, filter func(*corev1.ResourceList)) (corev1.ResourceList, error) {
 	paddingRes := make(corev1.ResourceList)
+
+	filter(&requiredRes)
+
 	for resName, resQty := range requiredRes {
 		zoneQty, ok := FindResourceAvailableByName(zone.Resources, string(resName))
 		if !ok {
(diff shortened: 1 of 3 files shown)


test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -225,8 +225,11 @@ func accumulateNodeAvailableResources(nrt nrtv1alpha2.NodeResourceTopology, reas
 	klog.Infof("resInfoList available %s: %s", reason, e2enrt.ResourceInfoListToString(resInfoList))
 	return resInfoList, nil
 }
-func SaturateZoneUntilLeft(zone nrtv1alpha2.Zone, requiredRes corev1.ResourceList) (corev1.ResourceList, error) {
+func SaturateZoneUntilLeft(zone nrtv1alpha2.Zone, requiredRes corev1.ResourceList, filter func(*corev1.ResourceList)) (corev1.ResourceList, error) {
```
ffromani, line 228: `corev1.ResourceList` is a map so no need to pass it by pointer. This is a language-level inconsistency we can only embrace and we can't fix

test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -246,6 +249,14 @@ func SaturateZoneUntilLeft(zone nrtv1alpha2.Zone, requiredRes corev1.ResourceLis
 	return paddingRes, nil
 }
 
+func AllowAll(res *corev1.ResourceList) {
+	return
+}
+
+func DropHostLevelRes(res *corev1.ResourceList) {
```
ffromani, line 256: please rename `DropHostLevelResources`

test/utils/noderesourcetopologies/noderesourcetopologies.go
```diff
@@ -225,8 +225,11 @@ func accumulateNodeAvailableResources(nrt nrtv1alpha2.NodeResourceTopology, reas
 	klog.Infof("resInfoList available %s: %s", reason, e2enrt.ResourceInfoListToString(resInfoList))
 	return resInfoList, nil
 }
-func SaturateZoneUntilLeft(zone nrtv1alpha2.Zone, requiredRes corev1.ResourceList) (corev1.ResourceList, error) {
+func SaturateZoneUntilLeft(zone nrtv1alpha2.Zone, requiredRes corev1.ResourceList, filter func(*corev1.ResourceList)) (corev1.ResourceList, error) {
 	paddingRes := make(corev1.ResourceList)
+
+	filter(&requiredRes)
```
ffromani, line 231: because this is a map, we should return the mutated `ResourceList` and use the mutated one

ffromani (commented):
I like the idea, but the implementation needs some refinement

test/utils/noderesourcetopologies/noderesourcetopologies.go
shajmakh: thanks for pointing out this, I have updated this block in https://github.com/openshift-kni/numaresources-operator/pull/942/commits/c4b438f3b2a9755c77ef7971f6a3f92ce95768ef because the proposed changes are needed by new tests

shajmakh:
closing this in favor https://github.com/openshift-kni/numaresources-operator/pull/942 because it is a prerequisite for some of the added checks

PR #813, opened 2024-01-10
tree: version: kickoff 4.16

ffromani:
trivial bump 4.15 -> 4.16

PR #946, opened 2024-06-18
tree: version: move to 4.17

ffromani:
start the 4.17 cycle


ffromani:
xref:

ffromani:
> xref: #813

fixed creating tags for 4.17 images on quay.io _before_ to do the test run, to break the catch-22 cycle. Proper builds will be automatically created once all bits fall into piece.

PR #949, opened 2024-07-01
numaresources: cleanup and modularize controller

ffromani:
internal cleanup of the reconciliation functions to enable future enhancements

xref: OCPBUGS-36314


Tal-or:
LGTM but tests are falling, i'll provide the label when tests will pass

PR #950, opened 2024-07-03
host-level and reboot-required tests stablization

shajmakh:
This PR goal is to fix common issues we see in the d/s CI. Please check the commits for more details.

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -202,38 +202,12 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
 				mcps, err := nropmcp.GetListByNodeGroupsV1(context.TODO(), e2eclient.Client, nroOperObj.Spec.NodeGroups)
 				Expect(err).ToNot(HaveOccurred())
 
-				var wg sync.WaitGroup
 				By("waiting for mcp to start updating")
-				for _, mcp := range mcps {
-					wg.Add(1)
-					klog.Infof("wait for mcp %q to start updating", mcp.Name)
-					go func(mcpool *machineconfigv1.MachineConfigPool) {
-						defer GinkgoRecover()
-						defer wg.Done()
-						err = wait.With(fxt.Client).
-							Interval(configuration.MachineConfigPoolUpdateInterval).
-							Timeout(configuration.MachineConfigPoolUpdateTimeout).
-							ForMachineConfigPoolCondition(context.TODO(), mcpool, machineconfigv1.MachineConfigPoolUpdating)
-						Expect(err).ToNot(HaveOccurred())
-					}(mcp)
-				}
-				wg.Wait()
+				waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdating)
 
 				By("wait for mcp to get updated")
-				for _, mcp := range mcps {
-					klog.Infof("wait for mcp %q to get updated", mcp.Name)
-					wg.Add(1)
-					go func(mcpool *machineconfigv1.MachineConfigPool) {
-						defer GinkgoRecover()
-						defer wg.Done()
-						err = wait.With(fxt.Client).
-							Interval(configuration.MachineConfigPoolUpdateInterval).
-							Timeout(configuration.MachineConfigPoolUpdateTimeout).
-							ForMachineConfigPoolCondition(context.TODO(), mcpool, machineconfigv1.MachineConfigPoolUpdated)
-						Expect(err).ToNot(HaveOccurred())
-					}(mcp)
-				}
-				wg.Wait()
+				waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdated)
+
 			}() //end of defer
 
 			// here we expect mcp-test and worker mcps to get updated.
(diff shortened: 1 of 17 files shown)


shajmakh:
/hold
for additional flaky reboot-required test's fix

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -277,7 +277,7 @@ var _ = Describe("[serial][disruptive][scheduler][resacct] numaresources workloa
 			By("waiting for the pod to be scheduled")
 			updatedPod, err := wait.With(fxt.Client).Timeout(3*time.Minute).ForPodPhase(context.TODO(), pod.Namespace, pod.Name, corev1.PodRunning)
 			if err != nil {
-				_ = objects.LogEventsForPod(fxt.K8sClient, updatedPod.Namespace, updatedPod.Name)
+				_ = objects.LogEventsForPod(fxt.K8sClient, pod.Namespace, pod.Name)
```
ffromani, line 280: why not use `updatedPod` here?

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -1369,15 +1369,12 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			Expect(err).ToNot(HaveOccurred())
 
 			By("verify the pod keep on pending")
-			err = wait.With(fxt.Client).Interval(10*time.Second).Steps(5).WhileInPodPhase(context.TODO(), pod.Namespace, pod.Name, corev1.PodPending)
+			_, err = wait.With(fxt.Client).Interval(10*time.Second).Steps(5).WhileInPodPhase(context.TODO(), pod.Namespace, pod.Name, corev1.PodPending)
```
ffromani, line 1372: I think we should get `updatedPod` here:
`updatedPod, err = wait.With(fxt.Client).Interval(10*time.Second).Steps(5).WhileInPodPhase(context.TODO(), pod.Namespace, pod.Name, corev1.PodPending)`

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -1401,14 +1398,14 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
...
+			}).WithTimeout(2*time.Minute).WithPolling(10*time.Second).Should(BeTrue(), "pod %s/%s doesn't contains the expected event error", pod.Namespace, pod.Name)
 
 			By("deleting the test pod")
-			err = fxt.Client.Delete(context.TODO(), updatedPod)
+			err = fxt.Client.Delete(context.TODO(), pod)
 			Expect(err).ToNot(HaveOccurred())
 
 			By("checking the test pod is removed")
-			err = wait.With(fxt.Client).Timeout(3*time.Minute).ForPodDeleted(context.TODO(), updatedPod.Namespace, updatedPod.Name)
+			err = wait.With(fxt.Client).Timeout(3*time.Minute).ForPodDeleted(context.TODO(), pod.Namespace, pod.Name)
```
ffromani, line 1408: all these changes seems unnecessary if we retrieve `updatedPod` like I suggested above. If we don't, I'm not sure why this code is more correct than before

test/e2e/serial/tests/configuration.go
```diff
@@ -906,3 +786,21 @@ func accumulateKubeletConfigNames(cms []corev1.ConfigMap) sets.Set[string] {
...
+			defer wg.Done()
+			err := wait.With(cli).
+				Interval(configuration.MachineConfigPoolUpdateInterval).
+				Timeout(configuration.MachineConfigPoolUpdateTimeout).
+				ForMachineConfigPoolCondition(ctx, mcpool, condition)
+			Expect(err).ToNot(HaveOccurred())
+		}(mcp)
+	}
+	wg.Wait()
+}
```
ffromani, line 806: thanks, this is nice. There's not a hard and fast rule about extracting code in test code, and I've been advocating for actually leaving copy/pasted code around (and I still am).
The reasons are mostly two:
- avoid subtle coupling: it's relatively common to have flows which are similar but different in tiny details. A helper function would need to have `if`s and flags. This tend to snowball overtime, leading to actually worse readability and maintainability. Rather than rush to extract helpers, usually better to step back and review the flow of code, and evaluate other solutions like tabular testing
- readability. Being able to follow the flow without jumping too much in/out functions is still a huge plus even in 2024 where AI-powered tools and powerful IDEs are a commodity.
The rule of thumb I'm following is when we have a really self-contained, well defined piece of code which embodies a check or a step which cannot really fail and/or than we heavily depend on, then it is a candidate to be factored out. Repetitiviness is also a factor: up to 2-3 times is OK

test/e2e/serial/tests/resource_accounting.go
shajmakh: because if there was an error for any reason we want to address a non empty pod definition, and "pod" here guranatees that. same reason for the rest of similar updates.

test/e2e/serial/tests/workload_placement_tmpol.go
shajmakh: there is no need for the updatedPod in the rest of the test, we mainly care about what ns and name of that pod to do the checks and it is the same for pod and updated one. if that was a deployment it would be different but for orphan pods I didn't see the need

test/e2e/serial/tests/configuration.go
shajmakh: Thanks for highlighting these useful points, I've updated the log to be more specific to the actual condition that this check is waiting for (line 794)

test/e2e/serial/tests/resource_accounting.go
ffromani: right, in the error path we should not use `updatedPod` indeed.

test/e2e/serial/tests/workload_placement_tmpol.go
ffromani: uhm, for my own personal taste it would be clearer to use the `updatedPod`, but this is indeed mostly a personal preference so let's keep it like you did

shajmakh:
47674 ran several times with the changes and passed. 
regarding host level tests they now run with enough waiting time for pending state:


```
I0708 18:16:53.417737   81744 images.go:27] using pause image: "registry.hlxcl9.lab.eng.tlv2.redhat.com:5000/numaresources-operator-tests:4.16.999-snapshot"
I0708 18:16:53.417795   81744 images.go:27] using pause image: "registry.hlxcl9.lab.eng.tlv2.redhat.com:5000/numaresources-operator-tests:4.16.999-snapshot"
I0708 18:16:58.449916   81744 pod.go:38] ensuring the pod e2e-test-resource-hostlevel-3505/testpod keep being in phase Pending 1/3
I0708 18:17:03.474927   81744 pod.go:38] ensuring the pod e2e-test-resource-hostlevel-3505/testpod keep being in phase Pending 2/3
I0708 18:17:08.500935   81744 pod.go:38] ensuring the pod e2e-test-resource-hostlevel-3505/testpod keep being in phase Pending 3/3
I0708 18:17:08.523658   81744 resource_hostlevel.go:330] pod e2e-test-resource-hostlevel-3505/testpod resources: cpu=3500m, ephemeral-storage=384Mi, memory=32Mi
```

/hold cancel

shajmakh:
@ffromani the PR should be ready, can you have another look please and let me know if you have more comments?

test/e2e/serial/tests/resource_accounting.go
```diff
@@ -659,7 +659,7 @@ var _ = Describe("[serial][disruptive][scheduler][resacct] numaresources workloa
 			// 3 minutes is plenty, should never timeout
 			updatedPod2, err := wait.With(fxt.Client).Timeout(3*time.Minute).ForPodPhase(context.TODO(), podGuanranteed.Namespace, podGuanranteed.Name, corev1.PodRunning)
 			if err != nil {
-				_ = objects.LogEventsForPod(fxt.K8sClient, updatedPod2.Namespace, updatedPod2.Name)
+				_ = objects.LogEventsForPod(fxt.K8sClient, podGuanranteed.Namespace, podGuanranteed.Name)
```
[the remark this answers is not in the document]
shajmakh: answered in https://github.com/openshift-kni/numaresources-operator/pull/950#discussion_r1668429757

test/e2e/serial/tests/workload_placement_tmpol.go
shajmakh: answered in https://github.com/openshift-kni/numaresources-operator/pull/950#discussion_r1668432441

ffromani (commented):
/lgtm
/approve

errgroup is nice! TIL!

PR #952, opened 2024-07-10
[release-4.15][manual] improve support to coexist with standard scheduler

ffromani:
manual backport of #801 to stable branch 4.15.

    react to any foreign pod (= pod not scheduled by this scheduler)
    consider all the pods when computing the expected node state

release-4.15 specific notes: please see commit https://github.com/openshift-kni/numaresources-operator/commit/21ab280d8a84647a5ff82fecd756da4202413391


ffromani:
/hold cancel

ready for review

test/e2e/serial/tests/scheduler_cache.go
```diff
@@ -611,13 +596,26 @@ var _ = Describe("[serial][scheduler][cache][tier1] scheduler cache", Label("sch
...
 
+func dumpFailedPodInfo(fxt *e2efixture.Fixture, failedPods []*corev1.Pod) {
+	if len(failedPods) == 0 {
+		return // not much to do here
+	}
+	nrtListFailed, _ := e2enrt.GetUpdated(fxt.Client, nrtv1alpha2.NodeResourceTopologyList{}, time.Minute)
+	klog.Infof("%s", e2enrtint.ListToString(nrtListFailed.Items, "post failure"))
+
+	for _, failedPod := range failedPods {
+		_ = objects.LogEventsForPod(fxt.K8sClient, failedPod.Namespace, failedPod.Name)
```
shajmakh, line 615: it would also be good to show the pod/container resources next to the nrt lists

controllers/numaresourcesscheduler_controller_test.go
```diff
@@ -421,7 +430,7 @@ var _ = ginkgo.Describe("Test NUMAResourcesScheduler Reconcile", func() {
 			_, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: key})
 			gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
-			expectCacheParams(reconciler.Client, depmanifests.InformerModeShared, depmanifests.CacheResyncAutodetect, depmanifests.ForeignPodsDetectAll)
+			expectCacheParams(reconciler.Client, depmanifests.CacheInformerDedicated, depmanifests.CacheResyncAutodetect, depmanifests.ForeignPodsDetectAll)
```
shajmakh, line 433: good catch

shajmakh (commented):
/lgtm

PR #954, opened 2024-07-10
[release-4.14][manual] improve support to coexist with standard scheduler

ffromani:
manual backport of #952  to stable branch 4.15.

- react to any foreign pod (= pod not scheduled by this scheduler)
- consider all the pods when computing the expected node state

release-4.14 specific notes: please see commit fa5ba29949a7120c33c80b5d0b3c2caf1140036d
release-4.14 specific notes: e2e tests not backported yet


shajmakh:
lgtm but let's wait for the ci to pass

ffromani:
Ci failures look legit. I'll setup a repro.

ffromani:
needs to be rebased on top of #846 once the latter merges

ffromani:
#846 dropped, changed accordingly. We should now be good to go

Tal-or:
/lgtm
trivial backport

PR #955, opened 2024-07-17
e2e: align tags and labels

ffromani:
reflect all tags but `testtype` and `test_id` into labels. From now on we should always add together labels and tags, if we add tags. Label is now the preferred way, but tags aren't forbidden yet, just deprecated.

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -70,7 +70,7 @@ import (
 	serialconfig "github.com/openshift-kni/numaresources-operator/test/e2e/serial/config"
 )
 
-var _ = Describe("[serial][disruptive][slow] numaresources configuration management", Serial, Label("disruptive", "slow"), func() {
+var _ = Describe("[serial][disruptive] numaresources configuration management", Serial, Label("disruptive"), func() {
 	var fxt *e2efixture.Fixture
 	var nrtList nrtv1alpha2.NodeResourceTopologyList
 	var nrts []nrtv1alpha2.NodeResourceTopology
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -70,7 +70,7 @@ import (
 	serialconfig "github.com/openshift-kni/numaresources-operator/test/e2e/serial/config"
 )
 
-var _ = Describe("[serial][disruptive][slow] numaresources configuration management", Serial, func() {
+var _ = Describe("[serial][disruptive][slow] numaresources configuration management", Serial, Label("disruptive", "slow"), func() {
 	var fxt *e2efixture.Fixture
 	var nrtList nrtv1alpha2.NodeResourceTopologyList
 	var nrts []nrtv1alpha2.NodeResourceTopology
@@ -105,7 +105,7 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
 	Context("cluster has at least one suitable node", func() {
 		timeout := 5 * time.Minute
 
-		It("[test_id:47674][reboot_required][slow][images][tier2] should be able to modify the configurable values under the NUMAResourcesOperator CR", func() {
+		It("[test_id:47674][reboot_required][slow][images][tier2] should be able to modify the configurable values under the NUMAResourcesOperator CR", Label("reboot_required", "slow", "images", "tier2"), func() {
 			fxt.IsRebootTest = true
 			nroOperObj := &nropv1.NUMAResourcesOperator{}
 			nroKey := objects.NROObjectKey()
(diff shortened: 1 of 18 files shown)


test/e2e/serial/tests/resource_hostlevel.go
```diff
@@ -65,7 +65,7 @@ var _ = Describe("[serial][hostlevel] numaresources host-level resources", Seria
 		// It should behave exactly like scope=pod. But we keep these tests as non-regression
 		// to have a signal the system is behaving as expected.
 		// This is the reason we don't filter for scope, but only by policy.
-		DescribeTable("[tier0][hostlevel] a pod should be placed and aligned on the node",
+		DescribeTable("[tier0][hostlevel] a pod should be placed and aligned on the node", Label("tier0"),
```
rshemtov13, line 68: hostlevel can be added to the label

test/e2e/serial/tests/resource_hostlevel.go
```diff
@@ -371,6 +380,7 @@ var _ = Describe("[serial][hostlevel] numaresources host-level resources", Seria
 				corev1.PodQOSBurstable,
 			),
 			Entry("[test_id:74255][tier3][qos:be][unsched] with ephemeral storage, multi-container",
+				Label("tier2", "qos:be", "unsched"),
```
rshemtov13, line 383: should be tier3 inside the Label not tier2

test/e2e/serial/tests/workload_placement_tmpol.go
```diff
@@ -791,6 +808,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			},
 		),
 		Entry("[tier1][testtype4][tmscope:container][hugepages2Mi] pod with two gu cnt land on a node with enough resources, containers should be spread on a different zone",
+			Label("tier1", "tmscope:cnt", "memory", "hugepages2Mi"),
```
rshemtov13, line 811: memory should be removed from this Label

rshemtov13 (changes requested):

test/e2e/serial/tests/resource_hostlevel.go
ffromani: yup, done. Thanks!

test/e2e/serial/tests/resource_hostlevel.go
ffromani: thanks, fixed

test/e2e/serial/tests/configuration.go
```diff
@@ -70,7 +70,7 @@ import (
 	serialconfig "github.com/openshift-kni/numaresources-operator/test/e2e/serial/config"
 )
 
-var _ = Describe("[serial][disruptive][slow] numaresources configuration management", Serial, func() {
+var _ = Describe("[serial][disruptive][slow] numaresources configuration management", Serial, Label("disruptive", "slow"), func() {
```
shajmakh, line 73: if we'd like to proceed with the labels as the preferred way we should enable interacting with them using the binary, like this: https://github.com/openshift-kni/numaresources-operator/pull/951/files (the shell script part only)

test/e2e/serial/tests/configuration.go
```diff
@@ -105,7 +105,7 @@ var _ = Describe("[serial][disruptive][slow] numaresources configuration managem
 	Context("cluster has at least one suitable node", func() {
 		timeout := 5 * time.Minute
 
-		It("[test_id:47674][reboot_required][slow][images][tier2] should be able to modify the configurable values under the NUMAResourcesOperator CR", func() {
+		It("[test_id:47674][reboot_required][slow][images][tier2] should be able to modify the configurable values under the NUMAResourcesOperator CR", Label("reboot_required", "slow", "images", "tier2"), func() {
```
shajmakh, line 108: some labels are duplicated in the nested specs (mentioned in Describe(..) and in It(..)) for example "slow", it sounds better to have one only for instance on the parent spec if all the tests under it are slow, or on the test spec if not all children tests are classified as slow.

shajmakh (commented):
Thank you for this!
two comments below

test/e2e/serial/tests/configuration.go
ffromani: yes, this can be done import your change in this PR or deferring the script changes to another PR. I'm fine both ways.

test/e2e/serial/tests/configuration.go
ffromani: This is true, but ginkgo is smart enough to de-duplicate labels, we can check with
```
./bin/e2e-nrop-serial.test -ginkgo.v -ginkgo.dry-run -ginkgo.no-color | less
```
I think it's fair to remove duplicate labels as we go, but if you feel it should be done in this PR I can do it as well

shajmakh: ack, no need for now then

shajmakh:
\lgtm

PR #958, opened 2024-07-18
e2e: manager: add features introspection

shajmakh:
As we proceed to maintain one tests image exported from the main branch,
some tests verify specific features that are active on
specific releases only.

Classify the tests of serial suite by features using ginkgo Labels on
specs so that later one can filter tests based on these features to run
only the active features per release.

### DIFF
--- a/.github/workflows/support-tools.yml
+++ b/.github/workflows/support-tools.yml
@@ -1,0 +1,43 @@
+name: Build support tools binaries
+
+on:
+  workflow_dispatch:
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout sources
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: setup golang
+        uses: actions/setup-go@v2
+        id: go
+        with:
+          go-version: 1.22.4
+
+  build:
+    needs: [setup]
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout sources
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: build tools' binaries
+        run: |
+          make bin/mkginkgolabelfilter
+          make bin/catkubeletconfmap
+
+      - name: release the binaries
+        uses: ncipollo/release-action@v1
+        with:
+          artifacts: bin/*
+          tag: support-tools
--- a/Makefile
+++ b/Makefile
@@ -386,7 +386,7 @@ goversion:
 build-tools: goversion bin/buildhelper bin/envsubst bin/lsplatform
 
 .PHONY: build-tools-all
-build-tools-all: goversion bin/buildhelper bin/envsubst bin/lsplatform bin/catkubeletconfmap bin/watchnrtattr
+build-tools-all: goversion bin/buildhelper bin/envsubst bin/lsplatform bin/catkubeletconfmap bin/watchnrtattr bin/mkginkgolabelfilter
 
 bin/buildhelper:
 	@go build -o bin/buildhelper tools/buildhelper/buildhelper.go
@@ -398,10 +398,15 @@ bin/lsplatform:
 	@go build -o bin/lsplatform tools/lsplatform/lsplatform.go
 
 bin/catkubeletconfmap:
-	@go build -o bin/catkubeletconfmap tools/catkubeletconfmap/catkubeletconfmap.go
+	LDFLAGS="-static"
+	CGO_ENABLED=0 go build -o bin/catkubeletconfmap -ldflags "$$LDFLAGS" tools/catkubeletconfmap/catkubeletconfmap.go
 
 bin/watchnrtattr:
 	@go build -o bin/watchnrtattr tools/watchnrtattr/watchnrtattr.go
+
+bin/mkginkgolabelfilter:
+	LDFLAGS="-static"
+	@go build -o bin/mkginkgolabelfilter -ldflags "$$LDFLAGS" tools/mkginkgolabelfilter/mkginkgolabelfilter.go
 
 verify-generated: bundle generate
 	@echo "Verifying that all code is committed after updating deps and formatting and generating code"
--- a/doc/features/README.md
+++ b/doc/features/README.md
@@ -1,0 +1,52 @@
+Features introspection in e2e tests
+======================================
+
+**Goal**
+
+Provide a way to identify the active features for every supported operator version.
+
+**Design**
+
+- Provide the possibility to group a set of tests that verify a common feature using ginkgo Label("feature:<topic>").
+  A Topic should describe the main feature being tested or the unique effect of the test on the cluster. For instance:
+  `feature:schedrst`: indicates that the test is expected to restart the schedular, be it a pod restart or a complete removal of the scheduler and recreation.
+  `feature:rtetols`: indicates that the test is testing the new feature RTE tolerations, which involves specific non-default configuration on the operator CRs hence functional effects.
+  `feature:unsched`: indicates that the test workload is anticipated to be un-schedulable for any reason for example: insufficient resources (pod is pending), TAE (failed)..
+  `feature:wlplacement`: indicates that the test challenges the scheduler to place a workload on a node. This is a common label for most of the functional tests.
+    - Rules for creating a new topic
+        - A topic should point out briefly the new feature being tested or the unique effect of the test on the cluster
+        - The topic should be lowercase
+        - For composite topics use `_` between the words, e.g no_nrt
+        - Should not consist of any of `&|!,()/`
+- Define the list of supported features for each version.
+- Allow the user to list the supported features by adding new flag, `-inspect-features` which if passed the numaresources-operator binary on the controller pod will list the active features for the deployed version of the operator.
+- For automated bugs' scenarios that their fix is not back-ported to all versions use keywords tags that briefly describes the bug's main check.
+
+**List active features**
+
+Once the operator is installed, perform the following on the controller manager pod:
+
+```azure
+oc exec -it numaresources-controller-manager-95dd55c6-k6428 -n openshift-numaresources -- /bin/numaresources-operator --inspect-features
+ ```
+
+Example output:
+```
+{"active":["config","nonreg","hostlevel","resacct","cache","stall","rmsched","rtetols","overhead","wlplacement","unsched","nonrt","taint","nodelabel","byres","tmpol"]}
+```
+
+**Use case example: run tests for supported features only**
+
+To build the filter label query of the supported features on a specific version, a binary called `mkginkgolabelfilter` exists in `releases/support-tools` which expects a JSON input showing the supported features from the operator controller pod as follows:
+
+To use the helper tool perform the following:
+
+```
+# curl -L -o mkginkgolabelfilter https://github.com/openshift-kni/numaresources-operator/releases/download/support-tools/mkginkgolabelfilter
+# chmod 755 mkginkgolabelfilter
+# ./mkginkgolabelfilter # This will wait to read the input which should be the output of the --inspect-features above
+{"active":["config","nonreg","hostlevel","resacct","cache","stall","rmsched","rtetols","overhead","wlplacement","unsched","nonrt","taint","nodelabel","byres","tmpol"]}
+feature: consistAny {config,nonreg,hostlevel,resacct,cache,stall,rmsched,rtetols,overhead,wlplacement,unsched,nonrt,taint,nodelabel,byres,tmpol}
+```
+
+Then later in the podman command of running the tests use `--filter-label` with the output of the tool to run tests of supported features only.  
--- a/internal/api/features/topics.go
+++ b/internal/api/features/topics.go
@@ -1,0 +1,31 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2024 Red Hat, Inc.
+ */
+
+package features
+
+import (
+	_ "embed"
+	"encoding/json"
+)
+
+//go:embed _topics.json
+var topicsData string
+
+func GetTopics() TopicInfo {
+	var tp TopicInfo
+	_ = json.Unmarshal([]byte(topicsData), &tp)
+	return tp
+}
--- a/internal/api/features/topics_test.go
+++ b/internal/api/features/topics_test.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package features
+
+import (
+	"strings"
+	"testing"
+
+	"k8s.io/utils/strings/slices"
+)
+
+func TestGetTopics(t *testing.T) {
+	type testcase struct {
+		name       string
+		topicsList string
+		isPositive bool
+	}
+	testcases := []testcase{
+		{
+			name:       "with list of well known topics that every release operator must support",
+			topicsList: "config,nonreg,hostlevel,resacct,cache,stall,schedrst,overhead,wlplacement,unsched,taint,nodelabel,byres,tmpol",
+			isPositive: true,
+		},
+		{
+			name:       "with inactive topics included",
+			topicsList: "config,nonreg,hostlevel,notfound,cache",
+			isPositive: false,
+		},
+	}
+
+	actual := GetTopics()
+	for _, tc := range testcases {
+		topicsStr := strings.TrimSpace(tc.topicsList)
+		topics := strings.Split(topicsStr, ",")
+		tp := findMissingTopics(actual.Active, topics)
+
+		if len(tp) > 0 && tc.isPositive {
+			t.Errorf("expected to include topic(s) %v but it didn't, list found: %v", tp, actual.Active)
+		}
+
+		if len(tp) == 0 && !tc.isPositive {
+			t.Errorf("active topics included unsupported topic(s) %v, list found: %v", tp, actual.Active)
+		}
+	}
+}
+
+func findMissingTopics(agianstList, sublist []string) []string {
+	missing := []string{}
+	for _, el := range sublist {
+		if !slices.Contains(agianstList, el) {
+			missing = append(missing, el)
+		}
+	}
+	return missing
+}
--- a/internal/api/features/types.go
+++ b/internal/api/features/types.go
@@ -1,0 +1,59 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2024 Red Hat, Inc.
+ */
+
+package features
+
+import (
+	"fmt"
+
+	goversion "github.com/aquasecurity/go-version/pkg/version"
+)
+
+const (
+	Version = "v4.17.0"
+)
+
+type Metadata struct {
+	// semantic versioning vX.Y.Z, e.g. v1.0.0
+	Version string `json:"version"`
+}
+
+type TopicInfo struct {
+	Metadata Metadata `json:"metadata"`
+	// unsorted list of topics active (supported)
+	Active []string `json:"active"`
+}
+
+func NewTopicInfo() TopicInfo {
+	return TopicInfo{
+		Metadata: Metadata{
+			Version: Version,
+		},
+	}
+}
+
+func (tp TopicInfo) Validate() error {
+	if tp.Metadata.Version == "" {
+		return fmt.Errorf("metadata: missing version")
+	}
+	if _, err := goversion.Parse(tp.Metadata.Version); err != nil {
+		return fmt.Errorf("metadata: malformed version: %w", err)
+	}
+	if len(tp.Active) == 0 {
+		return fmt.Errorf("missing active topics")
+	}
+	return nil
+}
--- a/main.go
+++ b/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
@@ -55,6 +56,7 @@ import (
 	nropv1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1"
 	nropv1alpha1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1alpha1"
 	"github.com/openshift-kni/numaresources-operator/controllers"
+	"github.com/openshift-kni/numaresources-operator/internal/api/features"
 	"github.com/openshift-kni/numaresources-operator/pkg/hash"
 	"github.com/openshift-kni/numaresources-operator/pkg/images"
 	rteupdate "github.com/openshift-kni/numaresources-operator/pkg/objectupdate/rte"
@@ -117,6 +119,7 @@ type Params struct {
 	enableHTTP2           bool
 	enableMCPCondsForward bool
 	image                 ImageParams
+	inspectFeatures       bool
 }
 
 func (pa *Params) SetDefaults() {
@@ -132,6 +135,7 @@ func (pa *Params) FromFlags() {
 	flag.StringVar(&pa.platformName, "platform", pa.platformName, "platform to deploy on - leave empty to autodetect")
 	flag.StringVar(&pa.platformVersion, "platform-version", pa.platformVersion, "platform version to deploy on - leave empty to autodetect")
 	flag.BoolVar(&pa.detectPlatformOnly, "detect-platform-only", pa.detectPlatformOnly, "detect and report the platform, then exits")
+	flag.BoolVar(&pa.inspectFeatures, "inspect-features", pa.inspectFeatures, "outputs the supported features as JSON, then exits")
 	flag.BoolVar(&pa.renderMode, "render", pa.renderMode, "outputs the rendered manifests, then exits")
 	flag.BoolVar(&pa.render.NRTCRD, "render-nrt-crd", pa.render.NRTCRD, "outputs only the rendered NodeResourceTopology CRD manifest, then exits")
 	flag.StringVar(&pa.render.Namespace, "render-namespace", pa.render.Namespace, "outputs the manifests rendered using the given namespace")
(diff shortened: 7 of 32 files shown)


test/e2e/serial/tests/tolerations.go
```diff
@@ -51,7 +51,7 @@ import (
 	serialconfig "github.com/openshift-kni/numaresources-operator/test/e2e/serial/config"
 )
 
-var _ = Describe("[serial][disruptive][rtetols] numaresources RTE tolerations support", Serial, Label("disruptive", "rtetols"), func() {
+var _ = Describe("[serial][disruptive][rtetols] numaresources RTE tolerations support", Serial, Label("disruptive"), Label("feature:rtetols"), func() {
```
shajmakh, line 54: made by mistake while resolving conflicts, will restore the label

pkg/features/_topics.json
```diff
@@ -0,0 +1,20 @@
...
+    "rtetols",
+    "overhead",
+    "wlplacement",
+    "unsched",
+    "nonrt",
+    "taint",
+    "nodelabel",
+    "byres",
+    "tmpol"
+  ]
```
ffromani, line 19: we need to sketch a schema for this document. Would you mind start a (VERY) simple design doc? should stay in the `doc/` folder of this repo.
At very minimum we need some forward-looking features like document versioning, documentation of the fields and a way to deprecate stuff

pkg/features/types.go
```diff
@@ -0,0 +1,27 @@
...
+
+type Topics struct {
+	Active []string `json:"active"`
+}
+
+func (tp Topics) Clone() Topics {
+	return Topics{
+		Active: append([]string{}, tp.Active...),
+	}
+}
```
ffromani, line 27: let's add a unit test here

pkg/features/topics.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+//go:embed _topics.json
+var topicsData string
+
+func GetTopics() Topics {
+	var tp Topics
+	_ = json.Unmarshal([]byte(topicsData), &tp)
+	return tp
+}
```
ffromani, line 31: let's add a unit test here

.github/workflows/ginkgo-label-filter-tool.yml
```diff
@@ -0,0 +1,41 @@
...
+    needs: [setup]
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout sources
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: build ginkgo label-filter generator tool
+        run: make bin/mkginkgolabelfilter
```
ffromani, line 36: we probably need static buillds (will require Makefile changes).
In addition, we should probably release other tools, let's generalize this workflow to cover generically "support tools"

pkg/features/_topics.json
shajmakh: right, will handle that

.github/workflows/support-tools.yml
```diff
@@ -0,0 +1,41 @@
...
+          fetch-depth: 0
+
+      - name: build tools' binaries
+        run: make build-tools-all
+
+      - name: release the binaries
+        uses: ncipollo/release-action@v1
+        with:
+          artifacts: bin/*
+          tag: support-tools
```
ffromani, line 41: did we check that this captures only the support tools and not unnecessary binaries?

Makefile
```diff
@@ -389,22 +389,28 @@ build-tools: goversion bin/buildhelper bin/envsubst bin/lsplatform
 build-tools-all: goversion bin/buildhelper bin/envsubst bin/lsplatform bin/catkubeletconfmap bin/watchnrtattr bin/mkginkgolabelfilter
 
 bin/buildhelper:
-	@go build -o bin/buildhelper tools/buildhelper/buildhelper.go
+	LDFLAGS="-static"
+	CGO_ENABLED=0 go build -o bin/buildhelper tools/buildhelper/buildhelper.go
```
ffromani, line 393: this is used only by the build process itself and should not be released.
Plus: how is `LDFLAGS` consumed? implicitely?

Makefile
```diff
@@ -389,22 +389,28 @@ build-tools: goversion bin/buildhelper bin/envsubst bin/lsplatform
...
 
 bin/envsubst:
-	@go build -o bin/envsubst tools/envsubst/envsubst.go
+	LDFLAGS="-static"
+	CGO_ENABLED=0 go build -o bin/envsubst tools/envsubst/envsubst.go
```
ffromani, line 397: ditto

Makefile
```diff
@@ -389,22 +389,28 @@ build-tools: goversion bin/buildhelper bin/envsubst bin/lsplatform
...
 
 bin/lsplatform:
-	@go build -o bin/lsplatform tools/lsplatform/lsplatform.go
+	LDFLAGS="-static"
+	CGO_ENABLED=0 go build -o bin/lsplatform tools/lsplatform/lsplatform.go
```
ffromani, line 401: this may be useful outside the build process, so OK to release. Ditto for `LDFLAGS`.

Makefile
```diff
@@ -389,22 +389,28 @@ build-tools: goversion bin/buildhelper bin/envsubst bin/lsplatform
...
 
 bin/catkubeletconfmap:
-	@go build -o bin/catkubeletconfmap tools/catkubeletconfmap/catkubeletconfmap.go
+	LDFLAGS="-static"
+	CGO_ENABLED=0 go build -o bin/catkubeletconfmap tools/catkubeletconfmap/catkubeletconfmap.go
```
ffromani, line 405: this may be useful outside the build process, so OK to release. Ditto for `LDFLAGS`.

Makefile
```diff
@@ -389,22 +389,28 @@ build-tools: goversion bin/buildhelper bin/envsubst bin/lsplatform
...
 
 bin/watchnrtattr:
-	@go build -o bin/watchnrtattr tools/watchnrtattr/watchnrtattr.go
+	LDFLAGS="-static"
+	CGO_ENABLED=0 go build -o bin/watchnrtattr tools/watchnrtattr/watchnrtattr.go
```
ffromani, line 409: this is a demo tool which should not be released (yet). Ditto for `LDFLAGS`.

doc/features/README.md
```diff
@@ -0,0 +1,39 @@
+Classifying tests with feature labels
```
ffromani, line 1: this is good but too narrowly focused on the (main) use case. We should focus on the features inspection alone and document the purpose of the  inspection and document the expected schema of the output. Only then we can illustrate how this information can be consumed with the testing use case.

pkg/features/topics.go
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2024 Red Hat, Inc.
+ */
+
+package features
```
ffromani, line 17: let's move this code under `internal/api/features` and let's document this is the first lightweight internal (semi-private) API. For this reason is not under `API`, not meant to be consumed but from tools inside this project (e.g. brave 3rd parties, if ever, are on their own with no guarantees).

pkg/features/types.go
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2024 Red Hat, Inc.
+ */
+
+package features
+
+type Topics struct {
```
ffromani, line 19: Let's add a Metadata struct with at least the `Version` in it. Plus, let's rename this `TopicInfo`. API objects should _preferably_ be singular unless they are collections (this is not)

sketch:
```
const (
	Version = "v1.0.0"
)

type Metadata struct {
	// semantic versioning vX.Y.Z, e.g. v1.0.0
	Version string `json:"version"`
}

type TopicInfo struct {
	Metadata Metadata `json:"metadata"`
	// unsorted list of topics active (supported)
	Active []string `json:"active"`
}

func NewTopicInfo() TopicInfo {
	return TopicInfo{
		Metadata: Metadata{
			Version: Version,
		},
	}
}
```

pkg/features/types.go
```diff
@@ -0,0 +1,30 @@
...
+	Active []string `json:"active"`
+}
```
ffromani, line 21: once we added `Metadata` (which we need), we can add a `func (tp TopicInfo) Validate() error` function (with tests)

pkg/features/_topics.json
```diff
@@ -0,0 +1,20 @@
+{
+  "active": [
+    "config",
```
ffromani, line 3: let's also document what a "topic" is and give some guidelines about creating new topics (we can chat offline about this)

Makefile
shajmakh: removed unneeded tools here and below, and fixed the ldflags, which was missing in the go build cmd. thanks for catching

.github/workflows/support-tools.yml
shajmakh: thanks for the comment, from what I [tested](https://github.com/shajmakh/numaresources-operator/releases/tag/support-tools) it should release only binaries that were built using the imported make target. It would be interesting to make sure though. I agree we should not assume that though, so I updated that to point exactly to the 2 helper tools that can be released.

ffromani: explicitely allowlisting seems to be the best option, thanks for updating

shajmakh:
cc @rshemtov13

test/e2e/tools/tools_test.go
```diff
@@ -48,5 +63,77 @@ var _ = Describe("[tools] Auxiliary tools", func() {
 			_, ok := platform.ParsePlatform(text)
 			Expect(ok).To(BeTrue(), "cannot recognize detected platform: %s", text)
 		})
+
+		It("[api][mkginkgolabelfilter] should expose correct active features and create filter", Label("api", "mkginkgolabelfilter"), func(ctx context.Context) {
```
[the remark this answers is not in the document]
ffromani: note the `mkginkgolabelfilter` tests can be added later. So the change I'm suggesting here is meant to _reduce_ the scope of this PR rather than expand it further

pkg/objectnames/objectnames.go
```diff
@@ -19,8 +19,10 @@ package objectnames
 import "fmt"
 
 const (
-	DefaultNUMAResourcesOperatorCrName  = "numaresourcesoperator"
-	DefaultNUMAResourcesSchedulerCrName = "numaresourcesscheduler"
+	DefaultNUMAResourcesOperatorCrName   = "numaresourcesoperator"
+	DefaultNUMAResourcesSchedulerCrName  = "numaresourcesscheduler"
+	DefaultNUMAResourcesControllerDpName = "numaresources-controller-manager"
```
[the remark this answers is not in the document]
ffromani: xref: #963

test/e2e/tools/tools_test.go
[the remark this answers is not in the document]
shajmakh: thanks for your comment, I did thought about splitting the API check and the tool one but I would need to duplicate the API check so combined that as one test. Your suggestion sounds better though so ill adopt it. thanks!

test/e2e/tools/tools_test.go
```diff
@@ -48,5 +63,77 @@ var _ = Describe("[tools] Auxiliary tools", func() {
...
+			Expect(err).ToNot(HaveOccurred(), "err=%v", err)
+			Expect(stderr).ToNot(HaveOccurred(), "stderr=%s", stderr)
+			Expect(stdoutVersion).ToNot(BeEmpty())
+			klog.Infof("deployed version: %s", string(stdoutVersion))
+
+			re := regexp.MustCompile("numaresources-operator\\s+([0-9]+.[0-9]+.[0-9]+)\\s+")
+			match := re.FindStringSubmatch(string(stdoutVersion))
+			Expect(len(match)).To(Equal(1), "different pattern of version was found:\n%s", string(stdoutVersion))
+			tp.Metadata.Version = match[0]
+			klog.Infof("version to validate: %s", tp.Metadata.Version)
```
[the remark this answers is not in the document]
shajmakh: I want to validate the version of the nrop binary vs the version of the expected features info

pkg/objectnames/objectnames.go
```diff
@@ -19,8 +19,10 @@ package objectnames
...
+	NUMAResourcesNamespace               = "openshift-numaresources"
```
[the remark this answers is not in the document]
shajmakh: ack will use #963  to discover this instead, thanks for raising this

test/e2e/tools/tools_test.go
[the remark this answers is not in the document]
ffromani: fine for me, but let's split this very specific validation to its own test, let's not sneak into the main test

test/e2e/tools/tools_test.go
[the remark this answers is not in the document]
shajmakh: regarding the first point,  if we want to go with local binary we’ll miss the version check so either we drop its check in the validate method or set it manually to match the expected and use validate as it is, which is better than completely dropping it as we are consuming binaries of the same code version, I'll adjust this anyway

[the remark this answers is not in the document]
ffromani: I'm not against running against the cluster if it is helpful. Let's try to require cluster in the tools lane _if we can_, but if cluster access is needed, so be it.

main.go
```diff
@@ -56,7 +56,7 @@ import (
 	nropv1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1"
 	nropv1alpha1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1alpha1"
 	"github.com/openshift-kni/numaresources-operator/controllers"
-	"github.com/openshift-kni/numaresources-operator/pkg/features"
+	"github.com/openshift-kni/numaresources-operator/internal/api/features"
```
ffromani, line 59: bad rebase? :)

test/e2e/tools/mkginkgolabelfilter_test.go
```diff
@@ -158,7 +158,7 @@ var _ = Describe("[tools][mkginkgolabelfilter] Auxiliary tools", Label("tools",
 })
 
 func getListFromGinkgoQuery(q string) ([]string, error) {
-	re := regexp.MustCompile(`consistAny\\s+{(.*)*}`)
+	re := regexp.MustCompile(`consistAny\s+\{(.*)*\}`)
```
ffromani, line 161: bad rebase?

test/e2e/tools/tools_test.go
```diff
@@ -17,6 +17,8 @@
 package tools
 
 import (
```
ffromani, line 19: `controller-manager` is the ugly default name for operator-sdk projects that yours truly forgot to change initially and now we're kinda stuck with it. Let's call it "operator" or "manager" if we feel fancy. Please **DO NOT** resubmit just to fix this silly nit I'm raising

test/e2e/tools/tools_test.go
```diff
@@ -25,12 +27,16 @@ import (
...
 	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
+
+	"github.com/openshift-kni/numaresources-operator/internal/api/features"
 )
 
-var _ = Describe("[tools] Auxiliary tools", func() {
+var _ = Describe("[tools] Auxiliary tools", Label("tools"), func() {
 	Context("with the binary available", func() {
-		It("[lsplatform] lsplatform should detect the cluster", func() {
+		It("[lsplatform] lsplatform should detect the cluster", Label("inspectfeatures"), func() {
```
ffromani, line 39: this specific test is not pertaining inspect features though (it predates the change)

test/e2e/tools/tools_test.go
```diff
@@ -48,5 +54,30 @@ var _ = Describe("[tools] Auxiliary tools", func() {
 			_, ok := platform.ParsePlatform(text)
 			Expect(ok).To(BeTrue(), "cannot recognize detected platform: %s", text)
 		})
+
+		It("[api][manager][inspectfeatures] should expose correct active features", Label("api", "inspectfeatures"), func(ctx context.Context) {
+			By("inspect active features from local binary")
+			cmdline := []string{
+				filepath.Join(BinariesPath, "manager"),
+				"-inspect-features",
```
ffromani, line 62: nit (please DO NOT resubmit just to fix this): albeit `golang` apps accept both ways, let's use `--double-dashes` for long options and `-s`ingle dash for short options

internal/api/features/topics.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+	"strings"
+)
+
+//go:embed _topics.json
+var topicsData string
+
+func GetTopics() TopicInfo {
+	tp := NewTopicInfo()
+	var tmp TopicInfo
```
ffromani, line 31: you can unmarshal straight into `tp`

internal/api/features/topics_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"encoding/json"
+	"os"
+	"reflect"
+	"testing"
+)
+
+func TestGetTopics(t *testing.T) {
+	var expected, actual TopicInfo
+
+	topicsData, err := os.ReadFile("_topics.json")
```
ffromani, line 29: bad test, because we mess with the internal implementation details (which we should not).
I'd just test the "active" topic list includes well known topics which every released operator must support. I'd actually hardcode  that list in the test code.

internal/api/features/topics.go
shajmakh: indeed, fixed

test/e2e/tools/mkginkgolabelfilter_test.go
shajmakh: fixed, thanks!

test/e2e/tools/tools_test.go
shajmakh: by mistake, the label should be lsplatform

test/e2e/tools/tools_test.go
shajmakh: right indeed, I followed the help, I'll fix it

shajmakh: right indeed, I followed the help, I'll fix that

test/e2e/tools/tools_test.go
shajmakh: no problem, fixed, thanks for raising

internal/api/features/topics_test.go
shajmakh: done, thanks

test/e2e/tools/mkginkgolabelfilter_test.go
```diff
@@ -0,0 +1,174 @@
...
+	. "github.com/onsi/gomega"
+
+	"k8s.io/klog/v2"
+
+	"github.com/openshift-kni/numaresources-operator/internal/api/features"
+)
+
+const FeatureKey = "feature"
+
+var _ = Describe("[tools][mkginkgolabelfilter] Auxiliary tools", Label("tools", "mkginkgolabelfilter"), Ordered, func() {
```
ffromani, line 41: do we need `Ordered` here?

test/e2e/tools/mkginkgolabelfilter_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+		It("should create a filter that matches the a valid input", func(ctx context.Context) {
+			input := "{\"active\":[" +
+				"\"foo\"," +
+				"\"bar\"," +
+				"\"foobar\"" +
+				"]}"
+
+			var inputTp features.TopicInfo
+			_ = json.Unmarshal([]byte(input), &inputTp)
```
ffromani, line 60: this is not wrong, but e2e test _preferably_ should not depend on library code. I'd actually hardcode the expected input/output pairs, human-crafted and/or human-validated

test/e2e/tools/mkginkgolabelfilter_test.go
```diff
@@ -0,0 +1,174 @@
...
+			expectedToolOutput, _ := features.BuildFilterConsistAny(FeatureKey, inputTp.Active)
+			expectedToolOutput = fmt.Sprintf("%s\n", expectedToolOutput)
+
+			klog.Infof("running: %v\n", cmdline)
+			toolCmd := exec.Command(cmdline[0])
+			r, w := io.Pipe()
+			toolCmd.Stdin = r
```
ffromani, line 67: I'm pretty sure you don't _need_ a pipe and you can just feed a buffer to the Stdin

test/e2e/tools/mkginkgolabelfilter_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+			topics, err := getListFromGinkgoQuery(string(out))
+			Expect(err).ToNot(HaveOccurred())
+			sort.Strings(topics)
+			sort.Strings(inputTp.Active)
+			Expect(reflect.DeepEqual(topics, inputTp.Active)).To(BeTrue(), "different active topics are displayed in the query:\n%v\nexpected:\n%v", topics, inputTp.Active)
+		})
+
+		It("should fail on an invalid json format", func(ctx context.Context) {
+			input := "(\"active\"=[" +
```
ffromani, line 87: let's use \`\` to simplify the input: https://www.digitalocean.com/community/tutorials/an-introduction-to-working-with-strings-in-go#string-literals

doc/features/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+- Provide the possibility to group a set of tests that verify a common feature using ginkgo Label("feature:<topic>").
+  A Topic should describe the main feature being tested or the unique effect of the test on the cluster. For instance:
+  `feature:schedrst`: indicates that the test is expected to restart the schedular, be it a pod restart or a complete removal of the scheduler and recreation.
+  `feature:rtetols`: indicates that the test is testing the new feature RTE tolerations, which involves specific non-default configuration on the operator CRs hence functional effects.
+  `feature:unsched`: indicates that the test workload is anticipated to be un-schedulable for any reason for example: insufficient resources (pod is pending), TAE (failed)..
+  `feature:wlplacement`: indicates that the test challenges the scheduler to place a workload on a node. This is a common label for most of the functional tests.
+    - Rules for creating a new topic
+        - A Topic should point out briefly the new feature being tested or the unique effect of the test on the cluster
+        - The topic should consist of small letters
```
ffromani, line 18: I think you mean `lowercase`?

doc/features/README.md
```diff
@@ -0,0 +1,64 @@
...
+        - should not consist of any of `&|!,()/`
```
ffromani, line 19: let's add a rule for composite words. Is it gonna be `composite_word`? or `composite-word`? or maybe `compositeWord`?

doc/features/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+- Define the list of supported features for each version: this is documented in a json format in pkg/features/_topics.json under the corresponding branch.
```
ffromani, line 21: the `_topics.json` should be an implementation detail consumers of the API should not depend on (and possibly even be aware of)

doc/features/README.md
```diff
@@ -0,0 +1,64 @@
...
+  Similarly, pausing refreshing the NRT info is supported starting 4.16, and such tests are associated with `Label("feature:rtetols")` should not run below 4.16.
+
+- Allow the user to list the supported features by adding new flag, `-inspect-features` which if passed the numaresources-operator binary on the controller pod will list the active features for the deployed version of the operator.
+
+- For automated bugs' scenarios that their fix is not back-ported to all versions:
+    - **TBD**
+
+**Deprecation**
+
+Tagging tests to a specific feature functionality were done by adding tags in the spec description like `It("[rtetols][distruptive]...")`. Starting Jul 2024, while the master branch is pointing to 4.17, the new preferred way to tag tests is using ginkgo `Label("")` such as `Context("should run pods requesting host-level resources", Label("hostlevel","distruptive"), func(){..})`. Tags aren't forbidden yet; they are just deprecated, so whenever a tag is added, a label is also required.
```
ffromani, line 33: this is true and correct, but doesn't belong to the design of the features inspection

doc/features/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+```azure
+oc exec -it numaresources-controller-manager-95dd55c6-k6428 -n openshift-numaresources -- /bin/sh 
+```
+
+Then from within the pod run the binary with the required flag:
+
+```
+sh-5.1$ bin/numaresources-operator --inspect-features
+{"active":["config","nonreg","hostlevel","resacct","cache","stall","rmsched","rtetols","overhead","wlplacement","unsched","nonrt","taint","nodelabel","byres","tmpol"]}
```
ffromani, line 47: good.
Better:
```
oc exec -it numaresources-controller-manager-95dd55c6-k6428 -n openshift-numaresources -- /bin/numaresources-operator --inspect-features
```

internal/api/features/topics.go
```diff
@@ -0,0 +1,40 @@
...
+	_ = json.Unmarshal([]byte(topicsData), &tp)
+	return tp
+}
+
+func BuildFilterConsistAny(key string, values []string) (string, error) {
+	if strings.TrimSpace(key) == "" {
+		return "", fmt.Errorf("empty key: %q", key)
+	}
+	return fmt.Sprintf("%s: consistAny {%s}", key, strings.Join(values, ",")), nil
+}
```
ffromani, line 40: this is fine, but doesn't belong in the `api` package. The `api` package must include the API types definition and the most essential support code (e.g. validation, deepcopy...)

internal/api/features/types.go
```diff
@@ -0,0 +1,61 @@
...
+		return actual, fmt.Errorf("version mismatch: %+v vs expected %+v", tp.Metadata, actual.Metadata)
+	}
+
+	sort.Strings(tp.Active)
+	sort.Strings(actual.Active)
+	if !reflect.DeepEqual(tp.Active, actual.Active) {
+		return actual, fmt.Errorf("features mismatch: %+v vs expected %+v", tp.Active, actual.Active)
+	}
+
+	return actual, nil
```
ffromani, line 60: this validates semantically, which is too strict. I think we're better served about syntax/scheme validation. The semantics should not be validated this way, because there's more liberty in the semantic layer (e.g. multiple correct representations of the same state).
I'd do something like
```
...
import goversion "github.com/aquasecurity/go-version/pkg/version"
...
func (tp TopicInfo) Validate() error {
	if tp.Metadata.Version == "" {
		return fmt.Errorf("metadata: missing version")
	}
	if _, err := goversion.Parse(tp.Metadata.Version); err != nil {
		return fmt.Errorf("metadata: malformed version: %w", err)
	}
	if len(tp.Active) == 0 {
		return fmt.Errorf("missing active topics")
	}
	return nil
}
```

ffromani (commented):
Almost there! I think this is the latest round of changes needed.
Nice work so far!

internal/api/features/topics.go
ffromani: If the problem is testing, or making testing easier, we can still solve the problem convenientely using e2e tests for the `mkginkgolabelfilter` binary. The binary is going to be very easy and simple to test (no options! no state! can run locally!) so e2e tests will be fast and convenient.

shajmakh: makes sense, the tool e2e tests cover that indeed, which is the core of the tool. I'll remove this from here

test/e2e/tools/mkginkgolabelfilter_test.go
shajmakh: because there were multiple It()s that I wanted to run this inside each one, so running it inside beforeAll requires Ordered container. I adjusted that part of the code anyhow so now we have one test

test/e2e/tools/mkginkgolabelfilter_test.go
shajmakh: note taken, done. thanks!

test/e2e/tools/mkginkgolabelfilter_test.go
shajmakh: done, thanks for this!

internal/api/features/types.go
shajmakh: I thought we need to check validation against the OCP version and verify the active topics are as embedded. good to learn this, thank you for pointing it out!

doc/features/README.md
shajmakh: thought about this but since we've been using no separators (for example hostlevel,nonrt, rtetols..) I didn't include that. allowing a new rule while it doesn't apply to all the current available topics feels inconsistent, so we will need to update them to fulfill the rule, which is fine but want to confirm if this is the rule we want to proceed with and apply to all topics

ffromani: it's ok to update later. The current state is sadly inconsistent anyway (mostly my fault) so we will need revisions in the future.

ffromani: it is also OK, actually better, to fix all the topics in a consistent way. It's more work, but starting from clean slate is better if we are willing to endure it.

shajmakh: got it, I'll add this and address the needed adjustments in another commit/PR

shajmakh:
@ffromani Thank you for your valuable comments and time for review, the comments should be addressed now, I greatly appreciate if you can have another pass and let me know if there are more enhancements to do or if we are good to go

ffromani:
/approve
/lgtm

thanks @shajmakh for the great work and for addressing all the feedback. There are some minor improvements here and there and some nits, but let's address them in followup PRs.

PR #959, opened 2024-07-18
makefile: build serial bin with race detector

shajmakh:
Tests in serial often fails and there is no clear reason from the logs, although elaborated, on why the failure happened.

Build the serial suite binary with race detector so that when tests run a warning is shown pointing to occured race.

This is meant to help debug the failing tests but not actually fixing them or preventing race conditions.

xref: https://go.dev/doc/articles/race_detector


shajmakh:
/hold 
would like to test prow ci with similar change

ffromani:
/approve

I'm all for using more tools, including race detector, but I'm skeptical the issue is a race inside the test binary.

shajmakh:
ci fails because:
`go: -race requires cgo; enable cgo by setting CGO_ENABLED=1`
in fact for the d/s ci we need cgo disabled, so this is not an option

PR #961, opened 2024-07-24
e2e: add test scheduler picks TM policy update without need to restart

rshemtov13:
This test add automation coverage for OCPBUGS-34583

The numaresources secondary-scehduler doesn't account for topology-manager new configuration unless it gets deleted and restarted again. This means TAE can still occur if topology manager policy was updated to single-numa-node and the scheduler not restarted after that update.
This test checks that scenario by ensuring the TM policy is set to none then restarting the scheduler, setting back the TM policy to single-numa-node and then creating a TAE deployment to check if the deployment is pending

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -36,6 +37,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	k8swait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
+	"k8s.io/kubelet/config/v1beta1"
 
 	"sigs.k8s.io/yaml"
 
@@ -50,6 +52,7 @@ import (
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 
 	nropmcp "github.com/openshift-kni/numaresources-operator/internal/machineconfigpools"
+	"github.com/openshift-kni/numaresources-operator/internal/podlist"
 	"github.com/openshift-kni/numaresources-operator/internal/relatedobjects"
 	"github.com/openshift-kni/numaresources-operator/pkg/kubeletconfig"
 	rteconfig "github.com/openshift-kni/numaresources-operator/rte/pkg/config"
@@ -63,6 +66,7 @@ import (
 	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
 	"github.com/openshift-kni/numaresources-operator/test/utils/configuration"
 	e2efixture "github.com/openshift-kni/numaresources-operator/test/utils/fixture"
+	"github.com/openshift-kni/numaresources-operator/test/utils/images"
 	e2enrt "github.com/openshift-kni/numaresources-operator/test/utils/noderesourcetopologies"
 	"github.com/openshift-kni/numaresources-operator/test/utils/nrosched"
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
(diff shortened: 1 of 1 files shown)


test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
 				return kcCmNamesCur
 			}).WithContext(ctx).WithTimeout(5 * time.Minute).WithPolling(10 * time.Second).Should(Equal(kcCmNamesPre))
 		})
+
+		It("[reboot_required][slow] should be able to correctly identify topology manager policy without scheduler restarting", Label("reboot_required", "slow"), func() {
```
shajmakh, line 701: let's also add tags/labels of unsched and schedrst

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			nroOperObj := &nropv1.NUMAResourcesOperator{}
```
shajmakh, line 702: let's add a By() step here for fetching the matching kubeletconfig for the numaresources-operator CR

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+					}
+				}
+			}
+			Expect(targetedKC).ToNot(BeNil(), "there should be at least one kubeletconfig.machineconfiguration object")
+
+			kcObj, err := kubeletconfig.MCOKubeletConfToKubeletConf(targetedKC)
+			Expect(err).ToNot(HaveOccurred())
+			initialTopologyManagerPolicy := kcObj.TopologyManagerPolicy
+
+			By("modifying Topology Manager Policy under kubeletconfig to none")
```
shajmakh, line 732: just a side note, no action needed now, the kubeletconfig can be owned by a performanceprofile (d/s we support both) this need to be adjusted to detect on which object it should be done. I'll post a PR soon for this

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+
+				applyNewTopologyManagerPolicy(&kcObj.TopologyManagerPolicy, v1beta1.NoneTopologyManagerPolicy)
+				err = kubeletconfig.KubeletConfToMCKubeletConf(kcObj, targetedKC)
+				Expect(err).ToNot(HaveOccurred())
+
+				err = fxt.Client.Update(context.TODO(), targetedKC)
+				g.Expect(err).ToNot(HaveOccurred())
+			}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(Succeed())
+			waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdated)
+
```
shajmakh, line 748: after any kubeletconfig update the nodes reboot so need to wait for mcp to get updated, basically similar to what you did in lines 773-778.

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			defer func() {
+				By("reverting kubeletconfig changes")
+				kcObj.TopologyManagerPolicy = initialTopologyManagerPolicy
```
shajmakh, line 751: not sure why is this needed as kcObj is overridden in 756

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+
+				By("waiting for mcp to start updating")
+				waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdating)
+
+				By("wait for mcp to get updated")
+				waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdated)
+			}()
+
+			var schedulerName string
+			nroSchedObj := &nropv1.NUMAResourcesScheduler{}
```
shajmakh, line 781: defining the type `var nroSchedObj &nropv1.NUMAResourcesSchedule` can be also be enough

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			schedPods, err := podlist.With(fxt.Client).ByDeployment(context.TODO(), *schedDp)
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(len(schedPods)).To(Equal(1))
+
+			schedulerName = nroSchedObj.Status.SchedulerName
+			Expect(schedulerName).ToNot(BeEmpty(), "cannot autodetect the TAS scheduler name from the cluster")
+
+			// restart the secondary-scheduler pod
+			By(fmt.Sprintf("deleting the NRO Scheduler object: %s", nroSchedObj.Name))
```
shajmakh, line 798: the test is actually deleting the pod

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			nrosched.CheckNROSchedulerAvailable(fxt.Client, nroSchedObj.Name)
+			err = fxt.Client.Delete(context.TODO(), &schedPods[0])
+			Expect(err).ToNot(HaveOccurred())
+			nrosched.CheckNROSchedulerAvailable(fxt.Client, nroSchedObj.Name)
```
shajmakh, line 802: the scheduler object will still be there what we need to make sure of is that the new pod is restarted and is running

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+
+			By("modifying Topology Manager Policy under kubeletconfig to none")
+			Eventually(func(g Gomega) {
+				err = fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(targetedKC), targetedKC)
+				Expect(err).ToNot(HaveOccurred())
+
+				kcObj, err = kubeletconfig.MCOKubeletConfToKubeletConf(targetedKC)
+				Expect(err).ToNot(HaveOccurred())
+
+				applyNewTopologyManagerPolicy(&kcObj.TopologyManagerPolicy, v1beta1.NoneTopologyManagerPolicy)
```
shajmakh, line 740: let's add a comment why to change to "none" specifically or it can be other policy?

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			// getting the number of cpus that is required for a numa zone to make a TAE deployment
+			const (
+				NUMAZonesRequired     = 2
+				hostsRequired         = 2
+				cpuResourcePercentage = 1.2 // 1.2 meaning 120 percent
+			)
+			var nrtCandidates []nrtv1alpha2.NodeResourceTopology
+			By(fmt.Sprintf("filtering available nodes with at least %d NUMA zones", NUMAZonesRequired))
+			nrtCandidates = e2enrt.FilterZoneCountEqual(nrtList.Items, NUMAZonesRequired)
+			Expect(len(nrtCandidates)).ToNot(BeNumerically("<", hostsRequired))
```
shajmakh, line 830: this can be moved to the very first of the test so if not enough nodes with 2 numas the test should be skipped thus avoid updating the kubeletcofig

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+
+				err = fxt.Client.Update(context.TODO(), targetedKC)
+				g.Expect(err).ToNot(HaveOccurred())
+			}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(Succeed())
+			waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdated)
+
+			// getting the number of cpus that is required for a numa zone to make a TAE deployment
+			const (
+				NUMAZonesRequired     = 2
+				hostsRequired         = 2
```
shajmakh, line 824: for this test one node is also enough

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+				cpuResourcePercentage = 1.2 // 1.2 meaning 120 percent
+			)
+			var nrtCandidates []nrtv1alpha2.NodeResourceTopology
+			By(fmt.Sprintf("filtering available nodes with at least %d NUMA zones", NUMAZonesRequired))
+			nrtCandidates = e2enrt.FilterZoneCountEqual(nrtList.Items, NUMAZonesRequired)
+			Expect(len(nrtCandidates)).ToNot(BeNumerically("<", hostsRequired))
+
+			referenceNode := nrtCandidates[0]
```
shajmakh, line 832: choosing the first node is fine considering all nodes should be similar in topology let's just add a comment about that to explain the reason behind our decision

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+				applyNewTopologyManagerPolicy(&kcObj.TopologyManagerPolicy, v1beta1.SingleNumaNodeTopologyManagerPolicy)
+				err = kubeletconfig.KubeletConfToMCKubeletConf(kcObj, targetedKC)
+				Expect(err).ToNot(HaveOccurred())
+
+				err = fxt.Client.Update(context.TODO(), targetedKC)
+				g.Expect(err).ToNot(HaveOccurred())
+			}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(Succeed())
+			waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdated)
+
+			// getting the number of cpus that is required for a numa zone to make a TAE deployment
```
shajmakh, line 821: let's make this a By()

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			referenceNode := nrtCandidates[0]
+			referenceZone := referenceNode.Zones[0]
+			cpuQty, ok := e2enrt.FindResourceAvailableByName(referenceZone.Resources, string(corev1.ResourceCPU))
+			Expect(ok).To(BeTrue(), "no CPU resource in zone %q node %q", referenceZone.Name, referenceNode.Name)
+
+			cpuNum, ok := cpuQty.AsInt64()
+			Expect(ok).To(BeTrue(), "invalid CPU resource in zone %q node %q: %v", referenceZone.Name, referenceNode.Name, cpuQty)
+
+			cpuResources := strconv.Itoa(int(float64(cpuNum) * cpuResourcePercentage))
+
```
shajmakh, line 841: lets log out the avialable per numa on the selected node vs what is the cpuResources to be requested in the dp. add that as klog.Infof()

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,195 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+				time.Sleep(10 * time.Second)
+
+				By(fmt.Sprintf("ensuring the deployment %q keep being pending %d/%d", deployment.Name, step+1, maxStep))
+
+				err = fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(deployment), updatedDeployment)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(wait.IsDeploymentComplete(deployment, &updatedDeployment.Status)).To(BeFalse(), "deployment %q become ready", deployment.Name)
+			}
+		})
```
shajmakh, line 857: the deployment might be unready/pending because several reasons we want to make sure that the reason here is TAE, you can use something like https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/serial/tests/workload_unschedulable.go#L308

test/e2e/serial/tests/configuration.go
shajmakh: also let's logout the current policy and if not single-numa-node let's report a warning , and if the same as `none` then break with a klog.Info log

shajmakh (commented):
Thanks for this! 
added few comments below for initial review

test/e2e/serial/tests/configuration.go
rshemtov13: okay np

test/e2e/serial/tests/configuration.go
rshemtov13: got it

test/e2e/serial/tests/configuration.go
rshemtov13: yes you're right thanks for that

test/e2e/serial/tests/configuration.go
rshemtov13: ok np

test/e2e/serial/tests/configuration.go
rshemtov13: got it

test/e2e/serial/tests/configuration.go
rshemtov13: yes thanks

test/e2e/serial/tests/configuration.go
ffromani, line 701: let's make sure this is `[tagged]` and `Label()`ed `disruptive`, as it restarts the kubelet

ffromani, line 701: please use `It("description", Label(), func(ctx context.Context) { ...`  aka let gingko pass a context, and use it everywhere in the test function body, so we don't need to sprinkle `context.TODO()` everywhere

test/e2e/serial/tests/configuration.go
ffromani, line 740: this `applyNewTopologyManagerPolicy` is a one-liner, so i'd drop it and inline the assignment everywhere

test/e2e/serial/tests/configuration.go
rshemtov13: that's right moved it to the top section

test/e2e/serial/tests/configuration.go
rshemtov13: np

ffromani: this check should be done at the beginning of the test code, to avoid as much as we can any unnecessary churn and keep the test execution time at bay

test/e2e/serial/tests/configuration.go
rshemtov13: np

test/e2e/serial/tests/configuration.go
rshemtov13: got it

ffromani (commented):
this is an impressive first contribution, kudos

test/e2e/serial/tests/configuration.go
rshemtov13: It can be any other policy but making it none and then changing to single-numa-node and then creating a tae deployment is the best option in my opinion to check this scenario, but it can be any policy though

test/e2e/serial/tests/configuration.go
rshemtov13: All the tests in this file is under 1 Describe and it already has disruptive should I put another it also under It() or its considered a duplicate?

ffromani: no, if the parent `Describe` is already tagged and labeled `disruptive` then we're good.

test/e2e/serial/tests/configuration.go
rshemtov13: yes np

test/e2e/serial/tests/configuration.go
rshemtov13: ok got it

test/e2e/serial/tests/configuration.go
rshemtov13: yes fixed

test/e2e/serial/tests/configuration.go
rshemtov13: The topology policy at this point will be `single-numa-node` so the pod will keep on pending if the scheduler got the change correctly so we are expecting here for the pod to be pending because there is not enough cpus on a numa level to schedule the pod and it will fail on failed scheduling rather than TAE so there's no need to check that but there is a need to check for `FailedScheduling`.

shajmakh: that's right indeed, thanks for pointing this out

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,215 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			By("getting the number of cpus that is required for a numa zone to make a TAE deployment")
+			const (
+				NUMAZonesRequired     = 2
+				hostsRequired         = 1
+				cpuResourcePercentage = 1.5 // 1.5 meaning 150 percent
+			)
+			var nrtCandidates []nrtv1alpha2.NodeResourceTopology
+			By(fmt.Sprintf("filtering available nodes with at least %d NUMA zones", NUMAZonesRequired))
+			nrtCandidates = e2enrt.FilterZoneCountEqual(nrtList.Items, NUMAZonesRequired)
+			Expect(len(nrtCandidates)).ToNot(BeNumerically("<", hostsRequired))
```
shajmakh, line 711: let's make the assertion here a Skip instead of failing. It is okay that the environment is not relevant for a test, we'ld just skip instead of failing, for example:
https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/serial/tests/resource_accounting.go#L111

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,215 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			Eventually(func(g Gomega) {
+				err = fxt.Client.Get(ctx, client.ObjectKeyFromObject(targetedKC), targetedKC)
+				Expect(err).ToNot(HaveOccurred())
+
+				kcObj, err = kubeletconfig.MCOKubeletConfToKubeletConf(targetedKC)
+				Expect(err).ToNot(HaveOccurred())
+
+				// Before restarting the scheduler, we change the topology manager policy to 'none'.
+				// This ensures that if we later change it to 'single NUMA node' and attempt to create a deployment that should fail on TAE,
+				// the deployment will fail as expected. If the policy were still set to 'none', it would allow the deployment to proceed.
```
shajmakh, line 768: let's clarify here "..create a deployment that should fail on TAE," that we expect that fail to happen only when the bug occur

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,215 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+				err = kubeletconfig.KubeletConfToMCKubeletConf(kcObj, targetedKC)
+				Expect(err).ToNot(HaveOccurred())
+
+				err = fxt.Client.Update(ctx, targetedKC)
+				g.Expect(err).ToNot(HaveOccurred())
+			}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(Succeed())
+
+			// TODO find a solution to wait for mcp to start updating only if the tm policy
+			// has changed otherwise it will be stuck here
+			waitForMcpUpdate(fxt.Client, ctx, mcps)
```
shajmakh, line 779: you can verify that even before by checking if `initialTopologyManagerPolicy != NoneTopologyManagerPolicy` is true only then execute the eventually and wait for the update to happen

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,215 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+
+			pod := &pods[0]
+			Expect(pod.Status.Phase).To(Equal(corev1.PodPending))
+
+			schedulerName = nroSchedObj.Status.SchedulerName
+			Expect(schedulerName).ToNot(BeEmpty(), "cannot autodetect the TAS scheduler name from the cluster")
+
+			isFailed, err := nrosched.CheckPODSchedulingFailed(fxt.K8sClient, pod.Namespace, pod.Name, schedulerName)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(isFailed).To(BeFalse(), "pod %s/%s with scheduler %s failed to schedule", pod.Namespace, pod.Name, schedulerName)
```
shajmakh, line 882: CheckPODSchedulingFailed returns true if the pod indeed failed. if it returned false it either means there was an error or that the pod went running, which then the test should fail: https://github.com/openshift-kni/numaresources-operator/blob/main/test/utils/nrosched/nrosched.go#L88

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,215 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
 				return kcCmNamesCur
 			}).WithContext(ctx).WithTimeout(5 * time.Minute).WithPolling(10 * time.Second).Should(Equal(kcCmNamesPre))
 		})
+
+		It("[reboot_required][slow][unsched][schedrst] should be able to correctly identify topology manager policy without scheduler restarting", Label("reboot_required", "slow", "unsched", "schedrst"), func(ctx context.Context) {
```
shajmakh, line 701: for reboot tests we need to add an additional flag at the beginning of the test to avoid falling on NRT mismatch after reboot. this is a limitation in which kernel might distribute the resources (mainly memory) differently after node reboot. see here for example: https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/serial/tests/configuration.go#L109

test/e2e/serial/tests/configuration.go
rshemtov13: yes you're right

test/e2e/serial/tests/configuration.go
rshemtov13: np will do

test/e2e/serial/tests/configuration.go
rshemtov13: got it fixed

test/e2e/serial/tests/configuration.go
rshemtov13: got it ! added it

Tal-or (commented):
IIUC this test should verify a behavior that was considered to be a bug and now it was fixed right?
If this is the case can we add a reference to the original bug?

test/e2e/serial/tests/configuration.go
rshemtov13: right thanks!

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,219 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+				Eventually(func(g Gomega) {
+					err = kubeletconfig.KubeletConfToMCKubeletConf(kcObj, targetedKC)
+					Expect(err).ToNot(HaveOccurred())
+
+					// changing the topology manager policy back to the initial state when the test has finished or failed
+					kcObj.TopologyManagerPolicy = initialTopologyManagerPolicy
+
+					err = fxt.Client.Update(ctx, targetedKC)
+					g.Expect(err).ToNot(HaveOccurred())
+				}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(Succeed())
```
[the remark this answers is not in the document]
rshemtov13: understood and addressed!

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,216 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+						kcObj, err = kubeletconfig.MCOKubeletConfToKubeletConf(targetedKC)
+						Expect(err).ToNot(HaveOccurred())
+
+						err = kubeletconfig.KubeletConfToMCKubeletConf(kcObj, targetedKC)
+						Expect(err).ToNot(HaveOccurred())
+
+						// changing the topology manager policy back to the initial state when the test has finished or failed
+						kcObj.TopologyManagerPolicy = initialTopologyManagerPolicy
+
+						err = fxt.Client.Update(ctx, targetedKC)
```
shajmakh, line 801: this code is making the required change (`kcObj.TopologyManagerPolicy = initialTopologyManagerPolicy`) after converting the kcObj to targetedKC which is not going to make any difference at all in the targetedKC, lines 798-799 should be switched with 792-793, same for similar instances

rshemtov13: you're right! thanks!

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,215 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			schedPods, err := podlist.With(fxt.Client).ByDeployment(ctx, *schedDeployment)
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(len(schedPods)).To(Equal(1))
+
+			schedulerName = nroSchedObj.Status.SchedulerName
+			Expect(schedulerName).ToNot(BeEmpty(), "cannot autodetect the TAS scheduler name from the cluster")
+
+			// deleting the NRO scheduler
+			By(fmt.Sprintf("deleting the NRO Scheduler object: %s", nroSchedObj.Name))
```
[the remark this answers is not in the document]
rshemtov13: okay np

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,215 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+
+	podLabels := map[string]string{
+		"test": "test-dp",
+	}
+	nodeSelector := map[string]string{}
+	deployment := objects.NewTestDeployment(replicas, podLabels, nodeSelector, fxt.Namespace.Name, name, images.GetPauseImage(), []string{images.PauseCommand}, []string{})
+	deployment.Spec.Template.Spec.SchedulerName = schedulerName
+	deployment.Spec.Template.Spec.Containers[0].Resources.Limits = corev1.ResourceList{
+		corev1.ResourceCPU:    resource.MustParse(cpus),
+		corev1.ResourceMemory: resource.MustParse("16Gi"),
```
shajmakh, line 899: from the run logs looks like this is the reason the pod is unschedualable:

```

I0729 11:01:06.244631 3703411 nrosched.go:84] -> found relevant scheduling event for pod e2e-test-configuration-7485/testdp-85f4c8ccf7-v6v4h: [type="Warning" action="Scheduling" message="0/5 nodes are available: 2 Insufficient memory, 3 node(s) had untolerated taint {node-role.kubernetes.io/master: }. preemption: 0/5 nodes are available: 2 No preemption victims found for incoming pod, 3 Preemption is not helpful for scheduling."
```
let's adjust this to 256Mi

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,215 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(pods)).To(Equal(1))
+
+			pod := &pods[0]
+			Expect(pod.Status.Phase).To(Equal(corev1.PodPending))
+
+			schedulerName = nroSchedObj.Status.SchedulerName
+			Expect(schedulerName).ToNot(BeEmpty(), "cannot autodetect the TAS scheduler name from the cluster")
+
+			isFailed, err := nrosched.CheckPODSchedulingFailed(fxt.K8sClient, pod.Namespace, pod.Name, schedulerName)
```
shajmakh, line 880: to avoid false negative we can replace this general failure function with something like https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/serial/tests/resource_hostlevel.go#L340 but using the right message about insufficient cpus, you can check that from the failing pod's events

test/e2e/serial/tests/configuration.go
rshemtov13: Got it! Changed.

test/e2e/serial/tests/configuration.go
rshemtov13: got it np, will change that

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,220 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+
+					err = fxt.Client.Update(ctx, targetedKC)
+					g.Expect(err).ToNot(HaveOccurred())
+				}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(Succeed())
+
+				waitForMcpUpdate(fxt.Client, ctx, mcps)
+			}
+
+			defer func() {
+				err = fxt.Client.Get(ctx, client.ObjectKeyFromObject(targetedKC), targetedKC)
```
shajmakh, line 787: let's use local vars as much as possible in the implemented functions of defer and eventually like err and kcObj, for example the first initialization of each is associated with := or

rshemtov13:
after PR is merged will add test_id

test/e2e/serial/tests/configuration.go
rshemtov13: Has been changed thank you!

shajmakh:
Thank you for addressing all the comments and for all the efforts on this!
/lgtm
/approve
/hold
for @ffromani 's ack or others review

test/e2e/serial/tests/configuration.go
```diff
@@ -693,9 +697,222 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			// choosing the first node because the nodes should be similar in topology so it doesnt matter which one we choose
+			referenceNode := nrtCandidates[0]
+			referenceZone := referenceNode.Zones[0]
+
+			cpuQty, ok := e2enrt.FindResourceAvailableByName(referenceZone.Resources, string(corev1.ResourceCPU))
+			Expect(ok).To(BeTrue(), "no CPU resource in zone %q node %q", referenceZone.Name, referenceNode.Name)
+
+			cpuNum, ok := cpuQty.AsInt64()
+			Expect(ok).To(BeTrue(), "invalid CPU resource in zone %q node %q: %v", referenceZone.Name, referenceNode.Name, cpuQty)
+			klog.Infof("available CPUs per numa on the node: %d, ", cpuNum)
```
ffromani, line 726: nit: let's use https://kubernetes.io/blog/2020/09/04/kubernetes-1-19-introducing-structured-logs/ in new code (Infof -> InfoS)

ffromani (commented):
/lgtm
/unhold

thanks, very nice first contribution!

test/e2e/serial/tests/configuration.go
rshemtov13: np will do!

PR #962, opened 2024-07-24
e2e: adapt performance profile existance

shajmakh:
Considering that d/s ci runs on env with PP configured
according to the RDS, when the test updates kubeletconfig directly, which is owned by another object, the update is not applied; hence, the test fails.

Extend the test to detect if the kubeletconfig belongs to a
performanceprofile, and if so update the performance profile instead so that the updates actually apply.

This is the first time NTO dependencies are needed so also update the
dependencies. Note that the NTO version used in the dependencies is from 4.16 branch because starting 4.17 NTO is consuming Go 1.22, and using the latest NTO will require bumping the Go version for NROP as well, which is something we are not clear on yet, thus sticking with 1.21, since all we need is just use performance profile type.

So far the serial suite has 2 tests that apply update to the
kubeletconfig. This PR targets one test (id-47585), another PR for
the other will follow.


shajmakh:
/hold
for deps to be merged:
https://github.com/openshift-kni/numaresources-operator/pull/968: to get the new version of the mcp update method
https://github.com/openshift-kni/numaresources-operator/pull/971: to get the bumped deps, so expected less dependencies update here

test/e2e/serial/tests/configuration.go
```diff
@@ -453,25 +486,40 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+							types.JSONPatchType,
+							[]byte(fmt.Sprintf(`[{ "op": "replace", "path": "/metadata/annotations", "value": %s }]`, initialAnnotations)),
+						),
+					)).ToNot(HaveOccurred())
+				}
 				By("waiting for mcp to start updating")
 				waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdating)
 
-				By("wait for mcp to get updated")
 				waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdated)
```
rshemtov13, line 523: These lines are not inside the if statements and therefore if there is no change that happens to the initial state of the kubeletconfig or under the performance profile `waitForMcpCondition` will get stuck here and fail eventually.
So a better option would probably be to check if there even needs to be a change to the initial state if not it can all be skipped.

shajmakh: right, I'll update this. Thanks for catching

shajmakh: in fact getNewTopologyManagerScopeValue() ensures to return scope different from the initial one, nevertheless I added an assertion

shajmakh:
needs a rebase on top of https://github.com/openshift-kni/numaresources-operator/pull/968 once merged

test/utils/clients/clients.go
```diff
@@ -25,9 +25,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
 	nrtv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
-	nropv1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1"
+	ntov2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
```
ffromani, line 28: nto `nto`, this is `perfprof`

test/e2e/serial/tests/configuration.go
```diff
@@ -38,31 +38,30 @@ import (
...
 
 	"github.com/google/go-cmp/cmp"
 	"golang.org/x/sync/errgroup"
 
 	nrtv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
-	nropv1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
+	ntov2 "github.com/openshift/cluster-node-tuning-operator/pkg/apis/performanceprofile/v2"
```
ffromani, line 51: not `nto`, this is `perfprof`

test/utils/clients/clients.go
shajmakh: fixed, thanks!

test/e2e/serial/tests/configuration.go
shajmakh: fixed, thanks!

shajmakh:
test passed after the rebase and additional minor adjustments

ffromani (commented):
/lgtm

PR #963, opened 2024-07-25
manifests: label: add `app` label

ffromani:
In order to discover the deployment and/or the operator pod, filtering by `app` label is the recommended way.
We add the label to enable so.


Tal-or:
/lgtm
/approve
Thanks!

shajmakh:
Thanks for this!

PR #968, opened 2024-08-01
e2e: simplify reboot test 47674 and minor fix

shajmakh:
The PR is mainly targetting test 47674 stabilization distributed across few commits. so for more details please see each commit. 

The test goal is to check if updating configurable values under the
numaresourcesoperator CR is done successfully with the anticipated
affects.

The test is flaky in the ci because of mainly 2 issues:
1. waiting improperly for the mcp updates (see commit )
2. not handling the mcp errors on time but instead continuing with the
   test steps

In this commit, we reorder the steps and simplify the test while
retaining the goal of it.

The initial scenario was like this:
1. pick target node
        1.1. label it with new role "mcp-test"
        1.2. remove the worker role
2. create new mcp targeting both roles: mcp-test & worker
        2.1. wait on the mcp mcp-test to get updated
        2.2. wait for mcp worker to start updating and finally reach updated
3. update mcp label on the NROP CR to point to mcp-test
        3.1. verify anticipated results of the CR updates
Now the recovery:
4. undo the NROP CR updates
        4.1. wait for worker mcp to start updating and reach updated
5. delete the mcp mcp-test
6. restore the labels on the target node as follows:
        6.1. remove mcp-test label
        6.2. label it back woth worker

In the scenario above, one can see that already the first step is risky
as it is leaving the target node with unknown mapping for mcp because
mcp `mcp-test` is not yet created.
The second inaccurate thing is that mcp `mcp-test` is targetting two roles:
worker & mcp-test while the intention was to have the target node with 1 label and role only. The difference this does is that if an mcp is
targetting 2 roles the RTE pods will be created on all nodes having
one of these roles, so RTE will still be created on the non-target
nodes, which means mcp worker will not be updated because NROP settings
will still target nodes with worker role, hence step 4 will fail.
The failure however will not stop the test from running, thus it will
continue and do the verifications, and the verification will pass because the test waited long enough on step 4 to update the RTEs on the nodes.
When the test recovers, the recovery is also done incorrectly.
Starting the defers in LIFO order, starting from step 5, it deletes mcp
which holds one machine that has only one role, so after deletion that
one machine will be looking for a machine config but it won't find it
because it was deleted. The other wrong step is that again leaving the
node with no labels (unlabel then label) while it should be done (label
then unlabel). This leaves the environment messed up for the next tests
as mcp is stuck updating.

To fix this we simplify the test (wait for less updates) to make it more stable by following the
below steps:
1. create mcp `mcp-test` targetting one role `mcp-test`
2. pick target node and label it with mcp-test role (keep the old roles)
        2.1. this will trigger mcp update on mcp-test, and it will be
very quick so we allow to check the condition every 2 seconds (see
previous commit)
3. update NROP CR to use mcp-test. this will trigger an update on worker
   mcp because NROP settings will be removed from the non-target nodes.
Nodes associated with mcp-test will not observe an update because NROP is
already configured on them and the only anticipated update would be on
the RTE ds -> restarted with new node-selector
        3.1 wait for mcp update on worker mcp
4. verify nrop cr updates results
Recovery:
5. undo NROP updates
        5.1. wait for mcp update on worker mcp only. remember target
node has worker role.
6. remove mcp-test label off the target node
        6.1 wait for mcp update on worker mcp. this should be quick.
7. delete the mcp-test mcp. no waiting is needed for mcp update.

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -75,6 +75,25 @@ import (
 	serialconfig "github.com/openshift-kni/numaresources-operator/test/e2e/serial/config"
 )
 
+/*
+MCP gets updated=false & updating=true in 2 cases:
+ 1. mc is being updated with new configuration -> new CONFIG: all nodes associated to the pool will be updated
+ 2. machine count is increasing: meaning new nodes joins to the pool by matching their roles to the machineConfigSelector.matchExpressions of the mcp.
+    so the same CONFIG on the pool but different machine count. This will trigger reboot on the new nodes thus mcp Updating contidtion will be true.
+*/
+type MCPUpdateType string
+
+const (
+	MachineConfig MCPUpdateType = "MachineConfig"
+	MachineCount  MCPUpdateType = "MachineCount"
+)
+
+type mcpInfo struct {
+	obj           *machineconfigv1.MachineConfigPool
+	initialConfig string
+	sampleNode    corev1.Node
+}
+
 var _ = Describe("[serial][disruptive] numaresources configuration management", Serial, Label("disruptive"), Label("feature:config"), func() {
 	var fxt *e2efixture.Fixture
 	var nrtList nrtv1alpha2.NodeResourceTopologyList
@@ -120,14 +139,25 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 
 			workers, err := depnodes.GetWorkers(fxt.DEnv())
 			Expect(err).ToNot(HaveOccurred())
+			if len(workers) < 2 {
+				e2efixture.Skipf(fxt, "the test requires at least 2 worker nodes, found %d", len(workers))
+			}
 
 			targetIdx, ok := e2efixture.PickNodeIndex(workers)
 			Expect(ok).To(BeTrue())
 			targetedNode := workers[targetIdx]
 
+			// we need to save a node that will still be associated to the initial mcps so later we conduct the MachineConfig check on it
+			initialMcpSampleNode := workers[len(workers)-targetIdx-1]
+
 			// save the initial nrop mcp to use it later while waiting for mcp to get updated
 			initialMcps, err := nropmcp.GetListByNodeGroupsV1(context.TODO(), e2eclient.Client, nroOperObj.Spec.NodeGroups)
 			Expect(err).ToNot(HaveOccurred())
+
+			if len(initialMcps) > 1 {
+				e2efixture.Skip(fxt, "the test supports single node group")
+			}
+			initialMcp := initialMcps[0]
 
 			mcp := objects.TestMCP()
 			By(fmt.Sprintf("creating new MCP: %q", mcp.Name))
@@ -162,7 +192,16 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 
 			}()
 			//so far 0 machine count for mcp-test -> no nodes -> no updates status
-			newMcps := []*machineconfigv1.MachineConfigPool{mcp}
+			newMcpInfo := mcpInfo{
+				obj:           mcp,
+				initialConfig: mcp.Status.Configuration.Name,
+				sampleNode:    targetedNode,
+			}
+			initialMcpInfo := mcpInfo{
+				obj:           initialMcp,
+				initialConfig: initialMcp.Status.Configuration.Name,
+				sampleNode:    initialMcpSampleNode,
+			}
 
 			By(fmt.Sprintf("Label node %q with %q", targetedNode.Name, getLabelRoleMCPTest()))
 			unlabelFunc, err := labelNode(fxt.Client, getLabelRoleMCPTest(), targetedNode.Name)
@@ -170,14 +209,19 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 
 			defer func() {
 				By(fmt.Sprintf("CLEANUP: restore initial labels of node %q with %q", targetedNode.Name, getLabelRoleWorker()))
+				var updatedMcp machineconfigv1.MachineConfigPool
+				Expect(fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(initialMcpInfo.obj), &updatedMcp)).To(Succeed())
+				initialMcpInfo.initialConfig = updatedMcp.Status.Configuration.Name
+				initialMcpInfo.sampleNode = targetedNode
+
 				err = unlabelFunc()
 				Expect(err).ToNot(HaveOccurred())
 
-				//this will trigger node reboot as the NROP settings will be reapplied to the unlabelled node
-				waitForMcpUpdate(fxt.Client, context.TODO(), initialMcps)
+				//this will trigger node reboot as the NROP settings will be reapplied to the unlabelled node, so new node is added under the old mcp hence the MachineCount update type
+				waitForMcpUpdate(fxt.Client, context.TODO(), []mcpInfo{initialMcpInfo}, MachineCount)
 			}()
 
-			waitForMcpUpdate(fxt.Client, context.TODO(), newMcps)
+			waitForMcpUpdate(fxt.Client, context.TODO(), []mcpInfo{newMcpInfo}, MachineConfig)
 
 			By(fmt.Sprintf("modifying the NUMAResourcesOperator nodeGroups field to match new mcp: %q labels %q", mcp.Name, mcp.Labels))
 			Eventually(func(g Gomega) {
@@ -194,6 +238,10 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 
 			defer func() {
 				By("CLEANUP: reverting the changes under the NUMAResourcesOperator object")
+				var updatedMcp machineconfigv1.MachineConfigPool
+				Expect(fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(initialMcpInfo.obj), &updatedMcp)).To(Succeed())
+				initialMcpInfo.initialConfig = updatedMcp.Status.Configuration.Name
+
 				// see https://pkg.go.dev/github.com/onsi/gomega#Eventually category 3
 				nroOperObj := &nropv1.NUMAResourcesOperator{}
 				Eventually(func(g Gomega) {
@@ -206,17 +254,15 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 					g.Expect(err).ToNot(HaveOccurred())
 				}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(Succeed())
 
-				mcps, err := nropmcp.GetListByNodeGroupsV1(context.TODO(), e2eclient.Client, nroOperObj.Spec.NodeGroups)
-				Expect(err).ToNot(HaveOccurred())
-
 				By("waiting for mcps to start updating")
-				// this will trigger mcp update only for the initial mcps because the mcp-test nodes are still labeled with the old labels
-				waitForMcpUpdate(fxt.Client, context.TODO(), mcps)
+				// this will trigger mcp update only for the initial mcps because the mcp-test nodes are still labeled
+				// with the old labels, so worker mcp will switch back to the NROP mc
+				waitForMcpUpdate(fxt.Client, context.TODO(), []mcpInfo{initialMcpInfo}, MachineConfig)
 			}() //end of defer
 
-			By("waiting for the mcps to start updating")
-			// on old mcp because the ds will no longer include the worker node that is not labeled with mcp-test
-			waitForMcpUpdate(fxt.Client, context.TODO(), initialMcps)
+			By("waiting for the mcps to update")
+			// on old mcp because the ds will no longer include the worker node that is not labeled with mcp-test, so returning to MC without NROP settings
+			waitForMcpUpdate(fxt.Client, context.TODO(), []mcpInfo{initialMcpInfo}, MachineConfig)
 
 			By(fmt.Sprintf("Verify RTE daemonsets have the updated node selector matching to the new mcp %q", mcp.Name))
 			Eventually(func() (bool, error) {
@@ -384,6 +430,10 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 			initialNrtList, err = e2enrt.GetUpdated(fxt.Client, initialNrtList, timeout)
 			Expect(err).ToNot(HaveOccurred(), "cannot get any NodeResourceTopology object from the cluster")
 
+			mcpsInfo, err := buildMCPsInfo(fxt.Client, context.TODO(), *nroOperObj)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(mcpsInfo).ToNot(BeEmpty())
+
 			mcps, err := nropmcp.GetListByNodeGroupsV1(context.TODO(), fxt.Client, nroOperObj.Spec.NodeGroups)
 			Expect(err).ToNot(HaveOccurred(), "cannot get MCPs associated with NUMAResourcesOperator %q", nroOperObj.Name)
 
@@ -428,10 +478,14 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 			}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(Succeed())
 
 			By("waiting for mcp to update")
-			waitForMcpUpdate(fxt.Client, context.TODO(), mcps)
+			waitForMcpUpdate(fxt.Client, context.TODO(), mcpsInfo, MachineConfig)
 
 			defer func() {
 				By("reverting kubeletconfig changes")
+				mcpsInfo, err := buildMCPsInfo(fxt.Client, context.TODO(), *nroOperObj)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(len(mcpsInfo)).To(BeNumerically(">", 0))
+
 				Eventually(func(g Gomega) {
 					err = fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(targetedKC), targetedKC)
 					Expect(err).ToNot(HaveOccurred())
@@ -448,7 +502,7 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 				}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(Succeed())
 
 				By("waiting for mcp to update")
-				waitForMcpUpdate(fxt.Client, context.TODO(), mcps)
+				waitForMcpUpdate(fxt.Client, context.TODO(), mcpsInfo, MachineConfig)
 			}()
 
 			By("checking that NUMAResourcesOperator's ConfigMap has changed")
@@ -737,6 +791,10 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 
 			if initialTopologyManagerPolicy != v1beta1.NoneTopologyManagerPolicy {
 				By("modifying topology manager policy under kubeletconfig to none")
+				mcpsInfo, err := buildMCPsInfo(fxt.Client, ctx, nroOperObj)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(len(mcpsInfo)).To(BeNumerically(">", 0))
+
 				Eventually(func(g Gomega) {
 					err := fxt.Client.Get(ctx, client.ObjectKeyFromObject(targetedKC), targetedKC)
 					Expect(err).ToNot(HaveOccurred())
@@ -756,7 +814,7 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 					g.Expect(err).ToNot(HaveOccurred())
 				}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(Succeed())
 
-				waitForMcpUpdate(fxt.Client, ctx, mcps)
+				waitForMcpUpdate(fxt.Client, ctx, mcpsInfo, MachineConfig)
 			}
 
 			defer func() {
@@ -768,6 +826,10 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 
 				if initialTopologyManagerPolicy != kcObj.TopologyManagerPolicy {
 					By("reverting kubeletconfig changes")
+					mcpsInfo, err := buildMCPsInfo(fxt.Client, ctx, nroOperObj)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(len(mcpsInfo)).To(BeNumerically(">", 0))
+
 					Eventually(func(g Gomega) {
 						err := fxt.Client.Get(ctx, client.ObjectKeyFromObject(targetedKC), targetedKC)
 						Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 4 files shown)


test/e2e/serial/tests/workload_placement.go
```diff
@@ -246,6 +243,9 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			By("wait for NRT data to settle")
 			e2efixture.MustSettleNRT(fxt)
 
+			nrtPostCreateDeploymentList, err := e2enrt.GetUpdated(fxt.Client, nrtInitialList, time.Minute)
+			Expect(err).ToNot(HaveOccurred())
+
```
ffromani, line 248: we had a similar bug in a different test, didn't we?

test/e2e/serial/tests/configuration.go
```diff
@@ -1018,7 +1026,5 @@ func waitForMcpsCondition(cli client.Client, ctx context.Context, mcps []*machin
 			return err
 		})
 	}
-	if err := eg.Wait(); err != nil {
-		fmt.Printf("An error occurred: %v\n", err)
-	}
```
ffromani, line 1023: yet another reminder that swallowing errors is almost always a bad idea, and that some italian reviewer should be stricter about it

test/e2e/serial/tests/configuration.go
```diff
@@ -1014,13 +1014,21 @@ func accumulateKubeletConfigNames(cms []corev1.ConfigMap) sets.Set[string] {
...
 func waitForMcpsCondition(cli client.Client, ctx context.Context, mcps []*machineconfigv1.MachineConfigPool, condition machineconfigv1.MachineConfigPoolConditionType) error {
 	var eg errgroup.Group
+	interval := configuration.MachineConfigPoolUpdateInterval
+	if condition == machineconfigv1.MachineConfigPoolUpdating {
+		// the transition from updated to updating to updated can be very fast sometimes. so if
+		// the status changed to updating and then to updated while on wait it will miss the updating
+		// status, and it will keep waiting and lastly fail the test. to avoid that decrease the interval
+		// to allow more often checks for the status
+		interval = 2 * time.Second
+	}
```
ffromani, line 1001: this suggests we may be waiting for the wrong condition. Rather than matching the exact (fragile) sequence of events we should probably do stricter check when we detect `Updated` to make sure we updated to the desired state.

test/e2e/serial/tests/configuration.go
```diff
@@ -170,12 +158,26 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			unlabelFunc, err := labelNode(fxt.Client, nodes.GetLabelRoleMCPTest(), targetedNode.Name)
 			Expect(err).ToNot(HaveOccurred())
 
+			defer func() {
+				By(fmt.Sprintf("CLEANUP: restore initial labels of node %q with %q", targetedNode.Name, nodes.GetLabelRoleWorker()))
+				err = unlabelFunc()
+				Expect(err).ToNot(HaveOccurred())
+
+				//this will trigger node reboot as the NROP settings will be reapplied to the unlabelled node
+				waitForMcpUpdate(fxt.Client, context.TODO(), initialMcps)
```
ffromani, line 176: considering what we fixed like 1-2 commits before, do we need to check the error here?
`waitForMcpuUpdate` would need the changes to not swallow the errors

shajmakh:
the test passed once but
/hold
for another round of testing

ffromani (commented):
that seems to be a nasty one to catch, so nice work. But let's be through now. Few questions inside

test/e2e/serial/tests/configuration.go
```diff
@@ -446,10 +427,12 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
 
 			By("waiting for mcp to start updating")
-			waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdating)
+			err = waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdating)
+			Expect(err).ToNot(HaveOccurred())
 
 			By("wait for mcp to get updated")
-			waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdated)
+			err = waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdated)
+			Expect(err).ToNot(HaveOccurred())
```
rshemtov13, line 435: Here you can also use `waitForMcpUpdate` and some other places below

test/e2e/serial/tests/workload_placement.go
shajmakh: yes indeed, fixed below, but hard to catch since it isn't always failing:
https://github.com/openshift-kni/numaresources-operator/commit/129621266defbcb71483a8d39e1b8c3826428ed7

test/e2e/serial/tests/configuration.go
shajmakh: please allow me to take responsibility for that, the goal for moving from wait group to errorgroup was mainly for handling the error properly, it really was a bad implementation for the intended new alternative just reporting the error

test/e2e/serial/tests/configuration.go
shajmakh: this is done inside https://github.com/openshift-kni/numaresources-operator/pull/968/commits/ff19d4bae8dbe48ef2f4b70347a00c0193a0d1b7#diff-ff13df3dd632af928cfdb92fae23b8e0f6de4cfe44a77c0d3a5d1a5a2eacbf5dR786 `waitForMcpUpdate()` is basically waiting for complete updates (updating -> updated) for the **same** mcp group

test/e2e/serial/tests/configuration.go
shajmakh: correct, I need to look into a way to do that in addition. thanks for raising this

test/e2e/serial/tests/configuration.go
```diff
@@ -207,11 +207,12 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
 				Expect(err).ToNot(HaveOccurred())
 
 				By("waiting for mcp to start updating")
-				waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdating)
+				err = waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdating)
```
Tal-or: nit:
More elegant way for handling such functions is:
```
Expect(waitForMcpsCondition(fxt.Client, context.TODO(), mcps, machineconfigv1.MachineConfigPoolUpdating)).To(Succeed())
```

test/e2e/serial/tests/configuration.go
Tal-or: I think in such extreme cases we can also watch for the `node`'s annotations:
```
machineconfiguration.openshift.io/currentConfig: rendered-worker-5eb756bcf4dcfb1c4b160b4b1270ab4d
machineconfiguration.openshift.io/desiredConfig: rendered-worker-5eb756bcf4dcfb1c4b160b4b1270ab4d
```
We can have the following steps:
1. Check node desired
2. `Update(MCP)`
3. Wait for node desired and current to change
4. Wait for MCP to move to `Updated` state
This grantees that nodes were actually updated into new revision and MCP is done.

This should be done on all nodes associated with the MCP in parallel in order to save time.

ffromani: yep, very good suggestion. This is a much better form of what I was thinking about, kudos!

test/e2e/serial/tests/configuration.go
shajmakh: indeed I'll update these, thanks for catching

test/e2e/serial/tests/configuration.go
shajmakh: right, I'll adjust the instances, thanks!

test/e2e/serial/tests/configuration.go
shajmakh: thanks for your comments, sounds reasonable, I'll verify this scenario

shajmakh: implemented in last commit & ready for review

shajmakh:
/unhold
affected tests passed & ready for review

test/e2e/serial/tests/configuration.go
```diff
@@ -123,11 +136,18 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
 			targetIdx, ok := e2efixture.PickNodeIndex(workers)
 			Expect(ok).To(BeTrue())
 			targetedNode := workers[targetIdx]
+			// node that will still be associated to initial mcps
+			initialMcpSampleNode := workers[len(workers)-targetIdx-1]
```
ffromani, line 140: can it be `targetNode`? what's the benefit of picking some other node?
are we guaranteed `len(workers)-targetIdx-1` is always in the interval `[0, len(workers))` ?

test/e2e/serial/tests/configuration.go
```diff
@@ -383,6 +419,10 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
 			initialNrtList, err = e2enrt.GetUpdated(fxt.Client, initialNrtList, timeout)
 			Expect(err).ToNot(HaveOccurred(), "cannot get any NodeResourceTopology object from the cluster")
 
+			mcpsInfo, err := buildMCPsInfo(fxt.Client, context.TODO(), *nroOperObj)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(mcpsInfo)).To(BeNumerically(">", 0))
```
ffromani, line 424: nit: `ToNot(BeEmpty())`?

test/e2e/serial/tests/tolerations.go
```diff
@@ -716,6 +726,40 @@ var _ = Describe("[serial][disruptive][rtetols] numaresources RTE tolerations su
 	})
 })
 
+func buildMCPsInfo(cli client.Client, ctx context.Context, nroObj nropv1.NUMAResourcesOperator) ([]mcpInfo, error) {
+	mcpsInfo := []mcpInfo{}
+	var updatedNroObj nropv1.NUMAResourcesOperator
+
+	if err := cli.Get(ctx, client.ObjectKeyFromObject(&nroObj), &updatedNroObj); err != nil {
+		return mcpsInfo, err
+	}
```
ffromani, line 735: we should be consistent; either always
`if err := action(); err != nil {`
or always
```
err := action()
if err != nil {
```
I'm biased towards the latter but I don't mind the former if we have at least source-code-level consistency

test/e2e/serial/tests/tolerations.go
```diff
@@ -783,10 +827,51 @@ func sriovToleration() corev1.Toleration {
...
-func waitForMcpUpdate(cli client.Client, ctx context.Context, mcps []*machineconfigv1.MachineConfigPool) {
-	By("waiting for mcp to start updating")
-	Expect(waitForMcpsCondition(cli, ctx, mcps, machineconfigv1.MachineConfigPoolUpdating)).To(Succeed())
+func waitForMcpUpdate(cli client.Client, ctx context.Context, mcpsInfo []mcpInfo, updateType MCPUpdateType) {
+	for _, info := range mcpsInfo {
+		By(fmt.Sprintf("verify updates for mcp %q", info.obj.Name))
+		klog.Info("waiting for mcp to start updating")
+		if err := waitForMcpsCondition(cli, ctx, []*machineconfigv1.MachineConfigPool{info.obj}, machineconfigv1.MachineConfigPoolUpdating); err != nil {
+			klog.Warningf("failed to find mcps while in updating status")
+		}
```
ffromani, line 836: I'd drop this part. Checking we updated to the desired state should be sufficient

test/e2e/serial/tests/tolerations.go
```diff
@@ -783,10 +827,51 @@ func sriovToleration() corev1.Toleration {
...
+		if updateType == MachineConfig {
+			Expect(updatedConfig).ToNot(Equal(info.initialConfig))
+		}
+
+		// MachineConfig update type will also update the node currentConfig so check that anyway
+		klog.Info("verify mcp config is updated by ensuring the sample node has updated MC")
+		ok, err := verifyUpdatedMCOnNodes(cli, ctx, info.sampleNode, updatedConfig)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(ok).To(BeTrue())
+	}
```
ffromani, line 857: This should be done in a loop, which should periodically poll the given MCP and exit succesfully if all the following are true:
- condition is Updated
- the desired config matches the expected value or, equivalently but less strong (but way easier to check) is different from the initial state

ffromani (commented):
partial review, I need a deeper dive in this code

test/e2e/serial/tests/configuration.go
shajmakh: > can it be targetNode? what's the benefit of picking some other node?

we chose targetNode for the new mcp. we need another sample node for the old mcps

> are we guaranteed len(workers)-targetIdx-1 is always in the interval [0, len(workers)) ?

good catch, let me add a filter

ffromani: >we chose targetNode for the new mcp. we need another sample node for the old mcps

ffromani:
please rebase on top of current main branch

test/e2e/serial/tests/tolerations.go
shajmakh: I don't think that would be right in case the node didn't flip faster than the automation. additionally we check only a sample/random node, we need at least some waiting time before making sure the mcp got updated so we stick with a meaningful wait on the updating state

test/e2e/serial/tests/tolerations.go
shajmakh: the loop is done on the Updated condition `waitForMcpsCondition` (and before that another loop is done on Updating). the desired config on the node being updated to the new state is a result of fulfilling the Updated condition. so if the mcp Updated==true only, then it is reasonable to verify the node's current/desired status. Even if the specific node status is updated, we still have work to do, so the deciding factor here is the mcp Updated condition. We can put it in the loop alongside the Updated condition, but if the node status passed, it will not make any difference if, on the other hand, the Updated is still false, so what we mainly check for here is Updated=true, and only then the other conditions would be legit to check. 
If the concern is that the node's current vs desired config would take time to be updated after the mcp is done updating, we can make verifyUpdatedMCOnNodes functionality in eventually but as far as I learned mcp will not turn Updated unless all nodes got the correct annotations, otherwise the node should be reported as degraded and the reason would not be empty, which will make the mcp status Updated=false

test/e2e/serial/tests/configuration.go
shajmakh: > I think we can make it work with just one node but is not too important

we must pass node to be examined for the desired vs current config. the node must be associated to the mcp that we are checking it's updates, if not the check would fail because we compare against the mcp.status.configuration.name. we can exclude the target node for the old mcp check, but we need eventually to work with a different node or node list (for now one node is sufficient maybe later we can extend the new function to check a node list as the tests require)

test/e2e/serial/tests/tolerations.go
shajmakh: ack, changed that, thanks for pointing out

test/e2e/serial/tests/tolerations.go
```diff
@@ -784,10 +830,63 @@ func sriovToleration() corev1.Toleration {
...
-	By("waiting for mcp to start updating")
-	waitForMcpsCondition(cli, ctx, mcps, machineconfigv1.MachineConfigPoolUpdating)
+func waitForMcpUpdate(cli client.Client, ctx context.Context, mcpsInfo []mcpInfo, updateType MCPUpdateType) {
+	for _, info := range mcpsInfo {
+		/*
+				For every mcp check the following:
+				1. soft requirement: loop over until condition Updating==true
+			    2. loop over until condition Updated==true, is a must
+				3. check the sample node is updated with new config in its annotations, both for desired and current, is a must
+		*/
```
ffromani, line 840: good. Better:
1. fix spacing, it is unnecessarily misaligned and harder to read than it can
2. consistency. The first bullet starts with required/desired, the second and third bullet but the requirement at the end.
Example of how it can read:
```
/*
    procedure:
    1. soft requirement: loop over until condition Updating==true
    2. hard requirement: loop over until condition Updated==true
    3. hard requirement: check the sample node is updated with new config in its annotations, both for desired and current
*/
```

test/e2e/serial/tests/tolerations.go
```diff
@@ -784,10 +830,63 @@ func sriovToleration() corev1.Toleration {
 	}
 }
 
-func waitForMcpUpdate(cli client.Client, ctx context.Context, mcps []*machineconfigv1.MachineConfigPool) {
-	By("waiting for mcp to start updating")
-	waitForMcpsCondition(cli, ctx, mcps, machineconfigv1.MachineConfigPoolUpdating)
+func waitForMcpUpdate(cli client.Client, ctx context.Context, mcpsInfo []mcpInfo, updateType MCPUpdateType) {
```
ffromani, line 833: we can move the variadic part at the end, and use an actual variadic argument:
```
func waitForMcpUpdate(cli client.Client, ctx context.Context, updateType MCPUpdateType, mcpsInfo ...mcpInfo) {
```

so we can write
```
waitForMcpUpdate(cli, ctx, myUpdate, mcp1)
...
waitForMcpUpdate(cli, ctx, myUpdate, mcp1, mcp2, mcp3)
...
waitForMcpUpdate(cli, ctx, myUpdate, mcpList...)
```

ffromani: ditto for `waitForMcpsCondition`
nit: we should always UPPERCASE ACRONYMS, so it should be MCPs, but the codebase is already inconsistent

ffromani (commented):
ok, what you added makes sense and the rationale is solid. We talked offline about a different approach, but you correctly identified gaps we didn't seen at time. Nevertheless, I think we can rearrange some code flows to make the solution even simpler and more robust. I'll sketch a proposal which can be implemented in a later PR.

ffromani:
we can change `internal/wait` like this
```
$ git diff internal/wait/
diff --git a/internal/wait/machineconfig.go b/internal/wait/machineconfig.go
index dd286de8..2b46f856 100644
--- a/internal/wait/machineconfig.go
+++ b/internal/wait/machineconfig.go
@@ -55,6 +55,12 @@ func (wt Waiter) ForKubeletConfigDeleted(ctx context.Context, kc *machineconfigv
 }
 
 func (wt Waiter) ForMachineConfigPoolCondition(ctx context.Context, mcp *machineconfigv1.MachineConfigPool, condType machineconfigv1.MachineConfigPoolConditionType) error {
+       return wt.ForMachineConfigPoolFunc(ctx, mcp, func(updatedMcp machineconfigv1.MachineConfigPool) (bool, error) {
+               return IsMachineConfigPoolConditionMatching(updatedMcp, condType), nil
+       })
+}
+
+func (wt Waiter) ForMachineConfigPoolFunc(ctx context.Context, mcp *machineconfigv1.MachineConfigPool, checker func(updatedMcp machineconfigv1.MachineConfigPool) (bool, error)) error {
        immediate := false
        err := k8swait.PollUntilContextTimeout(ctx, wt.PollInterval, wt.PollTimeout, immediate, func(aContext context.Context) (bool, error) {
                updatedMcp := machineconfigv1.MachineConfigPool{}
@@ -63,18 +69,22 @@ func (wt Waiter) ForMachineConfigPoolCondition(ctx context.Context, mcp *machine
                if err != nil {
                        return false, err
                }
-               for _, cond := range updatedMcp.Status.Conditions {
-                       if cond.Type == condType {
-                               if cond.Status == corev1.ConditionTrue {
-                                       return true, nil
-                               } else {
-                                       klog.Infof("mcp: %q condition type: %q status is: %q expected status: %q", updatedMcp.Name, cond.Type, cond.Status, corev1.ConditionTrue)
-                                       return false, nil
-                               }
-                       }
-               }
-               klog.Infof("mcp: %q condition type: %q was not found", updatedMcp.Name, condType)
-               return false, nil
+               return checker(updatedMcp)
        })
        return err
 }
+
+func IsMachineConfigPoolConditionMatching(updatedMcp machineconfigv1.MachineConfigPool, condType machineconfigv1.MachineConfigPoolConditionType) bool {
+       for _, cond := range updatedMcp.Status.Conditions {
+               if cond.Type == condType {
+                       if cond.Status == corev1.ConditionTrue {
+                               return true
+                       } else {
+                               klog.Infof("mcp: %q condition type: %q status is: %q expected status: %q", updatedMcp.Name, cond.Type, cond.Status, corev1.ConditionTrue)
+                               return false
+                       }
+               }
+       }
+       klog.Infof("mcp: %q condition type: %q was not found", updatedMcp.Name, condType)
+       return false
+}
```

then we can reimplement like this
```
func waitForMcpUpdate(cli client.Client, ctx context.Context, updateType MCPUpdateType, mcpInfos ...mcpInfo) {
	ginkgo.GinkgoHelper()

	var wg WaitGroup

	for _, info := range mcpInfos {
		wg.Add(1)
		func(cli client.Client, ctx context.Context, updateType MCPUpdateType, info mcpInfo) {
			defer GinkgoRecover()
			defer wg.Done()
			waitMCPUpdateByType(cli, ctx, updateType, info)
		}(cli, ctx, updateType, info)
	}

	wg.Wait()
}

func waitMCPUpdateByType(cli client.Client, ctx context.Context, updateType MCPUpdateType, info mcpInfo) {
	ginkgo.GinkgoHelper()

	/*
	   For every mcp check the following:
	   1. soft requirement: loop over until condition Updating==true
	   2. hard requirement: loop over until condition Updated==true
	   3. hard requirement: check the sample node is updated with new config in its annotations, both for desired and current
	*/

	By(fmt.Sprintf("verify updates for mcp %q", info.obj.Name))
	klog.Info("waiting for mcp to start updating")

	// the transition from updated to updating to updated can be very fast sometimes. so if
	// the status changed to updating and then to updated while on wait it will miss the updating
	// status, and it will keep waiting and lastly fail the test. to avoid that decrease the interval
	// to allow more often checks for the status
	klog.Infof("wait for mcp %q to meet condition %q", mcp.Name, condition)
	err := wait.With(cli).Interval(2*time.Second).Timeout(configuration.MachineConfigPoolUpdateTimeout).ForMachineConfigPoolCondition(ctx, mcpInfo.obj, machineconfigv1.MachineConfigPoolUpdating)
	if err != nil {
		// just warn here because the switch between the mcp conditions: updated->updating->updated can be faster
		// and may be missed while the condition was actually met at some point
		klog.Warningf("failed to find mcps while in updating status")
	}

	klog.Info("wait for mcp to get updated")
	// here we must fail on errors

	err = wait.With(cli).Interval(configuration.MachineConfigPoolUpdateInterval).Timeout(configuration.MachineConfigPoolUpdateTimeout).ForMachineConfigPoolFunc(ctx, info.obj, func(updatedMcp machineconfigv1.MachineConfigPool) (bool, error) {
		if !IsMachineConfigPoolConditionMatching(updatedMcp, machineconfigv1.MachineConfigPoolUpdated) {
			return false, nil
		}
		updatedConfig := updatedMcp.Status.Configuration.Name
		if updateType == MachineConfig && updatedConfig != info.initialConfig {
			return false, nil
		}
		// MachineConfig update type will also update the node currentConfig so check that anyway
		klog.Info("verify mcp config is updated by ensuring the sample node has updated MC")
		return verifyUpdatedMCOnNodes(cli, ctx, info.sampleNode, updatedConfig)
	})
}
```

ffromani:
/lgtm

it's a good enough improvement for now, if we don't want further refinements we're good; if we want, better to do them in a followup.

shajmakh:
@ffromani thank you for the valuable review comments and lgtm, I'll address the improvements in a later PR and link it to this one

PR #969, opened 2024-08-02
tests: add fake entry point

ffromani:
this is ONLY to make
`ginkgo labels -r`
`gingko labels ./test/e2e/serial/tests`
work

the build tak is supposed to prevent the bogus entry to be consumed in the test binaries, so no ill effects or any change in the e2e serial test binary is expected.


ffromani:
/hold

do we want this at all?

shajmakh:
Thank you for looking into this, I tested locally and indeed, lists all the labels. we don't need it though so far but I see no harm either, do as you like

ffromani:
if this is no longer necessary, let's keep the codebase as simple and linear as possible

PR #971, opened 2024-08-02
upgrade ginkgo and additional feature introspection followup

shajmakh:
This is a followup for https://github.com/openshift-kni/numaresources-operator/pull/958, please see commits for more details.

### DIFF
--- a/internal/api/features/_topics.json
+++ b/internal/api/features/_topics.json
@@ -15,6 +15,7 @@
     "taint",
     "nodelabel",
     "byres",
-    "tmpol"
+    "tmpol",
+    "schedattrwatch"
   ]
 }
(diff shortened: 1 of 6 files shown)


ffromani (commented):
the only thing I don't like too much is the topic `auto_sched_update_on_kc_change` but I don't have better suggestions anyway

shajmakh:
/hold
@ffromani, if you prefer, including a bug ID for the latest version is also an option; what do you think?

internal/api/features/_topics.json
```diff
@@ -15,6 +15,7 @@
     "taint",
     "nodelabel",
     "byres",
-    "tmpol"
+    "tmpol",
+    "auto_sched_update_on_kc_change"
```
ffromani, line 19: maybe `schedattrwatch`?

ffromani (commented):
I think topics should preferably be different than bug ids because we talk about something approaching more features than bufixes, but I'm not strongly against using  them.

internal/api/features/_topics.json
shajmakh: sounds better, will use that

ffromani (commented):
/approve
/lgtm

PR #972, opened 2024-08-02
topics: add target and machinery to generate offline

ffromani:
enable offline generation of topics.json, to enable bundle it into the tests image, or anywhere it is useful.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -211,6 +211,9 @@ runner-e2e-serial: bin/envsubst
 	hack/render-e2e-runner.sh
 	hack/test-e2e-runner.sh
 
+build-topics:
+	mkdir -p bin && go run tools/lstopics/lstopics.go > bin/topics.json
+
 build: generate fmt vet binary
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -211,6 +211,9 @@ runner-e2e-serial: bin/envsubst
 	hack/render-e2e-runner.sh
 	hack/test-e2e-runner.sh
 
+build-topics:
+	mkdir -p bin && go run tools/lstopics/lstopics.go > bin/topics.json
```
shajmakh, line 215: can ship (COPY) the file in the tests image https://github.com/openshift-kni/numaresources-operator/blob/main/Dockerfile.tests or to the operator image https://github.com/openshift-kni/numaresources-operator/blob/main/Dockerfile?  otherwise I don't understand how to use this offline

ffromani: it is going to sit in the tests image: Dockerfile PR pending because depends on this one

Issue #715, opened 2023-10-14
Replace local pointer-ing mechanism with the known k8s.io/utils/pointer


shajmakh:
there are other places in the codebase which can benefit to move to `k8s.io/utils/pointer`. I actually welcome this change. But it should be done all in once: the only thing I'm not fine with is having both the old clunky way I started (reinventing `k8s.io/utils/pointer` locally) and using `k8s.io/utils/pointer`.  Either you submit a PR to convert to `k8s.io/utils/pointer` everywhere and get that PR merged before to move on with this one (fine) or you use the old way in this new PR and convert to `k8s.io/utils/pointer` with a later PR

_Originally posted by @ffromani in https://github.com/openshift-kni/numaresources-operator/pull/700#discussion_r1357781930_

PR #975, opened 2024-08-05
cleanup: remove `internal/nodes`

ffromani:
`deployer` >= 0.19 provides most of the functionalities in `internal/nodes`. So we dissolve the latter and we move the code to use the former, to reduce the duplication, even if the API of the deployer package can use some refinement.
In addition, we get rid of another internal duplication by moving to `k8s/utils/ptr`

Fixes: #715 (probably)


ffromani:
/hold cancel
we don't have LGTM anyway

Issue #974, opened 2024-08-05
scheduler: set priorityClass


ffromani:
We currently don't set the priority for the secondary scheduler pod:
```
 oc get pod -n numaresources secondary-scheduler-8955769bc-c5zxh -o json | jq .spec.priority
0
```
but we should. We need to evaluate, the correct priority is probably `system-node-critical`
This change should be backported to all the active branches.

ffromani:
> We currently don't set the priority for the secondary scheduler pod:
> 
> ```
>  oc get pod -n numaresources secondary-scheduler-8955769bc-c5zxh -o json | jq .spec.priority
> 0
> ```
> 
> but we should. We need to evaluate, the correct priority is probably `system-node-critical` This change should be backported to all the active branches.

random thoughts:
- e2e test not necessary, a controller test and/or even a unit test which ensures rhe `proprityClass{Name}` field is set should be sufficient
- need some reasearch to figure out if we want to make it system-node-critical or system-cluster-critical. We should probably set the same as (OCP?) scheduler pods
- we may want to bump the priority if replica==1 (the current default) but this is debatable.

shajmakh:
addressed in 4.17 (https://github.com/openshift-kni/numaresources-operator/pull/979) and 4.16 (https://github.com/openshift-kni/numaresources-operator/pull/980) so far

PR #979, opened 2024-08-12
sched: controller: set scheduler priority

shajmakh:
So far the scheduler priority is set to default which is 0 this is risky especially when the preemtion of pods is needed to fit more important pods.

The NRS is important enough to deserve the most critical priority class system-node-critical which is the same priority for the kube-scheduler.

addresses https://github.com/openshift-kni/numaresources-operator/issues/974

### DIFF
--- a/controllers/numaresourcesscheduler_controller.go
+++ b/controllers/numaresourcesscheduler_controller.go
@@ -56,6 +56,7 @@ import (
 
 const (
 	leaderElectionResourceName = "numa-scheduler-leader"
+	schedulerPriorityClassName = "system-node-critical"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


controllers/numaresourcesscheduler_controller.go
```diff
@@ -55,7 +55,8 @@ import (
 )
 
 const (
-	leaderElectionResourceName = "numa-scheduler-leader"
+	leaderElectionResourceName      = "numa-scheduler-leader"
+	SystemNodeCriticalPriorityClass = "system-node-critical"
```
ffromani, line 59: please make it private and rename to convey the intent, not the value. Maybe `schedulerPriorityClassName`

ffromani: In addition, please record in the commit message why `system-node-critical` and not `system-cluster-critical` (I'm fine with the decision, but let's record the reason)

shajmakh: updated, thanks for raising

ffromani (commented):
/approve
/lgtm

thanks!

PR #981, opened 2024-08-13
e2e: adapt performance profile existence and stabilize test 

rshemtov13:
This PR follows #962 for adding the performance profile dependencies for reboot test cases.
It adapts to changes in the kubeletconfig through performance profile if it exists. This PR also adds test id to that test which is 75354.
This test is a reboot test case that is changing the kubeletconfig configuration. The changes here are for adapting the test to make the necessary changes to the kubeletconfig through performance profile if it exists in the first place otherwise if the system which is testing this has a performance profile and it tries to change it directly through kubeletconfig the test will fail.

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -464,7 +464,7 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 			Expect(initialTMScope).ToNot(Equal(newTMScope))
 
 			By("verify owner reference of kubeletconfig")
-			Expect(len(targetedKC.OwnerReferences)).To(BeNumerically("<", 2), "") //so 0 or 1
+			Expect(len(targetedKC.OwnerReferences)).To(BeNumerically("<", 2)) //so 0 or 1
 			if len(targetedKC.OwnerReferences) == 0 {
 				By("modifying Topology Manager Scope under kubeletconfig")
 				Eventually(func(g Gomega) {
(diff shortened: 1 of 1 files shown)


rshemtov13:
This PR is also a follow up of #977 which was closed for some rebasing issues.
The #977 PR was the one that most of the changes here were reviewed, hence the mentioning here of that.

test/e2e/serial/tests/configuration.go
```diff
@@ -1067,7 +1067,15 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
 			By("checking the deployment pod has failed scheduling and its at the pending status")
 			pods, err := podlist.With(fxt.Client).ByDeployment(ctx, updatedDeployment)
 			Expect(err).ToNot(HaveOccurred())
-			Expect(len(pods)).To(Equal(1))
+			Expect(len(pods)).To(BeNumerically("<", 10), "")
```
shajmakh, line 1070: `, "")` is used to add details upon failure of the Expect, if you don't want to add more details please delete it

shajmakh: also please add a comment explaining the reasoning for using the number 10, you can tell that pods with TAE may be created but their count should be always bounded and no more than 10 based on the results of several runs

test/e2e/serial/tests/configuration.go
```diff
@@ -845,37 +849,66 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
 			Expect(err).ToNot(HaveOccurred())
 			initialTopologyManagerPolicy := kcObj.TopologyManagerPolicy
 
+			// Before restarting the scheduler, we change here the Topology Manager Policy to 'none'.
+			// This ensures that if we later change it to 'single NUMA node' and attempt to create
+			// a deployment that is expected to fail due to TopologyAffinityError it will otherwise fail for failed scheduling.
+			// We expect that failure to occur only when the bug is present.
+			// Changes are made through the kubeletconfig directly or through performance profile.
+			By("verify owner reference of kubeletconfig")
+			Expect(numTargetedKCOwnerReferences).To(BeNumerically("<", 2), "") // so 0 or 1
```
shajmakh, line 864: same here regarding the failure msg, I see one instance that I missed in previous PR would be great if you can fix that too. https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/serial/tests/configuration.go#L467

test/e2e/serial/tests/configuration.go
ffromani: +1, make it a constants in this source file if possible

test/e2e/serial/tests/configuration.go
rshemtov13: Sure np will fix that too and the others

test/e2e/serial/tests/configuration.go
rshemtov13: alright will fix that now thanks for noticing that

shajmakh:
Thanks for all the updates!
/lgtm
/approve

shajmakh:
@rshemtov13 did we make sure the test passes after the updates?

rshemtov13:
The changes in this PR have been tested using an environment with a performance profile and also
an environment with a kubletconfig and passed successfully

PR #983, opened 2024-08-13
Dockerfile: align OCP/non-OCP

ffromani:
PR https://github.com/openshift-kni/numaresources-operator/pull/686 updated the non-OCP Dockerfile only. To preserve/increase image compatibility, we do now the same change for OCP Dockerfiles.


shajmakh:
thanks for this!
/lgtm

Issue #978, opened 2024-08-10
support for RTE >= 0.17


ffromani:
RTE 0.17 revamped its config support to a much regulare and handy format. We need to catch up.
xref: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/releases/tag/v0.17.0

ffromani:
fixed by #1095

PR #988, opened 2024-08-20
deps: bump RTE version to v0.19.0

shajmakh:
Update dependency version to consume latest updates.

addresses: https://github.com/openshift-kni/numaresources-operator/issues/978

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/k8stopologyawareschedwg/deployer v0.20.1
 	github.com/k8stopologyawareschedwg/noderesourcetopology-api v0.1.2
 	github.com/k8stopologyawareschedwg/podfingerprint v0.2.2
-	github.com/k8stopologyawareschedwg/resource-topology-exporter v0.16.1
+	github.com/k8stopologyawareschedwg/resource-topology-exporter v0.19.0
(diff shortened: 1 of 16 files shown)


ffromani:
this will require evaluating the transition to the new APIs (e.g. move to config files rather than the bunch of options we are using) and regression testing

shajmakh:
/hold
for transitioning to configfiles style

shajmakh:
> > this will require evaluating the transition to the new APIs (e.g. move to config files rather than the bunch of options we are using) and regression testing
> > example of how it could look like: [k8stopologyawareschedwg/resource-topology-exporter#256](https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/256)
> > [EDIT] we can stay with the old-style commandline option and defer the transition to the configfile, but we should make a plan/timeline for the transition to configfile.
> 
> I recommend to the take the opportunity of supporting NROP on HCP, and:
> 
> 1. bump dependent packages (deployer, RTE) to latest and greatest k8s and go versions.
> 2. rebase kni-scheduler to latest commit on sig-scheduler.
> 3. bump RTE, deployer and scheduler versions to latest and greatest on NROP.

thanks for sharing, good idea indeed for the 4.18 branch. I'll note that to apply it when we branch.

PR #994, opened 2024-09-03
tree: version: move to 4.18

shajmakh:
Redirect main to 4.18.

### DIFF
--- a/Dockerfile.must-gather
+++ b/Dockerfile.must-gather
@@ -1,4 +1,4 @@
-FROM quay.io/openshift/origin-must-gather:4.16.0 AS builder
+FROM quay.io/openshift/origin-must-gather:4.18.0 AS builder
 
 FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
 RUN microdnf install -y procps-ng tar rsync
--- a/test/utils/objects/objects.go
+++ b/test/utils/objects/objects.go
@@ -54,7 +54,7 @@ func TestNROScheduler() *nropv1.NUMAResourcesScheduler {
 			Name: objectnames.DefaultNUMAResourcesSchedulerCrName,
 		},
 		Spec: nropv1.NUMAResourcesSchedulerSpec{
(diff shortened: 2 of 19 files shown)


Dockerfile.must-gather
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/openshift/origin-must-gather:4.16.0 AS builder
+FROM quay.io/openshift/origin-must-gather:4.17.0 AS builder
```
ffromani, line 1: should be 4.18 if the image is available (it usually is not when we update)

test/utils/objects/objects.go
```diff
@@ -54,7 +54,7 @@ func TestNROScheduler() *nropv1.NUMAResourcesScheduler {
 			Name: objectnames.DefaultNUMAResourcesSchedulerCrName,
 		},
 		Spec: nropv1.NUMAResourcesSchedulerSpec{
-			SchedulerImage: "quay.io/openshift-kni/scheduler-plugins:4.17-snapshot",
+			SchedulerImage: "quay.io/openshift-kni/scheduler-plugins:4.18-snapshot",
```
ffromani, line 57: because of the scheduler references, we need to branch out and fix the scheduler (operand) image first, then the operator

ffromani (commented):
it's usually better to do the git branching first, then to merge this PR

ffromani:
/hold

let's do the scheduler first

ffromani:
ok, now that we branched out 4.17 we can merge this, but again let's fix the scheduler first

Dockerfile.must-gather
shajmakh: it is:
https://quay.io/repository/openshift/origin-must-gather?tab=tags
updated it. Thanks!

ffromani:
tests won't pass until we have 4.18 scheduler builds online

ffromani:
/retest

4.18 scheduler plugin images have been published

ffromani:
had to do a bogus tag on quay because chicken/egg problem. It's a known issue.

PR #995, opened 2024-09-03
manifests: labels: don't match using new labels

ffromani:
In 055d665ea54a3aeadc24 we added new labels AND we immediately used them to match  components in our deployment. Turns out this breaks upgrades because the old image getting replaced can't possibly have the matching label, for starters. So the proper way is to add labels, let them sit for few releases, then start consuming them. This should work in all the flows, including upgrade.
In this commit we unblock upgrades removing the matching requirement, which will be introduced again later in the future.


shajmakh:
/lgtm
Thanks for this!

PR #1005, opened 2024-09-05
[release-4.16] internal: align topics version with the branch

shajmakh:
Mistakenly backported with 4.17 value, realign it with the 4.16 branch.

### DIFF
--- a/internal/api/features/types.go
+++ b/internal/api/features/types.go
@@ -23,7 +23,7 @@ import (
 )
 
 const (
-	Version = "v4.17.0"
+	Version = "v4.16.0"
 )
 
 type Metadata struct {


internal/api/features/types.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 const (
-	Version = "v4.17.0"
+	Version = "v4.16.0"
```
ffromani, line 26: The version is meant to control the data format of this API, not of the operator. Up until we change/add fields to this API, we should use the same Version. `v4.17.0` is not worse than `v1.0.0` in this regard, albeit perhaps less usual. All the branches use the same **API** version, so they should all have the same value for `Version`

shajmakh: I saw the expected version in the API test is mismatched below 4.17 branches so it didn't look right. Thanks for the clarification. I'll close this and the similar PRs.

PR #1006, opened 2024-09-11
WIP: nro controller: enable NodeSelector under NodeGroup

shajmakh:
So far the controller configures the operator RTE daemonset on node groups according to the MCP selector provided under NodeGroup[]. The determination of nodes based on that MCP selector boils down to the node selector provided under that MCP spec.

This PR enables configuring the nodeGroups of the NROP CR via NodeSelector directly, while still retaining the MCP selector option.

The only restriction for having these both options configured is that if both are found in the CR they should point to the same MCPs, otherwise the operator's status will be in a degraded condition.


shajmakh:
bad approach. new PR will follow.

PR #1007, opened 2024-09-12
CNF-10142: Enable NROP metrics to be to scraped securely by Prometheus   

rbaturov:
This PR encompasses all the required changes to reintegrate NROP metrics with Prometheus. It introduces a kube-rbac-proxy sidecar to establish a secure communication channel. The majority of the changes in this PR follow the guidelines outlined in the following guide:
https://rhobs-handbook.netlify.app/products/openshiftmonitoring/collecting_metrics.md/

The only difference between this implementation and the guide's is that we use a `bearerTokenFile` for Prometheus authentication instead of tls.crt and tls.key. This approach uses TLS but does not implement mTLS.

A follow-up PR will be issued to ensure we implement this for RTE metrics as well.
Moreover, will issue a PR adding an e2e test to the CI, for this functionality.

2. run: `make deploy` 
4. run:
  ```
curl -v \
  https://numaresources-controller-manager-metrics-service.numaresources.svc:8443/metrics
```

### DIFF
--- a/config/prometheus/monitor.yaml
+++ b/config/prometheus/monitor.yaml
@@ -1,20 +1,24 @@
 
-# Prometheus Monitor Service (Metrics)
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
-  labels:
-    control-plane: controller-manager
-  name: controller-manager-metrics-monitor
+  name: controller-manager
   namespace: system
 spec:
   endpoints:
-    - path: /metrics
-      port: https
-      scheme: https
-      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
-      tlsConfig:
-        insecureSkipVerify: true
+  - interval: 30s
+    # Matches the name of the service's port.
+    targetPort: 8443
+    path: /metrics
+    scheme: https
+    bearerTokenFile: "/var/run/secrets/kubernetes.io/serviceaccount/token"
+    tlsConfig:
+      # The CA file used by Prometheus to verify the server's certificate.
+      # It's the cluster's CA bundle from the service CA operator.
+      caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+      # The name of the server (CN) in the server's certificate.
+      serverName: numaresources-controller-manager-metrics-service.numaresources.svc
+      insecureSkipVerify: false
   selector:
     matchLabels:
       control-plane: controller-manager
--- a/config/prometheus/rbac.yaml
+++ b/config/prometheus/rbac.yaml
@@ -1,0 +1,31 @@
+# creates Role and RoleBinding for prometheus-k8s service account to access our namespace
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: prometheus-k8s
+  namespace: system
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - services
+  - endpoints
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: prometheus-k8s
+  namespace: system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s
+subjects:
+- kind: ServiceAccount
+  name: prometheus-k8s
+  namespace: openshift-monitoring
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -12,7 +12,7 @@ resources:
 # Comment the following 4 lines if you want to disable
 # the auth proxy (https://github.com/brancz/kube-rbac-proxy)
 # which protects your /metrics endpoint.
-#- auth_proxy_service.yaml
-#- auth_proxy_role.yaml
-#- auth_proxy_role_binding.yaml
-#- auth_proxy_client_clusterrole.yaml
+- auth_proxy_service.yaml
+- auth_proxy_role.yaml
+- auth_proxy_role_binding.yaml
+- auth_proxy_client_clusterrole.yaml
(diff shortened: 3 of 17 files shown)

### CODE REVIEW

File: main.go [hunk 1]
```diff
@@ -71,11 +71,11 @@ const (
...
 const (
-	defaultWebhookPort = 9443
-	defaultMetricsAddr = ":8080"
+	defaultWebhookPort    = 9443
+	defaultMetricsAddr    = ":8080"
 	defaultMetricsEnabled = true
-	defaultProbeAddr   = ":8081"
-	defaultNamespace   = "numaresources-operator"
+	defaultProbeAddr      = ":8081"
+	defaultNamespace      = "numaresources-operator"
```
ffromani, line 78: where do we use these?

ffromani (commented):
at glance LGTM, but I'll have a proper review later on

ffromani:
let's merge https://github.com/openshift-kni/numaresources-operator/pull/1008 before

ffromani:
please rebase on top of current main branch

main.go
rbaturov: I only added `defaultMetricsEnabled` which is used [here](https://github.com/rbaturov/numaresources-operator/blob/enable-prometheus/main.go#L133)

ffromani (commented):
/hold
/lgtm

need to be tested d/s before merge. Looks good, but we need the due diligence and I don't have enough bandwidth atm.

rbaturov:
@ffromani A question regarding this work:
How do we ensure the image (gcr.io/kubebuilder/kube-rbac-proxy:v0.15.0) we use for the sidecar is CVE's free/remains secured? Will CVP notify us of it?

main.go
```diff
@@ -73,6 +73,7 @@ const (
 const (
 	defaultWebhookPort = 9443
 	defaultMetricsAddr = ":8080"
+	defaultMetricsEnabled = true
```
ffromani, line 76: let's rename 'defaultMetricsSupport'

config/prometheus/monitor.yaml
```diff
@@ -1,20 +1,24 @@
 
-# Prometheus Monitor Service (Metrics)
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
-  labels:
-    control-plane: controller-manager
-  name: controller-manager-metrics-monitor
+  name: controller-manager
```
ffromani, line 5: this is a pretty generic name for the `system` namespace
we can getaway with `controller-manager` generic name ONLY if we are in a numaresources namespace

config/prometheus/rbac.yaml
```diff
@@ -0,0 +1,31 @@
+# creates Role and RoleBinding for prometheus-k8s service account to access our namespace
```
ffromani, line 1: do we need this only in CI or in production in general?

config/rbac/kustomization.yaml
```diff
@@ -12,7 +12,7 @@ resources:
...
 # the auth proxy (https://github.com/brancz/kube-rbac-proxy)
 # which protects your /metrics endpoint.
-#- auth_proxy_service.yaml
-#- auth_proxy_role.yaml
-#- auth_proxy_role_binding.yaml
-#- auth_proxy_client_clusterrole.yaml
+- auth_proxy_service.yaml
+- auth_proxy_role.yaml
+- auth_proxy_role_binding.yaml
+- auth_proxy_client_clusterrole.yaml
```
ffromani, line 18: same question, are those for CI or for production?

config/prometheus/rbac.yaml
rbaturov: For CI, which don't have prometheus installed we don't need this.
However, for production (OCP) if we opt to allow prometheus to scrape merics by default, we should apply these RBAC's.

config/prometheus/monitor.yaml
rbaturov: But the `system` namespace will always be replaced with `numaresources`.
Do you wish to return to `controller-manager-metrics-monitor`?

config/rbac/kustomization.yaml
rbaturov: `auth_proxy_role.yaml `and `auth_proxy_role_binding.yaml` are mandatory for the sidecar operation. Meaning these two needed for CI (for curl tests) but also for production.
However, `auth_proxy_service.yaml` is needed for production usage, (it is part of the infrastructure to allow Prometheus to scrape metrics)

rbaturov: However, the `auth_proxy_client_clusterrole.yaml` is not being used at all

config/prometheus/monitor.yaml
ffromani: ok, if this is gonna be sitting in the `numaresources` namespace, it's good

config/rbac/kustomization.yaml
ffromani: ok, so let's not add `auth_proxy_client_clusterrole.yaml` or unnecessary stuff in general

ffromani: > `auth_proxy_role.yaml, `auth_proxy_service.yaml`and`auth_proxy_role_binding.yaml` are mandatory for the sidecar operation. Meaning these three needed for CI (for curl tests) but also for production. without them sidecar won't be ready. The service deployment is mandatory as its annotation responsible for creating a secret consumed as a volume by the sidecar.

Could you elaborate about "not be ready"? I'd expect the sidecars to be up and running, but not be accessible by anyone in the cluster without these RBAC rules.

rbaturov: If the service for example won't be created, the secret that we mount as a volume won't be created. Therefore, this would result with an error and the sidecar state won't be ready.

ffromani: ok, thanks. This is a possible problem for the backports, because makes them more invasive than expected.

main.go
rbaturov: Updated code to consume this change

PR #1009, opened 2024-09-12
doc: examples: use ctrreschk releases

ffromani:
let's use releases instead of snapshots, and
the first available release is v0.0.3.

### DIFF
--- a/doc/examples/pod.ctrreschk.yaml
+++ b/doc/examples/pod.ctrreschk.yaml
@@ -6,7 +6,7 @@ spec:
   schedulerName: topo-aware-scheduler
   containers:
   - name: ctrreschk
-    image: quay.io/fromani/ctrreschk:v0.0.2024080204
+    image: quay.io/fromani/ctrreschk:v0.0.3
     imagePullPolicy: Always
     command: ["/usr/local/bin/ctrreschk", "-S", "align"]
     resources:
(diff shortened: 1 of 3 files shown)


Tal-or:
> /test ci-install-e2e

I think there's a problem with the operator sdk cli image I recently added.
Although it surprising, because the other lanes are using the same image.

You welcome to open a thread on the test platform channel, I'll have a deeper look on Sunday

ffromani:
> > /test ci-install-e2e
> 
> I think there's a problem with the operator sdk cli image I recently added. Although it surprising, because the other lanes are using the same image.
> 
> You are welcome to open a thread on the test platform channel, I'll have a deeper look on Sunday

thanks for catching!

Tal-or:
> > > /test ci-install-e2e
> > 
> > 
> > I think there's a problem with the operator sdk cli image I recently added. Although it surprising, because the other lanes are using the same image.
> > You are welcome to open a thread on the test platform channel, I'll have a deeper look on Sunday
> 
> thanks for catching!
> 
> the retry is already past that point though: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-kni_numaresources-operator/1009/pull-ci-openshift-kni-numaresources-operator-main-ci-install-e2e/1834477655957704704

Then it was just a glitch with image and that explains why the other lanes passed.

PR #1011, opened 2024-09-15
 Makefile: bump tooling version

Tal-or:
Bump the following tooling version to the latest stable:
* `operator-sdk`
* `controller-gen`
* `kustomize`

In addition addressed all deprecation messages and other dependencies that
were resulted from the use of the latest versions of the tools.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ CONTAINER_ENGINE ?= docker
 
 BIN_DIR="bin"
 
-OPERATOR_SDK_VERSION="v1.25.0"
+OPERATOR_SDK_VERSION="v1.36.1"
 OPERATOR_SDK_BIN="operator-sdk_$(GOOS)_$(GOARCH)"
 OPERATOR_SDK="$(BIN_DIR)/$(OPERATOR_SDK_BIN)"
 
@@ -289,11 +289,11 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 
 CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
-	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0)
+	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.16.3)
 
 KUSTOMIZE = $(shell pwd)/bin/kustomize
 kustomize: ## Download kustomize locally if necessary.
-	$(call go-install-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.5.4)
+	$(call go-install-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v5@v5.4.3)
 
 # use the last version before the bump golang 1.20 -> 1.22. We want to avoid the go.mod version format changes - for now.
 ENVTEST = $(shell pwd)/bin/setup-envtest
--- a/Dockerfile.bundle
+++ b/Dockerfile.bundle
@@ -7,7 +7,7 @@ LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=numaresources-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
 LABEL operators.operatorframework.io.bundle.channel.default.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.25.0
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.36.1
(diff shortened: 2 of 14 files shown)


ffromani:
/approve

Generally ok as long as the versions are known-compatible with each other, the prject and the kubernetes libs we use.
Granted, staying with the same tooling for too long breaks compatibility as well - backward compatibility in k8s world is not infinite, but still.

Tal-or:
> /approve
> 
> Generally ok as long as the versions are known-compatible with each other, the prject and the kubernetes libs we use. Granted, staying with the same tooling for too long breaks compatibility as well - backward compatibility in k8s world is not infinite, but still.

The three are tight together. I used the latest stable  version for each

Makefile
```diff
@@ -289,11 +289,11 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 
 CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
-	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0)
+	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.16.3)
```
ffromani, line 292: `controller-gen` depends on kube 1.31 (https://github.com/kubernetes-sigs/controller-tools/blob/v0.16.3/go.mod), which is still OK as we are stuck on 1.29.

Makefile
```diff
@@ -60,7 +60,7 @@ CONTAINER_ENGINE ?= docker
 
 BIN_DIR="bin"
 
-OPERATOR_SDK_VERSION="v1.25.0"
+OPERATOR_SDK_VERSION="v1.36.1"
```
ffromani, line 63: this depends on kube 1.29 so it seems it's fine: https://github.com/operator-framework/operator-sdk/releases/tag/v1.36.0

Makefile
```diff
@@ -289,11 +289,11 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 
 CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
-	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0)
+	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.16.3)
 
 KUSTOMIZE = $(shell pwd)/bin/kustomize
 kustomize: ## Download kustomize locally if necessary.
-	$(call go-install-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.5.4)
+	$(call go-install-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v5@v5.4.3)
```
ffromani, line 296: this should be fine and pull no deps I reckon - we use it as command to process YAMLs.

PROJECT
```diff
@@ -1,6 +1,6 @@
 domain: openshift.io
 layout:
-- go.kubebuilder.io/v3
+- go.kubebuilder.io/v4
```
ffromani, line 3: I'm not thrilled about migrating to the new layout (https://book.kubebuilder.io/migration/manually_migration_guide_gov3_to_gov4#changes-to-the-layout) in a future PR of course, but I won't oppose it either.

ffromani (commented):
I like the fact RBAC rules are more compact now, but I need to doublecheck them before I can add my LGTM

PROJECT
Tal-or: > I'm not thrilled about migrating to the new layout (https://book.kubebuilder.io/migration/manually_migration_guide_gov3_to_gov4#changes-to-the-layout) in a future PR of course, but I won't oppose it either.

It can be removed from the current PR.
The reason I added it here is because kubebuilder is complaining about the fact that gov3 is deprecated

ffromani: Yes, let's keep it v3 for now to reflect the actual laout of the project

ffromani:
Flake:
```
   * could not run steps: step [release:latest] failed: failed to get CLI image: unable to tag the 'cli' image into the stable stream: ImageStreamTag.image.openshift.io "stable:cli" is invalid: metadata: Invalid value: "": may not update fields other than metadata.annotations 
INFO[2024-09-16T06:29:12Z] Reporting job state 'failed' with reason 'executing_graph:step_failed:importing_release' 
```

not retesting because we agreed to drop the v3->v4 change, so a resubmit is incoming

Tal-or:
Dropped back from v4 -> v3

ffromani:
interesting. Dropping v4 prevents us to update kustomize :\

Tal-or:
> interesting. Dropping v4 prevents us to update kustomize :\

Nope. this is not the reason. there's a change in the Dockerfile.bundle file because we're using v3 now. Will fix that in a sec

ffromani (commented):
/lgtm

PR #1012, opened 2024-09-16
api: support NodeSelector for NodeGroup

shajmakh:
As a preparation step for supporting TAS on HCP (hosted control planes) known as Hypershift, this PR introduces a new API for selecting nodes on which the RTE will be running, exposing the nodes' topologies.

On OCP:
 The way that cluster nodes are grouped is by using Machine Config Pools (MCPs); On the operator side,
 umaresourcesoperator CR defines an MCP selector that, deep down, uses the node selector on the corresponding MCP and sets it under the RTE daemonset's node selector.

On HCP:
 The MCP term does not exist on hypershift, but instead, the hypershift platform has a node-pool, which essentially groups the nodes based on a specific node-pool label that is added under the nodes' labels.

This PR proposes to enable additional options for selecting node groups to run the RTE pods that will work on both OCP and HCP platforms. The new API does not change how MCP selector works, so it is backward compatible, as it only adds a NodeSelector option under the NodeGroup which will be reflected in the numaresourcesoperator spec. Values for this field are valid node labels.

Restrictions and limitations:
- Only one selector should be specified per node group; more than one will not be accepted and will cause a degraded state.
- No extra validations are applied whether nodes of one node group that specifies MCP Selector correlate with nodes from another that specifies NodeSelector. The user takes responsibility for providing nonconflicting nodes per selector group.
- Like was for MCP selectors, there is no validation on the selected nodes whether or not they have the correct machine configuration needed for TAS to be operational.


api/numaresourcesoperator/v1/helper/nodegroup/nodegroup.go
```diff
@@ -49,9 +49,22 @@ func FindTrees(mcps *mcov1.MachineConfigPoolList, nodeGroups []nropv1.NodeGroup)
...
+			klog.Errorf("only one node group selector is allowed: %+v", nodeGroup)
+			continue
+		}
+
+		if nodeGroup.NodeSelector != nil {
+			result = append(result, Tree{
+				NodeGroup: nodeGroup,
+			})
+			continue
+		}
```
ffromani, line 66: this is good, but seems misplaced. I'd move (a slightly amended version of) this logic in https://github.com/openshift-kni/numaresources-operator/blob/main/pkg/validation/validation.go .

The validation logic is called much earlier in each reconciliation step. Feel free to add this logic into an existing function, or to create a new one.

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup_test.go
```diff
@@ -185,27 +181,127 @@ func TestFindTrees(t *testing.T) {
 				},
 			},
 		},
+		{
```
ffromani, line 184: likewise, good to add these tests (actually: must to have these tests :) ) but they probably better fit in `pkg/validation`

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -117,11 +117,22 @@ type NodeGroup struct {
 	// MachineConfigPoolSelector defines label selector for the machine config pool
 	// +optional
 	MachineConfigPoolSelector *metav1.LabelSelector `json:"machineConfigPoolSelector,omitempty"`
+	// NodeSelector defines label selector for the nodes that will host RTE pods
+	// +optional
+	NodeSelector *NodeSelector `json:"nodeSelector,omitempty"`
```
ffromani, line 122: can this be just a `metav1.LabelSelector`?

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -117,11 +117,22 @@ type NodeGroup struct {
...
 	// Config defines the RTE behavior for this NodeGroup
 	// +optional
 	Config *NodeGroupConfig `json:"config,omitempty"`
 }
 
+type NodeSelector struct {
+	// Name the name of the machine config pool
+	Name string `json:"name"`
```
ffromani, line 130: is not obvious to me why do we need a name here and most notably the "machine config pool" reference

internal/objects/objects.go
```diff
@@ -29,14 +29,7 @@ import (
...
 
-func NewNUMAResourcesOperator(name string, labelSelectors []*metav1.LabelSelector) *nropv1.NUMAResourcesOperator {
-	var nodeGroups []nropv1.NodeGroup
-	for _, selector := range labelSelectors {
-		nodeGroups = append(nodeGroups, nropv1.NodeGroup{
-			MachineConfigPoolSelector: selector,
-		})
-	}
-
+func NewNUMAResourcesOperator(name string, nodeGroups []nropv1.NodeGroup) *nropv1.NUMAResourcesOperator {
```
ffromani, line 32: since this is an internal API, and strictly while this is an internal API, what about
```
func NewNUMAResourcesOperator(name string, nodeGroups ...nropv1.NodeGroup) *nropv1.NUMAResourcesOperator {
// ...
```

ffromani (commented):
The general direction seems promising, but we can use some changes

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
shajmakh: I wanted a way to name the nodeselector daemonset. For MCP selector nodegroup the ds name is unique for every matching MCP because it ends with the MCP name.
I'm against depending on the labelSelector key-vals because it will turn out very long and unpractical, especially when having more than a single key-value pair. 
I am open for suggestions, we can make it serial (ngns1,ngns2...), name it based on date/time or keep it up to the user to define it as suggested here to choose the name of the nodes group.

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
shajmakh: the machine config pool is a copy-paste mistake. 
the name however is needed because we need to identify the ds for such node group. for MCPs we use the value of the nodeslector label of the corresponding MCP. I'm open for suggestions though as stated earlier

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup_test.go
shajmakh: I'll remove redundant tests for the function in nodegroup.go and cover what's missing in validation_test.go

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup.go
shajmakh: removed the reported errors but still kept the conditions

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
ffromani: I think you are right, but the Name should identify a NodeGroup regardless about how we match the nodes (using NodeSelector or MCP), this is why I think it belongs a level above.

Tal-or:
Thank you for working on this.

I'll review it in the following days. 

Next time when a PR with this amount of changes is issued, I would consider decompose it to smaller  commits to make the review process easier.

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup_test.go
```diff
@@ -185,27 +181,106 @@ func TestFindTrees(t *testing.T) {
...
+						NodeSelector: &nropv1.NodeSelector{
+							Name: "ns1",
+							LabelSelector: &metav1.LabelSelector{
+								MatchLabels: map[string]string{
+									"label-1": "test1",
+								},
+							},
+						},
+					},
+					MachineConfigPools: nil,
```
ffromani, line 249: if you explicitely set `MachineConfigPools` nil (which I'm not opposing...)

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup_test.go
```diff
@@ -185,27 +181,106 @@ func TestFindTrees(t *testing.T) {
...
+				},
+				{
+					MachineConfigPools: []*mcov1.MachineConfigPool{
+						{
+							ObjectMeta: metav1.ObjectMeta{
+								Name: "mcp2",
+							},
+						},
+					},
```
ffromani, line 258: ...then you should set `NodeSelector` to nil also explicitely

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
ffromani: In addition, we need to add the new fields in `NUMAResourcesOperatorStatus`

ffromani (commented):
the API side is more complex than expected. Let's pause for a sec and let's review the new API first. This is mostly my fault: I underestimated the effort.

shajmakh:
> In addition, we need to add the new fields in NUMAResourcesOperatorStatus

@ffromani correct. I'll add the complete status reporting to the preparation PR without exceptions (like MCP only).

PR #1018, opened 2024-09-20
api: add `PoolName` and `NodeGroupStatus`

shajmakh:
This PR introduces a new API option for selecting nodes on which the RTE will be running, exposing the nodes' topologies, as a preamble to supporting TAS on HCP (hosted control planes) known as Hypershift.

This PR has two main parts 
Part 1: new `PoolName` field under NodeGroup
Part 2: new `NodeGroupStatus` type and appending it under the operator
status.
For more details on each part, please look at the corresponding commits.


api/numaresourcesoperator/v1/helper/nodegroup/nodegroup.go
```diff
@@ -29,17 +29,14 @@ import (
 )
 
 type Tree struct {
-	NodeGroup          *nropv1.NodeGroup
-	MachineConfigPools []*mcov1.MachineConfigPool
+	NodeGroup         *nropv1.NodeGroup
+	MachineConfigPool *mcov1.MachineConfigPool
```
ffromani, line 33: I'd like to merge this change so much, but we need to make sure len(MachineConfigPools) == 1 always holds true. Are we fully confident this is the case?

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -141,10 +140,6 @@ type NUMAResourcesOperatorStatus struct {
 type MachineConfigPool struct {
 	// Name the name of the machine config pool
 	Name string `json:"name"`
-	// Conditions represents the latest available observations of MachineConfigPool current state.
```
ffromani, line 145: we can't remove fields from the API until we bump the version. The best we can do is not set them (which we already do)

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -120,32 +119,32 @@ type NodeGroup struct {
...
+
 // NUMAResourcesOperatorStatus defines the observed state of NUMAResourcesOperator
 type NUMAResourcesOperatorStatus struct {
-	// DaemonSets of the configured RTEs, one per node group
-	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="RTE DaemonSets"
-	DaemonSets []NamespacedName `json:"daemonsets,omitempty"`
-	// MachineConfigPools resolved from configured node groups
-	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="RTE MCPs from node groups"
-	MachineConfigPools []MachineConfigPool `json:"machineconfigpools,omitempty"`
-	// Conditions show the current state of the NUMAResourcesOperator Operator
```
ffromani, line 133: we can't remove fields from the API until we bump the version. The best we can do is to not set any values to them. In this case we need to keep setting these values for backward compatibility.

ffromani (commented):
I think we can use some trajectory adjustements, but looks a nice preparation step and I agree with the general direction

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
shajmakh: I appreciate your comment. These are in the operator status which the operator user does not use to configure the operator. I have updated the conversion between the API versions to match what's needed in the new fields. Do you think these modifications are not enough to use the new instance of the version? 
It will help if you could provide an example on how the compatibility is affected by this change

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup.go
ffromani: Offline I wrote (and I'm reiterating here) I can't imagine a usecase on which len(MachineConfigPools) > 1 and I can't imagine a valid usecase for this. But still, this is what we had till now so if we want to make stricter (which I'd VERY MUCH like to do!) we need to take extra care to not break possible users.
If we manage to demonstrate len(MachineConfigPools) == 1 alwyas carefully analyzing the code and the possible flows, I'm all to merge this change.
I think this should be done in a separate (and welcome) PR though

ffromani: I want this change very badly, but we need to tread carefully. So what I'd do is

1. Add a `NodeGroupName` back in the `MachineConfigPool` struct so we can now backlink MachineConfigPools to their NodeGroup should a NodeGroup match more than a MachineConfigPool. It really should not, if does it's unintended and a bug, but still
2.  *obviously in another PR* keep the NodeGroupTree as-is, but if we discover a NodeGroup matches more than a MachineConfigPool, do everything like today, but set the status.Conditions as Degraded, reporting that `NodeGroup XYZ matches N MachineConfigPool, which is unsupported. Please fix the config.
3. Optionally (but I'm inclined to do it) add a magic annotation to silence the warning.
4. After 2 releases (4.20), actually apply this change right here and force 1:1 between NodeGroups and MachineConfigPools

this is clumsy and unfortunate, but in order to maximize the backward compatibility and fix our code while keeping users happy, it seems the best we can do atm.

Dockerfile.bundle
```diff
@@ -7,7 +7,7 @@ LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=numaresources-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
 LABEL operators.operatorframework.io.bundle.channel.default.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.36.1
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.25.0
```
ffromani, line 10: minor: why going back?

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup.go
```diff
@@ -72,6 +75,13 @@ func FindTrees(mcps *mcov1.MachineConfigPoolList, nodeGroups []nropv1.NodeGroup)
 			return nil, fmt.Errorf("failed to find MachineConfigPool for the node group with the selector %q", nodeGroup.MachineConfigPoolSelector.String())
 		}
 
+		// generate name if not provided
+		trimmedName := strings.TrimSpace(nodeGroup.Name)
+		trimmedName = strings.Replace(trimmedName, " ", "", -1)
+		if trimmedName == "" {
+			nodeGroup.Name = guuid.New().String()
+		}
+
```
ffromani: let's do this in the normalize package. Once the code is moved should be OK conceptually. I'm not sure we would use a uuid here.

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup_test.go
```diff
@@ -198,6 +198,12 @@ func TestFindTrees(t *testing.T) {
 				t.Errorf("Trees mismatch: got=%v expected=%v", gotNames, expectedNames)
 			}
 
+			for _, tree := range got {
+				if tree.NodeGroup.Name == "" {
+					t.Errorf("node group name must be filled for all node groups")
+				}
+			}
+
```
ffromani, line 251: conceptually OK but this should be moved in the validation package

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -114,6 +114,10 @@ type NodeGroupConfig struct {
 // NodeGroup defines group of nodes that will run resource topology exporter daemon set
 // You can choose the group of node by MachineConfigPoolSelector or by NodeSelector
 type NodeGroup struct {
+	// Name is the name used to identify this node group. Has to be unique among node groups.
+	// If not provided, is determined by the system.
+	// +optional
+	Name string `json:"name,omitempty"`
```
ffromani, line 120: let's use a string pointer so it's obvious when user set it and when not (also, "" is illegal: if user bothers to set, it should set a valid value. Using a pointer we can distinguish it)

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -150,6 +154,8 @@ type MachineConfigPool struct {
 	// +optional
 	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="Optional configuration enforced on this NodeGroup"
 	Config *NodeGroupConfig `json:"config,omitempty"`
+	// NodeGroupName the name of the node group this MCP belongs to
+	NodeGroupName string `json:"nodeGroupName,omitempty"`
```
ffromani, line 191: ok, this is generated by the system so if we get this far it MUST be present. But still we need to mark as optional because it was not present in the earlier v1 API, so a well-behaving client is not required to consume it. Hence must be optional even if in practice it will always be present starting 4.18

config/crd/bases/nodetopology.openshift.io_numaresourcesoperators.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.16.3
+    controller-gen.kubebuilder.io/version: v0.14.0
```
ffromani, line 6: let's no go back in time please

ffromani (commented):
partial review

Dockerfile.bundle
shajmakh: this was generated automatically by running `make generate bundle manifests`. I'll check and fix it, thanks for raising.

ffromani:
the `security` lane failure is real, the other are due to a known issue @Tal-or is already working on

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup.go
```diff
@@ -59,31 +59,44 @@ func (ttr Tree) Clone() Tree {
 // One of the key design assumptions in NROP is the 1:1 mapping between NodeGroups and MCPs.
 // This historical accident should be fixed in future versions.
 func FindTrees(mcps *mcov1.MachineConfigPoolList, nodeGroups []nropv1.NodeGroup) ([]Tree, error) {
+	// node groups are validated by the controller before getting to this phase, so for sure all node groups will be valid at this point.
+	// a valid node group has either PoolName OR MachineConfigPoolSelector, not both. Getting hear means operator is deployed on Openshift thus processing MCPs
```
ffromani, line 63: typo: hear -> here

pkg/validation/validation.go
```diff
@@ -75,18 +79,21 @@ func NodeGroups(nodeGroups []nropv1.NodeGroup) error {
 }
 
 // TODO: move it under the validation webhook once we will have one
-func nodeGroupsMachineConfigPoolSelector(nodeGroups []nropv1.NodeGroup) error {
+func nodeGroupPools(nodeGroups []nropv1.NodeGroup) error {
 	for _, nodeGroup := range nodeGroups {
-		if nodeGroup.MachineConfigPoolSelector == nil {
-			return fmt.Errorf("one of the node groups does not have machineConfigPoolSelector")
+		if nodeGroup.MachineConfigPoolSelector == nil && nodeGroup.PoolName == nil {
+			return fmt.Errorf("one of the node groups does not set a pool specifier")
```
ffromani, line 87: I'd say "nodegroup missing any pool specifier"
we can do
```
for idx, nodeGroup := range nodeGroups {
    if nodeGroup.MachineConfigPoolSelector == nil && nodeGroup.PoolName == nil {
			return fmt.Errorf("node group %d missing any pool specifier", idx)
    }
```

to perhaps help user a bit

pkg/validation/validation.go
```diff
@@ -75,18 +79,21 @@ func NodeGroups(nodeGroups []nropv1.NodeGroup) error {
...
+		}
+		if nodeGroup.MachineConfigPoolSelector != nil && nodeGroup.PoolName != nil {
+			return fmt.Errorf("one of the node groups specify more than one pool specifier while only one is allowed")
```
ffromani, line 90: same, let's rephrase as "node group %d must have only a single specifier set"

pkg/validation/validation.go
```diff
@@ -114,6 +121,35 @@ func nodeGroupsDuplicates(nodeGroups []nropv1.NodeGroup) error {
 	return nil
 }
 
+// TODO: move it under the validation webhook once we will have one
```
ffromani, line 126: all these TODOs can be removed, adding a webhook is not in our plans

pkg/validation/validation.go
```diff
@@ -114,6 +121,35 @@ func nodeGroupsDuplicates(nodeGroups []nropv1.NodeGroup) error {
...
+func nodeGroupsDuplicatesByPoolName(nodeGroups []nropv1.NodeGroup) error {
+	duplicates := map[string]int{}
+	for _, nodeGroup := range nodeGroups {
+		if nodeGroup.PoolName == nil {
+			continue
```
ffromani, line 131: this must never happen so it's worth a log (`klog.InfoS`)

pkg/validation/validation.go
```diff
@@ -114,6 +121,35 @@ func nodeGroupsDuplicates(nodeGroups []nropv1.NodeGroup) error {
...
+
+	var duplicateErrors []string
+	for name, count := range duplicates {
+		if count > 1 {
+			duplicateErrors = append(duplicateErrors, fmt.Sprintf("the pool name %q has duplicates", name))
+		}
+	}
+
+	if len(duplicateErrors) > 0 {
+		return errors.New(strings.Join(duplicateErrors, "; "))
```
ffromani, line 149: I think there's some `multiError` code somewhere. But can be improved later.

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup_test.go
```diff
@@ -420,3 +461,20 @@ func findListByNodeGroups(mcps *mcov1.MachineConfigPoolList, nodeGroups []nropv1
 
 	return result, nil
 }
+
+// containedItemsFromSlice returns true if all of b items are part of slice a
+func containsSliceItems(a, b []string) bool {
```
ffromani, line 466: we can use the `slices` package  I think: https://pkg.go.dev/slices@go1.22.0
(and yes, in the future we should scan our codebase and move possibly all of the places to `slices`)

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -125,6 +125,28 @@ type NodeGroup struct {
...
+// which in turn correctly references unambiguously a set of nodes in the cluster.
+// Hence, if a NodeGroupStatus is published, its `Name` must be present, because it refers back to a NodeGroup whose
+// config was correctly processed in the Spec. And its DaemonSet will be nonempty, because matches correctly a set
+// of nodes in the cluster. The Config is best-effort always represented, possibly reflecting the system defaults.
+// If the system cannot process a NodeGroup correctly from the Spec, it will report Degraded state in the top-level
+// condition, and will provide details using the aforementioned conditions.
+type NodeGroupStatus struct {
+	// DaemonSet of the configured RTEs, for this node group
+	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="RTE DaemonSets"
+	DaemonSet *NamespacedName `json:"daemonsets,omitempty"`
```
ffromani, line 139: not completely sure. If we publish a `NodeGroupStatus`, it MUST have a valid DaemonSet.

controllers/numaresourcesoperator_controller.go
```diff
@@ -295,6 +297,30 @@ func (r *NUMAResourcesOperatorReconciler) syncDaemonSetsStatuses(ctx context.Con
...
+		for idx, ng := range instance.Status.NodeGroups {
+			if ng.PoolName == extractedPoolName {
+				instance.Status.NodeGroups[idx].DaemonSet = copyOfDS
+				break
+			}
+		}
+	}
+	return instance.Status.NodeGroups
+}
+
```
ffromani, line 323: this looks not wrong but we can probably set it straight away instead of searching it back.
But it's an optimization/cleanup we can do later.

pkg/objectnames/objectnames.go
```diff
@@ -30,3 +33,7 @@ func GetMachineConfigName(instanceName, mcpName string) string {
 func GetComponentName(instanceName, mcpName string) string {
 	return fmt.Sprintf("%s-%s", instanceName, mcpName)
 }
+
+func ExtractPoolNameFromRTEDaemonset(dsName, instanceName string) string {
+	return strings.Replace(dsName, fmt.Sprintf("%s-", instanceName), "", 1)
```
ffromani, line 38: `strings.Cut` ? https://pkg.go.dev/strings#Cut

ffromani (commented):
partial review, but looks nice.

controllers/numaresourcesoperator_controller.go
ffromani: there's another way which includes changes to `reconcileResourcesDaemonSet` and which I'd like. We can generate quite easily the mapping between nodegroups and daemonsets while we build them. We can start with something like
```
diff --git a/controllers/numaresourcesoperator_controller.go b/controllers/numaresourcesoperator_controller.go
index 6e2be478..46e16937 100644
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -462,8 +462,21 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
        }
 
+       dssByMCP := make(map[string]nropv1.NamespacedName)
+       processor := func(mcpName string, gdm *rtestate.GeneratedDesiredManifest) error {
+               err := daemonsetUpdater(mcpName, gdm)
+               if err != nil {
+                       return err
+               }
+               dssByMCP[mcpName] = nropv1.NamespacedName{
+                       Namespace: gdm.DaemonSet.GetNamespace(),
+                       Name:      gdm.DaemonSet.GetName(),
+               }
+               return nil
+       }
+
        existing := rtestate.FromClient(ctx, r.Client, r.Platform, r.RTEManifests, instance, trees, r.Namespace)
-       for _, objState := range existing.State(r.RTEManifests, daemonsetUpdater) {
+       for _, objState := range existing.State(r.RTEManifests, processor) {
                if objState.Error != nil {
                        // We are likely in the bootstrap scenario. In this case, which is expected once, everything is fine.
                        // If it happens past bootstrap, still carry on. We know what to do, and we do want to enforce the desired state.
```
with a mapping like` dssByMCP` (trivial to extend to `dssByNodeGroups`, you name it, this code should become simpler

This can be cleaned up later, though

pkg/validation/validation.go
shajmakh: Thanks for the comment; depending on the direction we went with, the user should set either MCPselector or PoolName, so one of them should be nil. We do not map MCPSelector to PoolName (and we can't at the moment because matching mcps can be > one).
This function is checking the NodeGroupSpec where PoolName is a pointer. however on NodeGroupStatus PoolName must be set (derived from the trees). can you elaborate more please on what I'm missing for this condition to never be true?

ffromani: Right, we must still support identification by MCPSelector. Hence the code looks correct as it is

pkg/validation/validation.go
```diff
@@ -169,3 +171,16 @@ func nodeGroupMachineConfigPoolSelector(nodeGroups []nropv1.NodeGroup) error {
 
 	return nil
 }
+
+// EqualNamespacedDSSlicesByName validates two slices of type NamespacedName are equal in Names
+func EqualNamespacedDSSlicesByName(s1, s2 []nropv1.NamespacedName) error {
```
Tal-or, line 176: Wouldn't it be simpler to store the DaemonSets in map?

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
shajmakh: I'm not sure I understand where this type can be improved. Can you explain more please?

ffromani: The type is fine, but this is likely (not 100% sure yet) mandatory, so can be a field, not a pointer

shajmakh: correct, DS isn't optional, I'll make this a struct type instead. thanks!

ffromani: please hold on a sec, decision not yet final and I want to take some time to review the API to avoid more churn

pkg/validation/validation.go
ffromani: no maps in the API, but we can use them freely internally if helps

pkg/objectnames/objectnames.go
shajmakh: no longer needed,but thanks for the info

pkg/validation/validation.go
shajmakh: No longer needed, it was replaceable with reflect.DeepEqual()

shajmakh:
Controller tests were lost; I'm working on fixing

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup.go
```diff
@@ -59,31 +59,44 @@ func (ttr Tree) Clone() Tree {
...
+		if nodeGroup.PoolName != nil {
+			for i := range mcps.Items {
+				mcp := &mcps.Items[i]
+				if mcp.Name == *nodeGroup.PoolName {
+					treeMCPs = append(treeMCPs, mcp)
+					// MCO ensures there are no mcp name duplications
+					break
+				}
+			}
+			// it is valid if no match was found
```
shajmakh, line 78: @ffromani is this how we want to continue or should we return an error here like 
https://github.com/openshift-kni/numaresources-operator/pull/1018/commits/168e8ed3ea965f71baeece9e2b0b74aa00f1cb3f#diff-31ede3f56b1f873ccf421371942426d8ad0c30160d467d790dd7e050802d25d7R97-R98 ?

ffromani: yes please

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup_test.go
```diff
@@ -420,3 +462,13 @@ func findListByNodeGroups(mcps *mcov1.MachineConfigPoolList, nodeGroups []nropv1
...
+	for _, item := range b {
```
ffromani, line 468: ok, this can be optimized/cleaned up further later on

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup.go
```diff
@@ -75,7 +75,10 @@ func FindTrees(mcps *mcov1.MachineConfigPoolList, nodeGroups []nropv1.NodeGroup)
 					break
 				}
 			}
-			// it is valid if no match was found
+			// is it valid if no match was found?
+			//if len(treeMCPs) == 0 {
+			//	return nil, fmt.Errorf("failed to find MachineConfigPool for the node group with the PoolName %q", nodeGroup.PoolName)
+			//}
```
ffromani, line 81: let's fix this, commented code should be removed (or uncommented!)

controllers/numaresourcesoperator_controller_test.go
```diff
@@ -106,36 +106,94 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
...
+			verifyDegradedCondition(nro, validation.NodeGroupsError)
+		})
+		//It("should update the CR condition to degraded when PoolName set", func() {
+		//	pn := "pn-1"
+		//	ng := nropv1.NodeGroup{
+		//		PoolName: &pn,
+		//	}
+		//	nro := testobjs.NewNUMAResourcesOperator(objectnames.DefaultNUMAResourcesOperatorCrName, ng)
+		//	verifyDegradedCondition(nro, validation.NodeGroupsError)
+		//})
```
ffromani, line 151: likewise, let's fix this commented code

ffromani (commented):
I'm inclined to leave all the fields but `PoolName` optional in the `NodeGroupStatus` even if we will always populate them. There are pending comments to address, so let's think one last time about it before to make the final decision and than merge.

Changing from/to pointer to value fields should be a quick change, so hopefully it won't be too much churn if we change direction.

Everything else LGTM. Once the few pending comments are addresed and once we settle the status fields types conversation, we can merge.

controllers/numaresourcesoperator_controller_test.go
shajmakh: sure, will remove it, I added them like this because I wasn't clear what direction we want here: https://github.com/openshift-kni/numaresources-operator/pull/1018#discussion_r1786005282
but now I'll remove them. thanks for the review and comments

controllers/numaresourcesoperator_controller.go
shajmakh: thanks for raising, addressed!

shajmakh:
> I'm inclined to leave all the fields but PoolName optional in the NodeGroupStatus even if we will always populate them. There are pending comments to address, so let's think one last time about it before to make the final decision and than merge.

I agree with this direction (i.e. making all NodeGroupStatus fields as optional except the PoolName) for mainly below two reasons:
1. if we set a NodeGroupStatus it has to have a PoolName which we in fact copy from the MCPs status
2. having a PoolName set doesn't assume that the config and the daemonset would be set, because according to the reconcile loop this would require the MC to be updated as the resource of the corresponding MCP.

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup_test.go
shajmakh: updated, thanks!

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup.go
```diff
@@ -59,31 +59,44 @@ func (ttr Tree) Clone() Tree {
...
+				mcpLabels := labels.Set(mcp.Labels)
+				if !selector.Matches(mcpLabels) {
+					continue
+				}
+				treeMCPs = append(treeMCPs, mcp)
+			}
+
+			if len(treeMCPs) == 0 {
+				return nil, fmt.Errorf("failed to find MachineConfigPool for the node group with the selector %q", nodeGroup.MachineConfigPoolSelector.String())
+			}
```
ffromani, line 99: sorry, I got your previous question wrong. We need to be consistent in both branches. If we do this validation we need to do some validation also in the other branch. Or, equally good, we need a common validation towards the end of the function.

ffromani (commented):
can we remove WIP from the PR title?
possibly the last comment inside. 
Still making my mind about the status, reviewing the recommendations

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup.go
shajmakh: got it, I'll update that, thank for another review round!

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup_test.go
```diff
@@ -203,9 +244,9 @@ func TestFindTrees(t *testing.T) {
 			if err != nil {
 				t.Errorf("unexpected error checking backward compat: %v", err)
 			}
-			compatNames := mcpNamesFromList(gotMcps)
-			if !reflect.DeepEqual(gotNames, compatNames) {
-				t.Errorf("Trees mismatch (non backward compatible): got=%v compat=%v", gotNames, compatNames)
+			compatibleNames := mcpNamesFromList(gotMcps)
+			if !containsSliceItems(gotNames, compatibleNames) {
```
Tal-or, line 248: You can use `set.Set` for finding the matches. 
```
A := set.New(gotNames...)
if !A.HasAll(compatibleNames...)
...
```

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup_test.go
```diff
@@ -420,3 +461,51 @@ func findListByNodeGroups(mcps *mcov1.MachineConfigPoolList, nodeGroups []nropv1
 
 	return result, nil
 }
+
+func TestContainsSliceItems(t *testing.T) {
```
Tal-or, line 465: if you're using `set.Set` you don't need the `containsSliceItems` function and this unit test either

controllers/numaresourcesoperator_controller.go
```diff
@@ -274,25 +287,48 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
...
-		dsStatuses = append(dsStatuses, nname)
+		dssWithReadyStatus[mcp] = nname
 	}
-	return dsStatuses, true, nil
+	return dssWithReadyStatus, true, nil
+}
+
+func syncNodeGroupsStatusPerMCPWithOperatorStatus(instance *nropv1.NUMAResourcesOperator, syncConfig bool) []nropv1.NodeGroupStatus {
+	ngStatuses := []nropv1.NodeGroupStatus{}
+	for _, mcp := range instance.Status.MachineConfigPools {
```
Tal-or, line 313: so this logic is only valid for OpenShift and we'll add the HyperShift logic later on?

controllers/numaresourcesoperator_controller.go
```diff
@@ -222,29 +222,34 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResourceMachineConfig(ctx con
 	// It can take a while.
 	mcpStatuses, allMCPsUpdated := syncMachineConfigPoolsStatuses(instance.Name, trees, r.ForwardMCPConds)
 	instance.Status.MachineConfigPools = mcpStatuses
+	instance.Status.NodeGroups = syncNodeGroupsStatusPerMCPWithOperatorStatus(instance, false)
```
Tal-or, line 225: sorry for my lack of understanding, but why do we need to call this function twice (here and in line 232) and one time with `false` and then with `true`? 
this deserve at least a comment IMO

ffromani: good catch. This deserves a comment and another honest attempt at simplifying the code.

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
ffromani: So, the guiding principle here is https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required .

If we publish a `NodeGroupStatus`, is because the operator is managing that `NodeGroup` with a RTE.
While NROP exists and manages `NUMAResourcesOperator` objects, is meant to manage RTEs through daemonset. This field is thus required, not optional, thus it should *not* be a pointer *nor* it should be `omitempty`.

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -125,6 +125,28 @@ type NodeGroup struct {
...
+// condition, and will provide details using the aforementioned conditions.
+type NodeGroupStatus struct {
+	// DaemonSet of the configured RTEs, for this node group
+	// +optional
+	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="RTE DaemonSets"
+	DaemonSet *NamespacedName `json:"daemonsets,omitempty"`
+	// NodeGroupConfig represents the latest available configuration applied to this NodeGroup
+	// +optional
+	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="Optional configuration enforced on this NodeGroup"
+	Config *NodeGroupConfig `json:"config,omitempty"`
```
ffromani, line 144: Likewise, every `NodeGroup` managed by the Operator because it is reconciling a `NUMAResourcesOperator` object must have a `NodeGroupConfig`. If not supplied by the user, we are using defaults, which we must publish.
This field is thus required, not optional, thus it should _not_ be a pointer _nor_ it should be `omitempty`.

ffromani (commented):
I reviewed again https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md and come to a conclusion regarding the new API types.

I realized we (mostly me) made some mistakes in the previous API iteration, but this is no reason to keep making the same mistakes. Let's do better from now on.

controllers/numaresourcesoperator_controller.go
shajmakh: correct, this will be handled here based on our conversations to separate the content of the OCP vs HCP:
https://github.com/openshift-kni/numaresources-operator/pull/1027

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup_test.go
shajmakh: This indeed looks better and shorter, if I understand correctly this is the set util you are referring to:
https://github.com/kubernetes/client-go/blob/v1.5.2/1.5/pkg/util/sets/string.go#L69 
I see that ultimately the method is using similar principles of map[string]struct{}, and my question is whether we want to add a dependency or reimplement this as suggested in this PR? 
I'm still working on how to identify in which cases we need to add a dependency and in which we would go with reimplementation, but for this I can say that is not required in other places in the project and the implementation and testing is very simple so why to increase our dependencies?.
Please share your thoughts on this I'd like to learn more.

ffromani: the question is very correct, but in this case we already depend on `client-go`, and we use sets already in the codebase. You can use `k8s.io/apimachinery/pkg/util/sets`, same reasoning. please see 
`pkg/validator/validator.go` and `internal/schedcache/synced.go` for example.

controllers/numaresourcesoperator_controller.go
shajmakh: good point. I was following the rules by which the MCP status would get updated, which is only if the MC is updated then update the corresponding NodeGroupConfig in the status, hence the forwarded boolean value.
I agree the code in this area needs can be improved and also adding a way to test the config status update in the controller tests. I'll open an issue to track this work.

shajmakh: Glancing at the open issues, I see this:https://github.com/openshift-kni/numaresources-operator/issues/1031 which sounds to me it includes the subject raised here?

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup_test.go
ffromani: > I'm still working on how to identify in which cases we need to add a dependency and in which we would go with reimplementation, but for this I can say that is not required in other places in the project and the implementation and testing is very simple so why to increase our dependencies?

Unfortunately there's no hard rule and ideas vary wildly. Everyone has to make their own mind.
Few questions may give some ideas
1. is the new proposed dep part of a module we pull already? (e.g. apimachinery/pkg/util/set while we already depend on apimachinery)
2. how happy are we with the same module? are we happy to stay with that dep or we have a general sentiment to migrate? (if not completely happy let's first stop adding, so the transition is easier)
3. how widespread is that new dep? (e.g. in the k8s ecosystem world everyone is using it, so all other factors being equal using the same code makes our project easier to understand because it looks like the others)
4. (from now on the usual suspects) how hard is to reimplement that dep? if we use 2 functions but these 2 functions are heavily optimized compression or crypto algos, well :)
5. how heavy (code size) is the new dep? Do we need to pull in 10 megs and 100k LoCs just for 2 functions?
6. how well maintained is the new dep?

shajmakh: Thanks @ffromani for the tips that should be handy going forward to help me decide in similar situation.
I'll be using the apimachinary packages in this case, thank you for the references.

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
shajmakh: makes sense, adjusted in latest updates and updated controller logic when publishing NodeGroupStatus. Thanks!

api/numaresourcesoperator/v1/helper/nodegroup/nodegroup_test.go
shajmakh: adjusted in the latest updates. Thanks!

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
shajmakh: Adjusted in the latest updates and updated controller logic when publishing NodeGroupStatus to be after all DSs and MCP are in sync. Thanks!

shajmakh:
failed to pull upi-installer image
/retest

ffromani (commented):
/approve
/lgtm

ffromani:
/retest

the candidate fix should have been merged few mins ago

shajmakh:
@ffromani could you please share what fixed the CI issue?

PR #1020, opened 2024-09-24
api: docs: start documenting nodegroup and tree

ffromani:
doc-only change


ffromani:
a doc only change can't break anything, and the CI is currently failing for known reasons

PR #1025, opened 2024-09-24
CNF-14084: controller: update MachineConfig reconciliation

Tal-or:
Starting from 4.18 RTE pods can use a built-in SELinux policy instead
of custom one.

This means that MachineConfig deployment is not mandatory anymore.
However, it is still an option to deploy a custom policy using MachineConfig.

This commit updates the MahcineConfig reconciliation logic.

By default when upgrade is done from 4.1X -> 4.18 the controller
tries to remove the redundant MahcineConfig, unless user state
explicitly that custom policy is needed.

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -336,7 +336,7 @@ func (r *NUMAResourcesOperatorReconciler) syncMachineConfigs(ctx context.Context
 	// so the MachineConfig which applies the custom policy is no longer necessary.
 	// In case of operator upgrade from 4.1X → 4.18, it's necessary to remove the old MachineConfig,
 	// unless an emergency annotation is provided which forces the operator to use custom policy
-	if annotations.IsCustomPolicyEnabled(instance.Annotations) {
+	if !annotations.IsCustomPolicyEnabled(instance.Annotations) {
 		for _, objState := range objStates {
 			if !objState.IsNotFoundError() {
 				klog.V(4).InfoS("delete Machine Config", "MachineConfig", objState.Desired.GetName())
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -22,10 +22,12 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/k8stopologyawareschedwg/deployer/pkg/assets/selinux"
 	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
 	"github.com/k8stopologyawareschedwg/deployer/pkg/manifests"
 	apimanifests "github.com/k8stopologyawareschedwg/deployer/pkg/manifests/api"
 	rtemanifests "github.com/k8stopologyawareschedwg/deployer/pkg/manifests/rte"
+	k8swgrteupdate "github.com/k8stopologyawareschedwg/deployer/pkg/objectupdate/rte"
 	securityv1 "github.com/openshift/api/security/v1"
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	"github.com/pkg/errors"
@@ -52,6 +54,7 @@ import (
 
 	nropv1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1"
 	nodegroupv1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1/helper/nodegroup"
+	"github.com/openshift-kni/numaresources-operator/internal/api/annotations"
 	"github.com/openshift-kni/numaresources-operator/internal/relatedobjects"
 	"github.com/openshift-kni/numaresources-operator/pkg/apply"
 	"github.com/openshift-kni/numaresources-operator/pkg/hash"
@@ -80,6 +83,8 @@ type NUMAResourcesOperatorReconciler struct {
 	Recorder        record.EventRecorder
 	ForwardMCPConds bool
 }
+
+type mcpWaitForUpdatedFunc func(string, *machineconfigv1.MachineConfigPool) bool
 
 // TODO: narrow down
 
@@ -211,16 +216,16 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResourceAPI(ctx context.Conte
 func (r *NUMAResourcesOperatorReconciler) reconcileResourceMachineConfig(ctx context.Context, instance *nropv1.NUMAResourcesOperator, trees []nodegroupv1.Tree) (bool, ctrl.Result, string, error) {
 	// we need to sync machine configs first and wait for the MachineConfigPool updates
 	// before checking additional components for updates
-	_, err := r.syncMachineConfigs(ctx, instance, trees)
+	mcpUpdatedFunc, err := r.syncMachineConfigs(ctx, instance, trees)
 	if err != nil {
 		r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedMCSync", "Failed to set up machine configuration for worker nodes: %v", err)
 		return true, ctrl.Result{}, status.ConditionDegraded, errors.Wrapf(err, "failed to sync machine configs")
 	}
 	r.Recorder.Eventf(instance, corev1.EventTypeNormal, "SuccessfulMCSync", "Enabled machine configuration for worker nodes")
 
-	// MCO need to update SELinux context and other stuff, and need to trigger a reboot.
+	// MCO needs to update the SELinux context removal and other stuff, and need to trigger a reboot.
 	// It can take a while.
-	mcpStatuses, allMCPsUpdated := syncMachineConfigPoolsStatuses(instance.Name, trees, r.ForwardMCPConds)
+	mcpStatuses, allMCPsUpdated := syncMachineConfigPoolsStatuses(instance.Name, trees, r.ForwardMCPConds, mcpUpdatedFunc)
 	instance.Status.MachineConfigPools = mcpStatuses
 	if !allMCPsUpdated {
 		// the Machine Config Pool still did not apply the machine config, wait for one minute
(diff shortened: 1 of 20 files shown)


ffromani:
> Starting from 4.18 RTE pods can use a built-in SELinux policy instead of custom one.
> 
> This means that MachineConfig deployment is not mandatory anymore. However, it is still an option to deploy a custom policy using MachineConfig.
> 
> This commit updates the MahcineConfig reconciliation logic.
> 
> By default when upgrade is done from 4.1X -> 4.18 the controller tries to remove the redundant MahcineConfig, unless user state explicitly that custom policy is needed.

this could be the bit that pushes us to v2.
New deployments should use the new builtin policy
Old deployments should keep using the old one till they upgrade explicitely

But then we have conflicting defaults :\

Tal-or:
> > Starting from 4.18 RTE pods can use a built-in SELinux policy instead of custom one.
> > This means that MachineConfig deployment is not mandatory anymore. However, it is still an option to deploy a custom policy using MachineConfig.
> > This commit updates the MahcineConfig reconciliation logic.
> > By default when upgrade is done from 4.1X -> 4.18 the controller tries to remove the redundant MahcineConfig, unless user state explicitly that custom policy is needed.
> 
> this could be the bit that pushes us to v2. New deployments should use the new builtin policy Old deployments should keep using the old one till they upgrade explicitely
> 
> But then we have conflicting defaults :\

Let's discuss it more deeply offline

internal/api/features/types.go
```diff
@@ -23,7 +23,8 @@ import (
 )
 
 const (
-	Version = "v4.18.0"
+	Version                            = "v4.18.0"
+	ForceCustomSELinuxPolicyAnnotation = "force-custom-selinux-policy"
```
Tal-or, line 27: The API/name is open for discussion.

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -189,3 +190,8 @@ func (ngc *NodeGroupConfig) ToString() string {
 	}
 	return ""
 }
+
+func (nro *NUMAResourcesOperator) IsCustomPolicyEnabled() bool {
+	_, ok := nro.Annotations[features.ForceCustomSELinuxPolicyAnnotation]
+	return ok
+}
```
ffromani, line 197: this should be a free function in a different package, perhaps `internal/api/annotations` again

ffromani: The rationale is that all the magic annotation constants should not be exposed in the public packages. Users will see only the magic strings, but they won't be part of the API contracts now (preferibily, never)

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -22,6 +22,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/openshift-kni/numaresources-operator/internal/api/features"
```
ffromani, line 25: let's avoid this, will make the package non importable: https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit

controllers/numaresourcesoperator_controller.go
```diff
@@ -318,41 +320,54 @@ func (r *NUMAResourcesOperatorReconciler) syncNodeResourceTopologyAPI(ctx contex
...
-		if err2 := controllerutil.SetControllerReference(instance, objState.Desired, r.Scheme); err2 != nil {
-			err = errors.Wrapf(err2, "failed to set controller reference to %s %s", objState.Desired.GetNamespace(), objState.Desired.GetName())
-			break
+	// Since 4.18 we're using a built-in SELinux policy,
+	// so the MachineConfig which applies the custom policy is no longer necessary.
+	// In case of operator upgrade from 4.1X → 4.18, it's necessary to remove the old MachineConfig,
+	// unless an emergency annotation is provided which forces the operator to use custom policy
+	if !instance.IsCustomPolicyEnabled() {
+		for _, objState := range objStates {
+			if !objState.IsNotFoundError() {
```
ffromani, line 340: this looks fragile. We want to target only the `MachineConfig`. So it's probably better to add new functionalities to `pkg/objectstate/rte`

controllers/numaresourcesoperator_controller.go
```diff
@@ -318,41 +320,54 @@ func (r *NUMAResourcesOperatorReconciler) syncNodeResourceTopologyAPI(ctx contex
...
+			}
 
-		_, updated, err2 := apply.ApplyObject(ctx, r.Client, objState)
-		if err2 != nil {
-			err = errors.Wrapf(err2, "could not apply (%s) %s/%s", objState.Desired.GetObjectKind().GroupVersionKind(), objState.Desired.GetNamespace(), objState.Desired.GetName())
-			break
-		}
-		if !updated {
-			continue
+			_, _, err2 := apply.ApplyObject(ctx, r.Client, objState)
```
ffromani, line 361: let's see if we can add `apply.DeleteObject`. I'll think about it

controllers/numaresourcesoperator_controller.go
```diff
@@ -318,41 +320,54 @@ func (r *NUMAResourcesOperatorReconciler) syncNodeResourceTopologyAPI(ctx contex
...
+			if err2 != nil {
+				err = errors.Wrapf(err2, "could not apply (%s) %s/%s", objState.Desired.GetObjectKind().GroupVersionKind(), objState.Desired.GetNamespace(), objState.Desired.GetName())
+				break
+			}
 		}
-		updatedCount++
+		mcpUpdatedFunc = IsMachineConfigPoolUpdated
```
ffromani, line 367: nice idea

pkg/objectstate/rte/rte.go
```diff
@@ -91,19 +91,17 @@ func (em *ExistingManifests) MachineConfigsState(mf rtemanifests.Manifests) []ob
...
-			ret = append(ret,
-				objectstate.ObjectState{
-					Existing: existingMachineConfig.machineConfig,
-					Error:    existingMachineConfig.machineConfigError,
-					Desired:  desiredMachineConfig,
-					Compare:  compare.Object,
-					Merge:    merge.ObjectForUpdate,
-				},
-			)
+			ret = append(ret, os)
```
ffromani, line 102: why this change?

controllers/numaresourcesoperator_controller.go
```diff
@@ -771,6 +773,15 @@ func daemonsetUpdater(mcpName string, gdm *rtestate.GeneratedDesiredManifest) er
 		klog.V(5).InfoS("DaemonSet update: cannot update config", "mcp", mcpName, "daemonset", gdm.DaemonSet.Name, "error", err)
 		return err
 	}
+	if gdm.ClusterPlatform != platform.Kubernetes {
+		if instance.IsCustomPolicyEnabled() && gdm.ClusterPlatform == platform.OpenShift {
```
ffromani, line 778: I'd rather move the flag into the `GeneratedDesiredManifest` struct

pkg/objectstate/rte/rte.go
```diff
@@ -130,13 +130,13 @@ type GeneratedDesiredManifest struct {
...
 
-type GenerateDesiredManifestUpdater func(mcpName string, gdm *GeneratedDesiredManifest) error
+type GenerateDesiredManifestUpdater func(mcpName string, gdm *GeneratedDesiredManifest, instance *nropv1.NUMAResourcesOperator) error
 
 func SkipManifestUpdate(gdm *GeneratedDesiredManifest) error {
 	return nil
 }
 
-func (em *ExistingManifests) State(mf rtemanifests.Manifests, updater GenerateDesiredManifestUpdater) []objectstate.ObjectState {
+func (em *ExistingManifests) State(mf rtemanifests.Manifests, instance *nropv1.NUMAResourcesOperator, updater GenerateDesiredManifestUpdater) []objectstate.ObjectState {
```
ffromani, line 139: `FromClient` already gets a `instance` parameter, I think we can safely compute the flag there and carry it in the `ExistingManifests` struct

test/e2e/install/install_test.go
```diff
@@ -288,7 +288,7 @@ var _ = Describe("[Install] durability", func() {
 			By("checking there are no leftovers")
 			// by taking the ns from the ds we're avoiding the need to figure out in advanced
 			// at which ns we should look for the resources
-			mf, err := rte.GetManifests(configuration.Plat, configuration.PlatVersion, ds.Namespace, true)
+			mf, err := rte.GetManifests(configuration.Plat, configuration.PlatVersion, ds.Namespace, true, true)
```
ffromani, line 294: time to cleanup this signature if time allows

test/utils/deploy/deploy.go
```diff
@@ -37,6 +37,7 @@ import (
 	"github.com/openshift-kni/numaresources-operator/controllers"
 	nropmcp "github.com/openshift-kni/numaresources-operator/internal/machineconfigpools"
 	"github.com/openshift-kni/numaresources-operator/pkg/status"
+	"github.com/openshift-kni/numaresources-operator/test/e2e/serial/tests"
```
ffromani, line 40: please no: code in `test/e2e/serial/tests` may depend on these utilities but not the other way around

controllers/numaresourcesoperator_controller.go
```diff
@@ -478,6 +478,7 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 		}
 		rteupdate.DaemonSetHashAnnotation(r.RTEManifests.DaemonSet, cmHash)
 	}
+	rteupdate.SecurityContextConstraintSetSELinuxOption(r.RTEManifests.SecurityContextConstraint, instance.IsCustomPolicyEnabled())
```
ffromani, line 481: maybe just `rteupdate.SecurityContextConstraint()`

ffromani (commented):
most comments are about code organization. The overall direction is reasonnable, but we would need some work still

controllers/numaresourcesoperator_controller.go
ffromani: I'm also thinking about `deleteUnusedMachineConfigs`. We can perhaps reuse that function - or even better get rid of it.

ffromani: ...or not? https://github.com/openshift-kni/numaresources-operator/pull/1029

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
Tal-or: Here: https://github.com/openshift-kni/numaresources-operator/pull/1025#discussion_r1784578995 you mentioned you want it non-importable, I'm a bit confused :)

ffromani: I could have been clearer. The magic annotations should not be part of the API packages, nor of its helpers.
Thus, let's move the helpers away from the api packages: rather than import the internal package in the api package, let's move the helper in the internal package.

controllers/numaresourcesoperator_controller.go
Tal-or: What you did in #1029 is true, but we still need to remove the machine config explicitly when moving to built-in policy, because the CR remain in the cluster.
And as long as it persist on the cluster the machine-config won't go anywhere

ffromani: OK, let's try to remove them in the reconciliation loop rather than explicitely in the `deleteXXX` functions.

Tal-or: > this looks fragile. We want to target only the MachineConfig. So it's probably better to add new functionalities to pkg/objectstate/rte

We do. you have the `objStates := existing.MachineConfigsState(r.RTEManifests)` that filters the MachineConfigs out of all the objects for you.

> OK, let's try to remove them in the reconciliation loop rather than explicitely in the deleteXXX functions.

This code is part of the reconciliation loop
it's called in `syncMachineConfigs`

controllers/numaresourcesoperator_controller.go
Tal-or: Thanks!

controllers/numaresourcesoperator_controller.go
ffromani: yes, I commented before to starting the effort in #1029 . It seems this of yours is the best approach on the table, and the removal of the `deleteXXX` functions should be tried (and preferably done!) on the side

internal/api/features/types.go
Tal-or: Done.

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
Tal-or: Done. Thanks!

pkg/objectstate/rte/rte.go
Tal-or: residues of an old implementation effort. Removed the changes

controllers/numaresourcesoperator_controller.go
Tal-or: Done.

pkg/objectstate/rte/rte.go
Tal-or: Done.

test/e2e/install/install_test.go
Tal-or: Will do, but I wouldn't block this PR for that.

test/utils/deploy/deploy.go
Tal-or: Done. I moved the MCP wait logic to `test/utils/deploy/deploy.go` and 
now `github.com/openshift-kni/numaresources-operator/test/e2e/serial/tests` consumes it

test/e2e/install/install_test.go
ffromani: of course no, this must to be a separate independent effort

ffromani (commented):
/approve
/lgtm

I'm not completely happy with how the code would look like, but I prefer to merge the functionality and improve in followup PRs.

@Tal-or please unhold if you're happy with this work and if you think the test coverage is sufficient

Tal-or:
> /approve /lgtm
> 
> I'm not completely happy with how the code would look like, but I prefer to merge the functionality and improve in followup PRs.
> 
> @Tal-or please unhold if you're happy with this work and if you think the test coverage is sufficient

We can wait and have another iteration if you're not happy with code flow. I don't mind at all

ffromani:
> > /approve /lgtm
> > I'm not completely happy with how the code would look like, but I prefer to merge the functionality and improve in followup PRs.
> > @Tal-or please unhold if you're happy with this work and if you think the test coverage is sufficient
> 
> We can wait and have another iteration if you're not happy with code flow. I don't mind at all

for starters, I'd love to push the Delete logic in the `apply` package, for consistency. Something like https://github.com/openshift-kni/numaresources-operator/pull/1038

PR #1029, opened 2024-10-02
WIP: nrop: controller: don't delete unused resources explictly

ffromani:
We add Controller References, so we expect the system to take care of orphaned resources.
Hence, the explicit delete logic can be removed. That logic was always suspicious, it should never be needed and it's a sweeping yet effective fix that could have masked real issues. It's time to address these issues and had real ownership and object cleanup instead of using this bandaid.

Besides simplification, we now are down two List() per reconcile loop


ffromani:
e2e tests:
- extend install suite (runs on CI)
- add explicit check to verify all expected manifests are present
- add explicit check to ensure all expected manifests are removed once we remove the NROP object (including when we delete/recreate, probably modify existing test + add explicit test)
- add explicit using a NROP object with 2 nodegroups. Edit the NROP object to remove 1, make sure all the objects pertaining to the second are removed, while the objects pertaining to the one left are not gone

ffromani:
xref: CNF-3740

shajmakh:
if not explicitly deleted (or at least ownerRefernce is removed) wouldn't they still be under the related objects of the NRO CR?

ffromani:
> if not explicitly deleted (or at least ownerRefernce is removed) wouldn't they still be under the related objects of the NRO CR?

this is one of the questions we need to answer. The rationale for explicit delete is fuzzy and likely obsolete (since few releases)

ffromani:
good, the expected failures mean that so far the suites are working as expected. We can proceed adding the real new e2e tests.

ffromani:
needs to be rebased on top of #1045

ffromani:
/hold cancel

tests are good enough now

ffromani:
/hold

I think finally I managed to remember. The scenario the explicit delete is supposed to cover is:
1. NodeGroups are edited or deleted
2. Due to item 1, Nodes (MCPs) previously managed are no longer managed
3. The MC/DaemonSet of the previous nodegroup, now no longer relevant, are left lingering
4. Since the NRO object still exists, these objects are never cleaned up

ffromani:
much ado about nothing. There are good ideas in this PR, but the overall concept turned out wrong. I'll post later more targeted PRs to recycle the good bits in there.

PR #1032, opened 2024-10-03
cleanup and streamline status computation

ffromani:
The way we computed the `NUMAResourcesOperator` status was messy, relying on inner functions and helpers to reported somehow if something changed in the object, abusing the name of the conditions (leading to awkward function signatures), doing and undoing checks and so forth.

Besides messy and unecessarily hard to read code, the outcome was that we both sometimes missed to update the status, leading to not major bugs (yet) but to less than ideal experience.

To untangle this mess, the new approach is to just mutate the status freely during the reconciliation loop in the reconciliation sub-step. Every step is free to mutate the status just reporting error or not, and the top-level reconciliation code will detect semantic differences (e.g. if only timestamps changed but say nothing else, this is not a semantically relevant difference so no actual update should be sent).

Detecting changes this way needs nested comparations of objects but it's a very major simplifications; if we implement carefully the comparison code (coming soon), benchmarks show good numbers (= still some theoretical slowdowns, but the overall state is not terrible and better than expected) wrt the current implementation, so win-win.


ffromani:
/hold

need to wait for all the HCP work to go in anyway

ffromani:
@shajmakh hey! let's discuss the cleanups in the last 2 commits to see if they can help you

controllers/controllers.go
```diff
@@ -16,7 +16,49 @@
...
+		Message: message,
+		Reason:  reason,
+	}
+}
+
+func degradedConditionInfoFromError(err error) conditionInfo {
+	return conditionInfo{
+		Type:    status.ConditionDegraded,
+		Message: messageFromError(err),
+	}
```
shajmakh, line 85: should we add reason here? 
`Reason: reasonFromError(err)`

controllers/numaresourcesoperator_controller.go
```diff
@@ -179,21 +179,22 @@ func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctr
...
+	conditions, ok := status.UpdateConditions(instance.Status.Conditions, cond.Type, cond.Reason, cond.Message)
 	if ok {
 		instance.Status.Conditions = conditions
 	}
 }
 
 func (r *NUMAResourcesOperatorReconciler) degradeStatus(ctx context.Context, instance *nropv1.NUMAResourcesOperator, reason string, stErr error) (ctrl.Result, error) {
-	message := messageFromError(stErr)
+	info := degradedConditionInfoFromError(stErr)
+	info.Reason = reason
```
shajmakh, line 195: if `reason` is empty it will override `info.Reason` (keep it as "InternalError"?) if we set it above `Reason: reasonFromError(err)`

controllers/numaresourcesoperator_controller.go
```diff
@@ -207,25 +208,27 @@ func (r *NUMAResourcesOperatorReconciler) degradeStatus(ctx context.Context, ins
 	return ctrl.Result{}, nil
 }
 
-func (r *NUMAResourcesOperatorReconciler) reconcileResourceAPI(ctx context.Context, instance *nropv1.NUMAResourcesOperator, trees []nodegroupv1.Tree) (bool, ctrl.Result, string, error) {
+func (r *NUMAResourcesOperatorReconciler) reconcileResourceAPI(ctx context.Context, instance *nropv1.NUMAResourcesOperator, trees []nodegroupv1.Tree) (bool, ctrl.Result, conditionInfo, error) {
```
shajmakh: I don't see a reason to keep the returned error, it is already part of the returned `conditionInfo`

controllers/numaresourcesoperator_controller.go
```diff
@@ -207,25 +208,27 @@ func (r *NUMAResourcesOperatorReconciler) degradeStatus(ctx context.Context, ins
...
 	applied, err := r.syncNodeResourceTopologyAPI(ctx)
 	if err != nil {
 		r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedCRDInstall", "Failed to install Node Resource Topology CRD: %v", err)
-		return true, ctrl.Result{}, status.ConditionDegraded, fmt.Errorf("FailedAPISync: %w", err)
+		err = fmt.Errorf("FailedAPISync: %w", err)
+		return true, ctrl.Result{}, degradedConditionInfoFromError(err), err
```
shajmakh: if not mistaken I think we can even waive the returned boolean value, it is `false` only when all updates are done  -> condition is empty/available

shajmakh: same for all similar substeps of the reconciliation

controllers/numaresourcesoperator_controller.go
shajmakh: same below in the other reconciliation substeps

controllers/controllers.go
```diff
@@ -16,7 +16,49 @@
...
+}
+
+func fillConditionInfoFromError(cond conditionInfo, err error) conditionInfo {
+	if cond.Message == "" {
+		cond.Message = messageFromError(err)
+	}
+	if cond.Reason == "" {
+		cond.Reason = reasonFromError(err)
+	}
+	return cond
```
shajmakh, line 95: the only path the reconciliation will return an error is when the condition is degraded, do we still need this function or can we just integrate the reason in degradedConditionInfoFromError?

controllers/controllers.go
```diff
@@ -18,10 +18,45 @@ package controllers
...
 	"errors"
+	"time"
+
+	ctrl "sigs.k8s.io/controller-runtime"
 
 	"github.com/openshift-kni/numaresources-operator/pkg/status"
 )
 
+type reconcileStep struct {
+	Done          bool // is reconciliation process done after this step?
```
shajmakh, line 29: do we really need this considering the condition type says it all and there is 1:1 relation for when done is false (condition is available)?

shajmakh: same for the error

controllers/controllers.go
```diff
@@ -18,10 +18,45 @@ package controllers
...
+	Result        ctrl.Result
+	ConditionInfo conditionInfo
+	Error         error
+}
```
shajmakh, line 33: I think it looks better with the new struct and easy to use

controllers/controllers.go
ffromani: seems unnecessary indeed, removed

ffromani:
/hold
need to test

controllers/numaresourcesoperator_controller.go
ffromani: right! handled in the latest commits

controllers/numaresourcesoperator_controller.go
ffromani: that's a good point. What it should go is not the top level error though, which is used to report to the upper layers, but the inner error in `ConditionInfo`. Let's see what I can do.

controllers/numaresourcesoperator_controller.go
ffromani: indeed, good catch. I think I fixed in the later commits by rearranging the flow.

controllers/controllers.go
ffromani: we probably should, yes. It should be helpful and harmless.

ffromani:
know controller test failure. It's legit. I'll have a look and fix ASAP

ffromani:
/hold cancel

ready, but require more testing (e.g. running the serial suite)

controllers/controllers.go
```diff
@@ -18,10 +18,53 @@ package controllers
...
+	return rs.ConditionInfo.Type == status.ConditionAvailable
+}
+
+// EarlyStop returns true if we should shortcut after a reconcile substep,
+// false otherwise
+func (rs reconcileStep) EarlyStop() bool {
+	return rs.ConditionInfo.Type != status.ConditionAvailable
+}
+
+func reconcileStepSuccess() reconcileStep {
```
Tal-or, line 46: Since we're kinda inventing new abstraction for the different reconciliation I think we should add comments explaining when each step returned.

ffromani: done!

Tal-or:
/hold
/lgtm
feel free to unhold if you're happy or if other reviews are done

controllers/numaresourcesoperator_controller.go
```diff
@@ -164,18 +164,27 @@ func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctr
...
 
 	result, condition, err := r.reconcileResource(ctx, instance, trees)
-	if condition != "" {
-		if condition == status.ConditionAvailable && multiMCPsErr != nil {
-			_, _ = r.updateStatus(ctx, instance, status.ConditionDegraded, validation.NodeGroupsError, multiMCPsErr)
-		} else {
-			_, _ = r.updateStatus(ctx, instance, condition, reasonFromError(err), err)
-		}
+
+	_ = updateStatusConditionsIfNeeded(instance, condition, reasonFromError(err), messageFromError(err))
```
shajmakh: we need to ensure we don't lose the error data when parsing the message. this is related to how the errors are built, so unwrapping a non-nil error can result in a nil error which then will cause the main reason of the degraded state to be lost. We can double verify that and handle this in a cleanup PR

controllers/numaresourcesoperator_controller.go
```diff
@@ -164,18 +164,27 @@ func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctr
...
+
+	if condition == status.ConditionAvailable && multiMCPsErr != nil {
+		_, _ = r.updateStatus(ctx, instance, status.ConditionDegraded, validation.NodeGroupsError, multiMCPsErr)
 	}
+
+	updErr := r.Client.Status().Update(ctx, instance)
+	if updErr != nil {
```
shajmakh, line 179: do we still want to update if there are no updates?

controllers/numaresourcesoperator_controller.go
```diff
@@ -171,6 +173,10 @@ func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctr
 		_, _ = r.degradeStatus(ctx, instance, validation.NodeGroupsError, multiMCPsErr)
 	}
 
+	if !status.IsUpdatedNUMAResourcesOperator(curStatus, &instance.Status) {
+		return result, err
+	}
```
shajmakh: okay I see it's validated here that we don't need the status update

controllers/controllers.go
```diff
@@ -18,16 +18,70 @@ package controllers
...
 )
 
+type reconcileStep struct {
+	Result        ctrl.Result
+	ConditionInfo conditionInfo
+	Error         error
+}
+
+// Done returns true if a reconciliation step is fully completed,
+// meaning all the substeps have been performed successfully, false otherwise
```
shajmakh, line 35: a reasonable definition of done, thanks for this

shajmakh (commented):
Thanks for working on this, the code looks much better with these updates.
My only concern is that we are likely going to lose the status message parse from the error, and since we don't have controller tests (unit tests we do have) for checking the error message is reflected, we will only see that when we need to debug. This however can be handled in a different PR

controllers/numaresourcesoperator_controller.go
shajmakh: I'll verify and post the needed changes

ffromani: we can probably fix this by fixing `messageFromError` and `reasonFromError`. I'd be very interested in learning on which cases we can lose errors.

ffromani:
/hold cancel

LGTM is gone anyway

pkg/status/status.go
```diff
@@ -151,7 +151,7 @@ func MessageFromError(err error) string {
 	}
 	unwErr := errors.Unwrap(err)
 	if unwErr == nil {
-		return ""
+		return err.Error()
```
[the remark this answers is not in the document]
shajmakh: yes, this should behave better. Thanks for fixing

pkg/status/status_test.go
```diff
@@ -159,3 +161,37 @@ func TestIsUpdatedNUMAResourcesOperator(t *testing.T) {
...
+		{
+			name:     "simple error",
+			err:      fmt.Errorf("testing error with simple message"),
+			expected: "testing error with simple message",
+		},
+		{
+			name:     "wrapped error",
+			err:      fmt.Errorf("outer error: %w", errors.New("inner error")),
+			expected: "inner error",
+		},
```
shajmakh, line 186: nit: we can also add test with errors.New

controllers/numaresourcesoperator_controller.go
shajmakh: thanks for addressing

ffromani:
/hold

the serial suite runs are finding some actual bugs :\

PR #1033, opened 2024-10-04
[release-4.15] Makefile: binary-all: support build with -tags

shajmakh:
Introduce GOTAGS environment variable. This is needed by d/s builds to allow passing tags like `-tags strictfipsruntime`. If the env variable is not set, passing empty string as in `-tags ""` (not empty value as in `-tags `) will run the command as if nothing added with no disturbing.


shajmakh:
manual backport of https://github.com/openshift-kni/numaresources-operator/pull/1030

PR #1035, opened 2024-10-06
CNF-11234: Enable RTE metrics to be scraped securely  by Prometheus

rbaturov:
This is a follow-up PR for [Enable NROP metrics to be to scraped securely by Prometheus ](https://github.com/openshift-kni/numaresources-operator/pull/1007).

This PR encompasses and integrates all the needed infrastructure to enable secured communication for scraping metrics by Prometheus, now for **RTE**.

A follow-up PR will be posted for e2e tests later.

 To validate that this PR is functioning correctly, please follow these steps:
1. build image of the operator (`make docker-build docker-push`)
2. run: `make deploy` 
3. Attach to one of the prometheus pods `oc exec -it prometheus-k8s-0 -n openshift-monitoring /bin/bash`
4. run:
  ```
 curl -v \
  --cacert /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt \
  -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
  https://rte-metrics-service.numaresources.svc:8443/metrics
```

PR #1036, opened 2024-10-07
Added rules to ClusterRole

rbaturov:
We need to modify the ClusterRole to grant the required permissions for the manager and RTE-worker pods to access the /metrics endpoint. This is essential for e2e testing, as we will run curl commands from within the manager and RTE-worker containers to interact with the endpoint. Since this ClusterRole is also used by the RTE service account (as referenced in PR #1035), this change will grant access to both pods.

This is needed to support:
https://issues.redhat.com/browse/CNF-10142


bundle/manifests/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -62,7 +62,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2024-09-25T14:08:22Z"
+    createdAt: "2024-10-07T12:09:23Z"
```
Tal-or, line 65: Not needed. you may commit without this file

Tal-or (commented):
I don't see that we're consuming the `auth_proxy_role.yaml`

```
[titzhak@fedora numaresources-operator]$ grep -R auth_proxy_role.yaml
config/rbac/kustomization.yaml:#- auth_proxy_role.yaml
```

Anyway, if this file is supposed to be consumed only at testing environment I would make sure it deployed only there and not on production.

rbaturov:
> I don't see that we're consuming the `auth_proxy_role.yaml`
> 
> ```
> [titzhak@fedora numaresources-operator]$ grep -R auth_proxy_role.yaml
> config/rbac/kustomization.yaml:#- auth_proxy_role.yaml
> ```
> 
> Anyway, if this file is supposed to be consumed only on testing environment I would make sure it deployed only there and not on production.

Thanks for the fast reply.
I decided to add this to the default cluster role that is used for the operator, rather than the proxy ClusterRole as this is a more correct approach.

Tal-or:
> > I don't see that we're consuming the `auth_proxy_role.yaml`
> > ```
> > [titzhak@fedora numaresources-operator]$ grep -R auth_proxy_role.yaml
> > config/rbac/kustomization.yaml:#- auth_proxy_role.yaml
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Anyway, if this file is supposed to be consumed only on testing environment I would make sure it deployed only there and not on production.
> 
> Thanks for the fast reply. I decided to add this to the default cluster role that is used for the operator, rather than the proxy ClusterRole as this is a more correct approach.

But you said it's being used only for e2e tests isn't it?

rbaturov:
> > > I don't see that we're consuming the `auth_proxy_role.yaml`
> > > ```
> > > [titzhak@fedora numaresources-operator]$ grep -R auth_proxy_role.yaml
> > > config/rbac/kustomization.yaml:#- auth_proxy_role.yaml
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Anyway, if this file is supposed to be consumed only on testing environment I would make sure it deployed only there and not on production.
> > 
> > 
> > Thanks for the fast reply. I decided to add this to the default cluster role that is used for the operator, rather than the proxy ClusterRole as this is a more correct approach.
> 
> But you said it's being used only for e2e tests isn't it?

Yes you are right. now that I think of this, @ffromani what do you guys think if I'll just use the ClusterRole `auth_proxy_client_clusterrole` which does the same and add a clusterorlebinindg for this under `kind-ci` and close this PR?

Tal-or:
This or you can add the the specific content you need using MergePatch:
https://stackoverflow.com/questions/71496560/kustomize-using-strategic-merge-patch-on-helmreleases

rbaturov:
> This or you can add the specific content you need using MergePatch: https://stackoverflow.com/questions/71496560/kustomize-using-strategic-merge-patch-on-helmreleases
Sounds good to me.
@ffromani WDYT about this idea?

ffromani:
@rbaturov this is the _only_ missing RBAC changeset to enable the testing per our offline discussion?

rbaturov:
> @rbaturov this is the _only_ missing RBAC changeset to enable the testing per our offline discussion?

Yes, this is the only change needed to test the operator metrics endpoint for **(manager container)**.
For RTE metrics endpoint, (which is a separate discussion) we need to think of a different resolution to allow the worker pod to access its endpoint. here are two ideas I have:
1. Add and apply a `clusterrole` and `clusterorlebinding` solely for accessing endpoint  in the metrics test iteself
2. Alternatively, apply both (`clusterrole` and `clusterorlebinding`) by default in `pkg/metrics/manifests/yaml`

ffromani:
> > @rbaturov this is the _only_ missing RBAC changeset to enable the testing per our offline discussion?
> 
> Yes, this is the only change needed to test the operator metrics endpoint for **(manager container)**. For RTE metrics endpoint, (which is a separate discussion) we need to think of a different resolution to allow the worker pod to access its endpoint. here are two ideas I have:
> 
>     1. Add and apply a `clusterrole` and `clusterorlebinding` solely for accessing endpoint  in the metrics test iteself
> 
>     2. Alternatively, apply both (`clusterrole` and `clusterorlebinding`) by default in `pkg/metrics/manifests/yaml`

ok then this is good to go in main branch for everyone, not just for CI/testing. If upgrade is no bother (as I expect, honestly) than this change can also be backported.

rbaturov:
> > > @rbaturov this is the _only_ missing RBAC changeset to enable the testing per our offline discussion?
> > 
> > 
> > Yes, this is the only change needed to test the operator metrics endpoint for **(manager container)**. For RTE metrics endpoint, (which is a separate discussion) we need to think of a different resolution to allow the worker pod to access its endpoint. here are two ideas I have:
> > ```
> > 1. Add and apply a `clusterrole` and `clusterorlebinding` solely for accessing endpoint  in the metrics test iteself
> > 
> > 2. Alternatively, apply both (`clusterrole` and `clusterorlebinding`) by default in `pkg/metrics/manifests/yaml`
> > ```
> 
> ok then this is good to go in main branch for everyone, not just for CI/testing. If upgrade is no bother (as I expect, honestly) than this change can also be backported.

So can we merge this?

PR #1040, opened 2024-10-14
api: fix v1alpha1 backward compatibility

ffromani:
Back in 4.16.0 we changed defaults and added options to both operator and scheduler objects. Due to the sharing of the config objects between spec and status and due to the fact we still serve v1alpha1, we may end up with fields values in the status which are not declared in the API definition.

The most obvious offender is the enum values, but this actually affects all the values we set explicit defaults for and we expose in the status.

This is what OLM complained about.
To fix this, we align the status definition from v1 to v1alpha1, until we deprecate the v1alpha1 and we stop serving it.

xref: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/dependency-resolution.md

### DIFF
--- a/api/numaresourcesoperator/v1alpha1/numaresourcesoperator_types.go
+++ b/api/numaresourcesoperator/v1alpha1/numaresourcesoperator_types.go
@@ -43,21 +43,24 @@ type NUMAResourcesOperatorSpec struct {
 	PodExcludes []NamespacedName `json:"podExcludes,omitempty"`
 }
 
-// +kubebuilder:validation:Enum=Enabled;Disabled
+// +kubebuilder:validation:Enum=Disabled;Enabled;EnabledExclusiveResources
 type PodsFingerprintingMode string
 
-var (
+const (
 	// PodsFingerprintingDisabled disables the pod fingerprinting reporting.
 	PodsFingerprintingDisabled PodsFingerprintingMode = "Disabled"
 
 	// PodsFingerprintingEnabled enables the pod fingerprint considering all the pods running on nodes. It is the default.
 	PodsFingerprintingEnabled PodsFingerprintingMode = "Enabled"
+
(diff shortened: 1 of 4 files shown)


api/numaresourcesoperator/v1alpha1/numaresourcesoperator_types.go
```diff
@@ -43,21 +44,35 @@ type NUMAResourcesOperatorSpec struct {
...
+// +kubebuilder:validation:Enum=Disabled;Enabled
+type InfoRefreshPauseMode string
+
+const (
+	// InfoRefreshPauseDisabled enables RTE and NRT sync
+	InfoRefreshPauseDisabled InfoRefreshPauseMode = "Disabled"
+
+	// InfoRefreshPauseEnabled pauses RTE and disables the NRT sync
+	InfoRefreshPauseEnabled InfoRefreshPauseMode = "Enabled"
 )
```
shajmakh, line 70: I understand that PodsFingerprintingMode was present in v1alpha1 before and starting from 4.16 it had new option, but why to include new fields like InfoRefreshPause in v1alpha1 that were never present (meant to be supported) in v1alpha1?

ffromani: Good point. I was not completely sure about this, so let's go incrementally and let's minimize this change.

ffromani:
for our future selves, the investigation started with OLM not being able to upgrade - the new CSV was stuck in `pending` state.
Then, the key debugging steps where
```
oc get csv -o yaml -n openshift-numaresources
oc get installplan -n openshift-numaresources install-6tfnt -o yaml
```
and _carefully_ looking at the complains in the status. The key insight was that the message was complaining about the _status_ (not the spec!) and that the `NodeGroupConfig` is shared in both spec and status, and *the system makes the default explicit in the status*, defaults which changed starting in 4.16.

PR #1050, opened 2024-10-18
nrop: cleanup dangling objects removal

ffromani:
Move the dangling object removal code in its own package. This is useful because
- we have now a clear recorded explicit rationale about why we need this code
- this code is in essence a workaround for a questionable design choice, and we now isolate from the regular controller code.
- [future] we enable future reuse (maybe) in the validator.

The main benefit is the added `doc.go`.

this PR salvages the good parts of #1029 
Adding validation support is also a nice idea, but is deferred for later.


ffromani:
unit test failures are expected atm, no time to look into them and fix the problem

ffromani:
the failure here is reproducible, will debug later

PR #1052, opened 2024-10-21
controller: use slice instead of map to conserve consistent order

shajmakh:
Using map to store the daemonset of the targeted MCPs turned out to be a not the best data structure especially when we would like to conserve a consistent order of the elements that aligns with the rest of the slices in the operator status.
The problem with map is that it cannot be guaranteed that the elements will be ordered in the same order they were placed. To fix that, define a new struct of an MCP and its respective created RTE daemonset and replace the map with a slice of this struct which ensures that same data will produce same output every time.

ref: https://groups.google.com/g/golang-nuts/c/YfDxpkI34hY/m/4pktJI2ytusJ

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -73,6 +73,12 @@ import (
 
 const numaResourcesRetryPeriod = 1 * time.Minute
 
+// poolDaemonSet a struct to hold the target MCP of a configured node group and its created respective RTE daemonset
+type poolDaemonSet struct {
+	PoolName  string
+	DaemonSet nropv1.NamespacedName
+}
+
 // NUMAResourcesOperatorReconciler reconciles a NUMAResourcesOperator object
 type NUMAResourcesOperatorReconciler struct {
 	client.Client
@@ -238,7 +244,7 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResourceMachineConfig(ctx con
 	return false, ctrl.Result{}, "", nil
 }
 
-func (r *NUMAResourcesOperatorReconciler) reconcileResourceDaemonSet(ctx context.Context, instance *nropv1.NUMAResourcesOperator, trees []nodegroupv1.Tree) (map[string]nropv1.NamespacedName, bool, ctrl.Result, string, error) {
+func (r *NUMAResourcesOperatorReconciler) reconcileResourceDaemonSet(ctx context.Context, instance *nropv1.NUMAResourcesOperator, trees []nodegroupv1.Tree) ([]poolDaemonSet, bool, ctrl.Result, string, error) {
 	daemonSetsInfoPerMCP, err := r.syncNUMAResourcesOperatorResources(ctx, instance, trees)
 	if err != nil {
 		r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedRTECreate", "Failed to create Resource-Topology-Exporter DaemonSets: %v", err)
@@ -286,13 +292,13 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
 	return ctrl.Result{}, status.ConditionAvailable, nil
 }
 
-func (r *NUMAResourcesOperatorReconciler) syncDaemonSetsStatuses(ctx context.Context, rd client.Reader, daemonSetsInfo map[string]nropv1.NamespacedName) ([]nropv1.NamespacedName, bool, error) {
+func (r *NUMAResourcesOperatorReconciler) syncDaemonSetsStatuses(ctx context.Context, rd client.Reader, daemonSetsInfo []poolDaemonSet) ([]nropv1.NamespacedName, bool, error) {
 	dssWithReadyStatus := []nropv1.NamespacedName{}
-	for _, nname := range daemonSetsInfo {
+	for _, dsInfo := range daemonSetsInfo {
 		ds := appsv1.DaemonSet{}
 		dsKey := client.ObjectKey{
-			Namespace: nname.Namespace,
-			Name:      nname.Name,
+			Namespace: dsInfo.DaemonSet.Namespace,
+			Name:      dsInfo.DaemonSet.Name,
 		}
 		err := rd.Get(ctx, dsKey, &ds)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


controllers/numaresourcesoperator_controller.go
```diff
@@ -302,20 +308,26 @@ func (r *NUMAResourcesOperatorReconciler) syncDaemonSetsStatuses(ctx context.Con
...
+			if mcp.Name != info.MCPName {
+				continue
+			}
+
+			status := nropv1.NodeGroupStatus{
+				PoolName:  mcp.Name,
+				Config:    *mcp.Config,
+				DaemonSet: info.Daemonset,
+			}
+			ngStatuses = append(ngStatuses, status)
```
ffromani, line 329: thanks for the fix. We can use another fix later: this function should create the `NodeGroupStatus` using `dsPerMCP` as master source (iterating over there) and not depend on other status fields like `instance.Status.MachineConfigPools`.

controllers/numaresourcesoperator_controller.go
```diff
@@ -73,6 +73,12 @@ import (
 
 const numaResourcesRetryPeriod = 1 * time.Minute
 
+// MCPDaemonSetPair a struct to hold the target MCP of a configured node group and its created respective RTE daemonset
+type MCPDaemonSetPair struct {
+	MCPName   string
+	Daemonset nropv1.NamespacedName
+}
+
```
ffromani, line 81: if we call this
```
type poolDaemonSet struct {
    PoolName string
    DaemonSet nropv1.NamespacedName
}
```
we generalize it for free.
In addition, this type is (should be) used only internally, so should not be exported (type should start lowercase).

controllers/numaresourcesoperator_controller.go
```diff
@@ -462,28 +474,28 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 	}
 
 	var err error
-	dssByMCP := make(map[string]nropv1.NamespacedName)
-
+	// using a slice of MCPDaemonSetPair instead of a map because Go maps assignment order is not consistent and indeterministic
```
ffromani, line 477: no need to reiterate, the rationale is already well explained in the commit message. But not blocking either.

ffromani (commented):
the general direction is fine, thanks for the fix!
some improvements and followups as inline comments.

controllers/numaresourcesoperator_controller.go
shajmakh: right, I'll update that. Thanks for catching it.

controllers/numaresourcesoperator_controller.go
shajmakh: I thought it would be handy here to avoid the thoughts of "this scenario fits into a map so let's change this to use map". I can adjust the comment to focus more on what slice provides and its importance, or completely remove that. what do you see better fit?

ffromani: right, good point. Let's keep it.

controllers/numaresourcesoperator_controller.go
shajmakh: we need to be aligned with the status update on MCP, currently the config won't be updated there if MCP is not completely ready. I'll look into this anyhow, we can use an improvement here indeed

controllers/numaresourcesoperator_controller.go
```diff
@@ -73,6 +73,12 @@ import (
 
 const numaResourcesRetryPeriod = 1 * time.Minute
 
+// poolDaemonSet a struct to hold the target MCP of a configured node group and its created respective RTE daemonset
+type poolDaemonSet struct {
+	PoolName  string
+	DaemonSet nropv1.NamespacedName
```
ffromani, line 79: for the future: let's add a methof in the `api/v1` package:
```
func (nn NamespacedName) AsKey() client.ObjectKey {
	return client.ObjectKey{
		Namespace: nn.Namespace,
		Name:      nn.Name,
	}
}
```

controllers/numaresourcesoperator_controller.go
```diff
@@ -286,13 +292,13 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
 	return ctrl.Result{}, status.ConditionAvailable, nil
 }
 
-func (r *NUMAResourcesOperatorReconciler) syncDaemonSetsStatuses(ctx context.Context, rd client.Reader, daemonSetsInfo map[string]nropv1.NamespacedName) ([]nropv1.NamespacedName, bool, error) {
+func (r *NUMAResourcesOperatorReconciler) syncDaemonSetsStatuses(ctx context.Context, rd client.Reader, daemonSetsInfo []poolDaemonSet) ([]nropv1.NamespacedName, bool, error) {
 	dssWithReadyStatus := []nropv1.NamespacedName{}
-	for _, nname := range daemonSetsInfo {
+	for _, dsInfo := range daemonSetsInfo {
 		ds := appsv1.DaemonSet{}
 		dsKey := client.ObjectKey{
```
ffromani, line 299: this will then become: `dsKey := nname.AsKey()` (maybe inlined, seems to be used only once)

controllers/numaresourcesoperator_controller.go
```diff
@@ -462,28 +474,28 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 	}
 
 	var err error
-	dssByMCP := make(map[string]nropv1.NamespacedName)
-
+	// using a slice of poolDaemonSet instead of a map because Go maps assignment order is not consistent and indeterministic
```
ffromani, line 477: nit: `nondeterministic`

controllers/numaresourcesoperator_controller.go
```diff
@@ -494,10 +506,10 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 		if err != nil {
 			return err
 		}
-		dssByMCP[mcpName] = nropv1.NamespacedName{
+		dsMCPPairs = append(dsMCPPairs, poolDaemonSet{mcpName, nropv1.NamespacedName{
 			Namespace: gdm.DaemonSet.GetNamespace(),
 			Name:      gdm.DaemonSet.GetName(),
-		}
+		}})
```
ffromani, line 512: for the future: let's add our clone of `client.ObjectKeyFromObject` (will become `NamespacedNameFromObject`)
in addition, please always use named initializers everywhere

ffromani (commented):
/approve
/lgtm

controllers/numaresourcesoperator_controller.go
ffromani: actually bad idea, we will bring the client package as dep into our api package, which is something we don't want.

controllers/numaresourcesoperator_controller.go
ffromani: same as https://github.com/openshift-kni/numaresources-operator/pull/1052#discussion_r1808484818

controllers/numaresourcesoperator_controller.go
shajmakh: ack

ffromani: *in another PR* we can add another helper package though

controllers/numaresourcesoperator_controller.go
shajmakh: used metav.Object since we already call it in v1.

controllers/numaresourcesoperator_controller.go
shajmakh: PR: https://github.com/openshift-kni/numaresources-operator/pull/1053

PR #1053, opened 2024-10-21
add namespacedname helper pkg

shajmakh:
Minor adjustments and cleanups that wrap helper functions for
NamespacedName under api/v1/helper/namespacedname.

### DIFF
--- a/api/numaresourcesoperator/v1/helper/namespacedname/namespacedname.go
+++ b/api/numaresourcesoperator/v1/helper/namespacedname/namespacedname.go
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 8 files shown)


shajmakh:
depends on: https://github.com/openshift-kni/numaresources-operator/pull/1052

internal/api/helper.go
```diff
@@ -0,0 +1,22 @@
+package api
```
ffromani, line 1: we can and should move this in 
```
api/numaresourcesoperator/v1/helper/
```
we can call it `namespacedname` (`package namespacedname`)
packages in `.../helper` build on top of the main api package, so we can add deps (in a reasonnable fashion) here.

internal/api/helper.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func AsObjectKey(nn nropv1.NamespacedName) client.ObjectKey {
+	return client.ObjectKey{
+		Namespace: nn.Namespace,
+		Name:      nn.Name,
+	}
+}
+
+func NamespacedNameFromObject(obj metav1.Object) nropv1.NamespacedName {
```
ffromani, line 17: if we move as suggested, this can be just `FromObject` (and should probably take a `client.Object`)

internal/api/helper.go
```diff
@@ -0,0 +1,22 @@
...
+	return nropv1.NamespacedName{
+		Namespace: obj.GetNamespace(),
+		Name:      obj.GetName(),
+	}
+}
```
ffromani, line 22: please add a simple test which does roundtrip(s). Should be sufficient to prove the correctness.

internal/api/helper.go
ffromani: please add the copyright boilerplate (`hack/boilerplate.go.txt`)

ffromani (commented):
Nice!

this make me realize I need to add docs to our package to express the intent of these

ffromani (commented):
/approve
/lgtm

nice little cleanup

PR #1056, opened 2024-10-23
WIP: DEMO: make dangling unwanted MachineConfig

ffromani:
is this better than the current approach?

### DIFF
--- a/internal/dangling/dangling.go
+++ b/internal/dangling/dangling.go
@@ -39,6 +39,8 @@ import (
 	"github.com/openshift-kni/numaresources-operator/pkg/objectnames"
 )
 
+type KeepDanglingObjectFunc func(expected bool, obj client.Object) (bool, string)
+
 func DeleteUnusedDaemonSets(cli client.Client, ctx context.Context, instance *nropv1.NUMAResourcesOperator, trees []nodegroupv1.Tree) []error {
 	klog.V(3).Info("Delete dangling Daemonsets start")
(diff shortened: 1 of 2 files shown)


ffromani:
this can be clever, but I think it's better to delete explicitely, to really emphasize that we want the extra MachineConfig to be gone.

From user perspective the overall behavior should be the same.

internal/dangling/dangling.go
```diff
@@ -117,6 +119,21 @@ func DeleteUnusedMachineConfigs(cli client.Client, ctx context.Context, instance
 	return errors
 }
 
+func keepMachineConfig(expected bool, mc machineconfigv1.MachineConfig, filterObjects ...KeepDanglingObjectFunc) bool {
+	if len(filterObjects) == 0 {
+		return expected
+	}
+	for _, filterObject := range filterObjects {
+		keep, reason := filterObject(expected, &mc)
+		if keep {
```
Tal-or, line 128: I don't see any good reason for keeping a dangling object.
Moreover we should expose a way for the user to specify whether he wants that or not.

Tal-or (commented):
Could you please explain the rationale here? I'm probably missing something

ffromani:
this was a demo to explore if it's better to use the dangling objects handling flow or the current explicit delete flow.
While the former is feasible, the current solution seems better as it is more explicit.
So no longer need for this PR.

PR #1058, opened 2024-10-25
cleanup PR: reflect SELinux policy in RTE DS and report Progressing condition message

shajmakh:
Cleanup PR to enhance trouble shooting. Please see the commits for additional info.

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -396,8 +396,6 @@ func syncMachineConfigPoolsStatuses(instanceName string, trees []nodegroupv1.Tre
 	pendingMCPs := []string{}
 	for _, tree := range trees {
 		for _, mcp := range tree.MachineConfigPools {
-			mcpStatuses = append(mcpStatuses, extractMCPStatus(mcp, forwardMCPConds))
-
 			isUpdated := updatedFunc(instanceName, mcp)
 			klog.V(5).InfoS("Machine Config Pool state", "name", mcp.Name, "instance", instanceName, "updated", isUpdated)
 
@@ -405,6 +403,7 @@ func syncMachineConfigPoolsStatuses(instanceName string, trees []nodegroupv1.Tre
 				pendingMCPs = append(pendingMCPs, mcp.Name)
 				continue
 			}
+			mcpStatuses = append(mcpStatuses, extractMCPStatus(mcp, forwardMCPConds))
 		}
 	}
 	return mcpStatuses, pendingMCPs
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -490,7 +490,15 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 		}
 		rteupdate.DaemonSetHashAnnotation(r.RTEManifests.DaemonSet, cmHash)
 	}
-	rteupdate.SecurityContextConstraint(r.RTEManifests.SecurityContextConstraint, annotations.IsCustomPolicyEnabled(instance.Annotations))
+
+	customPolicyEnabled := annotations.IsCustomPolicyEnabled(instance.Annotations)
+
+	delete(r.RTEManifests.DaemonSet.Annotations, annotations.SELinuxPolicyConfigAnnotation)
+	if customPolicyEnabled {
+		rteupdate.DaemonSetAnnotation(r.RTEManifests.DaemonSet, annotations.SELinuxPolicyConfigAnnotation, instance.Annotations[annotations.SELinuxPolicyConfigAnnotation])
+	}
+
+	rteupdate.SecurityContextConstraint(r.RTEManifests.SecurityContextConstraint, customPolicyEnabled)
 
 	processor := func(mcpName string, gdm *rtestate.GeneratedDesiredManifest) error {
 		err := daemonsetUpdater(mcpName, gdm)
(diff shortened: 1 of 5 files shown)


ffromani (commented):
good idea to expose the selinux policy in use but
1. logging unconditionally will spam the logs
2. logs are good only for developers, and we have a way to infer it anyway checking the daemonset, there's 1:1 mapping between security context constraint and selinux policy in use.

What about adding an annotation to the RTE daemonset, which the NROP owns anyway?

shajmakh:
/hold
under testing

pkg/objectupdate/rte/rte.go
```diff
@@ -129,6 +129,13 @@ func DaemonSetHashAnnotation(ds *appsv1.DaemonSet, cmHash string) {
 	template.Annotations[hash.ConfigMapAnnotation] = cmHash
 }
 
+func DaemonSetAnnotation(ds *appsv1.DaemonSet, annotKey string, annotValue string) {
```
ffromani, line 132: this can be an overkill, but let's roll with it for now

controllers/numaresourcesoperator_controller.go
```diff
@@ -206,80 +210,81 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResourceAPI(ctx context.Conte
 	return false, ctrl.Result{}, "", nil
 }
 
-func (r *NUMAResourcesOperatorReconciler) reconcileResourceMachineConfig(ctx context.Context, instance *nropv1.NUMAResourcesOperator, trees []nodegroupv1.Tree) (bool, ctrl.Result, string, error) {
+func (r *NUMAResourcesOperatorReconciler) reconcileResourceMachineConfig(ctx context.Context, instance *nropv1.NUMAResourcesOperator, trees []nodegroupv1.Tree) (bool, ctrl.Result, string, string, error) {
```
ffromani, line 213: can we abuse the error instead of adding a new return value?

shajmakh: we can, progressing condition never produces a real error anyway, but I don't think it is right to abuse it because it'll create a confusion on the reconciliation result. 
my take is that we can define a struct to group the common values returned by the substeps used in the reconciliation procedure. OR if we want to keep it as similar to the current state we can create another type to store additional info (includes and error and a string for msg) and replace the returned error

shajmakh:
note: the RTE annotation addition/removal is still buggy

pkg/objectupdate/rte/rte.go
shajmakh: we can add it under the spec annotations; or inject a controller event (shown in the operator CR events) which policy is being used. If it doesn't make a difference why do you think this is an overkill?

controllers/numaresourcesoperator_controller.go
ffromani: Returning a new struct seems the best approach, but this needs a larger refactoring. I'll attempt something in my #1032

pkg/objectupdate/rte/rte.go
ffromani: adding a new function in `pkg/objectupdate/rte` seems overkilly to me, but not enough to change course just yet

shajmakh:
/hold 
rebase on top of https://github.com/openshift-kni/numaresources-operator/pull/1032 when merged

pkg/objectupdate/rte/rte.go
Tal-or: We tend to use annotation to reflect some useful information which is not exposed officially by the API.

In this case the SELinux context policy exposed explicitly in the daemonset/pod spec so it's part of the official API, hence annotation is not needed.

pkg/objectstate/merge/merge.go
```diff
@@ -55,44 +55,5 @@ func MetadataForUpdate(current, updated client.Object) (client.Object, error) {
 	updated.SetManagedFields(current.GetManagedFields())
 	updated.SetFinalizers(current.GetFinalizers())
 
-	_ = Annotations(current, updated)
```
Tal-or, line 58: This involve an actual change in the behavior - it won't keep the current labels, only the updated one.
What's the rational behind this change?
According to the commit message:
```
if the key is found in the current map it will
still exist in the updated one, while the required is to omit it.
```
But that means the user can override the labels, because the operator is no longer maintaining the existing ones.

shajmakh: > it won't keep the current labels, only the updated one.

That's exactly the issue. The problem in the current flow of updating the annotations is that in the case where the user intentionally didn't add the annotation (equivalent to deleting it if it exists) the code won't update that in the next results the annotation will be removed. It'll still exist. Then how the user is supposed to delete that specific annotation?
The annotations are built at every reconciliation iteration, and the RTE ObjectState carry this data until it gets ignored in this MetadataForUpdate.

shajmakh: As far as I have checked, updating the annotations never happened so far that is why this problem wasn't visible before this PR

Tal-or: I see your point, but this how we check for the annotation:
https://github.com/openshift-kni/numaresources-operator/blob/27a91fd43bd576f719a935fe2a5d54ed18a5e071/internal/api/annotations/annotations.go#L29 
so in case we want to disable it, we can change the value to be different than `custom`

Second thing, assuming we're not keeping the existing one like this PR suggests, and one day we'll add another annotation (to enable some other functionality).
If we would like to change one of the annotations we'll have to send all the other annotations as well, so the operator won't delete the existing ones.

@ffromani Could you please provide your POV here?

pkg/objectupdate/rte/rte.go
shajmakh: I might have missed this is already reflected. Can you point where this is set in the DS/pod spec please?

controllers/numaresourcesoperator_controller.go
```diff
@@ -490,7 +490,15 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 		}
 		rteupdate.DaemonSetHashAnnotation(r.RTEManifests.DaemonSet, cmHash)
 	}
-	rteupdate.SecurityContextConstraint(r.RTEManifests.SecurityContextConstraint, annotations.IsCustomPolicyEnabled(instance.Annotations))
+
+	customPolicyEnabled := annotations.IsCustomPolicyEnabled(instance.Annotations)
+
+	delete(r.RTEManifests.DaemonSet.Annotations, annotations.SELinuxPolicyConfigAnnotation)
```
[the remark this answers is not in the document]
shajmakh: The plan was to reset the annotations to allow discarding it from the DS if required; or in other words to set the annotation in the updated DS only if it is required to set a value to it

shajmakh:
RTE SELInux annotation is no longer relevant, 
populating the progressing condition reason and is fixed in https://github.com/openshift-kni/numaresources-operator/pull/1093 as well as the MCP status cleanup. Closing this.

PR #1061, opened 2024-11-04
manifests: relax node affinity for hypershift

shajmakh:
For local deployment (not via OLM), we configure the controller manager using yaml files instead of CSV. These yamls use requiredDuringSchedulingIgnoredDuringExecution affinity but on hypershift it was agreed that controller and scheduler be scheduled on the hosted cluster which contains worker nodes only.

related to
https://github.com/openshift-kni/numaresources-operator/pull/1059.


ffromani (commented):
/approve
/lgtm

PR #1062, opened 2024-11-06
CI: Update github actions versions

ffromani:
update CI github action versions for base images and for golang toolchain


ffromani:
prow failures are unaffected by this change which only affects github actions ci

PR #1063, opened 2024-11-07
mustgather: Dockerfile: use ocp CLI base image

shajmakh:
Initially, must-gather for NROP was built by grabbing the latest openshift-ose-must-gather image, copying the required binaries for running must gather and override the collection scripts and base that on rhel-minimal.

While this works, there is no need for extra dependencies on images when all we need is the openshift clients to gather the must-gather collection which is available in the CLI image.

### DIFF
--- a/Dockerfile.must-gather
+++ b/Dockerfile.must-gather
@@ -1,16 +1,4 @@
-FROM quay.io/openshift/origin-must-gather:4.18.0 AS builder
-
-FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
-RUN microdnf install -y procps-ng tar rsync
-RUN microdnf clean all
-
-# Copy must-gather required binaries
-COPY --from=builder /usr/bin/openshift-must-gather /usr/bin/openshift-must-gather
-COPY --from=builder /usr/bin/oc /usr/bin/oc
-
-# Save original gather script
-COPY --from=builder /usr/bin/gather* /usr/bin/
-RUN mv /usr/bin/gather /usr/bin/gather_original
+FROM quay.io/openshift/origin-cli:4.18
 
 COPY must-gather/collection-scripts/* /usr/bin/
 


Dockerfile.must-gather
```diff
@@ -1,17 +1,9 @@
-FROM quay.io/openshift/origin-must-gather:4.18.0 AS builder
+FROM quay.io/openshift/origin-cli:latest
```
ffromani, line 1: we should pick the latest _of the corrisponding Y stream_ not the latest in general

ffromani (commented):
conceptually fine, but let's narrow down the image

Dockerfile.must-gather
```diff
@@ -1,16 +1,4 @@
...
-RUN microdnf clean all
-
-# Copy must-gather required binaries
-COPY --from=builder /usr/bin/openshift-must-gather /usr/bin/openshift-must-gather
-COPY --from=builder /usr/bin/oc /usr/bin/oc
-
-# Save original gather script
-COPY --from=builder /usr/bin/gather* /usr/bin/
-RUN mv /usr/bin/gather /usr/bin/gather_original
+FROM quay.io/openshift/origin-cli:4.18.0
```
ffromani, line 1: ok, this is now me nitpicking, but still: is this tag pointing to the latest build of 4.18.z or to the latest build of 4.18.0?

must-gather/collection-scripts/gather
```diff
@@ -10,7 +10,7 @@ resources+=(nodes machineconfigs machineconfigpools featuregates kubeletconfigs)
 
 # run the collection of resources using must-gather
 for resource in ${resources[@]}; do
-  /usr/bin/oc adm inspect --dest-dir must-gather --all-namespaces ${resource}
+  oc adm inspect --dest-dir must-gather --all-namespaces ${resource}
```
ffromani, line 13: please keep the abspath (may need some amends)

Dockerfile.must-gather
ffromani: in addition, can you please link how the `origin-must-gather` image is built (its Dockerfile)?

shajmakh:
MG test passes with some adjustments, I'm post a fix in a different PR.

must-gather/collection-scripts/gather
shajmakh: oc is no longer explicitly copied because it exists in the new CLI base

Dockerfile.must-gather
shajmakh: 4.X and 4.X.0 have the same content. there are no 4.X.z images in quay:
https://quay.io/repository/openshift/origin-cli?tab=tags&tag=latest

shajmakh: I can't find the dockerfile for origin-must-gather but I know that the main must-gather uses the origin-cli base too:
https://github.com/openshift/must-gather/blob/master/Dockerfile.ocp

ffromani: thanks, this looks good enough!

ffromani: so can we use `origin-cli:4.18` like the linked Dockerfile does?

must-gather/collection-scripts/gather
shajmakh: we can but might cause issues d/s. why do you think we should keep it? 
A midstream MR depends on this actually, since there we use another dockerfile in terms of the base we aim to no longer copy the oc binary. I still need to see how m/s will end up. we can use an env var for the binary if that helps

ffromani: which kinds of issue you see here? I don't see the benefit in relaxing the path requirements here. Since we control (and we use a base image which we control) the base image, we know where `oc` must be, so we can use a absolute path

shajmakh: I haven't got the chance to check this yet, but using the m/s build was shouting that we don't have the oc binary installed in this path. I need to revisit the m/s batch and handle installing openshift clients properly first.

Dockerfile.must-gather
shajmakh: dropped this change we don't need it coupled with the base-image update

must-gather/collection-scripts/gather
shajmakh: dropped this change we don't need it coupled with the base-image update

Dockerfile.must-gather
ffromani: still unaddressed? can we use `4.18` (vs current `4.18.0`)?

ffromani:
/lgtm

thanks for the updates

PR #1064, opened 2024-11-11
controller: handle dangling ds when using poolname

Tal-or:
After the new API for PoolName, we are now should detect DaemonSet by `PoolName` and not only by MCP selector.

This will resolve the:
```
E1111 10:46:31.948164       1 controller.go:329] "Reconciler error" err="FailedRTESync: failed to apply (apps/v1, Kind=DaemonSet) numaresources/numaresourcesoperator-hostedcluster02: could not update object (apps/v1, Kind=DaemonSet) numaresources/numaresourcesoperator-hostedcluster02: Operation cannot be fulfilled on daemonsets.apps \"numaresourcesoperator-hostedcluster02\": StorageError: invalid object, Code: 4, Key: /kubernetes.io/daemonsets/numaresources/numaresourcesoperator-hostedcluster02, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: a2590c0a-e367-4422-81af-0658b7ccec74, UID in object meta: " controller="numaresourcesoperator" controllerGroup="nodetopology.openshift.io" controllerKind="NUMAResourcesOperator" NUMAResourcesOperator="numaresourcesoperator" namespace="" name="numaresourcesoperator" reconcileID="6abcc9d5-347f-4c75-9f96-9e4d9d532221"
```

issue

### DIFF
--- a/internal/dangling/dangling.go
+++ b/internal/dangling/dangling.go
@@ -50,12 +50,7 @@ func DeleteUnusedDaemonSets(cli client.Client, ctx context.Context, instance *nr
 		return err
 	}
 
-	expectedDaemonSetNames := sets.NewString()
-	for _, tree := range trees {
-		for _, mcp := range tree.MachineConfigPools {
-			expectedDaemonSetNames = expectedDaemonSetNames.Insert(objectnames.GetComponentName(instance.Name, mcp.Name))
-		}
-	}
+	expectedDaemonSetNames := buildDaemonSetNames(instance, trees)
 
 	var errs error
 	deleted := 0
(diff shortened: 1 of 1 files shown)


internal/dangling/dangling.go
```diff
@@ -127,3 +122,18 @@ func isOwnedBy(element metav1.Object, owner metav1.Object) bool {
 	}
 	return false
 }
+
+func buildDaemonSetNames(instance *nropv1.NUMAResourcesOperator, trees []nodegroupv1.Tree) sets.Set[string] {
+	expectedDaemonSetNames := sets.New[string]()
+	for _, tree := range trees {
+		poolName := tree.NodeGroup.PoolName
+		if poolName != nil && *poolName != "" {
```
shajmakh, line 130: no need to validate it's not empty string, we do that early at the validation phase for node group

shajmakh:
thanks for fixing this!
/lgtm

internal/dangling/dangling.go
Tal-or: I see.
As a general comment I would say that any contributor, that is not too familiar with the project and follow best practices, would add this trivial check since this is a pointer which is given to the function from unknown source (unknown in the context of the function per se)

ffromani: > I see. As a general comment I would say that any contributor, that is not too familiar with the project and follow best practices, would add this trivial check since this is a pointer which is given to the function from unknown source (unknown in the context of the function per se)

Not fully onboard (but not a problem in this case). I'd prefer adding testcases covering the flows first, which would help highlitghting cases like that. Reviews should also highlight this more.
Surely we should (my fault here) better docs/architecture docs/developer README to outline general guidelines like that.

PR #1065, opened 2024-11-12
hack:ci: remove trap

Tal-or:
The run-test-install-e2e.sh script is usually running as part of the u/s CI tests.
We don't want to catch the signal in trap because it's
masking the error which makes the test a false positive.

We want to check the error explicitly and fail loud instead,
because on CI there's no bother to revert the cluster back
to its initial state.

### DIFF
--- a/hack/run-test-install-e2e.sh
+++ b/hack/run-test-install-e2e.sh
@@ -10,12 +10,15 @@ fi
 
 setupreport
 
-# Make sure that we always properly clean the environment
-trap '{ echo "Running NRO uninstall test suite"; ${BIN_DIR}/e2e-nrop-uninstall.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.junit-report=${REPORT_DIR}/uninstall.xml; }' EXIT SIGINT SIGTERM SIGSTOP
-
 # Run install test suite
 echo "Running NRO install test suite"
-${BIN_DIR}/e2e-nrop-install.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.fail-fast --ginkgo.junit-report=${REPORT_DIR}/install.xml --ginkgo.focus='\[Install\] durability'
-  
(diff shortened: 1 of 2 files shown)


hack/run-test-install-e2e.sh
```diff
@@ -2,6 +2,8 @@
 
 source hack/common.sh
 
+ENABLE_CLEANUP="${ENABLE_CLEANUP:-false}"
+
```
ffromani, line 6: the install suite should always try the uninstall stage (there are legit tests here) BUT should never try to cleanup

hack/run-test-serial-e2e.sh
```diff
@@ -196,13 +196,15 @@ function runtests() {
 		${FOCUS}
 }
 
-# Make sure that we always properly clean the environment
```
ffromani, line 199: OTOH this should be optional because we still regularly run the serial suite in dev environments (contrary to every other suite)

ffromani (commented):
conceptually fine, but we should swap the flag in the modified cases :)

hack/run-test-serial-e2e.sh
Tal-or: then it's already optional here, because of:
https://github.com/openshift-kni/numaresources-operator/blob/bb5985c336609a958b39258b0594cca7f33f26c4/hack/run-test-serial-e2e.sh#L158
so no changes are needed here

hack/run-test-serial-e2e.sh
```diff
@@ -197,12 +197,16 @@ function runtests() {
 }
 
 # Make sure that we always properly clean the environment
-trap 'teardown' EXIT SIGINT SIGTERM SIGSTOP
+trap 'teardown' EXIT SIGINT SIGTERM
```
Tal-or, line 200: https://github.com/koalaman/shellcheck/wiki/SC2173

ffromani (commented):
/hold

feel free to either unhold or resubmit if you want to address the nits

hack/run-test-serial-e2e.sh
ffromani: this was better done in a separate commit for documentation purposes. Reason is git history is always avaialble and should be self-sufficient. Otherwise, we now need to access github and review the PR review history to learn the why for this change.

Not a big deal but let's try to move towards this direction in the future

Tal-or: You're right I was lazy. fixed that

ffromani:
unmasking the failures seems the most efficient way forward, so let's go option 1 and force-merge this PR.

PR #1067, opened 2024-11-12
deploy-mco-crds.sh: do not install mco crds on hypershift

Tal-or:
On HyperShift hostedcluster, the concepts MCO and KubeletConfig do not exist.
Henec, we should not install their CRDs while using calling `make deploy`


ffromani (commented):
/approve
/lgtm

PR #1027, opened 2024-09-27
ctrl: HCP adaptations for NodeGroup validation and RTE

shajmakh:
Adapt the possibility of running the operator on Hypershift platform by
replacing MCP context with PoolName where relevant; In addition to extra
NodeGroup validation on hypershift.

PR #1068, opened 2024-11-13
e2e: mustgather: fix test to verify TAS enabled nodes only

shajmakh:
So far the CI worked with the default node labels which is "worker" and
the NodeGroups of the numaresourcesoperator CR we expected to target the
`worker` mcp. However, this isn't always the case and the test should be
smarter to determine which nodes it should look for their folders in the
collected must-gather.

Adjust the step of building the expected CRs paths to include only nodes
that the operator targets.

### DIFF
--- a/test/e2e/must-gather/must_gather_test.go
+++ b/test/e2e/must-gather/must_gather_test.go
@@ -31,9 +31,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/yaml"
 
+	nodegroupv1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1/helper/nodegroup"
 	"github.com/openshift-kni/numaresources-operator/internal/wait"
-
 	e2eclient "github.com/openshift-kni/numaresources-operator/test/utils/clients"
+
+	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 )
 
 var _ = ginkgo.Describe("[must-gather] NRO data collected", func() {
@@ -79,7 +81,7 @@ var _ = ginkgo.Describe("[must-gather] NRO data collected", func() {
 			os.RemoveAll(destDir)
 		})
 
-		ginkgo.It("check NRO data files have been collected", func() {
+		ginkgo.It("check NRO data files have been collected", func(ctx context.Context) {
 			crdDefinitions := []string{
 				"cluster-scoped-resources/apiextensions.k8s.io/customresourcedefinitions/noderesourcetopologies.topology.node.k8s.io.yaml",
 				"cluster-scoped-resources/apiextensions.k8s.io/customresourcedefinitions/numaresourcesoperators.nodetopology.openshift.io.yaml",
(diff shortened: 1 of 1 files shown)


ffromani (commented):
the intent of the change is good, but these tests should NOT query the live cluster, but only  operate on the must-gather data. IOW, the tests should not assume there is a live cluster at all (there could be bugs on which we do query the live cluster: these are just bugs and need to be removed ASAP)

shajmakh:
> the intent of the change is good, but these tests should NOT query the live cluster, but only operate on the must-gather data. IOW, the tests should not assume there is a live cluster at all (there could be bugs on which we do query the live cluster: these are just bugs and need to be removed ASAP)


Ack, I'll re-add the deleted function and fetch TAS nodes from the collected info. Thanks for highlighting this.

Tal-or:
Can we use the new API AKA `PoolName` so there will be less work when we would like to adapt for HCP?

ffromani:
> Can we use the new API AKA `PoolName` so there will be less work when we would like to adapt for HCP?

Yes please let's try to do like this

shajmakh:
> 
Can we use the new API AKA PoolName so there will be less work when we would like to adapt for HCP?


same mind here, and that's where `GetListByLabels()` come in handy. However this depends on `FindTreesHypershift` from  https://github.com/openshift-kni/numaresources-operator/pull/1027/files#diff-31ede3f56b1f873ccf421371942426d8ad0c30160d467d790dd7e050802d25d7R109 so we need that to get merged first

shajmakh:
> the intent of the change is good, but these tests should NOT query the live cluster, but only operate on the must-gather data. IOW, the tests should not assume there is a live cluster at all (there could be bugs on which we do query the live cluster: these are just bugs and need to be removed ASAP)

 There are 2 places where we need a live cluster for this test:
 * get the MCP labels
 * get the daemonset namespace 
 
we can use a hardcoded namespace but the most important part is the gathering the node selectors from the MCPs. Only on hypershift we wouldn't need that because we know how we build the labels if PoolName is provided on HCP (part of  #1027). On OCP we'll always need MCPs even if PoolName is the node group specifier that was set.

shajmakh:
> > the intent of the change is good, but these tests should NOT query the live cluster, but only operate on the must-gather data. IOW, the tests should not assume there is a live cluster at all (there could be bugs on which we do query the live cluster: these are just bugs and need to be removed ASAP)
> 
> There are 2 places where we need a live cluster for this test:
> 
> * get the MCP labels
> * get the daemonset namespace
> 
> we can use a hardcoded namespace but the most important part is the gathering the node selectors from the MCPs. Only on hypershift we wouldn't need that because we know how we build the labels if PoolName is provided on HCP (part of #1027). On OCP we'll always need MCPs even if PoolName is the node group specifier that was set.

that can actually be done by checking the collected MCPs. I'll verify that is possible and update.

shajmakh:
> > > the intent of the change is good, but these tests should NOT query the live cluster, but only operate on the must-gather data. IOW, the tests should not assume there is a live cluster at all (there could be bugs on which we do query the live cluster: these are just bugs and need to be removed ASAP)
> > 
> > 
> > There are 2 places where we need a live cluster for this test:
> > 
> > * get the MCP labels
> > * get the daemonset namespace
> > 
> > we can use a hardcoded namespace but the most important part is the gathering the node selectors from the MCPs. Only on hypershift we wouldn't need that because we know how we build the labels if PoolName is provided on HCP (part of #1027). On OCP we'll always need MCPs even if PoolName is the node group specifier that was set.
> 
> that can actually be done by checking the collected MCPs. I'll verify that is possible and update.

Updated in latest push. We still need to depend on the live daemonset to fetch the namespace as it can be different depending on the deployment method as even mustgather uses that:
https://github.com/openshift-kni/numaresources-operator/blob/main/must-gather/collection-scripts/namespace

PR #1069, opened 2024-11-13
Add missing controller test for PoolName

shajmakh:
add missing test for empty pool name and extra verification for Available condition

### DIFF
--- a/controllers/numaresourcesoperator_controller_test.go
+++ b/controllers/numaresourcesoperator_controller_test.go
@@ -222,6 +222,12 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
 					result, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: nroKey})
 					Expect(err).ToNot(HaveOccurred())
 					Expect(result).To(Equal(reconcile.Result{}))
+
+					Expect(reconciler.Client.Get(context.TODO(), client.ObjectKeyFromObject(nro), nro)).ToNot(HaveOccurred())
+
+					availableCondition := getConditionByType(nro.Status.Conditions, status.ConditionAvailable)
+					Expect(availableCondition).ToNot(BeNil())
+					Expect(availableCondition.Status).To(Equal(metav1.ConditionTrue))
 				})
 
 				It("should create all CRDs and objects and operator status are updated from the first reconcile iteration", func() {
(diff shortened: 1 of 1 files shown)


shajmakh:
/unhold
the value for this is for OCP only and in that context shouldn't be blocked. HCP cases are anyway broken and needs another PR to fix: https://github.com/openshift-kni/numaresources-operator/pull/1070

ffromani (commented):
the additions here look good. IIRC, in #1064 we fixed the bug on which the DS _was_ created, but immediately removed because the bug in the `dangling` pkg. I'd like a test which specifically check for this condition. Is the second commit targeting explicitely that case? it seems related but not too explicit.

ffromani:
> > the additions here look good. IIRC, in #1064 we fixed the bug on which the DS was created, but immediately removed because the bug in the dangling pkg. I'd like a test which specifically check for this condition. Is the second commit targeting explicitely that case? it seems related but not too explicit.
> 
> Yes the intended test is in the second commit. The problem appeared in HCP but the test doesn't run on that platform, and would need fixes anyway. I haven't tested if the bug occurs on OCP manually, will do the checks and update the code as needed.

fair enough. Please add a comment above the addition explicitely mentioning that PR, mentioning the bug we had and how this check helps in this area. Then I think I can move on and merge

shajmakh:
/hold 
for https://github.com/openshift-kni/numaresources-operator/pull/1083 to be merged first as it will enable the tests on hypershift. Also to avoid rebase conflicts

ffromani:
/hold

let's wait for #1071 to go in

ffromani:
/hold cancel

#1071 merged

ffromani (commented):
/lgtm

/retest-required

ffromani:
can we please rebase on top of current main branch?

shajmakh:
Following further examination, the bug fixed in #1064 can't be caught in the controller tests but in the e2e instead, because we need to look at the readiness of the ds not only the creation. This is covered by the e2e/serial  e2e/rte and e2e/install. I adjusted the second commit message.

shajmakh:
/hold 
for https://github.com/openshift-kni/numaresources-operator/pull/1083 to avoid uneeded complex rebase conflicts

ffromani (commented):
/lgtm

shajmakh:
e2e-install failure aligns with the new default behavior
/hold

PR #1070, opened 2024-11-13
WIP: ctrl: tests: add env var TEST_PLATFORM to choose the platform

shajmakh:
TBD

This will make unit tests fail in the CI and is the anticipated result

### DIFF
--- a/controllers/numaresourcesoperator_controller_test.go
+++ b/controllers/numaresourcesoperator_controller_test.go
@@ -146,7 +146,7 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
 		})
 	})
 
-	Context("with NRO mutiple pool specifiers set on same node group", func() {
+	Context("with NRO multiple pool specifiers set on same node group", func() {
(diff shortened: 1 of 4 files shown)


controllers/numaresourcesoperator_controller_test.go
```diff
@@ -59,6 +61,26 @@ const (
 	defaultOCPVersion = "v4.14"
 )
 
+var testPlatform platform.Platform
+
+func init() {
```
Tal-or, line 66: Using init is problematic here, because it will trigger this function even on production code since it's part of the `controller` package.

Tal-or (commented):
The general intent is good, but I think using `DescribeTable` here is better here.
There should be an `Entry` for OCP and  an `Entry` for HCP

controllers/numaresourcesoperator_controller_test.go
shajmakh: although this is initializing TEST_PLATFORM which is for tests use only, it is redundant for the manager code. I renamed the function and called it the main describe

PR #1072, opened 2024-11-14
Bump golang build 1.23

ffromani:
bump golang builder images and versions to 1.23.3 in both github actions CI and Dockerfiles


shajmakh:
Thanks for this.
why not use the latest stable build 1.23.3?

ffromani:
> Thanks for this. why not use the latest stable build 1.23.3?

because I didn't notice! fixed.

PR #1071, opened 2024-11-13
e2e: refactor NRO deployment

Tal-or:
At the beginning of the u/s tests we're deploying the NRO/Scheduler objects.
This PR introduce a refactor of the deployment code, so each k8s flavor will be able to implement it for its need.

At the moment we're having 3 flavors:
* Kuberenetes (vanilla)
* OpenShift
* HyperShift

In this PR the HyperShift logic is not implemented because there're missing details from the CI perspective such as the `KUBECONFIG` of the MNG cluster. we'll implement the missing code in future PR.

PR #1078, opened 2024-11-18
e2e:hypershift: helper functions

Tal-or:
Adding some helper packages for interacting with `NodePools` object and talking to the HyperShift MNG cluster.
This work inspired by similar work done on: https://github.com/openshift/cluster-node-tuning-operator


ffromani (commented):
the problem I have with this code is just one:
if we add in `test/utils` adding tests becomes unnecessarily hard. It would be much better to add code in `internal/` and use it (only) in tests. How much of this code can we move in `internal/`?

Tal-or:
> if we add in `test/utils` adding tests becomes unnecessarily hard

Why?

> How much of this code can we move in internal/?
Should not be a big deal, but then we might need to remove all `ginkgo/gomega` mentions since those are supposed to be used only under test code

ffromani:
> > if we add in `test/utils` adding tests becomes unnecessarily hard
> 
> Why?

There's no inherent reason, we can fix with enough effort. But considering that `utils` is a bad pattern anyway, I'd rather keep moving code towards `internal` or `pkg` and/or dissolve them in the calling sites (tests) rather than adding stuff in `test/utils`. With some targeted exceptions of course.

> > How much of this code can we move in internal/?
> 
> Should not be a big deal, but then we might need to remove all `ginkgo/gomega` mentions since those are supposed to be used only under test code

That's fair. `clients` can stay there, the stuff filed under `nodepools` can probably be moved, at least a big chunk of it, in `internal/wait` (which is untested, but that's another problem for another day)

ffromani:
ok, to summarize:
- `clients`: no change needed
- `hypershift`: no change needed, we can check again later
- `nodepools`: please move most that you can in `internal/wait`

unit tests added in this process are a nice bonus but not required

Tal-or:
/hold cancel
#https://github.com/openshift/release/pull/58789 merged

Tal-or:
> ok, to summarize:
> 
> * `clients`: no change needed
> * `hypershift`: no change needed, we can check again later
> * `nodepools`: please move most that you can in `internal/wait`
> 
> unit tests added in this process are a nice bonus but not required

Done.
About unit tests - it's a bit unusual to write a unit tests for a test code.

Anyway, this code will be exercised as part of #1071, we can wait for #1071 to be completed and only if hypershift CI is passing we'll merge this one

ffromani:
> Done. About unit tests - it's a bit unusual to write a unit tests for a test code.

This is *exactly* why we should move code in` pkg` or `internal` :)

ffromani (commented):
/lgtm

PR #1079, opened 2024-11-19
sched: manifests: disable leader election by default

ffromani:
Defaults for leader election are questionable at best and likely harmful. So, we should default the leader election to off and enable it only if user sets so, because this will make us to tune all the leader election parameters and set sane values.


ffromani:
xref: 4ac0f7cb236b232eb7f0066854c490fa73bf8936

ffromani:
/override  ci/prow/ci-e2e-install-hypershift 

the lane is known broken *AND* we are actively working to fix it

PR #1080, opened 2024-11-19
sched: always set the leader election parameters

ffromani:
Because of a overlook, we used to set the leader election
params only if replicas > 1 was requested. This left the key *and default* corner case of replicas=1 with
compiled in defaults, which are questionable at best and most likely harmful for our use case.

Make sure to always set sane parameters, obviously taking into account the user desires from the spec
(Replicas field).

Add more logs to make troubleshooting easier

### DIFF
--- a/controllers/numaresourcesscheduler_controller.go
+++ b/controllers/numaresourcesscheduler_controller.go
@@ -45,20 +45,12 @@ import (
 	"github.com/openshift-kni/numaresources-operator/pkg/apply"
 	"github.com/openshift-kni/numaresources-operator/pkg/hash"
 	"github.com/openshift-kni/numaresources-operator/pkg/loglevel"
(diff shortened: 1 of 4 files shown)


ffromani:
/hold

we need controller tests

ffromani:
/hold cancel

tests added

Tal-or:
The waiting period for MCP to catch up has become flaky recently.

I wonder if it's somehow related to one of our recent changes or there's something broken in OCP

ffromani:
same thoughts here. I'll keep watching

ffromani:
/override ci/prow/ci-e2e-install-hypershift

fixes still pending, this is expected

ffromani:
/hold

let's wait for #1071 to go in

ffromani:
/hold cancel

#1071 merged

PR #1083, opened 2024-11-20
Support NROP controller tests on hypershift

shajmakh:
The PR includes cleanups and reorganizing the controller tests code by splitting the tests into groups based on what's relevant to be executed on what platforms. 
Please check the commits for more details.

### DIFF
--- a/controllers/numaresourcesoperator_controller_test.go
+++ b/controllers/numaresourcesoperator_controller_test.go
@@ -1005,7 +1005,7 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
 	)
 
 	Describe("Openshift only", func() {
-		Context("[openshift] without available machine config pools", func() {
+		Context("[openshift] without available machine config pools", Label("platform:openshift"), func() {
 			It("should update the CR condition to degraded when MachineConfigPoolSelector is set", func() {
 				ng := nropv1.NodeGroup{
 					MachineConfigPoolSelector: &metav1.LabelSelector{MatchLabels: map[string]string{"test": "test"}},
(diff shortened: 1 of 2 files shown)


shajmakh:
expected CI to fail. Blocked by #1027.
pending modification on the expected results for Hypershift (mainly ignore everything related to MCP)

shajmakh:
relevant commits start from https://github.com/openshift-kni/numaresources-operator/pull/1083/commits/7b0f95951b43220c0760b7fc8e78fc48ab7e3b3a

controllers/numaresourcesoperator_controller_test.go
```diff
@@ -133,7 +133,7 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
 		})
 	})
 
-	Context("without available machine config pools", func() {
+	Context("[openshift] without available machine config pools", func() {
```
ffromani: we should use labels also in the controller tests but too big of a deal yet

controllers/numaresourcesoperator_controller_test.go
```diff
@@ -186,7 +186,71 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
 		})
 	})
 
-	Context("with node group with MCP selector that matches more than one MCP", func() {
+	Context("with two node groups while both point to same pool using same pool specifier", func() {
```
ffromani: classifying existing tests and adding new tests should be preferably done in 2 separate commits

ffromani (commented):
/lgtm
/approve

this is another place when we need to review if/how we can abstract the platform instead of adding 100's `if platform == `, but first we need to have it working and let the dust settle.

controllers/numaresourcesoperator_controller_test.go
shajmakh: addressed in last commit, thanks!

controllers/numaresourcesoperator_controller_test.go
shajmakh: addressed in last commit, thanks!

controllers/numaresourcesoperator_controller_test.go
shajmakh: disunited the content in the first 2 commits, thanks for catching

ffromani (commented):
/lgtm

PR #1088, opened 2024-11-21
CNF-15733:kubeletcontroller:hypershift: configmap cleanup

Tal-or:
On hypershift we cannot rely on the automatic garbage collection mechanism provided by k8s controllers.

The reason is becuase we can not set the owner reference across namespaces.

Hence, we need to add some custom logic to cleanup the ConfigMap generated for RTE.

### DIFF
--- a/controllers/kubeletconfig_controller.go
+++ b/controllers/kubeletconfig_controller.go
@@ -20,10 +20,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	serializer "k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/apimachinery/pkg/types"
@@ -66,6 +68,9 @@ type KubeletConfigReconciler struct {
 	Recorder  record.EventRecorder
 	Namespace string
 	Platform  platform.Platform
+	// garbageCollectionFor is a list of objects
+	// that their dependent needs to be removed from the cluster.
+	garbageCollectionFor []*corev1.ConfigMap
 }
 
 type kubeletConfigHandler struct {
@@ -140,6 +145,15 @@ func (r *KubeletConfigReconciler) SetupWithManager(mgr ctrl.Manager, clusterPlat
 			_, ok := kubelet.Labels[HypershiftKubeletConfigConfigMapLabel]
 			return ok
 		})
+		p.DeleteFunc = func(e event.DeleteEvent) bool {
+			kubelet := e.Object.(*corev1.ConfigMap)
+			if _, ok := kubelet.Labels[HypershiftKubeletConfigConfigMapLabel]; !ok {
+				return false
+			}
+			klog.InfoS("KubeletConfig ConfigMap object got deleted", "KubeletConfig", kubelet.Name)
+			r.garbageCollectionFor = append(r.garbageCollectionFor, kubelet)
+			return true
+		}
 	}
 	return ctrl.NewControllerManagedBy(mgr).
 		For(o, builder.WithPredicates(p)).
@@ -161,6 +175,13 @@ func (e *InvalidKubeletConfig) Unwrap() error {
 }
 
 func (r *KubeletConfigReconciler) reconcileConfigMap(ctx context.Context, instance *nropv1.NUMAResourcesOperator, kcKey client.ObjectKey) (*corev1.ConfigMap, error) {
+	// first check if the ConfigMap should be deleted
+	// to save all the additional work related for create/update
+	cm, deleted, err := r.deleteConfigMap(ctx, instance, kcKey)
+	if deleted {
+		return cm, err
+	}
+
 	kcHandler, err := r.makeKCHandlerForPlatform(ctx, instance, kcKey)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


controllers/kubeletconfig_controller.go
```diff
@@ -140,6 +144,15 @@ func (r *KubeletConfigReconciler) SetupWithManager(mgr ctrl.Manager, clusterPlat
 			_, ok := kubelet.Labels[HypershiftKubeletConfigConfigMapLabel]
 			return ok
 		})
+		p.DeleteFunc = func(e event.DeleteEvent) bool {
+			kubelet := e.Object.(*corev1.ConfigMap)
+			if _, ok := kubelet.Labels[HypershiftKubeletConfigConfigMapLabel]; !ok {
+				return false
+			}
+			klog.InfoS("KubeletConfig ConfigMap object got deleted", "KubeletConfig", kubelet.Name)
+			r.GarbageCollectionFor = append(r.GarbageCollectionFor, *kubelet)
```
ffromani, line 153: maybe a deepcopy?

controllers/kubeletconfig_controller.go
```diff
@@ -66,6 +67,9 @@ type KubeletConfigReconciler struct {
 	Recorder  record.EventRecorder
 	Namespace string
 	Platform  platform.Platform
+	// GarbageCollectionFor is a list of objects
+	// that their dependent needs to be removed from the cluster.
+	GarbageCollectionFor []corev1.ConfigMap
```
ffromani, line 72: does this need to be public member?

ffromani (commented):
looks better than I anticipated. How hard is to actually run the logic only if running on hypershift?

Tal-or:
> looks better than I anticipated. How hard is to actually run the logic only if running on hypershift?

The interception of the deletion event is done only for HyperShift predicates

ffromani:
> > looks better than I anticipated. How hard is to actually run the logic only if running on hypershift?
> 
> The interception of the deletion event is done only for HyperShift predicates

Right, I should have been more careful in reading the code :P

controllers/kubeletconfig_controller.go
Tal-or: No. Let's change it

controllers/kubeletconfig_controller.go
```diff
@@ -275,6 +295,51 @@ func (r *KubeletConfigReconciler) makeKCHandlerForPlatform(ctx context.Context,
...
+		klog.V(2).InfoS("could not delete ConfigMap since it was not found", "configmapName", dependentCM.Name)
+	}
+	r.GarbageCollectionFor = removeDeletedOwner(kcKey, r.GarbageCollectionFor)
+	return cm, true, nil
+}
+
+func getDeletedOwner(kcKey client.ObjectKey, ownerConfigMaps []corev1.ConfigMap) *corev1.ConfigMap {
+	for i := range ownerConfigMaps {
+		cm := &ownerConfigMaps[i]
+		if client.ObjectKeyFromObject(cm).String() == kcKey.String() {
```
ffromani, line 326: let's check if there's a recommended way to compare keys. I don't see obvious issues with this one, but still

controllers/kubeletconfig_controller.go
Tal-or: `DeepCopy` still return a pointer I can make this a slice of pointers and save some redundant copying

test/utils/deploy/hypershift.go
```diff
@@ -91,6 +95,25 @@ func (h *HyperShiftNRO) Teardown(ctx context.Context, timeout time.Duration) {
 		Expect(wait.ForConfigToBeReady(ctx, e2eclient.MNGClient, np.Name, np.Namespace)).To(Succeed())
 
 		Expect(e2eclient.MNGClient.Delete(ctx, h.KcConfigMapObj)).To(Succeed())
+
+		By("checking that generated configmap has been deleted")
```
Tal-or, line 99: IMO this is not the ideal place for checking the deletion. 
It would be more appropriate in the `uninstall` suite.
But after checking the `uninstall` suite I see it completely redundant.
We're deleting the nro and kubeletconfig in the `uninstall` suite but these objects were already removed here:
https://github.com/openshift-kni/numaresources-operator/blob/7de8e4b52dcff68106fa0efa8f5b218acfdf423a/test/e2e/install/install_test.go#L203 

In the nutshell the `uninstall` need to go through some refactoring but this is another issue for another PR.

Tal-or:
/hold cancel
This PR is ready

controllers/kubeletconfig_controller.go
```diff
@@ -275,6 +295,51 @@ func (r *KubeletConfigReconciler) makeKCHandlerForPlatform(ctx context.Context,
...
+
+func removeDeletedOwner(kcKey client.ObjectKey, ownerConfigMaps []*corev1.ConfigMap) []*corev1.ConfigMap {
+	for i := range ownerConfigMaps {
+		cm := ownerConfigMaps[i]
+		if client.ObjectKeyFromObject(cm) == kcKey {
+			return append(ownerConfigMaps[:i], ownerConfigMaps[i+1:]...)
+		}
+	}
+	return ownerConfigMaps
+}
```
ffromani, line 341: This does not look obviously wrong but makes me wary. I'd go for the trivialest implementation:
```
func removeDeletedOwner(kcKey client.ObjectKey, ownerConfigMaps []*corev1.ConfigMap) []*corev1.ConfigMap {
        ret := make([]*corev1.ConfigMap, 0, len(ownerConfigMaps))
	for i := range ownerConfigMaps {
		cm := ownerConfigMaps[i]
		if client.ObjectKeyFromObject(cm) == kcKey {
			continue
		}
                ret = append(ret, cm)
	}
	return ret
}
```
perhaps something in the `slices` package can help?

ffromani: maybe https://pkg.go.dev/slices#DeleteFunc ?

test/utils/deploy/hypershift.go
ffromani: make sense to me with the catch that suites should not depend on each other as much as possible; IOW they should be indpedent and self-sufficient.

controllers/kubeletconfig_controller.go
Tal-or: Nice. I used https://pkg.go.dev/slices#Delete which is good enough

controllers/kubeletconfig_controller.go
```diff
@@ -275,6 +296,50 @@ func (r *KubeletConfigReconciler) makeKCHandlerForPlatform(ctx context.Context,
...
+			return cm
+		}
+	}
+	return nil
+}
+
+func removeDeletedOwner(kcKey client.ObjectKey, ownerConfigMaps []*corev1.ConfigMap) []*corev1.ConfigMap {
+	for i := range ownerConfigMaps {
+		if client.ObjectKeyFromObject(ownerConfigMaps[i]) == kcKey {
+			return slices.Delete(ownerConfigMaps, i, i+1)
```
ffromani, line 337: ok, let's start like this for now

ffromani (commented):
/approve
/lgtm

PR #1091, opened 2024-11-26
e2e: more robust wait for MCP updates

ffromani:
instead of chasing the MCP status transition updated -> updating -> updated, which is fragile and hard to get right both in theory and in practice, let's try a different approach.

When we change a MCP status (e.g. sending a new MachineConfig and/or pausing/unpausing)
1. we record the current MCP resourceversion
2. we ask for our desired state
3. we wait for the MCP desired state to be `updated` again *but with
   different resourceversion*

this should allow us to capture the conditions which describe our desired state while waiting for a stable state in a more robust way.

Add minorish cleanups along the way.

### DIFF
--- a/internal/wait/machineconfigpool.go
+++ b/internal/wait/machineconfigpool.go
@@ -52,6 +52,11 @@ func (wt Waiter) ForMachineConfigPoolCondition(ctx context.Context, mcp *machine
 		if err != nil {
 			return false, err
 		}
+		klog.Infof("mcp: updated resourceversion %q (reference: %q)", updatedMcp.ResourceVersion, mcp.ResourceVersion)
+		if updatedMcp.ResourceVersion == mcp.ResourceVersion {
+			// nothing yet changed, need to recheck later
+			return false, nil
+		}
 		for _, cond := range updatedMcp.Status.Conditions {
 			if cond.Type == condType {
 				if cond.Status == corev1.ConditionTrue {
--- a/internal/wait/machineconfigpool.go
+++ b/internal/wait/machineconfigpool.go
@@ -34,10 +34,10 @@ const (
 
 func (wt Waiter) ForMachineConfigPoolDeleted(ctx context.Context, mcp *machineconfigv1.MachineConfigPool) error {
 	immediate := false
-	err := k8swait.PollUntilContextTimeout(ctx, wt.PollInterval, wt.PollTimeout, immediate, func(aContext context.Context) (bool, error) {
+	err := k8swait.PollUntilContextTimeout(ctx, wt.PollInterval, wt.PollTimeout, immediate, func(ctx context.Context) (bool, error) {
 		updatedMcp := machineconfigv1.MachineConfigPool{}
 		key := ObjectKeyFromObject(mcp)
-		err := wt.Cli.Get(aContext, key.AsKey(), &updatedMcp)
+		err := wt.Cli.Get(ctx, key.AsKey(), &updatedMcp)
 		return deletionStatusFromError("MachineConfigPool", key, err)
 	})
 	return err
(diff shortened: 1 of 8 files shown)


ffromani:
```
Failed to create cluster	{"error": "could not complete platform specific options: failed to create infra: cannot list vpc endpoints: RequestLimitExceeded:
```

ffromani:
```
 Registry server Password: <<non-empty>>
error: build error: Failed to push image: trying to reuse blob sha256:3d252ab42824f0f833f0fcf3c660065bd878cb4de12b9c69b6a1758287e338e8 at destination: unable to retrieve auth token: invalid username/password: authentication required 
```

test/e2e/uninstall/uninstall_test.go
```diff
@@ -82,8 +82,7 @@ var _ = Describe("[Uninstall] clusterCleanup", Serial, func() {
 				klog.Warningf("failed to delete the kubeletconfigs %q", kcObj.Name)
 			}
 
-			err = unpause()
-			Expect(err).NotTo(HaveOccurred())
+			Eventually(unpause).WithTimeout(configuration.MachineConfigPoolUpdateTimeout).WithPolling(configuration.MachineConfigPoolUpdateInterval).Should(Succeed())
```
Tal-or, line 85: nit: Both the interval and timeout are too long in this case. Also the names are a bit out of context here.
this should be and interval and timeout for the `Update` call which makes it a bit unclear

internal/wait/machineconfigpool.go
```diff
@@ -52,6 +52,11 @@ func (wt Waiter) ForMachineConfigPoolCondition(ctx context.Context, mcp *machine
 		if err != nil {
 			return false, err
 		}
+		klog.Infof("mcp: updated RV=%q reference RV=%q", updatedMcp.ResourceVersion, mcp.ResourceVersion)
```
Tal-or, line 55: while only reading the logs (without the code context) it's hard to understand what `RV` is

test/utils/deploy/openshift.go
```diff
@@ -47,6 +47,10 @@ func (o *OpenShiftNRO) deployWithLabels(ctx context.Context, matchLabels map[str
 	kcObj, err := objects.TestKC(matchLabels)
 	Expect(err).To(Not(HaveOccurred()))
 
+	// need to get reference values before any mutations
+	mcps, err := nropmcp.GetListByNodeGroupsV1(ctx, e2eclient.Client, nroObj.Spec.NodeGroups)
+	Expect(err).NotTo(HaveOccurred())
+
 	unpause, err := e2epause.MachineConfigPoolsByNodeGroups(nroObj.Spec.NodeGroups)
```
Tal-or, line 54: maybe move to `Eventually` here as well

Tal-or (commented):
Looking good, thanks!
small comments inside

internal/wait/machineconfigpool.go
ffromani: good point, will expand

test/e2e/uninstall/uninstall_test.go
ffromani: fixing the out-of-context is a bit hard and would need a deep review of `objects/pause`. But I can quickly fix the names using aliases.

test/utils/deploy/openshift.go
ffromani: I think just wrapping into `Eventually` is more a band-aid fix. I'll look for a redesign of the code in a later PR.

ffromani:
> ```
> [Install] durability with a running cluster with all the components and overall deployment [It] should be able to delete NUMAResourceOperator CR and redeploy without polluting cluster state
> /go/src/github.com/openshift-kni/numaresources-operator/test/e2e/install/install_test.go:281
>   [FAILED] Unexpected error:
>       <*meta.NoKindMatchError | 0xc00076ecc0>: 
>       no matches for kind "MachineConfigPool" in version "machineconfiguration.openshift.io/v1"
>       {
>           GroupKind: {
>               Group: "machineconfiguration.openshift.io",
>               Kind: "MachineConfigPool",
>           },
>           SearchedVersions: ["v1"],
>       }
>   occurred
>   In [It] at: /go/src/github.com/openshift-kni/numaresources-operator/test/e2e/install/install_test.go:331 @ 11/27/24 11:07:41.37
> ```
> 
> Something has changed in the test logic that causes HCP flow to ask for MCP from the API server which leads to the error above

thanks, will check

test/e2e/install/install_test.go
```diff
@@ -326,13 +326,14 @@ var _ = Describe("[Install] durability", Serial, func() {
 			// TODO change to an image which is test dedicated
 			nroObjRedep.Spec.ExporterImage = e2eimages.RTETestImageCI
 
+			// need to get MCPs before the mutation
```
ffromani, line 329: this is the culprit for the hypershift lane failure

PR #1093, opened 2024-11-28
controller: nrop: always populate reason (progressing)

ffromani:
The API spec requires the condition.Reason to always contain a non-empty CamelCase string. We added the infra to do so but we forgot for the Progressing case.

This change fixes that.
This becomes evident if RTEs fail to start, otherwise it's easy to miss, like it probably happened till now.


ffromani:
more tests are coming with a followup PR

PR #1098, opened 2024-12-04
Separated NROP uninstall from install script

rbaturov:
Invoking make test-install-e2e currently handles both the installation and uninstallation of NROP and its components. We propose separating this into two distinct suites, introducing a dedicated script for the uninstallation process. This separation would be beneficial in scenarios where we need to install and test the operator, upgrade it, and only then proceed with its deletion.

For example, this work will be needed in the [NROP upgrade test lane](https://github.com/openshift/release/pull/58457),
As we wish to:
Install the operator with 4.17 version, run e2e install suite, upgrade to 4.18 and only then uninstall. 

Note, after merging this PR I'll create a PR in the release repo to ensure both calls ([first](https://github.com/openshift/release/blob/master/ci-operator/config/openshift-kni/numaresources-operator/openshift-kni-numaresources-operator-main.yaml#L102) and [second](https://github.com/openshift/release/blob/master/ci-operator/config/openshift-kni/numaresources-operator/openshift-kni-numaresources-operator-main.yaml#L181)) of `make test-install-e2e` will be replaced with make `test-install-e2e test-uninstall-e2e`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,6 +135,9 @@ test-e2e: build-e2e-all
 test-install-e2e: build-e2e-all
 	hack/run-test-install-e2e.sh
 
+test-uninstall-e2e: build-e2e-all
+	hack/run-test-uninstall-e2e.sh
+
 test-must-gather-e2e: build-must-gather-e2e
 	hack/run-test-must-gather-e2e.sh
 
--- a/hack/run-test-install-e2e.sh
+++ b/hack/run-test-install-e2e.sh
@@ -16,9 +16,3 @@ if ! "${BIN_DIR}"/e2e-nrop-install.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=
   echo "Failed to run NRO install test suite"
   exit 1
 fi
-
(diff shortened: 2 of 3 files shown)


ffromani (commented):
/approve
/lgtm

Makefile
```diff
@@ -135,6 +135,9 @@ test-e2e: build-e2e-all
 test-install-e2e: build-e2e-all
 	hack/run-test-install-e2e.sh
```
ffromani, line 136: do we need changes to this wrapper script, perhaps just changing the defaults?
it seems to me this PR is not separating, rather adding a new targeted suite.

rbaturov: My bad! forgot to delete that part from the install script.

rbaturov: Updated the code to reflect this.
Sorry for the confusion.

ffromani:
#1099 merged, please rebase on top of current master

ffromani (commented):
/lgtm

PR #1099, opened 2024-12-05
 e2e: test: introduce workarounds

ffromani:
formalize workarounds we need to add to unblock CI.
We should strive to keep this list short, but a sad
reality is that we need workarounds from time to time.

### DIFF
--- a/internal/workarounds/prom.go
+++ b/internal/workarounds/prom.go
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2024 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package workarounds
+
+import (
+	"context"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	k8swait "k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	promConfigMapNamespace = "openshift-monitoring"
+	promConfigMapName      = "cluster-monitoring-config"
+	promConfigMapKey       = "config.yaml"
+
+	promConfigKey       = "prometheusK8s"
+	promConfigVolumeKey = "volumeClaimTemplate"
+)
+
+type Prometheus struct {
+	PollInterval time.Duration
+	PollTimeout  time.Duration
+}
+
+func ForPrometheus() *Prometheus {
+	return &Prometheus{
+		PollInterval: 5 * time.Second,
+		PollTimeout:  5 * time.Minute,
+	}
+}
+
+func (prom *Prometheus) IssueReference() string {
+	return "https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1784"
+}
+
+func (prom *Prometheus) Describe() string {
+	return "Prometheus instance  cannot use storage and fails to go running; only one instance out of two will go running, and this later prevents node draining"
+}
+
+func (prom *Prometheus) Apply(ctx context.Context, cli client.Client) error {
+	var cm corev1.ConfigMap
+	key := client.ObjectKey{
+		Namespace: promConfigMapNamespace,
+		Name:      promConfigMapName,
+	}
+	immediate := true
+	return k8swait.PollUntilContextTimeout(ctx, prom.PollInterval, prom.PollTimeout, immediate, func(ctx2 context.Context) (bool, error) {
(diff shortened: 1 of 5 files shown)


ffromani:
counterproof this is still required: https://github.com/openshift-kni/numaresources-operator/pull/1098#issuecomment-2520551032

internal/workarounds/prom.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	immediate := true
+	return k8swait.PollUntilContextTimeout(ctx, prom.PollInterval, prom.PollTimeout, immediate, func(ctx2 context.Context) (bool, error) {
+		err := cli.Get(ctx2, key, &cm)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				klog.InfoS("configmap not found, nothing to do", "namespace", key.Namespace, "name", key.Name)
+				return true, nil
+			}
+			return true, err
```
shajmakh, line 76: do you want the loop to be aborted if you find another error? if not you need to return nil instead
and regardless why to return true? the condition isn't satisfied

internal/workarounds/prom.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		err = UpdateConfigMap(&cm)
+		if err != nil {
+			return true, err
```
shajmakh, line 81: same here, I believe this should be return false, nil and log out the error as warning unless I'm missing the point of retrying

shajmakh (commented):
cool implementation!

internal/workarounds/prom.go
ffromani: yes, this failure is critical and needs to make the process abort

internal/workarounds/prom.go
ffromani: again failure at this stage is critical, to put it differently, the only non-critical failure is on update, because tolerating failure on update is needed for optimistic concurrency

internal/workarounds/prom.go
```diff
@@ -0,0 +1,139 @@
...
+
+func ForPrometheus() *Prometheus {
+	return &Prometheus{
+		PollInterval: 5 * time.Second,
+		PollTimeout:  5 * time.Minute,
+	}
+}
+
+func (prom *Prometheus) IssueReference() string {
+	return "https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1784"
```
Tal-or, line 55: This issue is closed, then what is the indication for us that this workaround is not needed anymore?

internal/workarounds/prom.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (prom *Prometheus) Describe() string {
+	return "Prometheus instance  cannot use storage and fails to go running; only one instance out of two will go running, and this later prevents node draining"
```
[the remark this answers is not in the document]
ffromani: good point. I'll re-upload with an explanation comment. I want to have another full run to demonstrate the W/A is fixing (or at least non-hurting) and then I'll re-upload

Tal-or (approved):
I'll be honest, I think we're going too far with this one.
There are too many questions here:
1. How we're going to decide when an external issue deserves a workaround? every time CI failed? which lanes? 
 
2. When issues like these reoccur (and they will)  are we going to spent hours and days understanding the issue and workaround it? why overriding the issue and testing locally is not enough? 

3. How do we know when issue get solved and we don't need the workaround anymore? what if the workaround will break?

I can understand the rational here, but this is a rabbit hole of maintenance and extra work the IMVHO is too costly.

Tal-or:
/hold
Got comments from other reviewers

ffromani:
> I'll be honest, I think we're going too far with this one. There are too many questions here:

thing is: you are right, full stop, no "if" "but" or "however". I don't like this idea any more than you do (even though, I must say, the impl turned out quite nice!). But we're past time and we have urgent matters to cover. So this is a necessary evil.

>      1. How we're going to decide when an external issue deserves a workaround? every time CI failed? which lanes?

Never. We are out of options at this particular point in time, and I wanted some structure to convey the key details. I maybe went overboard so I'm fine with narrowing down, even though that won't solve the fundamental issue that we should not have any code at all to workaround issues.

>     2. When issues like these reoccur (and they will)  are we going to spent hours and days understanding the issue and workaround it? why overriding the issue and testing locally is not enough?

The first part yes, unfortunately: we will need to painfully spend hours and days to be able to report the core issue to the relevant team (like we are doing here). Workarounding is a necessary evil. Testing locally is not enough: things degrade fast and hard without any form of automated CI, and our CI is already bare bones (serial suite can't run here :\ )

>     3. How do we know when issue get solved and we don't need the workaround anymore? what if the workaround will break?

We will need to keep going the right route and file issues to get the root cause addressed. In parallel we need to unblock ourselves as we are out of time. The workaround can (and will) very much break, and at that point we will be back at square one.

> I can understand the rational here, but this is a rabbit hole of maintenance and extra work that IMVHO is too costly.

While I do agree, the problem is that *just* waiting (vs workarounding and waiting for the fix in parallel) is even costlier.

> EDIT: I approved the changes because I understand we're on the edge here.

Yes, it's a necessary evil. I'm not happier than you are, likely less happy.

internal/workarounds/prom.go
ffromani: the u/s issue is fixed, yet we are hitting the same issue in our CI, and I don't have yet a better bug to track.
Once this fixes and is merged, I'll post a revert to track the real fix.

shajmakh:
The implementation looks good and I understand the raised concerns, I am willing to maintain if needed
/lgtm 
feel free to unhold as you see fit

ffromani:
turns out the actual culprit can be crun 1.18.z (vs previous 1.17) but the investigation is still ongoing and we don't have a repro (yet alone a fix) yet.

I'm taking the responsability to unhold, because we need to move forward with other issues and tasks

ffromani:
https://issues.redhat.com/browse/OCPBUGS-45394

PR #1101, opened 2024-12-09
add static buildinfo information

ffromani:
add static buildinfo rather than determine that at last stage of the build process. This data is expected to be frozen when we build our artifacts, so collect all the infos when we start and add this to our artifacts is actually simpler and safe.
shipping `buildinfo.json` alongside `topics.json` in our test images would also enable better tests introspection, yet to be implemented.

### DIFF
--- a/tools/buildhelper/buildhelper.go
+++ b/tools/buildhelper/buildhelper.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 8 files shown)


tools/buildhelper/buildhelper.go
```diff
@@ -26,39 +27,117 @@ import (
...
+	cmd := exec.Command("git", "log", "-1", "--pretty=format:%h")
+	out, err := cmd.Output()
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(out)), nil
+}
+
+func getBranch() (string, error) {
+	if cm, ok := os.LookupEnv("NRO_BUILD_BRANCH"); ok {
```
shajmakh, line 64: where are these env vars set? for such data I don't see the use case where we need to allow overriding the real fetched values or I'm missing a detail

Dockerfile.openshift.tests
```diff
@@ -4,5 +4,6 @@ COPY run-e2e-nrop-*.sh /usr/local/bin
 COPY numacell /bin
 COPY pause /
 COPY topics.json /usr/local/share
+COPY buildinfo.json /usr/local/share
```
shajmakh, line 7: I believe we should have this in the operator image rather than the tests, because tests image is created always and only from the main branch and will be used to run on all releases while it would be more beneficial to get this data from the operator image itself instead

shajmakh (commented):
Thanks for filing this! 
Added two comments below
/approve

Dockerfile.openshift.tests
ffromani: that's a good point. It's useless in the test image.

ffromani:
so we need https://github.com/openshift/release/pull/59712 first

ffromani:
kindly ping, there is more work waiting for this one

PR #1104, opened 2024-12-10
[release-4.16] deps: bump deployer dep

ffromani:
consume u/s fixes


ffromani:
/hold

for @Tal-or to review

ffromani:
@Tal-or lanes are failing with the exact error this PR should solve. But this IIRC should be caused by the very old bug/misconfiguration which causes the lane run with images built against EXISTING codebase (not the patched codebase changed by the PR under test), right?

Tal-or (approved):

PR #1105, opened 2024-12-10
tree: version: move to 4.19

shajmakh:
Redirect main to 4.19.


ffromani:
it's possible that some images we depend on don;t have a 4.19 stream started yet. This is a chronic pain and likely unsolvable. We may need to keep references to 4.18 streams.

ffromani (commented):
/lgtm

assuming the 4.19 streams all exist

ffromani:
/retest

missed the most important tag, now should be good

PR #1106, opened 2024-12-10
CNF-10142: Enable NROP metrics to be to scraped securely by Prometheus

rbaturov:
This PR encompasses all the required changes to reintegrate NROP metrics with Prometheus.  This implementation uses the metrics server API from the controller-runtime.
The majority of the changes in this PR follow the guidelines outlined in the following guide:
https://rhobs-handbook.netlify.app/products/openshiftmonitoring/collecting_metrics.md/

The difference between this implementation and the guide's is that we use a `bearerTokenFile` for Prometheus authentication instead of tls.crt and tls.key. This approach uses TLS but does not implement mTLS.

A follow-up PR will be issued to ensure we implement this for RTE metrics as well.
Moreover, will issue a PR adding an e2e test to the CI, for this functionality.

2. run: `make deploy` 
4. run:
  ```
curl -v \
  https://numaresources-controller-manager-metrics-service.numaresources.svc:8080/metrics
```

### DIFF
--- a/config/prometheus/kustomization.yaml
+++ b/config/prometheus/kustomization.yaml
@@ -1,2 +1,3 @@
 resources:
+- rbac.yaml
 - monitor.yaml
--- a/config/prometheus/rbac.yaml
+++ b/config/prometheus/rbac.yaml
@@ -1,0 +1,31 @@
+# creates Role and RoleBinding for prometheus-k8s service account to access our namespace
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: prometheus-k8s
+  namespace: system
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - services
+  - endpoints
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: prometheus-k8s
+  namespace: system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s
+subjects:
+- kind: ServiceAccount
+  name: prometheus-k8s
+  namespace: openshift-monitoring
--- a/config/rbac/service.yaml
+++ b/config/rbac/service.yaml
@@ -1,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: metrics-service-cert
+  labels:
+    control-plane: controller-manager
+  name: controller-manager-metrics-service
+spec:
+  ports:
+  - name: https
+    port: 8080
+    protocol: TCP
+    targetPort: https
+  selector:
+    control-plane: controller-manager
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ Assuming you can push container images to a container registry and you are in th
 1. build and upload the manifest bundle container image: `make bundle bundle-build bundle-push`
 1. leverage `operator-sdk` to deploy the container: `operator-sdk run bundle ${REPO}/numaresources-operator-bundle:${VERSION}`. Note the build procedure typically downloads a local copy of `operator-sdk` in `bin/` which you can reuse
 
+Note that installing the operator using this method requires adding the openshift.io/cluster-monitoring: "true" label to the operator namespace to enable Prometheus cluster monitoring.
+
 For further details, please refer to the [operator-sdk documentation](https://sdk.operatorframework.io/docs/olm-integration/tutorial-bundle/)
 
 ## roadmap
--- a/bundle/manifests/numaresources-controller-manager-metrics-service_v1_service.yaml
+++ b/bundle/manifests/numaresources-controller-manager-metrics-service_v1_service.yaml
@@ -1,0 +1,19 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: metrics-service-cert
+  creationTimestamp: null
+  labels:
+    control-plane: controller-manager
+  name: numaresources-controller-manager-metrics-service
+spec:
+  ports:
+  - name: https
+    port: 8080
+    protocol: TCP
+    targetPort: https
+  selector:
+    control-plane: controller-manager
+status:
+  loadBalancer: {}
--- a/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
@@ -62,8 +62,9 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2024-11-14T15:38:47Z"
+    createdAt: "2024-12-18T20:58:55Z"
     olm.skipRange: '>=4.18.0 <4.19.0'
+    operatorframework.io/cluster-monitoring: "true"
     operators.operatorframework.io/builder: operator-sdk-v1.36.1
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
   name: numaresources-operator.v4.19.999-snapshot
@@ -532,6 +533,10 @@ spec:
                   initialDelaySeconds: 15
                   periodSeconds: 20
                 name: manager
+                ports:
+                - containerPort: 8080
+                  name: https
+                  protocol: TCP
                 readinessProbe:
                   httpGet:
                     path: /readyz
@@ -544,6 +549,10 @@ spec:
                     memory: 20Mi
                 securityContext:
                   allowPrivilegeEscalation: false
+                volumeMounts:
+                - mountPath: /certs
+                  name: metrics-tls
+                  readOnly: true
               securityContext:
                 runAsNonRoot: true
               serviceAccountName: numaresources-controller-manager
@@ -553,6 +562,10 @@ spec:
                 key: node-role.kubernetes.io/control-plane
               - effect: NoSchedule
                 key: node-role.kubernetes.io/master
+              volumes:
+              - name: metrics-tls
+                secret:
+                  secretName: metrics-service-cert
       permissions:
       - rules:
         - apiGroups:
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -43,6 +43,10 @@ spec:
                 matchExpressions:
                   - key: node-role.kubernetes.io/control-plane
                     operator: Exists
+      volumes: 
+      - name : metrics-tls
+        secret:
+          secretName: metrics-service-cert
       containers:
       - command:
         - /bin/numaresources-operator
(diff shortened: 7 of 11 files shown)

### CODE REVIEW

**Comment by ffromani** [comment]
/approve

LGTM, but how can we test this?

**Comment by rbaturov** [comment]
> /approve
> 
> LGTM, but how can we test this?

We can test manually using the instructions in the description.
[Automated tests](https://github.com/openshift-kni/numaresources-operator/pull/1037) are here, but we need both this and RTE metrics PR to be merged first

**Comment by ffromani** [comment]
> > /approve
> > LGTM, but how can we test this?
> 
> We can test manually using the instructions in the description. [Automated tests](https://github.com/openshift-kni/numaresources-operator/pull/1037) are here, but we need both this and RTE metrics PR to be merged first

ok, I'm inclined to merge but after the branching (this means we will have to backport later on once the dust settles): https://github.com/openshift-kni/numaresources-operator/pull/1105

File: config/prometheus/rbac.yaml [hunk 1]
```diff
@@ -0,0 +1,31 @@
+# creates Role and RoleBinding for prometheus-k8s service account to access our namespace
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: prometheus-k8s
+  namespace: system
```
ffromani, line 6: does it _have_ to be in this namespace with this name? these are generic names, I'm worried about conflicts with other components

ffromani (commented):
LGTM besides the possible conflicts caused by name clashes, which is admittedly a larger and older issue. Let's not make it even worse though.

config/prometheus/rbac.yaml
rbaturov: This is going to be deployed in the numaresources-operator namespace, as far as I understand.
We use "system" in the namespace field for other manifests too.

config/rbac/service.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: metrics-service-cert
+  labels:
+    control-plane: controller-manager
+  name: controller-manager-metrics-service
```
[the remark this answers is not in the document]
rbaturov: This is also expected to be deployed in the NROP namespace.

config/prometheus/rbac.yaml
```diff
@@ -0,0 +1,31 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s
+subjects:
+- kind: ServiceAccount
+  name: prometheus-k8s
+  namespace: openshift-monitoring
```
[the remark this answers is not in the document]
rbaturov: This is what is instructed in this [article](https://rhobs-handbook.netlify.app/products/openshiftmonitoring/collecting_metrics.md/).
If already exists,  no change will be reflected.

config/rbac/service.yaml
```diff
@@ -0,0 +1,18 @@
...
+spec:
+  ports:
+  - name: https
+    port: 8080
+    protocol: TCP
+    targetPort: 8080
+  selector:
+    control-plane: controller-manager
```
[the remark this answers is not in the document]
rbaturov: Ditto

config/prometheus/rbac.yaml
ffromani: ok, this is silly but still let's change the default to something like `numaresources` (not `system`). Plus, do we have evidence this goes in our namespace? a copy paste from the patched operator running on any cluster is good enough evidence.

config/prometheus/rbac.yaml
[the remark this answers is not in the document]
ffromani: again, likely silly but please let's rename all generic names to be prefixed with `nrop-` or anything. And yes, we have existing and larger issues, but we should start going in the right direction rather than hoping for non-worse. I think the rename should be little effort so let's just do this and we can merge. (same goes for the RTE PR)

config/rbac/service.yaml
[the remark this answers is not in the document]
ffromani: ok, please see above.

config/rbac/service.yaml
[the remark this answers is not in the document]
ffromani: ok, please see above.

ffromani:
ok, let's change generic names to include a reference to NROP so they aren't really generic anymore and we're good to go.
**PLEASE NOTE** alternatively (or, even better, in addition) if we can demonstrate with a screenshot or a copy/paste snippet these manifests _only and always_ go in our namespace, that's good enough as well.

ffromani:
/hold

@rbaturov on second thought let me check this PR I have a cluster handy. Perhaps no changes are needed

ffromani:
/hold cancel

I managed to play with the PR and read carefully enough (but still...) https://rhobs-handbook.netlify.app/products/openshiftmonitoring/collecting_metrics.md/
- we need to make sure to annotate our namespace following https://rhobs-handbook.netlify.app/products/openshiftmonitoring/collecting_metrics.md/#configuring-prometheus-to-scrape-metrics
- let's add this `operatorframework.io/cluster-monitoring=true` annotation to the operator’s CSV
- let's document somewhere (`README.md` is a good starting point) that users NEED to have the namespace annotated if they install differently than the web UI linking back https://rhobs-handbook.netlify.app/products/openshiftmonitoring/collecting_metrics.md/#configuring-prometheus-to-scrape-metrics
- we should NOT install manifests outside our namespace. So prometheus role and rolebinding are a no go unless they demonstrably break the flow, which would require us to know precisely where and when. If CI is broken, say, then we can install these manifests but this step has to be 1. explicit and 2. separate. By default we must assume these role exists already, as AFAICT they do in well formed OCP clusters

config/prometheus/rbac.yaml
rbaturov: I'll add a screenshot later.
But what I'm saying is that "system" has been used across all the repo as a placeholder (look at manager.yaml e.g). This however replaced with "numaresources" after `make deploy`. kustomize replaces it.

config/rbac/service.yaml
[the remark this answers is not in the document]
rbaturov: This is just a label selector ^

rbaturov:
> /hold cancel
> 
> I managed to play with the PR and read carefully enough (but still...) https://rhobs-handbook.netlify.app/products/openshiftmonitoring/collecting_metrics.md/
> 
> * we need to make sure to annotate our namespace following https://rhobs-handbook.netlify.app/products/openshiftmonitoring/collecting_metrics.md/#configuring-prometheus-to-scrape-metrics
> * let's add this `operatorframework.io/cluster-monitoring=true` annotation to the operator’s CSV
> * let's document somewhere (`README.md` is a good starting point) that users NEED to have the namespace annotated if they install differently than the web UI linking back https://rhobs-handbook.netlify.app/products/openshiftmonitoring/collecting_metrics.md/#configuring-prometheus-to-scrape-metrics

 Added the label for the namespace and the annotation for the CSV. Also, mentioned this in the readme 

> * we should NOT install manifests outside our namespace. So prometheus role and rolebinding are a no go unless they demonstrably break the flow, which would require us to know precisely where and when. If CI is broken, say, then we can install these manifests but this step has to be 1. explicit and 2. separate. By default we must assume these role exists already, as AFAICT they do in well formed OCP clusters

The Prometheus role and rolebinding are created in the **operator** namespace.
BTW, they are currently not deployed by default. for this we can unquote [this](https://github.com/openshift-kni/numaresources-operator/blob/main/config/default/kustomization.yaml#L25)

config/prometheus/kustomization.yaml
```diff
@@ -1,2 +1,3 @@
 resources:
+- rbac.yaml
```
ffromani, line 2: sorry I was not clear earlier: this should be gone, and so `config/prometheus/rbac.yaml`.
Now: this suggestion of mine is likely to break something. In that case, we need to know what breaks and when.

If the content of `rbac.yaml` is used just and only when we run `kustomize` in flows like `make deploy`, good, we can uncomment it again (in a separate commit preferably).

I'm insisting on this point because I want to be really sure we do NOT try to add stuff in the non-numaresources namespace in production flows.

For development flows (`make deploy` and the like) this is fine.

All the above is moot (and so we can merge this version) if we are fully confident the manifests are created just and only in the developer flows (and I can't verify this myself now)

config/rbac/service.yaml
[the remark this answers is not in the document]
ffromani: which can be generic as well. We need to verify it only matches our code. Which should be the case, but can't check right now.

config/prometheus/kustomization.yaml
rbaturov: Well I'm pretty sure they are created only in the developer flows, and only when we unquote [this](https://github.com/openshift-kni/numaresources-operator/blob/main/config/default/kustomization.yaml#L25)  (will verify this of course!).
If I have unquoted this as part of this PR and then run `make generate bundle manifests` only then this will be part of the OLM installation.

rbaturov: Anyhow, I need to figure out what will break ib case we delete this

ffromani: ok, this is the last concern, other than that LGTM. Once we are confident (preferably we can demonstrate) this affects only developer flows, we can merge. Otherwise we can undo this specific part and move to another PR.

Note all this discussion can possibly apply also to the RTE PR.

rbaturov: The RTE PR is a bit different. we are not adding this role again or any other manifest except the service for the daemonset.
The role service account rolebinding and all that stuff should be declared applied only once and utilized by both RTE and the manager.
IOW, the RTE PR don't need to address most of this discussion comments (but there are other things maybe)

rbaturov: @ffromani 
removing the _rbac.yaml_ will result in these errors in the Prometheus pods **when ServiceMonitor is applied** and instructs Prometheus pods to poll metrics:

```
\"system:serviceaccount:openshift-monitoring:prometheus-k8s\" cannot list resource \"pods\" in API group \"\" in the namespace \"numaresources\""
ts=2024-12-18T20:55:46.293Z caller=klog.go:118 level=warn component=k8s_client_runtime func=Warningf msg="github.com/prometheus/prometheus/discovery/kubernetes/kubernetes.go:554: failed to list *v1.Endpoints: endpoints is forbidden: User \"system:serviceaccount:openshift-monitoring:prometheus-k8s\" cannot list resource \"endpoints\" in API group \"\" in the namespace \"numaresources\""
ts=2024-12-18T20:55:48.126Z caller=klog.go:118 level=warn component=k8s_client_runtime func=Warningf msg="github.com/prometheus/prometheus/discovery/kubernetes/kubernetes.go:555: failed to list *v1.Service: services is forbidden: User \"system:serviceaccount:openshift-monitoring:prometheus-k8s\" cannot list resource \"services\" in API group \"\" in the namespace \"numaresources\""

```

But again we are not applying this by default, neither on developer and official installation flows.

ffromani: ok, this needs to be investigated further

config/prometheus/rbac.yaml
ffromani: ok, this addresses my concerns, thanks

ffromani:
/lgtm

thanks for demonstrating role and rolebinding are only deployed in our namespaces. This, plus the fact I managed eventually to check everything step by step, addresses all my remaining concerns.

In a later PR we will need to add ` servicemonitor.yaml` example in `doc/examples`. But for now we can merge.

PR #1107, opened 2024-12-10
CNF-11234: Enable RTE metrics to be scraped securely by Prometheus

rbaturov:
This is a follow-up PR for [Enable NROP metrics to be to scraped securely by Prometheus ](https://github.com/openshift-kni/numaresources-operator/pull/1106).

This PR encompasses and integrates all the needed infrastructure to enable secured communication for scraping metrics by Prometheus, now for **RTE**.

A follow-up PR will be posted for e2e tests later.

2. run: `make deploy` 
4. run:
  ```
 curl -v \
  https://numaresources-rte-metrics-service.numaresources.svc:2112/metrics
```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -9,6 +9,7 @@ rules:
   resources:
   - configmaps
   - serviceaccounts
+  - services
   verbs:
   - '*'
 - apiGroups:
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -61,6 +61,7 @@ import (
 	"github.com/openshift-kni/numaresources-operator/pkg/hash"
 	"github.com/openshift-kni/numaresources-operator/pkg/images"
 	"github.com/openshift-kni/numaresources-operator/pkg/loglevel"
+	rtemetricsmanifests "github.com/openshift-kni/numaresources-operator/pkg/metrics/manifests/monitor"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectnames"
 	apistate "github.com/openshift-kni/numaresources-operator/pkg/objectstate/api"
 	rtestate "github.com/openshift-kni/numaresources-operator/pkg/objectstate/rte"
@@ -83,15 +84,16 @@ type poolDaemonSet struct {
 // NUMAResourcesOperatorReconciler reconciles a NUMAResourcesOperator object
 type NUMAResourcesOperatorReconciler struct {
 	client.Client
-	Scheme          *runtime.Scheme
-	Platform        platform.Platform
-	APIManifests    apimanifests.Manifests
-	RTEManifests    rtemanifests.Manifests
-	Namespace       string
-	Images          images.Data
-	ImagePullPolicy corev1.PullPolicy
-	Recorder        record.EventRecorder
-	ForwardMCPConds bool
+	Scheme              *runtime.Scheme
+	Platform            platform.Platform
+	APIManifests        apimanifests.Manifests
+	RTEManifests        rtemanifests.Manifests
+	RTEMetricsManifests rtemetricsmanifests.Manifests
+	Namespace           string
+	Images              images.Data
+	ImagePullPolicy     corev1.PullPolicy
+	Recorder            record.EventRecorder
+	ForwardMCPConds     bool
 }
 
 // TODO: narrow down
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -120,6 +120,7 @@ type NUMAResourcesOperatorReconciler struct {
 //+kubebuilder:rbac:groups=nodetopology.openshift.io,resources=numaresourcesoperators,verbs=*
 //+kubebuilder:rbac:groups=nodetopology.openshift.io,resources=numaresourcesoperators/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=nodetopology.openshift.io,resources=numaresourcesoperators/finalizers,verbs=update
+//+kubebuilder:rbac:groups="",resources=services,verbs=*
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

File: config/rbac/role.yaml [hunk 1]
```diff
@@ -26,6 +26,18 @@ rules:
...
+  resources:
+  - serviceaccounts
+  verbs:
+  - '*'
+- apiGroups:
+  - ""
+  resources:
+  - services
+  verbs:
+  - '*'
```
ffromani: ok, this is a role, hence it is ONLY granted in the NROP namespace, right? Stll this is very broad, we need to narrow down AND document in the commit message why do we need these permissions

controllers/numaresourcesoperator_controller.go
```diff
@@ -581,6 +581,31 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 			return nil, fmt.Errorf("failed to apply (%s) %s/%s: %w", objState.Desired.GetObjectKind().GroupVersionKind(), objState.Desired.GetNamespace(), objState.Desired.GetName(), err)
 		}
 	}
+
+	for _, obj := range r.RTEMetricsManifests.ToObjects() {
+		// Check if the object already exists
+		existingObj := obj.DeepCopyObject().(client.Object)
```
ffromani, line 587: why do we need the cast here?

controllers/numaresourcesoperator_controller_test.go
```diff
@@ -23,6 +23,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	rtemetricsmanifests "github.com/openshift-kni/numaresources-operator/pkg/metrics/manifests/monitor"
```
ffromani, line 26: misplaced import

ffromani (commented):
I'm not sure we need a separate manifest set. I'. mot saying I'm against it but I'd like to see if/how we can merge it with the existing manifest set

controllers/numaresourcesoperator_controller.go
rbaturov: If I'll remove the casting:
I'll get this error when using `r.Client.Get`:
`cannot use existingObj (variable of type runtime.Object) as client.Object value in argument to r.Client.Get: runtime.Object does not implement client.Object (missing method GetAnnotations)`

rbaturov:
> I'm not sure we need a separate manifest set. I'. mot saying I'm against it but I'd like to see if/how we can merge it with the existing manifest set

I agree on this one.
But I coudln't figure out a more appropriate place to put this manifest.

config/rbac/role.yaml
rbaturov: I agree, you are right. we want in only in the NROP namespace.

rbaturov: @ffromani 
I've tried to create a separate role (not clusterRole) and a RoleBinding.
But adding the label 
//+kubebuilder:rbac:groups="",resources=services,verbs=*
followed by `make manifests` appends its in the clusterRole.
Not sure how we can avoide this and append this in the dedicated role.

controllers/numaresourcesoperator_controller.go
ffromani: and why do we use `DeepCopyObject()` and not `DeepCopy()` ?

rbaturov: @ffromani This because the client.Object implements DeepCopyObject() and not  DeepCopy()

rbaturov:
> /hold
> 
> need to check RBAC rules again

Regarding this, it is correct that now I have granted permission to the operator to manage service objects across the entire cluster and not narrowed it down to NROP namespace.
As I said before, I'm not sure how we can create a role not a clusterrole for this matter.

ffromani:
> /retest-required

the first retest it's a freebie, but past that we need to check why tests are failing and only when we are confident it's actually an infra issue we should retest

rbaturov:
> > /retest-required
> 
> the first retest it's a freebie, but past that we need to check why tests are failing and only when we are confident it's actually an infra issue we should retest

I saw there was a timeout. I'm not sure this is the infra indeed, I wanted to give it a try before deep diving into that

ffromani (commented):
/lgtm

we can narrow down permissions later on

controllers/numaresourcesoperator_controller.go
```diff
@@ -581,6 +583,43 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
...
+		if apierrors.IsNotFound(err) {
+			err := controllerutil.SetControllerReference(instance, obj, r.Scheme)
+			if err != nil {
+				return nil, fmt.Errorf("failed to set controller reference to %s %s: %w", obj.GetNamespace(), obj.GetName(), err)
+			}
+			err = r.Client.Create(ctx, obj)
+			if err != nil {
+				return nil, fmt.Errorf("failed to create %s/%s: %w", obj.GetNamespace(), obj.GetName(), err)
+			}
+		} else {
```
Tal-or, line 603: The only change I would suggest here to make the logic more clear is to encapsulate the `else` in:
`UpdateIfNeeded()`  function
The function won't call `Update()` if the objects are actually equals

```
err := r.Client.Get(ctx, client.ObjectKeyFromObject(obj), existingObj)
		if err != nil && !apierrors.IsNotFound(err) {
			return nil, fmt.Errorf("failed to get %s/%s: %w", obj.GetNamespace(), obj.GetName(), err)
		}
		if apierrors.IsNotFound(err) {
			err := controllerutil.SetControllerReference(instance, obj, r.Scheme)
			if err != nil {
				return nil, fmt.Errorf("failed to set controller reference to %s %s: %w", obj.GetNamespace(), obj.GetName(), err)
			}
			err = r.Client.Create(ctx, obj)
			if err != nil {
				return nil, fmt.Errorf("failed to create %s/%s: %w", obj.GetNamespace(), obj.GetName(), err)
			}
		} else {
                           if err := UpdateIfNeeded(ctx, obj, existingObj); err != nil {
                             ....
                           }
}

func UpdateIfNeeded()...
```

Tal-or:
/lgtm
Good enough for now. Thanks!

PR #1115, opened 2024-12-13
e2e: mustgather: add missing deployer initialization

shajmakh:
Without the deployer initialization, deployment.deployer is nil and causes the test to panic. Fix that by calling deploy.NewForPlatform().


test/e2e/must-gather/must_gather_suite_test.go
```diff
@@ -71,6 +71,7 @@ var _ = ginkgo.BeforeSuite(func() {
 		return
 	}
 	ginkgo.By("Setting up the cluster")
+	deployment.Deployer = deploy.NewForPlatform(configuration.Plat)
```
ffromani, line 74: thanks, this is a nice catch. I think I like more a fix like
```
$ git diff
diff --git a/test/e2e/must-gather/must_gather_suite_test.go b/test/e2e/must-gather/must_gather_suite_test.go
index 838cdc42..2da25052 100644
--- a/test/e2e/must-gather/must_gather_suite_test.go
+++ b/test/e2e/must-gather/must_gather_suite_test.go
@@ -62,12 +62,12 @@ var _ = ginkgo.BeforeSuite(func() {
        mustGatherTag = getStringValueFromEnv(envVarMustGatherTag, defaultMustGatherTag)
        ginkgo.By(fmt.Sprintf("Using must-gather image %q tag %q", mustGatherImage, mustGatherTag))
 
-       if _, ok := os.LookupEnv("E2E_NROP_INFRA_SETUP_SKIP"); ok {
-               ginkgo.By("Fetching up cluster data")
+       var err error
+       deployment, err = deploy.GetDeploymentWithSched(context.TODO())
+       gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
-               var err error
-               deployment, err = deploy.GetDeploymentWithSched(context.TODO())
-               gomega.Expect(err).ToNot(gomega.HaveOccurred())
+       if _, ok := os.LookupEnv("E2E_NROP_INFRA_SETUP_SKIP"); ok {
+               ginkgo.Expect(deployment.NroSchedObj).ToNot(gomega.BeNil(), "infra setup skip but Scheduler instance not found")
                return
        }
        ginkgo.By("Setting up the cluster")
```
Do you think the above could work?

shajmakh: nice idea, thanks for raising! I updated GetDeploymentWithSched() a bit to initialize the deployer so even if setup is skipped it won't hurt. let me know if you meant something else please

test/utils/deploy/deploy.go
```diff
@@ -71,7 +71,10 @@ func NewForPlatform(plat platform.Platform) Deployer {
 }
 
 func GetDeploymentWithSched(ctx context.Context) (NroDeploymentWithSched, error) {
-	sd := NroDeploymentWithSched{}
+	sd := NroDeploymentWithSched{
+		Deployer: NewForPlatform(configuration.Plat),
```
ffromani, line 75: this seems correct

test/e2e/must-gather/must_gather_suite_test.go
```diff
@@ -62,12 +62,13 @@ var _ = ginkgo.BeforeSuite(func() {
 	mustGatherTag = getStringValueFromEnv(envVarMustGatherTag, defaultMustGatherTag)
 	ginkgo.By(fmt.Sprintf("Using must-gather image %q tag %q", mustGatherImage, mustGatherTag))
 
+	ginkgo.By("Fetching up cluster data")
```
ffromani, line 65: this seems wrong

test/e2e/must-gather/must_gather_suite_test.go
```diff
@@ -62,12 +62,13 @@ var _ = ginkgo.BeforeSuite(func() {
...
+	var err error
+	// the error might be not nil we'll decide if that's fine or not depending on E2E_NROP_INFRA_SETUP_SKIP
+	deployment, err = deploy.GetDeploymentWithSched(context.TODO())
```
ffromani, line 68: this should never fail anyway so let's check err not occurred

ffromani (commented):
your approach can work! let's polish it up

test/e2e/must-gather/must_gather_suite_test.go
shajmakh: it can fail in case the cluster doesn't have the scheduler and we try to fetch it

test/e2e/must-gather/must_gather_suite_test.go
shajmakh: indeed. I'll make it "collect available data" does that sound better?

test/utils/deploy/deploy.go
shajmakh: we need to set it first thing in case the cluster didn't have the scheduler and fails to get it

test/e2e/must-gather/must_gather_suite_test.go
Tal-or: I think the question is why the sudden change? why are we always running the deployment, even if `E2E_NROP_INFRA_SETUP_SKIP` exists.

ffromani: my take is that we always want to have an instance of the current scheduler object. If the scheduler does already exist (skip setup), then we should just fetch the existing instance and move on; otherwise we need to setup it and then fetch the setupped instance. This is probably not too evident from the code and we likely need a bit larger refactoring.

test/e2e/must-gather/must_gather_suite_test.go
```diff
@@ -62,12 +62,13 @@ var _ = ginkgo.BeforeSuite(func() {
...
 	if _, ok := os.LookupEnv("E2E_NROP_INFRA_SETUP_SKIP"); ok {
-		ginkgo.By("Fetching up cluster data")
-
-		var err error
-		deployment, err = deploy.GetDeploymentWithSched(context.TODO())
-		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		gomega.Expect(err).ToNot(gomega.HaveOccurred(), "infra setup skip but Scheduler instance not found")
```
ffromani, line 70: doesn't make sense to conditionally check the error. If we need to do `GetDeploymentWithSched` it must not fail.

ffromani:
/close

we merged #1120

but kudos to find and fix the issue here. #1120 takes a step further and was merged only because handoff of PRs.

shajmakh:
> /close
> 
> we merged #1120
> 
> but kudos to find and fix the issue here. #1120 takes a step further and was merged only because handoff of PRs.


Thanks for handling this. much appreciated!

PR #1116, opened 2024-12-13
controllers: kubelet: cleanup platform code

ffromani:
move platform-specific code as much as possible at initialization stage vs peppering the code with `if platform == `.
In addition, remove small repetitions.

### DIFF
--- a/controllers/kubeletconfig_controller.go
+++ b/controllers/kubeletconfig_controller.go
@@ -77,7 +77,8 @@ type kubeletConfigHandler struct {
 	ownerObject client.Object
 	mcoKc       *mcov1.KubeletConfig
 	// mcp or nodePool name
-	poolName string
+	poolName   string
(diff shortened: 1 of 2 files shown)


controllers/kubeletconfig_controller.go
```diff
@@ -288,6 +285,12 @@ func (r *KubeletConfigReconciler) makeKCHandlerForPlatform(ctx context.Context,
 			ownerObject: cmKc,
 			mcoKc:       mcoKc,
 			poolName:    nodePoolName,
+			// the owner should be the KubeletConfig object and not the NUMAResourcesOperator CR
+			// this means that when KubeletConfig will get deleted, the ConfigMap gets deleted as well
+			// TODO on HyperShift there's a cross-namespaced owner references that need to be fixed.
```
Tal-or, line 290: We shall remove this line because we added https://github.com/openshift-kni/numaresources-operator/pull/1088

PR #1120, opened 2024-12-16
e2e: must-gather: fix the setup process

ffromani:
In case we are skipping the setup, make sure to fetch the existing scheduler data, failing loudly if that resource is not actually present.

Along the way, cleanup the deploy API and package to make sure context and timeout are properly passed down the call chain.

### DIFF
--- a/test/utils/deploy/deploy.go
+++ b/test/utils/deploy/deploy.go
@@ -43,10 +43,14 @@ import (
 	"github.com/openshift-kni/numaresources-operator/test/utils/objects"
 )
 
+const (
+	NROSchedulerPollingInterval = 10 * time.Second
+)
+
 type Deployer interface {
 	// Deploy deploys NUMAResourcesOperator object create other dependencies
 	// per the platform that implements it
-	Deploy(ctx context.Context) *nropv1.NUMAResourcesOperator
+	Deploy(ctx context.Context, timeout time.Duration) *nropv1.NUMAResourcesOperator
 	// Teardown Teardowns NUMAResourcesOperator object delete other dependencies
 	// per the platform that implements it
 	Teardown(ctx context.Context, timeout time.Duration)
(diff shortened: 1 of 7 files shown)


ffromani:
carrying forward #1115 because holidays

test/utils/deploy/deploy.go
```diff
@@ -133,20 +137,23 @@ func DeployNROScheduler() *nropv1.NUMAResourcesScheduler {
...
+func TeardownNROScheduler(ctx context.Context, nroSched *nropv1.NUMAResourcesScheduler, timeout time.Duration) {
 	GinkgoHelper()
 
-	if nroSched != nil {
-		err := e2eclient.Client.Delete(context.TODO(), nroSched)
-		Expect(err).ToNot(HaveOccurred())
-
-		err = wait.With(e2eclient.Client).Interval(10*time.Second).Timeout(timeout).ForNUMAResourcesSchedulerDeleted(context.TODO(), nroSched)
-		Expect(err).ToNot(HaveOccurred(), "NROScheduler %q failed to be deleted", nroSched.Name)
+	if nroSched == nil {
```
Tal-or, line 147: Is there a use case for that?
this might stems from another problem, maybe at least add a log here?

test/utils/deploy/deploy.go
```diff
@@ -133,20 +137,23 @@ func DeployNROScheduler() *nropv1.NUMAResourcesScheduler {
...
+		// nothing to do!
+		return
 	}
+
+	err := e2eclient.Client.Delete(ctx, nroSched)
```
Tal-or, line 152: `Expect(e2eclient.Client.Delete(ctx, nroSched)).To(Succeed())`

test/utils/deploy/deploy.go
```diff
@@ -133,20 +137,23 @@ func DeployNROScheduler() *nropv1.NUMAResourcesScheduler {
...
+	Expect(err).ToNot(HaveOccurred())
+
+	err = wait.With(e2eclient.Client).Interval(NROSchedulerPollingInterval).Timeout(timeout).ForNUMAResourcesSchedulerDeleted(ctx, nroSched)
```
Tal-or, line 155:
```
Expect( wait.With(e2eclient.Client).Interval(NROSchedulerPollingInterval).Timeout(timeout).ForNUMAResourcesSchedulerDeleted(ctx, nroSched).To(Suceed(), "NROScheduler %q failed to be deleted", nroSched.Name)
```

test/utils/deploy/deploy.go
ffromani: good point, let's add it

test/utils/deploy/deploy.go
ffromani: done!

test/utils/deploy/deploy.go
ffromani: thanks, done

Tal-or (approved):

PR #1122, opened 2024-12-17
buildhelper: try harder to get version

ffromani:
try first to fetch the version from the master location (Makefile) before to deduce from the branch name.
The immediate benefit is we have a much higher chance to always report a version number, because we must have a version number in the master location in the Makefile.

Also: report git raw branch name

### DIFF
--- a/tools/buildhelper/buildhelper.go
+++ b/tools/buildhelper/buildhelper.go
@@ -26,16 +26,32 @@ import (
 
 	"github.com/mdomke/git-semver/version"
 
-	"github.com/openshift-kni/numaresources-operator/internal/api/buildinfo"
+	apibuildinfo "github.com/openshift-kni/numaresources-operator/internal/api/buildinfo"
+	"github.com/openshift-kni/numaresources-operator/internal/buildinfo"
 )
 
 const (
 	develBranchName     = "devel"
 	releaseBranchPrefix = "release-"
+
+	versionFileName = "Makefile"
 )
+
+func parseVersion() string {
+	srcFile := versionFileName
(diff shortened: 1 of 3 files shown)


tools/buildhelper/buildhelper.go
```diff
@@ -32,13 +35,56 @@ import (
 const (
 	develBranchName     = "devel"
 	releaseBranchPrefix = "release-"
+
+	versionFileName  = "Makefile"
+	versionTagPrefix = `VERSION ?= `
```
Tal-or, line 40: A prefix with a space ` ` in the name?

tools/buildhelper/buildhelper.go
```diff
@@ -32,13 +35,56 @@ import (
...
+	// our versioning convention.
+	versionTagSuffix = `.999-snapshot`
 )
 
+// parseVersion find an annotation like
+// `VERSION ?= 4.19.999-snapshot`
+// intentionally swallows any error
+func parseVersion() string {
+	srcFile := versionFileName
+	if fileName, ok := os.LookupEnv("NRO_BUILD_VERSION_FILE"); ok {
```
Tal-or, line 53: can we add a comment what's the purpose of the `NRO_BUILD_VERSION_FILE` ?

tools/buildhelper/buildhelper.go
ffromani: this is admittedly fragile, but I'd like to avoid going down the full makefile parsing route or regexp parsing route just yet

ffromani:
needs https://github.com/openshift-kni/numaresources-operator/pull/1125 before

PR #1124, opened 2024-12-17
selinux: per-nodegroup SCC toggle

ffromani:
with this PR, we enable partial update of SCC in different nodegroups. We can now support a mixed cluster on which a nodegroup uses the legacy custom policy while another nodegroup uses the builtin one, and all the many variations.

This is meant as _extrema ratio_ should the upgrade go wrong, and to enable blue-green cluster updates, which weren't just possible before.

To enable this, we add per-nodegroup annotations. At this point in time we only support a strict limit on them, which can maybe lifted in the future.


Tal-or:
Testing locally, it seems the rte configmap is not generated by KubeletController. 
It might be completely unrelated, I'll look further

Tal-or:
If we're going with the direction of per-nodegroup I suggest to drop the global annotation because it's redundant. 
A version with the global API was not released yet, so we can still remove the API without dealing with backward compatibility.

Tal-or:
Tests locally, looking good:
custom policy is present
```
sh-5.1# ls /etc/selinux/rte.cil 
/etc/selinux/rte.cil
```
machineconfig is present

```
➜  numaresources-operator git:(main) ✗ oc get mc 51-numaresourcesoperator-worker-cnf 
NAME                                  GENERATEDBYCONTROLLER   IGNITIONVERSION   AGE
51-numaresourcesoperator-worker-cnf  
```
pods are running:
```
➜  numaresources-operator git:(main) ✗ oc get pods                                  
NAME                                                READY   STATUS    RESTARTS   AGE
numaresources-controller-manager-6f48967784-q5wtv   1/1     Running   0          116m
numaresourcesoperator-worker-cnf-4nj5r              2/2     Running   0          27m
```

Tal-or:
Deleting the annotation under the `nodeGroup` didn't took affect. 
The `MachineConfig` and is still present

ffromani:
> If we're going with the direction of per-nodegroup I suggest to drop the global annotation because it's redundant. A version with the global API was not released yet, so we can still remove the API without dealing with backward compatibility.

done!

Tal-or:
@ffromani 
did you look into https://github.com/openshift-kni/numaresources-operator/pull/1124#issuecomment-2562835557?

ffromani:
> @ffromani did you look into [#1124 (comment)](https://github.com/openshift-kni/numaresources-operator/pull/1124#issuecomment-2562835557)?

let me add a test to ensure that

ffromani:
/hold
depends on #1149

pkg/validation/validation.go
```diff
@@ -81,6 +81,10 @@ func NodeGroups(nodeGroups []nropv1.NodeGroup, platf platform.Platform) error {
 		return err
 	}
 
+	if err := nodeGroupsAnnotations(nodeGroups); err != nil {
```
Tal-or, line 84: can be simplified to
`return nodeGroupsAnnotations(nodeGroups)`

controllers/numaresourcesoperator_controller_test.go
```diff
@@ -1903,10 +1870,75 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
 				Expect(apierrors.IsNotFound(err)).To(BeTrue(), "MachineConfig %s expected to be deleted; err=%v", mc2Key.Name, err)
 			})
 		})
-
 	})
 })
 
+func checkSELinuxPolicyProcessing(nro *nropv1.NUMAResourcesOperator, mcp1, mcp2 *machineconfigv1.MachineConfigPool) *NUMAResourcesOperatorReconciler {
```
Tal-or, line 1958: let's pass context from caller

Tal-or (commented):
some minor comments from my side, LGTM otherwise

api/numaresourcesoperator/v1/numaresourcesoperator_types.go
```diff
@@ -111,6 +111,10 @@ type NodeGroupConfig struct {
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
 }
 
+const (
+	NodeGroupMaxAnnotations = 8
```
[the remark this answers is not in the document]
ffromani: yes, good point. Adding

controllers/numaresourcesoperator_controller_test.go
ffromani: good point, done

pkg/validation/validation.go
ffromani: rearranged the function to simplify further

Tal-or (approved):

ffromani:
backport to 4.18 requires https://github.com/openshift-kni/numaresources-operator/pull/1152

PR #1133, opened 2025-01-01
Narrow RBAC Service Permissions

rbaturov:
This PR introduces the following changes:
* Removed Service RBAC permissions from ClusterRole.
* Added a Role and RoleBinding for the manager pod, allowing the operator to manage Services exclusively within the NROP namespace.
* Renamed the manager cluster role and cluster role binding to a more appropriate names.
* Restricted caching to the NROP namespace, ensuring the informer can successfully watch and list Service objects.


bundle/manifests/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -600,6 +599,12 @@ spec:
           verbs:
           - create
           - patch
+        - apiGroups:
```
ffromani, line 602: ok, we move from cluster to namespace. This explains both changes here.

ffromani (commented):
/approve

yes, we should do this

controllers/numaresourcesoperator_controller.go
```diff
@@ -122,7 +122,6 @@ type NUMAResourcesOperatorReconciler struct {
 //+kubebuilder:rbac:groups=nodetopology.openshift.io,resources=numaresourcesoperators,verbs=*
 //+kubebuilder:rbac:groups=nodetopology.openshift.io,resources=numaresourcesoperators/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=nodetopology.openshift.io,resources=numaresourcesoperators/finalizers,verbs=update
-//+kubebuilder:rbac:groups="",resources=services,verbs=*
```
Tal-or, line 125: Don't we need to add services here under namespace scoped?

Tal-or (commented):
Good work, left a minor comment.

Next resubmit please try to split every logical group of changes into a separate commit to make the review easier. 

A good separation could be having a commit for every bullet in the PR description.

controllers/numaresourcesoperator_controller.go
rbaturov: Don't think so

rbaturov:
> Good work, left a minor comment.
> 
> Next resubmit please try to split every logical group of changes into a separate commit to make the review easier.
> 
> A good separation could be having a commit for every bullet in the PR description.

Done

controllers/numaresourcesoperator_controller.go
Tal-or: Why? it should generate the rbac permissions based on these kubebuilder annotations

rbaturov: The kubebuilder annotations are only supported for the operator cluster role. That's what I was doing previously

https://book.kubebuilder.io/reference/markers/rbac

main.go
```diff
@@ -235,7 +235,11 @@ func main() {
 	klog.InfoS("metrics server", "enabled", params.enableMetrics, "addr", params.metricsAddr)
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
-		Cache:  cache.Options{}, // TODO: restrict namespace here?
+		Cache: cache.Options{
+			DefaultNamespaces: map[string]cache.Config{
+				namespace: {},
```
Tal-or, line 240: Why leave empty? set the correct namespace like we're doing in NTO: 
https://github.com/openshift/cluster-node-tuning-operator/blob/57de920ca72f1e06cb78b8df3baeace1f4e64657/cmd/cluster-node-tuning-operator/main.go#L136

controllers/numaresourcesoperator_controller.go
Tal-or: weird, it worked for me:
`//+kubebuilder:rbac:groups="",resources=services,verbs=*,namespace="numaresources"`

main.go
rbaturov: This is not empty, the _namespace_ is the namespace string, like in NTO. The difference is that in NTO we added another one: _metav1.namespaceNone_ which I'm not sure if needed here

Tal-or: We need the operator namespace and the cluster scoped namespace

Tal-or: `metav1.NamespaceNone` is the empty namespace `""` which is refer for objects in cluster-scope

rbaturov: @Tal-or Understood why we need this

controllers/numaresourcesoperator_controller.go
rbaturov: @Tal-or Well it works for me as well. Thanks, Talor. Not sure what went wrong when I tried.

rbaturov:
> Good work, left a minor comment.
> 
> Next resubmit please try to split every logical group of changes into a separate commit to make the review easier.
> 
> A good separation could be having a commit for every bullet in the PR description.

@Tal-or @ffromani Thanks for the review.

controllers/numaresourcesoperator_controller.go
```diff
@@ -122,7 +122,7 @@ type NUMAResourcesOperatorReconciler struct {
...
+//+kubebuilder:rbac:groups="",resources=services,verbs=*,namespace="numaresources"
```
ffromani, line 125: please move this very line on line 104 above, replacing "TODO"

ffromani: **for future work in future PRs** we likely can narrow down futher: securitycontextconstraints, daemonsets, roles, rolebindings, serviceaccounts, pods

main.go
```diff
@@ -235,7 +236,12 @@ func main() {
 	klog.InfoS("metrics server", "enabled", params.enableMetrics, "addr", params.metricsAddr)
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
-		Cache:  cache.Options{}, // TODO: restrict namespace here?
+		Cache: cache.Options{
+			DefaultNamespaces: map[string]cache.Config{
+				namespace:            {},
```
Tal-or, line 241: We are still missing the Numaresources-operator namespace

rbaturov: Why? _namespace_ equals the NROP namepsace
This is done exactly like in NTO

Tal-or: Sorry my bad the API has changed and I was looking at an old revision of NTO

Tal-or (approved):
Thanks. Looking good

ffromani:
the caching commit is already backported as part as https://github.com/openshift-kni/numaresources-operator/pull/1140

PR #1143, opened 2025-01-07
makefile: add go files as a prerequisite

Tal-or:
Without declaring the go files as a prerequisite
the `make` won't complie the tool's binary with the latest changes. IOW the fact that the binary exists is enough and it won't rebuild it again in case of local changes nor new changes pull from u/s

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -420,29 +420,29 @@ pkg/version/_buildinfo.json: bin/buildhelper
 
 update-buildinfo: pkg/version/_buildinfo.json
 
-bin/buildhelper:
-	@go build -o bin/buildhelper tools/buildhelper/buildhelper.go
+bin/buildhelper: tools/buildhelper/buildhelper.go
+	@go build -o $@ $<
 
-bin/envsubst:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -420,27 +420,27 @@ pkg/version/_buildinfo.json: bin/buildhelper
 
 update-buildinfo: pkg/version/_buildinfo.json
 
-bin/buildhelper:
+bin/buildhelper: tools/buildhelper/buildhelper.go
```
ffromani, line 423: ok, this is more correct, but let's repeat ourselves just once then:
```
bin/buildhelper: tools/buildhelper/buildhelper.go
	@go build -o $@ $<
```
https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html

Tal-or: Nice. TIL

ffromani:
LGTM, but missing `Signed-off-by` line

Tal-or:
> LGTM, but missing `Signed-off-by` line

Yea emitted mistakenly during re-submission

ffromani (approved):
/lgtm

thanks!

PR #1144, opened 2025-01-07
manifests: drop configmap support

ffromani:
controller-runtime supports ComponentConfig (see:
https://github.com/kubernetes-sigs/controller-runtime/blob/main/designs/component-config.md) which is an interesting way to feed configuration data into projects using controller-runtime. Thing is: we don't use, nor plan to do, this feature, so let's remove unnecessary noise in our tree and manifests we carry on for no reason.

+++

In hindsight, this could have been interesting to solve the RTE configuration problem, but upstream re-implemented and solved that problem on its own already anyway.

More than that, to date the bottleneck was not using configuration files per se, but untangling the existing option spaghetti and do a orderly transition to config files.

So, somehow the silver lining is the problem would probably have looked the same even if we consumed a `ComponentConfig` (assuming is a good fit for the RTE specific problem, which is not proven yet).

A better approach for the RTE problem would have been to stop using commandline past the first 3 flags (or any other tiny amount), but that ship long sailed (and it's a non trivial take anyway: config files make sense when config proper is complex enough, but stopping when reaching the 3 flags is not complex by any relevant metric, so it's still a different bet on the future).


rbaturov:
@ffromani I'd like to understand  "the RTE configuration problem". I'm missing some context here. we can discuss this offline aswell

PR #1145, opened 2025-01-08
[release-4.18] nrop: controller: get rtestate from cluster just once

ffromani:
We used to get the current cluster state both when attempting to sync the machine config state in the OpenShift flow, and when attempting to sync the RTE daemonsets in both flows.

Due to the way `rtestate.FromClient` works, we end up reading the full cluster state twice in the openshift flow, which is unnecessary and wasteful.

The fix is to get the cluster state once in the higher level function and pass the state in the specific reconcile step functions. The machine config and daemonset functions touch not overlapping objects (bar bugs) so sharing the state between them is expected to be fine.

In other words, we happen to fetch the state twice most likely because of oversight, not because a deliberate decision about data sharing/unsharing (which, should that be the case, would need some extra refactoring and more explicit code).

(cherry picked from commit 835a9a8e1acaaa29cf28dd54e354f4eae4422753)

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -240,10 +240,10 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResourceAPI(ctx context.Conte
 	return intreconcile.StepSuccess()
 }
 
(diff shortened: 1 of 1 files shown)


swatisehgal (commented):
/lgtm

PR #1140, opened 2025-01-07
[release-4.18] Enable NROP metrics to be to scraped securely by Prometheus

rbaturov:
This PR is a combined backport of #1106 and #1133.

This PR encompasses all the required changes to reintegrate NROP metrics with Prometheus.  This implementation uses the metrics server API from the controller-runtime.
The majority of the changes in this PR follow the guidelines outlined in the following guide:
https://rhobs-handbook.netlify.app/products/openshiftmonitoring/collecting_metrics.md/

The difference between this implementation and the guide's is that we use a `bearerTokenFile` for Prometheus authentication instead of tls.crt and tls.key. This approach uses TLS but does not implement mTLS.

A follow-up PR will be issued to ensure we implement this for RTE metrics as well.

To validate that this PR is functioning correctly, please follow these steps:
1. build image of the operator (`make docker-build docker-push`)
2. run: `make deploy` 
3. Attach to one of the prometheus pods `oc exec -it prometheus-k8s-0 -n openshift-monitoring /bin/bash`
4. run:
  ```
curl -v \
  --cacert /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt \
  -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
  https://numaresources-controller-manager-metrics-service.numaresources.svc:8080/metrics
```

PR #1147, opened 2025-01-09
Remove cluster role service RBAC permission leftover

rbaturov:
This PR removes the Service-related RBAC permissions previously granted in the ClusterRole.
This should have been removed as part of the #1140 backport, but it seems to have been unintentionally left behind, likely due to rebase issues.
@ffromani @Tal-or

### DIFF
--- a/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
@@ -62,7 +62,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2025-01-07T08:34:09Z"
+    createdAt: "2025-01-09T10:45:35Z"
     olm.skipRange: '>=4.17.0 <4.18.0'
     operatorframework.io/cluster-monitoring: "true"
     operators.operatorframework.io/builder: operator-sdk-v1.36.1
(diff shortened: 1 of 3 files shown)


bundle/manifests/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -62,7 +62,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2025-01-07T08:34:09Z"
+    createdAt: "2025-01-09T10:45:35Z"
```
ffromani, line 65: we need to get rid of this ugly wasteful timestamp

ffromani (commented):
/approve
/lgtm

please backport as needed

bundle/manifests/numaresources-operator.clusterserviceversion.yaml
rbaturov: I could have chosen not to commit this part. I wasn't sure if that was ok since I'm updating this yaml

ffromani: every PR does that, yours is not different. We just need to stop updating the timestamp somehow.

rbaturov: Yes I know. All I'm saying is that we (devs) can choose not to commit that.

PR #1149, opened 2025-01-10
controller: introduce custom matcher

ffromani:
use custom matcher to check reconciliation result. This enable both more concise and more expressive (higher level) code.


ffromani:
prerequisite for #1124

Tal-or (approved):

PR #1155, opened 2025-01-16
CNF-16537:Adjust serial suite for running on HyperShift platform [tier0]

Tal-or:
This PR contains the basic changes that are needed for adjusting the serial e2e suite for running on HyperShift platform.

The general idea is to reduce the usage in `MachineConfigPool` objects and move toward the `PoolName` API which is a common concept for both HyperShift and OpenShift.

NOTE: The changes here were tested only against `tier0` tests.

More details are captured in each commit.

### DIFF
--- a/nrovalidate/validator/kubeletconfig.go
+++ b/nrovalidate/validator/kubeletconfig.go
@@ -20,11 +20,16 @@ import (
 	"context"
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
 	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
+	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform/detect"
 	deployervalidator "github.com/k8stopologyawareschedwg/deployer/pkg/validator"
+	hypershiftconsts "github.com/openshift-kni/numaresources-operator/internal/hypershift/consts"
 	"github.com/openshift-kni/numaresources-operator/pkg/kubeletconfig"
 	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 )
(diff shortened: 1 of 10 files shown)


test/utils/clients/clients.go
```diff
@@ -86,6 +86,7 @@ func init() {
 	if hypershift.IsHypershiftCluster() {
 		MNGClient, err = hypershift.BuildControlPlaneClient()
 		if err != nil {
+			klog.Info("Failed to initialize HyperShift management cluster client, check the HYPERSHIFT_MANAGEMENT_CLUSTER_KUBECONFIG env variable", err.Error())
```
ffromani, line 89: `klog.Error(S)`?

pkg/kubeletconfig/kubeletconfig.go
```diff
@@ -46,3 +49,12 @@ func KubeletConfToMCKubeletConf(kcObj *kubeletconfigv1beta1.KubeletConfiguration
 	kcAsMc.Spec.KubeletConfig.Raw = rawKc
 	return err
 }
+
+func DecodeFromData(data []byte, scheme *runtime.Scheme, mcoKc *mcov1.KubeletConfig) error {
```
ffromani, line 53: nit: best would be to fix the API so it would look like
```
func DecodeFromData(data []byte, scheme *runtime.Scheme) (*mcov1.KubeletConfig, error) {
}
```

pkg/nodegroups/nodegroups.go
```diff
@@ -0,0 +1,105 @@
...
+	"k8s.io/klog/v2"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
+
+	nropv1 "github.com/openshift-kni/numaresources-operator/api/v1"
+	"github.com/openshift-kni/numaresources-operator/internal/hypershift/consts"
+	"github.com/openshift-kni/numaresources-operator/internal/machineconfigpools"
+	"github.com/openshift-kni/numaresources-operator/test/utils/configuration"
```
ffromani, line 33: a general package (`pkg/...`) must not depend on anything imported from `internal` and surely not from `test/...`

nrovalidate/validator/kubeletconfig.go
```diff
@@ -20,20 +20,78 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
 	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
 	deployervalidator "github.com/k8stopologyawareschedwg/deployer/pkg/validator"
+	hypershiftconsts "github.com/openshift-kni/numaresources-operator/internal/hypershift/consts"
 	"github.com/openshift-kni/numaresources-operator/pkg/kubeletconfig"
+	"github.com/openshift-kni/numaresources-operator/test/utils/configuration"
```
ffromani, line 33: again, we must not depend on `test/...` outside the testing tree (if we did already before this PR, that's a huge bug we need to fix ASAP)

ffromani (commented):
most of the changes LGTM but we need to review how and where we move code around. Didn't review the last commit yet

ffromani:
the general rule is:

1. `test/utils/...`: code which used ONLY by tests. Try hard to resist to add code here (`utils` is an anti-pattern: https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common) use `internal` or `pkg`. Code in `test/utils/...` can depend on any package
2. `internal/...`: code which is used ONLY by other code in the repo. Code in `internal/...` can depend on any package
3. `pkg/...`: code which we are confident or we plain recommend to use from external client. We provide soft-stability guarantee: we try our hardest to NOT break backward compat, but we can't really promise and we _rarely_ break if we have strong reasons to (#1160 ). Code in `pkg/...` can depend **only** on other public-facing code, like 3rd party imports or `api/...`.

Tal-or:
> nrovalidate/validator/kubeletconfig.go

Thanks, and what about code which is not included in any of the above?
like `[nrovalidate/validator/kubeletconfig.go` for example

ffromani:
> > nrovalidate/validator/kubeletconfig.go
> 
> Thanks, and what about code which is not included in any of the above? like `nrovalidate/validator/kubeletconfig.go` for example

that's top-level code, so it can use `internal/...` and `pkg/...` but not `test/utils/...`. Nothing outside `tests/...` can import `test/utils/...`

Tal-or:
@ffromani Done.

I must admit it's hard keep tracking the all the above rule, maybe we should add some CI lane to force it.
Anyway this is an enhancement proposal for another PR and another day

ffromani (commented):
/approve
/lgtm

thanks!

PR #1156, opened 2025-01-20
add missing unit tests

shajmakh:
check the commits please


internal/kloglevel/kloglevel.go
```diff
@@ -65,9 +65,10 @@ func getKLogLevel() *klog.Level {
 	// Second, if the '-v' was not set but is still present in flags defined for the command, attempt to acquire it
 	// by visiting all flags.
 	flag.VisitAll(func(f *flag.Flag) {
-		if level != nil {
+		if level != nil || f.Name != "v" {
```
ffromani, line 68: this should be non-critical, but seems logically correct

ffromani (commented):
/approve
/lgtm

thanks!

internal/kloglevel/kloglevel.go
shajmakh: tru it is only filtering earlier

PR #1157, opened 2025-01-20
konflux: add a dockerfile to must gather

rbaturov:
This Dockerfile will be used by konflux.

### DIFF
--- a/.konflux/Dockerfile.must-gather
+++ b/.konflux/Dockerfile.must-gather
@@ -1,0 +1,33 @@
+FROM brew.registry.redhat.io/rh-osbs/openshift-ose-must-gather-rhel9:v4.19 as mgbuilder
+
+COPY . .
+
+# Save original gather script
+RUN mv /usr/bin/gather /usr/bin/gather_original
+
+RUN mkdir -p /usr/libexec/must-gather/numaresources-operator && \
+    cp /must-gather/collection-scripts/* /usr/libexec/must-gather/numaresources-operator/
+
+FROM registry.redhat.io/rhel9-4-els/rhel-minimal:9.4
+
+RUN microdnf install -y procps-ng tar rsync ; microdnf clean all
+
+# Copy must-gather required binaries
+COPY --from=mgbuilder /usr/bin/openshift-must-gather /usr/bin/openshift-must-gather
+COPY --from=mgbuilder /usr/bin/oc /usr/bin/oc
+
+COPY --from=mgbuilder /usr/libexec/must-gather/numaresources-operator/* /usr/bin/
+
+ENTRYPOINT ["/usr/bin/gather"]
+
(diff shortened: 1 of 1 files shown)


rbaturov:
@shajmakh @ffromani 
We need to decide if we want to place this in the root dir, or create a `.konflux ` dir to hold the dockerfiles used by konflux.

shajmakh:
/hold 
for @ffromani 's review

Dockerfile.must-gather.konflux
```diff
@@ -0,0 +1,33 @@
+FROM brew.registry.redhat.io/rh-osbs/openshift-ose-must-gather-rhel9:v4.19.0 as mgbuilder
```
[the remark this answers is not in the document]
rbaturov: Will update this to use 4.19

Dockerfile.must-gather.konflux
```diff
@@ -0,0 +1,33 @@
+FROM brew.registry.redhat.io/rh-osbs/openshift-ose-must-gather-rhel9:v4.19 as mgbuilder
+
+COPY / /
```
shajmakh, line 3: do we need the whole / directory? this can backfire and slow done the build if the content there is heavy, I believe `COPY . .` would be enough

ffromani: +1

Dockerfile.must-gather.konflux
```diff
@@ -0,0 +1,33 @@
...
+
+# Save original gather script
+RUN mv /usr/bin/gather /usr/bin/gather_original
+
+RUN mkdir -p /usr/libexec/must-gather/numaresources-operator && \
+    cp /must-gather/collection-scripts/* /usr/libexec/must-gather/numaresources-operator/
+
+FROM registry.redhat.io/rhel9-4-els/rhel-minimal:9.4
+
+RUN microdnf install -y procps-ng tar rsync ; microdnf clean all
```
ffromani, line 13: what's the reason to use `;` to join commands?

ffromani (commented):
can we move the konflux docker files in (any) konflux-specific directory?

Dockerfile.must-gather.konflux
rbaturov: Thanks @shajmakh . My intention was to copy the current dir, this is a mistake

Dockerfile.must-gather.konflux
rbaturov: I don't see a particular reason. This is how it was done in m/s

ffromani: I like `&&` more but in this case I don't have a strong argument to change

rbaturov:
> thanks for the updates! can we move the konflux docker files in (any) konflux-specific directory?

Done

ffromani:
/hold cancel

deferring the final lgtm to @shajmakh

PR #1158, opened 2025-01-20
internal: remove workaround pkg

shajmakh:
Now that the bug is resovled, remove its workaround reverting https://github.com/openshift-kni/numaresources-operator/commit/51bf49305ce97e9ecc1d3684af551619c50e1fdd (except Serial decorator)


ffromani:
/approve
/lgtm

thanks!

PR #1159, opened 2025-01-22
internal: add unit tests for dangling

shajmakh:
Add missing unit tests for dangling pkg with fake client.

### DIFF
--- a/internal/dangling/dangling_test.go
+++ b/internal/dangling/dangling_test.go
@@ -1,0 +1,232 @@
+/*
+ * Copyright 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dangling
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	nropv1 "github.com/openshift-kni/numaresources-operator/api/v1"
+	nodegroupv1 "github.com/openshift-kni/numaresources-operator/api/v1/helper/nodegroup"
+	"github.com/openshift-kni/numaresources-operator/pkg/objectnames"
+	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+)
+
+const (
+	// the nro constants below are used to point to the same nro test object across all
+	// the functions which is only legal to have a single object in a k8s cluster
+	nrouid  = "6e662cb5-9264-494d-bed8-8ddb817317cc"
+	nroName = "nro"
+	nroNs   = "nro-ns"
+)
+
+func TestDeleteUnusedMachineConfigs(t *testing.T) {
+	targetMCPName := "mcp1" // var because we need its pointer
+
+	nro := nropv1.NUMAResourcesOperator{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      nroName,
+			Namespace: nroNs,
+			UID:       nrouid,
+		},
+	}
+
+	trees := []nodegroupv1.Tree{
+		{
+			NodeGroup: &nropv1.NodeGroup{
+				PoolName: &targetMCPName,
+			},
+			MachineConfigPools: []*mcov1.MachineConfigPool{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: targetMCPName,
+					},
+				},
+			},
+		},
+	}
+
+	mcs := []mcov1.MachineConfig{
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: objectnames.GetMachineConfigName(nroName, targetMCPName),
+				OwnerReferences: []metav1.OwnerReference{
+					{
+						UID: nrouid,
+					},
+				},
+			},
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: objectnames.GetMachineConfigName(nroName, "mcp2"),
+				OwnerReferences: []metav1.OwnerReference{
+					{
+						UID: nrouid, // owned by NRO object but dangling
+					},
+				},
+			},
+		},
+		{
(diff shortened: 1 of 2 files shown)


shajmakh:
/hold
for additional tests covering the DS deletion

internal/dangling/dangling_test.go
```diff
@@ -0,0 +1,208 @@
...
+		return
+	}
+
+	err = isMachineConfigFound(fakeClient, context.TODO(), mcs[0], mcs[2])
+	if err != nil {
+		t.Errorf("failed to find owned non-dangling MachineConfigs: %v", err)
+	}
+	err = isMachineConfigFound(fakeClient, context.TODO(), mcs[1])
+	if err == nil {
+		t.Errorf("succeeded to get dangling mc1 from fake client while expected to be deleted")
```
swatisehgal, line 124: Sorry for being pedantic but the expected flow here is to get an error so I wouldn't use the work "succeeded" to convey unexpected behavior. Also, mc1 in the error message is not the correct name right? 

We should either make it generic `Found dangling machine config from fake client which was expected to be deleted`. I think that a better way would be  mention its name  t.Errorf("Found dangling machine config: %s from fake client which was expected to be deleted", mcs[1].Name)

swatisehgal (commented):
Thanks for this work, really appreciated.

Overall LGTM, just some minor NITs

internal/dangling/dangling_test.go
shajmakh: Thanks for your comment. I agree, updating right away.

internal/dangling/dangling_test.go
```diff
@@ -0,0 +1,208 @@
...
+					},
+				},
+			},
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: fmt.Sprintf("51-%s-mcp2", nroName),
+				OwnerReferences: []metav1.OwnerReference{
+					{
+						UID: nrouid, // owned but dangling
```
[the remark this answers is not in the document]
shajmakh: Thanks for this

swatisehgal (commented):
/lgtm

Thanks for addressing the review comments.

/hold
in case @ffromani has some comments

internal/dangling/dangling_test.go
```diff
@@ -0,0 +1,208 @@
...
+	nropv1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1"
+	nodegroupv1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1/helper/nodegroup"
+	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+)
+
+const (
+	nrouid  = "123"
+	ns      = "nro-ns"
+	nroName = "nro"
+)
```
ffromani, line 39: these are good to be factored out only when we do want some coupling between the users (users being piece of code which need these constants). We should be careful to factor out these values because of this.
It's fine to have a global constant meaning "I don't care about the specific value, I just need a legal value", though.
I think we can just add comments here to represent that these are (should be) generic legal values to be used when we don't care about the value itself as long as it legal

internal/dangling/dangling_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+var (
+	targetMCPName = "mcp1" // var because we need its pointer
```
ffromani, line 42: why can't it be local?

internal/dangling/dangling_test.go
```diff
@@ -0,0 +1,208 @@
...
+	ctx           = context.TODO()
```
ffromani, line 43: I was recently reminded (or made aware?) it's almost always wrong to reuse contexts across different call trees.
It's ok to pack context in structs as long as these contexts don't outlive the call tree, which is not the case here.
So we should actually inline the context on its users.

internal/dangling/dangling_test.go
```diff
@@ -0,0 +1,208 @@
...
+			},
+		},
+	}
+)
+
+func TestDeleteUnusedMachineConfigs(t *testing.T) {
+	mcs := []mcov1.MachineConfig{
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: fmt.Sprintf("51-%s-%s", nroName, targetMCPName),
```
ffromani, line 73: let's use the helper from `objectname`

internal/dangling/dangling_test.go
```diff
@@ -0,0 +1,208 @@
...
+					{
+						UID: "100", // not owned by NRO object
+					},
+				},
+			},
+		},
+	}
+
+	_ = mcov1.AddToScheme(scheme.Scheme)
+	fakeClient := fake.NewClientBuilder().WithScheme(scheme.Scheme).WithObjects(&mcs[0], &mcs[1], &mcs[2]).Build()
```
ffromani, line 104: if we turn `mcs` in a slice of pointers, we can just write `WithObjects(mcs...)`

ffromani (commented):
some minorish comments inside, overall looks good

internal/dangling/dangling_test.go
ffromani: please check if there are other instances we can change

internal/dangling/dangling_test.go
shajmakh: Thanks for sharing, fixed

internal/dangling/dangling_test.go
shajmakh: I should have chosen a better value to represent a true situation even though this doesn't matter for the test. I updated that and added a comment. Thanks for highlighting.

internal/dangling/dangling_test.go
shajmakh: we need it in the shared trees object which is also local:
https://github.com/openshift-kni/numaresources-operator/pull/1159/files/7dc9836f111e8f9b9affd75d66fb3e5aec705013#diff-62cf9676d59b182225fe67db37903c5d08981a360186170552ab10f6b0b7dd48R56

ffromani: this seems a great reason to have local variables instead of one shared global

shajmakh: I get your point and it is valid. we are not updating this variable's value but we still need it as a variable because we want its pointer. I'll change that to be local var instead to avoid confusion or misuse if it happens.

internal/dangling/dangling_test.go
shajmakh: that doesn't work actually with implemented interface of type client.Object

ffromani (commented):
/unhold
/lgtm

internal/dangling/dangling_test.go
```diff
@@ -0,0 +1,232 @@
...
+	mcov1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+)
+
+const (
+	// the nro constants below are used to point to the same nro test object across all
+	// the functions which is only legal to have a single object in a k8s cluster
+	nrouid  = "6e662cb5-9264-494d-bed8-8ddb817317cc"
+	nroName = "nro"
+	nroNs   = "nro-ns"
+)
```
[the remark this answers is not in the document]
shajmakh: it is allowed to have a single NROP object on a real cluster, and these are consts so even if shared within tests there is no way they'll be modified. I don't mind and it's no big deal to make them local but I'd like to understand your point of view more please. what can go wrong with this or are you referring to this from style perspective?

PR #1160, opened 2025-01-23
project: kubebuilder layout v3 -> v4

ffromani:
We have no good reason to get stuck on layout v3 and  make our time harder when consuming upcoming changes and fixes.
    
Most of the changes here stem from the fact we change the path of the API package.
We also consume as much as possible from the new Makefile features.

Done using https://book.kubebuilder.io/migration/migration_guide_gov3_to_gov4


ffromani:
expected to fail in multiple lanes

ffromani:
should not be horribly broken anymore

Tal-or (approved):
Thanks

PR #1161, opened 2025-01-23
ghactions: use `kubectl` instead of `oc`

Tal-or:
oc binary seems to missing under the runner:
https://github.com/openshift-kni/numaresources-operator/actions/runs/12930090387/job/36060828252\?pr\=1155

It probably removed in one of the vm's distro update. Let's use kubectl instead.


ffromani (commented):
/approve
/lgtm

This is a RH-ism we introduced by mistake. We should always have used `kubectl`

PR #1162, opened 2025-01-24
ci: numacell: fix dockerfile, add release flow

ffromani:
Fix the dockerfile to be self-contained (use builder step) and add gh action release flow to make updating the CI image easier.


Tal-or (approved):

swatisehgal (commented):
/lgtm

PR #1164, opened 2025-01-26
e2e: check all nodes have NRT object

Tal-or:
The serial suite has an hidden assumption that all nodes present in the cluster are associated with an NRT object.

Because of u/s compatability pods are allowed to be scheduled on nodes that doesn't have NRT object, which is not desired in the test context.

Hence, we should validate if the assumption is correct before the test begin and not relaying on the user's good intentions.


ffromani (commented):
/approve
/lgtm

PR #1165, opened 2025-01-26
serial:config: validate smt alignment is off

Tal-or:
Some of the test pods are asking for an odd number of CPUs.
if SMT alignment enabled the test will fail in the middle of the
run and it will requires extra time and effort to understand the
root cause.

Lets avoid that and fail loudly in the beginning of the test
explaining that SMT alignment should be disabled.

The serial suite has a unique characteristics and it should
be running on a dedicated environment,
so although smt alignment should usually be enable
for the test purposes it make a complete sense to require
its disabelement.

### DIFF
--- a/test/e2e/serial/config/config.go
+++ b/test/e2e/serial/config/config.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
 	v1 "k8s.io/api/core/v1"
@@ -168,6 +169,13 @@ func CheckNodesTopology(ctx context.Context) error {
 		return fmt.Errorf("an NRT object must be associated with each node: %v", err)
 	}
 
+	errorMap = validateSMTAlignmentIsOFF(kconfigs)
+	if len(errorMap) != 0 {
+		errText := errorMapToString(errorMap)
+		klog.Infof("SMT alignment is ON in some of the nodes and should be disabled: %v", errText)
+		return fmt.Errorf("Following nodes have SMT alignment enabled:\n%#v\n", errText)
+	}
+
 	singleNUMANodeNRTs := e2enrt.FilterByTopologyManagerPolicy(nrtList.Items, intnrt.SingleNUMANode)
 	if len(singleNUMANodeNRTs) < minNumberOfNodesWithSameTopology {
 		return fmt.Errorf("Not enough nodes with %q topology (found:%d). Need at least %d", intnrt.SingleNUMANode, len(singleNUMANodeNRTs), minNumberOfNodesWithSameTopology)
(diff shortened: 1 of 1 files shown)


test/e2e/serial/config/config.go
```diff
@@ -178,3 +185,13 @@ func errorMapToString(errs map[string]error) string {
 	}
 	return sb.String()
 }
+
+func validateSMTAlignmentIsOFF(kconfigs map[string]*kubeletconfigv1beta1.KubeletConfiguration) map[string]error {
+	errs := map[string]error{}
+	for nodeName, kconfig := range kconfigs {
+		if kconfig.CPUManagerPolicyOptions["full-pcpus-only"] != "false" {
```
ffromani, line 192: I would prefer to use `strconv.ParseBool` rather than make logic on the string

ffromani (commented):
conceptually great idea, let's refine a bit

test/e2e/serial/config/config.go
ffromani: other than that, we can't import the `cpumanager` package (is not public by kube standards) so the option name is probably best left defined here, or maybe we can have moved as const in some internal package.

The only closing thought is this is an accidental limitation, not something we actively and willingly choose to require.

Tal-or: > other than that, we can't import the cpumanager package (is not public by kube standards) so the option name is probably best left defined here, or maybe we can have moved as const in some internal package.

Yea I was looking for a the public API representing this option but couldn't find it

ffromani:
needs rebase :\

test/e2e/serial/config/config.go
```diff
@@ -200,3 +208,21 @@ func checkNRTsArePresentForAllNodes(ctx context.Context) error {
 	}
 	return nil
 }
+
+func validateSMTAlignmentIsOFF(kconfigs map[string]*kubeletconfigv1beta1.KubeletConfiguration) map[string]error {
+	errs := map[string]error{}
+	for nodeName, kconfig := range kconfigs {
+		v, ok := kconfig.CPUManagerPolicyOptions["full-pcpus-only"]
+		if !ok {
+			// SMT is ON by default
```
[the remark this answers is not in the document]
Tal-or: We were discussing about it earlier, that SMT is enabled by default in CPU manager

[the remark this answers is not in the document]
Tal-or: Yes. This is why I added the validation in the first place. It complains about SMT alignment error.
It start working as expected only after I disabled it via the PerformanceProfie explicitly:
```
     annotations:
          kubeletconfig.experimental: |
             {"cpuManagerPolicyOptions": {"full-pcpus-only":"false"}}
```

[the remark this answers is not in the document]
ffromani: this is because the performance profile enables it: https://github.com/openshift/cluster-node-tuning-operator/blob/master/pkg/performanceprofile/controller/performanceprofile/components/kubeletconfig/kubeletconfig.go#L150

the end result is almost the same in this case but the reason why makes quite some difference. And it's wrong to assume that it is enabled by default if the option is present.

test/e2e/serial/config/config.go
```diff
@@ -200,3 +208,20 @@ func checkNRTsArePresentForAllNodes(ctx context.Context) error {
...
+		v, ok := kconfig.CPUManagerPolicyOptions["full-pcpus-only"]
+		if !ok {
+			// SMT alignment is OFF by default
+			continue
+		}
+		// don't check for error, the value has to be valid, kubelet won't start otherwise
+		enable, _ := strconv.ParseBool(v)
+		if enable {
+			errs[nodeName] = fmt.Errorf("SMT alignment is ON for node %q", nodeName)
+		}
```
ffromani, line 224: nit: writing like this is a tad more consistent:
```
enable, _ := strconv.ParseBool(v)
if !enable {
    continue
}
errs[nodeName] = fmt.Errorf("SMT alignment is ON for node %q", nodeName)
```

ffromani (commented):
/approve
/lgtm

PR #1167, opened 2025-01-27
ghaction: define E2E_NAMESPACE_NAME env

Tal-or:
The E2E_NAMESPACE_NAME environment variable is used but never defined.

That would lead to empty value which causes to unexpected behavior.


ffromani:
/approve
/lgtm

ah it fails in the line 73 below

Tal-or:
We don't hit the problem when CI is happy

PR #1169, opened 2025-01-27
utils: move nodepool pkg under tests

Tal-or:
The package is using another package from `test/...` and is not being used by any other packages but packages under `test/...`.

so let's move it also under `test` to be fully compatible with the project packages layout.


ffromani:
thanks, this works.
Ideally, though, we should dissolve `test/utils/...` over time, not making it larger.
What if we try the other way around? It seems we depend only on a couple functions like `GetHostedClusterName`. What if we pull out this function and move into this package (or create another thin package in `internal/...` ?)

Tal-or:
> thanks, this works. Ideally, though, we should dissolve `test/utils/...` over time, not making it larger. What if we try the other way around? It seems we depend only on a couple functions like `GetHostedClusterName`. What if we pull out this function and move into this package (or create another thin package in `internal/...` ?)

Sure. Will do

Tal-or:
BTW the fact that it sits under `utils` is just for convenience, the package has its own unique name which is `nodepools` so it avoid from the unclarity of general names like `utils`

ffromani:
> BTW the fact that it sits under `utils` is just for convenience, the package has its own unique name which is `nodepools` so it avoid from the unclarity of general names like `utils`

even stronger reasons to move it away from the `test/utils` tree :)

ffromani:
In theory we can rename `test/utils` like `test/internal` and call it a day, but this is mostly a (welcome) cosmetic change because we're still saying there are some packages which not only _are used_ but _should be used_ only by test code.
I think this is one of the key questions we should ask ourselves when we create new packages in this project (please note creating new packages or rearranging them [merge, split...] is **great** and something we should do regularly): "who's the expected user of this code?" `test/utils`: only test code; `internal`: only this project; `pkg`: potentially everyone in the world

Tal-or:
Discussed offline, and agree to leave it under `utils` and rename `test/utils` -> `test/internal`

`nodepools`  only client is the tests code so this reason is good enough.

PR #1170, opened 2025-01-28
e2e: classifying tests using labels

Tal-or:
The label package contains a list of labels that can be used in
the e2e and serial tests to indicate and point certain behavior or characteristics
of the various tests.

Those can be filtered/focused by ginkgo before the test runs.

Example:
`ginkgo --label-filter="!slow"` means avoid any specs labelled `slow`

The types defined in the package are type aliases and not new defined
types. This is done in order to avoid casting every time the labels are
being used.

This package allow greater readability, prevent typos (what can easily
happen while using plain-text) and
a common place to hold and documents every label that is being
used in the tests.

### DIFF
--- a/test/e2e/install/install_test.go
+++ b/test/e2e/install/install_test.go
@@ -19,6 +19,7 @@ package install
 import (
 	"context"
 	"fmt"
+	"github.com/openshift-kni/numaresources-operator/test/e2e/label"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -70,7 +71,7 @@ var _ = Describe("[Install] continuousIntegration", Serial, func() {
 	})
 
 	Context("with a running cluster with all the components", func() {
-		It("[test_id:47574][tier0] should perform overall deployment and verify the condition is reported as available", func() {
+		It("[test_id:47574] should perform overall deployment and verify the condition is reported as available", Label(label.Tier0), func() {
 			deployer := deploy.NewForPlatform(configuration.Plat)
 			nroObj := deployer.Deploy(context.TODO(), configuration.MachineConfigPoolUpdateTimeout)
 			nname := client.ObjectKeyFromObject(nroObj)
@@ -203,7 +204,7 @@ var _ = Describe("[Install] durability", Serial, func() {
 			deployer.Teardown(context.TODO(), 5*time.Minute)
 		})
 
-		It("[test_id:47587][tier1] should restart RTE DaemonSet when image is updated in NUMAResourcesOperator", func() {
+		It("[test_id:47587] should restart RTE DaemonSet when image is updated in NUMAResourcesOperator", Label(label.Tier1), func() {
 			By("getting up-to-date NRO object")
 			nroKey := objects.NROObjectKey()
 
--- a/test/e2e/label/label.go
+++ b/test/e2e/label/label.go
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package label
+
+// The label package contains a list of labels that can be used in
+// the e2e and serial tests to indicate and point certain behavior or characteristics
+// of the various tests.
+// Those can be filtered/focused by ginkgo before the test runs.
+
+// Kind is a label indicates specific criteria that classify the test.
+// A Mixture of kinds can be used for the same test
+type Kind = string
+
+const (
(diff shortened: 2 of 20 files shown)


test/e2e/label/label.go
```diff
@@ -0,0 +1,72 @@
+package label
```
ffromani, line 1: missing copyright boilerplate (`hack/boilerplate.go.txt`)

test/e2e/label/label.go
```diff
@@ -0,0 +1,72 @@
...
+// Tier is a label to classify tests under specific grade/level
+// that should roughly describe the execution complexity, maintainer identity and processing criteria.
+type Tier = string
+
+const (
+	// Tier0 are automated unit tests
+	// Minimal time needed to execute (minutes to 1 hour)
+	// Process criteria:
+	// 100% automated, must-pass 100%
+	// Development maintains tests, and reviews result
```
ffromani, line 32: descriptions are nice but they don't reflect what our serial tests are and wants to be

shajmakh (commented):
Thanks for proposing this.
The goal sounds good to me, I think that this is a good opportunity to cleanup test labels and tags such that:
1. move into using only ginkgo labels and not text tags (=remove tags within the specs titles such as `[tierX]`); d/s ci should be ready for this change and still a premerge test image can be tested for verification
2. remove duplicated ginkgo labels that are part of the features pool https://github.com/openshift-kni/numaresources-operator/blob/main/internal/api/features/_topics.json 
An example how this would turn out:
old:
```
It("[test_id:47613][tier3][nonreg][unsched] should not schedule a pod requesting resources that are not allocatable at numa level", Label(label.Tier3, "nonreg", "unsched"), Label("feature:unsched"), func()
```
new:
```
It("[test_id:47613] should not schedule a pod requesting resources that are not allocatable at numa level", Label(label.Tier3), Label("feature:unsched"), func()
```
^ non-reg is completely removed because it is labeled in the top Describe spec level

/approve

Tal-or:
> Thanks for proposing this. The goal sounds good to me, I think that this is a good opportunity to cleanup test labels and tags such that:
> 
> 1. move into using only ginkgo labels and not text tags (=remove tags within the specs titles such as `[tierX]`); d/s ci should be ready for this change and still a premerge test image can be tested for verification
> 2. remove duplicated ginkgo labels that are part of the features pool https://github.com/openshift-kni/numaresources-operator/blob/main/internal/api/features/_topics.json
>    An example how this would turn out:
>    old:
> 
> ```
> It("[test_id:47613][tier3][nonreg][unsched] should not schedule a pod requesting resources that are not allocatable at numa level", Label(label.Tier3, "nonreg", "unsched"), Label("feature:unsched"), func()
> ```
> 
> new:
> 
> ```
> It("[test_id:47613] should not schedule a pod requesting resources that are not allocatable at numa level", Label(label.Tier3), Label("feature:unsched"), func()
> ```
> 
> ^ non-reg is completely removed because it is labeled in the top Describe spec level
> 
> /approve

@shajmakh thank you for your comment
so basically all labels which are under: https://github.com/openshift-kni/numaresources-operator/blob/main/internal/api/features/_topics.json should be removed?

test/e2e/label/label.go
Tal-or: Please check again i've updated the desc

Tal-or:
> > @shajmakh thank you for your comment so basically all labels which are under: https://github.com/openshift-kni/numaresources-operator/blob/main/internal/api/features/_topics.json should be removed?
> 
> yes under the condition that they are in tags forms like `[feature-name]` of in ginkgo `Label("feature-name"..)` that is not in the form of `Label("feature:<feature-name>")`. The only labels/tags to keep should be either part of `Label("feature:<feature-name>")` or `Label(X,Y,Z)` where X,Y,Z are not feature names. I suggest this knowing that labels and tags are only used d/s and Label("feature:") is what is being looked for. I am not aware of other lanes using the other form of labels/tags.

so the feature list is obsolete? if so shall we remove the file as well? 
I mean this one  https://github.com/openshift-kni/numaresources-operator/blob/main/internal/api/features/_topics.json

Tal-or:
> > > @shajmakh thank you for your comment so basically all labels which are under: https://github.com/openshift-kni/numaresources-operator/blob/main/internal/api/features/_topics.json should be removed?
> > 
> > 
> > yes under the condition that they are in tags forms like `[feature-name]` of in ginkgo `Label("feature-name"..)` that is not in the form of `Label("feature:<feature-name>")`. The only labels/tags to keep should be either part of `Label("feature:<feature-name>")` or `Label(X,Y,Z)` where X,Y,Z are not feature names. I suggest this knowing that labels and tags are only used d/s and Label("feature:") is what is being looked for. I am not aware of other lanes using the other form of labels/tags.
> 
> so the feature list is obsolete? if so shall we remove the file as well? I mean this one https://github.com/openshift-kni/numaresources-operator/blob/main/internal/api/features/_topics.json

No need to delete the file. discussed offline

test/e2e/install/install_test.go
```diff
@@ -19,6 +19,7 @@ package install
 import (
 	"context"
 	"fmt"
+	"github.com/openshift-kni/numaresources-operator/test/e2e/label"
```
shajmakh, line 22: misplaced import

ffromani: true and it's annoying, but I don't think is worth waiting more to fix this. I can send a cleanup later.
There's a proposal about adding a CI check for imports, but IIRC requires more work.

Tal-or:
/retest
infra issue

ffromani:
failed in winddown stage, after the tests actually run succesfully.

PR #1174, opened 2025-01-28
pkg: add unit tests for merge

shajmakh:
Add missing unit tests for the merge package.

### DIFF
--- a/pkg/objectstate/merge/merge.go
+++ b/pkg/objectstate/merge/merge.go
@@ -32,11 +32,11 @@ var (
 func ServiceAccountForUpdate(current, updated client.Object) (client.Object, error) {
 	curSA, ok := current.(*corev1.ServiceAccount)
 	if !ok {
-		return updated, ErrWrongObjectType
+		return nil, ErrWrongObjectType
 	}
 	updSA, ok := updated.(*corev1.ServiceAccount)
 	if !ok {
-		return updated, ErrMismatchingObjects
+		return nil, ErrMismatchingObjects
 	}
 	updSA.Secrets = curSA.Secrets
 	return MetadataForUpdate(current, updated)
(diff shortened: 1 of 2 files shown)


pkg/objectstate/merge/merge_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+import (
+	"reflect"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	v1 "github.com/openshift-kni/numaresources-operator/api/numaresourcesoperator/v1"
```
ffromani, line 28: needs to be updated. Let's keep the convention to rename import `nropv1`

pkg/objectstate/merge/merge_test.go
```diff
@@ -0,0 +1,264 @@
...
+)
+
+type input struct {
```
ffromani, line 33: why do we need this? Can be defined inline below?

pkg/objectstate/merge/merge.go
```diff
@@ -32,11 +32,11 @@ var (
 func ServiceAccountForUpdate(current, updated client.Object) (client.Object, error) {
 	curSA, ok := current.(*corev1.ServiceAccount)
 	if !ok {
-		return updated, ErrWrongObjectType
+		return nil, ErrWrongObjectType
```
ffromani, line 35: we should ignore the other return value anyway if `err != nil` , but using an explicit value (`nil`) is a tad better anyway

ffromani (commented):
/approve

LGTM once rebased
minor/nit comments inside

pkg/objectstate/merge/merge_test.go
shajmakh: this is the functions input which is common for all functions, so defined it once instead of in every test function

ffromani (commented):
/lgtm

ffromani:
LGTM
once the linter errors are addressed

PR #1176, opened 2025-01-29
validation: fix error string

shajmakh:
Instead of pointers display readable string.

```
found multiple pools matches for node group 0xc00467dca0 but expected one. Pools found [0xc00129aa08 0xc00129ac78]
```

To:

```
found multiple pools matches for node group &{&LabelSelector{MatchLabels:map[string]string{test: common,},MatchExpressions:[]LabelSelectorRequirement{},} <nil> <nil> map[]} but expected one. Pools found [test1 test2]
```

### DIFF
--- a/api/v1/numaresourcesoperator_types.go
+++ b/api/v1/numaresourcesoperator_types.go
@@ -236,5 +236,9 @@ func (ng *NodeGroup) ToString() string {
 	if ng == nil {
 		return ""
 	}
-	return fmt.Sprintf("PoolName: %s MachineConfigPoolSelector: %s Config: %s", *ng.PoolName, ng.MachineConfigPoolSelector.String(), ng.Config.ToString())
+	return fmt.Sprintf("PoolName: %s "+
+		"MachineConfigPoolSelector: %s "+
+		"Config: %s "+
+		"Annotations: %s",
+		*ng.PoolName, ng.MachineConfigPoolSelector.String(), ng.Config.ToString(), ng.Annotations)
 }
--- a/api/v1/numaresourcesoperator_types.go
+++ b/api/v1/numaresourcesoperator_types.go
@@ -226,7 +226,7 @@ func init() {
 
 func (ngc *NodeGroupConfig) ToString() string {
 	if ngc == nil {
-		return ""
+		return "nil"
 	}
 	ngc.SetDefaults()
 	return fmt.Sprintf("PodsFingerprinting mode: %s InfoRefreshMode: %s InfoRefreshPeriod: %s InfoRefreshPause: %s Tolerations: %+v", *ngc.PodsFingerprinting, *ngc.InfoRefreshMode, *ngc.InfoRefreshPeriod, *ngc.InfoRefreshPause, ngc.Tolerations)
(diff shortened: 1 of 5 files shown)


Tal-or:
/hold
We can implement the `String()` function for `NodeGroups` which I see you already begun here:
https://github.com/openshift-kni/numaresources-operator/blob/643c76049443a2e170f3feeb8519563ae7db6853/api/v1/numaresourcesoperator_types.go#L235

ffromani:
99% of the cases we add `ToString()` methods and not `String()` methods because we are not yet ready to commit to a string representation or we want to reserve the chance to change.
In general using the common interfaces like `String()` is the right thing to do with the caveat above.

Tal-or:
Then we shall at least call `nodeGroup.ToString()` instead of printing it using `%v`

shajmakh:
/hold
Thanks for the feedback folks, this sounds like a better alternative, I'll update the PR.

ffromani (commented):
/approve
/lgtm

thanks, we can improve the string representation of the annotations map later

api/v1/numaresourcesoperator_types.go
```diff
@@ -226,15 +226,25 @@ func init() {
 
 func (ngc *NodeGroupConfig) ToString() string {
 	if ngc == nil {
-		return ""
+		return "nil"
```
ffromani, line 229: let's see what's the string representation of `nil`

api/v1/numaresourcesoperator_types.go
```diff
@@ -226,15 +226,25 @@ func init() {
...
 	}
 	ngc.SetDefaults()
 	return fmt.Sprintf("PodsFingerprinting mode: %s InfoRefreshMode: %s InfoRefreshPeriod: %s InfoRefreshPause: %s Tolerations: %+v", *ngc.PodsFingerprinting, *ngc.InfoRefreshMode, *ngc.InfoRefreshPeriod, *ngc.InfoRefreshPause, ngc.Tolerations)
 }
 
 func (ng *NodeGroup) ToString() string {
 	if ng == nil {
-		return ""
+		return "nil"
```
ffromani, line 237: ditto

api/v1/numaresourcesoperator_types.go
```diff
@@ -226,15 +226,25 @@ func init() {
...
 	}
-	return fmt.Sprintf("PoolName: %s MachineConfigPoolSelector: %s Config: %s", *ng.PoolName, ng.MachineConfigPoolSelector.String(), ng.Config.ToString())
+
+	pn := "nil"
```
ffromani, line 240: ditto

api/v1/numaresourcesoperator_types.go
ffromani: https://go.dev/play/p/Jx-o4sG1-dy

shajmakh: I followed the k8s convention as in https://github.com/kubernetes/apimachinery/blob/v0.31.2/pkg/apis/meta/v1/generated.pb.go#L4387 (and other String() methods) 
On another hand, I think we can use stringbuilder or similar to construct the string instead of sprintf for such long string presentation.

ffromani: ok, this is better. No need to change

shajmakh: The reason for this is that when ToString is called with empty values,some of these (like the v1.Labelselector presenting the MCP selector of the node group) is shown as "nil" while the internal ones lie ng.Config empty value is "" which doesn't look good and may confuse the readers (us mostly).

PR #1177, opened 2025-02-03
pkg: add unit tests for machine config state

shajmakh:
Add another set of tests to cover pkg/objectstate/cfg package.


ffromani (commented):
/approve
/lgtm

thanks a lot!

PR #1179, opened 2025-02-04
OCPBUGS-49407: pkg: use go-cmp only in tests

shajmakh:
Similarly to https://github.com/kubernetes/kubernetes/issues/104821 , we use the go-cmp package in our non-test code. Problem is: this package reserves the right to panic, which is something we cannot accept in out non-test code. We need to replace to similar code which guarantee not to panic.

Replace go-cmp.Equal with reflect.DeepEqual in non-test code. more on reflect.DeepEqual: https://pkg.go.dev/reflect#DeepEqual.

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -18,10 +18,8 @@ package status
 
 import (
 	"errors"
+	"reflect"
 	"time"
-
-	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -51,12 +49,13 @@ const (
 )
 
 func IsUpdatedNUMAResourcesOperator(oldStatus, newStatus *nropv1.NUMAResourcesOperatorStatus) bool {
-	options := []cmp.Option{
-		cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime"),
-		cmpopts.IgnoreFields(metav1.Condition{}, "ObservedGeneration"),
-	}
+	os := oldStatus.DeepCopy()
+	ns := newStatus.DeepCopy()
 
-	return !cmp.Equal(newStatus, oldStatus, options...)
+	resetIncomparableConditionFields(os.Conditions)
+	resetIncomparableConditionFields(ns.Conditions)
+
+	return !reflect.DeepEqual(os, ns)
 }
 
 // UpdateConditions compute new conditions based on arguments, and then compare with given current conditions.
@@ -65,12 +64,13 @@ func IsUpdatedNUMAResourcesOperator(oldStatus, newStatus *nropv1.NUMAResourcesOp
 func UpdateConditions(currentConditions []metav1.Condition, condition string, reason string, message string) ([]metav1.Condition, bool) {
 	conditions := NewConditions(condition, reason, message)
 
-	options := []cmp.Option{
-		cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime"),
-		cmpopts.IgnoreFields(metav1.Condition{}, "ObservedGeneration"),
-	}
+	cond := clone(conditions)
+	curCond := clone(currentConditions)
 
-	if cmp.Equal(conditions, currentConditions, options...) {
+	resetIncomparableConditionFields(cond)
+	resetIncomparableConditionFields(curCond)
+
+	if reflect.DeepEqual(cond, curCond) {
 		return currentConditions, false
 	}
 	return conditions, true
@@ -150,4 +150,17 @@ func MessageFromError(err error) string {
 		return err.Error()
 	}
 	return unwErr.Error()
+}
+
+func resetIncomparableConditionFields(conditions []metav1.Condition) {
+	for idx := range conditions {
+		conditions[idx].LastTransitionTime = metav1.Time{}
+		conditions[idx].ObservedGeneration = 0
+	}
+}
+
+func clone(conditions []metav1.Condition) []metav1.Condition {
+	var c = make([]metav1.Condition, len(conditions))
+	copy(c, conditions)
+	return c
 }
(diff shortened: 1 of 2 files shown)


pkg/status/status.go
```diff
@@ -151,3 +151,20 @@ func MessageFromError(err error) string {
...
+}
+
+func resetIncomparableConditionFields(conditions *[]metav1.Condition) {
+	for idx, _ := range *conditions {
+		(*conditions)[idx].LastTransitionTime = metav1.Time{}
+		(*conditions)[idx].ObservedGeneration = 0
+	}
+}
+
+func getShortcut(conditions []metav1.Condition) []metav1.Condition {
```
ffromani, line 166: why is this a shortcut?

pkg/status/status.go
```diff
@@ -51,12 +49,13 @@ const (
 )
 
 func IsUpdatedNUMAResourcesOperator(oldStatus, newStatus *nropv1.NUMAResourcesOperatorStatus) bool {
-	options := []cmp.Option{
-		cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime"),
-		cmpopts.IgnoreFields(metav1.Condition{}, "ObservedGeneration"),
-	}
+	os := *oldStatus.DeepCopy()
```
ffromani, line 52: why do we need to deference the pointer here...

pkg/status/status.go
```diff
@@ -51,12 +49,13 @@ const (
...
+	ns := *newStatus.DeepCopy()
+
+	resetIncomparableFields(&os)
```
ffromani, line 55: ...and take the address here?

ffromani (commented):
thanks!

pkg/status/status.go
```diff
@@ -65,12 +64,13 @@ func IsUpdatedNUMAResourcesOperator(oldStatus, newStatus *nropv1.NUMAResourcesOp
 func UpdateConditions(currentConditions []metav1.Condition, condition string, reason string, message string) ([]metav1.Condition, bool) {
 	conditions := NewConditions(condition, reason, message)
 
-	options := []cmp.Option{
-		cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime"),
-		cmpopts.IgnoreFields(metav1.Condition{}, "ObservedGeneration"),
-	}
+	newCmp := getShortcut(conditions)
```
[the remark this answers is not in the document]
shajmakh: for "compare", I wanted short names for the clone objects to compare, is newCopy sounds better? I'm open for other names

pkg/status/status.go
```diff
@@ -151,3 +151,20 @@ func MessageFromError(err error) string {
 	}
 	return unwErr.Error()
 }
+
+func resetIncomparableFields(status *nropv1.NUMAResourcesOperatorStatus) {
```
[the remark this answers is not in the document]
shajmakh: Yes we can, let me fix that. Thanks for highlighting.

pkg/status/status.go
shajmakh: Thanks for the comment. Bad name, intended as in get a copy not really shortcut, will change that to clone instead does that sound better?

pkg/status/status.go
shajmakh: because I want to update the copied object without the need to return a value from resetIncomparableFields. It isn't different from passing the original status and constructing a new one from it, update it and return it, do you think the return would be more clear?

pkg/status/status.go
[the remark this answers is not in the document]
ffromani: just use `cond` and `curCond` or any variation

pkg/status/status.go
ffromani: just `DeepCopy()` maybe?

shajmakh: changed to clone but DeepCopy is fine and goes well with the convention. But why do we need to export it?

pkg/status/status.go
```diff
@@ -151,3 +151,16 @@ func MessageFromError(err error) string {
 	}
 	return unwErr.Error()
 }
+
+func resetIncomparableConditionFields(conditions []metav1.Condition) {
+	for idx := range conditions {
+		conditions[idx].LastTransitionTime = metav1.Time{}
+		conditions[idx].ObservedGeneration = 0
+	}
+}
```
ffromani, line 160: are we guaranteed this mutates the given objects?

ffromani (commented):
Did we explored the tradeoff between a deepcopy and rewriting a specialized version of `DeepEqual` ?

pkg/status/status.go
ffromani: no need, private `clone` is fine

pkg/status/status.go
shajmakh: yes because `conditions[idx]` will modify the underlying array element at index `idx` whereas if this was a var defined in the `for` statement (or for scoped) and is updated later then the array wouldn't be mutated because we updated a copy:
https://go.dev/play/p/TZwmvA6_nJH

ffromani:
> > Did we explored the tradeoff between a deepcopy and rewriting a specialized version of `DeepEqual` ?
> 
> just to be clear, do you mean `reflect.DeepEqual()` vs writing a specialized version of `DeepEqual` ?

yes, and just to be clear I'm really just asking

pkg/status/status.go
ffromani: thanks for writing an example on go playground! even better would have been point out to a testcase in our codebase.

ffromani (commented):
/approve
/lgtm

we can move on. Thanks!

shajmakh:
> > > Did we explored the tradeoff between a deepcopy and rewriting a specialized version of `DeepEqual` ?
> > 
> > 
> > just to be clear, do you mean `reflect.DeepEqual()` vs writing a specialized version of `DeepEqual` ?
> 
> yes, and just to be clear I'm really just asking

reflectt.deepequal compares the exported and unexported fields deeply, but I think we can use our own deepEqual especially for the NROP status because we own that and we better ensure the equality with more tests in that area. 
/hold
we can merge this or wait for the custom equality check, let me know what do you prefer

ffromani:
/hold cancel

let's merge and move on

PR #1182, opened 2025-02-04
e2e serial fixes: 20250204

ffromani:
fixes for issue running the serial suite against a classic OCP cluster. Please check each commit for details.

### DIFF
--- a/internal/nodegroups/nodegroups.go
+++ b/internal/nodegroups/nodegroups.go
@@ -18,6 +18,7 @@ package nodegroups
 
 import (
 	"context"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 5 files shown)


internal/nodegroups/nodegroups.go
```diff
@@ -18,6 +18,7 @@ package nodegroups
```
shajmakh, line 18: why do we have this and https://github.com/openshift-kni/numaresources-operator/blob/main/api/v1/helper/nodegroup/nodegroup.go
the function GetPoolNamesFrom is similar to https://github.com/openshift-kni/numaresources-operator/blob/main/api/v1/helper/nodegroup/nodegroup.go#L136 should we have one internal helper?

ffromani: yes, we should merge in a future PR.

internal/nodegroups/nodegroups.go
```diff
@@ -70,10 +71,10 @@ func GetNodesFrom(ctx context.Context, cli client.Client, nodeGroups []nropv1.No
 }
 
 func GetPoolNamesFrom(ctx context.Context, cli client.Client, nodeGroups []nropv1.NodeGroup) ([]string, error) {
-	poolNames := make([]string, 0)
+	poolNames := make([]string, 0, len(nodeGroups))
 	for _, nodeGroup := range nodeGroups {
 		// this should cover HyperShift cases and OpenShift cases when PoolName in use
-		if *nodeGroup.PoolName != "" {
+		if nodeGroup.PoolName != nil && *nodeGroup.PoolName != "" {
```
Tal-or, line 77: Good catch

Tal-or (approved):

PR #1184, opened 2025-02-05
WIP: e2e: serial tests fixes

shajmakh:
e2e tests fixes, please check commits for details.

### DIFF
--- a/test/e2e/serial/tests/non_regression.go
+++ b/test/e2e/serial/tests/non_regression.go
@@ -116,7 +116,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			numOfNodeToBePadded := len(nrts) - 1
 
 			rl := corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("1"),
+				corev1.ResourceCPU:    resource.MustParse("2"),
 				corev1.ResourceMemory: resource.MustParse("4G"),
 			}
 			By("padding the nodes before test start")
@@ -152,7 +152,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			testPod := objects.NewTestPodPause(fxt.Namespace.Name, "testpod")
 			pSpec := &testPod.Spec
 
-			By(fmt.Sprintf("explicitly mentioning which we want pod to land on node %q", targetNodeName))
+			By(fmt.Sprintf("explicitly mentioning which node we want the pod to land on: %q", targetNodeName))
 			pSpec.NodeName = targetNodeName
 
 			By("setting a fake schedule name under the pod to make sure pod not scheduled by any scheduler")
@@ -161,7 +161,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 
 			cnt := &testPod.Spec.Containers[0]
 			requiredRes := corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("4"),
+				corev1.ResourceCPU:    resource.MustParse("2"),
 				corev1.ResourceMemory: resource.MustParse("100Mi"),
 			}
 			cnt.Resources.Requests = requiredRes
--- a/test/e2e/serial/tests/resource_hostlevel.go
+++ b/test/e2e/serial/tests/resource_hostlevel.go
@@ -298,7 +298,7 @@ var _ = Describe("[serial][hostlevel] numaresources host-level resources", Seria
 					}
 
 					klog.Infof("pad node %s with:\n%s", nodeName, intreslist.ToString(paddingResources))
-					pod := newPaddingPod(nodeName, "*", fxt.Namespace.Name, paddingResources)
+					pod := newPaddingPod(nodeName, "all", fxt.Namespace.Name, paddingResources)
 					pod.Spec.NodeName = nodeName // TODO: pinPodToNode?
 
 					err = fxt.Client.Create(ctx, pod)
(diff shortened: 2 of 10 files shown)


shajmakh:
/hold 
for local testing

ffromani (commented):
/approve
/lgtm

shajmakh:
more fixes are needed, will send them all in one patch

test/e2e/serial/tests/resource_hostlevel.go
```diff
@@ -298,7 +298,7 @@ var _ = Describe("[serial][hostlevel] numaresources host-level resources", Seria
 					}
 
 					klog.Infof("pad node %s with:\n%s", nodeName, intreslist.ToString(paddingResources))
-					pod := newPaddingPod(nodeName, "*", fxt.Namespace.Name, paddingResources)
```
ffromani, line 301: please split this change in its own commit

test/e2e/serial/tests/tolerations.go
```diff
@@ -107,66 +107,6 @@ var _ = Describe("[serial][disruptive][rtetols] numaresources RTE tolerations su
 			Expect(err).ToNot(HaveOccurred(), "cannot get %q in the cluster", dsKey.String())
 		})
 
-		When("[tier2] invalid tolerations are submitted ", Label("tier2"), func() {
```
ffromani, line 110: is not clear how this removal relates to the commit message, please elaborate

ffromani: If you remove this block, you can also remove `isDegradedRTESync`

test/e2e/serial/tests/tolerations.go
```diff
@@ -108,6 +108,7 @@ var _ = Describe("[serial][disruptive][rtetols] numaresources RTE tolerations su
 		})
 
 		When("[tier2] invalid tolerations are submitted ", Label("tier2"), func() {
+			const unsupportedValue = "Unsupported value"
```
ffromani, line 111: let's make it a top level constant or even better let's reuse constants from the library packages we consume, if possible

test/e2e/serial/tests/non_regression.go
```diff
@@ -116,7 +116,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			numOfNodeToBePadded := len(nrts) - 1
 
 			rl := corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("1"),
+				corev1.ResourceCPU:    resource.MustParse("2"),
 				corev1.ResourceMemory: resource.MustParse("4G"),
```
ffromani, line 120: TAE means we don't have enough resources, as confirmed by the commit message. So we fix the lack of resources by... requiring more resources?

ffromani (commented):
if you run the tests, which is welcome:
1. please explain in the commit message how do you run
2. please explain how and why was it failing before
3. please thus connect the last dot by explaining how and why the current fix is fixing

test/e2e/serial/tests/non_regression.go
shajmakh: these are resources to leave after the padding on each numazone, so this is in the setup, my bad on not elaborating this in the commit and clarify this and recommit

ffromani: right right, scratch my comment then.

test/e2e/serial/tests/tolerations.go
```diff
@@ -785,15 +786,23 @@ func isRTEPodFoundOnNode(cli client.Client, ctx context.Context, nodeName string
...
-	return strings.Contains(cond.Message, "FailedRTESync") // TODO: magic constant
+	if !strings.Contains(cond.Message, msgContains) {
+		klog.InfoS("Degraded message is not as expected", "condition", cond.String(), "expected message to contain", msgContains)
+		return false
+	}
+	if cond.Reason != status.ReasonInternalError {
+		klog.InfoS("Degraded reason is not as expected", "condition", cond.String(), "expected", status.ReasonInternalError)
+		return false
+	}
+	return true
```
ffromani, line 805: yes, it seems it's my fault in #1032 

I'd prefer a fix to look like
```
diff --git a/test/e2e/serial/tests/tolerations.go b/test/e2e/serial/tests/tolerations.go
index c4c0a919..866606dd 100644
--- a/test/e2e/serial/tests/tolerations.go
+++ b/test/e2e/serial/tests/tolerations.go
@@ -793,7 +793,11 @@ func isDegradedRTESync(conds []metav1.Condition) bool {
        if cond.Status != metav1.ConditionTrue {
                return false
        }
-       return strings.Contains(cond.Message, "FailedRTESync") // TODO: magic constant
+       if cond.Reason != status.ReasonInternalError {
+               klog.InfoS("Degraded reason is not as expected", "condition", cond.String(), "expected", status.ReasonInternalError)
+               return false
+       }
+       return strings.Contains(cond.Message, "Unsupported value") // TODO: magic constant
 }
```

but I'm not too attached to this form

shajmakh: if we just return the bool value of strings.Contains(cond.Message, "Unsupported value") it will hide some details from us as reviewers of test results, I'd look for a way to know what is the message of the degraded status and without the klog.InfoS in line 798 I will need to run the test myself to know that, so I think adding the extra log won't hurt but it will add extra details on why it failed, enabling us to address the issue faster.

ffromani: right, you have a point. Logs will help.

ffromani: still please mention #1032 as the source of breakage

shajmakh:
@ffromani thank you for the valuable reviews. considering the reboot test will be rewritten and another cleanup is needed I split the ready fixes to another PR: https://github.com/openshift-kni/numaresources-operator/pull/1192

PR #1186, opened 2025-02-10
remove GinkgoHelper() for better debugging

shajmakh:
The need for this was derived when reboot tests were reporting failure while waiting for MCP to get updated but the log was displaying only that there was context timeout without extra details on where specifically the failure happened. Due to additional reasons, it was also hard to follow the correct sequence of the logs. 
What GinkgoHelper does is that wherever it is added to a function, in case a failure happen during the execution of the function body, ginkgo will exclude this function from the stack trace, a thing that we highly need especially for the serial tests and reboot ones to be able to troubleshoot issues effectively. 

The purpose of this PR is not to fix any of the tests failures but to enable logging the exact failing line even if in helper functions, after all these are part of the tests and they too perform some unique verifications, and if the internal verifications fail the test is very likely to fail. 
Unless we are completely positive that the helper function is straightforward and stable, remove all GinkgoHelper() calls under `/tests`.

xref: https://github.com/onsi/ginkgo/blob/v2.22.2/core_dsl.go#L171


shajmakh:
/hold 
hold on with this because it is a part of a larger cleanup

PR #1188, opened 2025-02-10
WIP Event recording overhaul

ffromani:
WIP TBD


ffromani:
/close

first, let's verify we actually send too many events, as we _suspect_ we do.

PR #1189, opened 2025-02-11
add linters: copyloopvar, errcheck, gocritic

shajmakh:
Add more linters: copyloopvar, errcheck, gocritic. Please see commits for detailed changes.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -51,6 +51,7 @@ linters:
     - wastedassign
     - whitespace
     - gocritic
+    - stylecheck
 linters-settings:
   revive:
     rules:
@@ -89,3 +90,11 @@ linters-settings:
   gocritic:
     disabled-checks:
       - commentFormatting
+  stylecheck:
+    checks:
+      # A function's error value should be its last return value.
+      # https://staticcheck.dev/docs/checks/#ST1008
+      - ST1008
+      # Importing the same package multiple times.
+      # https://staticcheck.dev/docs/checks/#ST1019
+      - ST1019
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -32,7 +32,7 @@ issues:
 linters:
   disable-all: true
   enable:
-#    - errcheck
+    - errcheck
     - bodyclose
     - copyloopvar
     - depguard
@@ -82,3 +82,6 @@ linters-settings:
   ginkgolinter:
     # Trigger warning for ginkgo focus containers like `FDescribe`, `FContext`, `FWhen` or `FIt`
     forbid-focus-container: true
+  errcheck:
+    exclude-functions:
+      - fmt.Fprintf
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -50,6 +50,7 @@ linters:
     - unused
     - wastedassign
     - whitespace
+    - gocritic
 linters-settings:
   revive:
     rules:
@@ -85,3 +86,6 @@ linters-settings:
   errcheck:
     exclude-functions:
       - fmt.Fprintf
+  gocritic:
+    disabled-checks:
+      - commentFormatting
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -34,7 +34,7 @@ linters:
   enable:
 #    - errcheck
     - bodyclose
-#    - copyloopvar
+    - copyloopvar
     - depguard
     - ginkgolinter
     - gosimple
--- a/test/e2e/must-gather/must_gather_test.go
+++ b/test/e2e/must-gather/must_gather_test.go
@@ -29,6 +29,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog"
 	"sigs.k8s.io/yaml"
 
 	nropv1 "github.com/openshift-kni/numaresources-operator/api/v1"
@@ -82,7 +83,10 @@ var _ = Describe("[must-gather] NRO data collected", func() {
 			if _, ok := os.LookupEnv("E2E_NROP_MUSTGATHER_CLEANUP_SKIP"); ok {
 				return
 			}
-			os.RemoveAll(destDir)
+			err := os.RemoveAll(destDir)
+			if err != nil {
+				klog.Warningf("unable to remove temporary directory %q: %v", destDir, err)
+			}
 		})
 
 		It("check NRO data files have been collected", func(ctx context.Context) {
--- a/test/e2e/must-gather/must_gather_test.go
+++ b/test/e2e/must-gather/must_gather_test.go
@@ -248,14 +248,14 @@ func getMachineConfigPools(folder string) (mcov1.MachineConfigPoolList, error) {
 	return retval, nil
 }
 
-// return true if B keys and values respectively are part of A, otherwise false
-func isMapSubsetOf(A map[string]string, B map[string]string) bool {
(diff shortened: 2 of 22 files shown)


shajmakh:
/unhold 
we can add more later

test/deviceplugin/pkg/numacell/cmd/numacell.go
```diff
@@ -38,9 +38,17 @@ import (
...
+			klog.Warningf("failed writing data to builder: %v\n", err)
+			return buf.String()
+		}
 		for _, core := range node.Cores {
-			fmt.Fprintf(&buf, "\t%s\n", core.String())
+			_, err := fmt.Fprintf(&buf, "\t%s\n", core.String())
+			if err != nil {
+				klog.Warningf("failed writing data to builder: %v\n", err)
+				return buf.String()
+			}
```
ffromani, line 51: this can't really happen. If it does the sky ha fell already. I'd ignore this specific linter check.

test/deviceplugin/pkg/numacell/cmd/numacell.go
```diff
@@ -62,7 +70,10 @@ func render(w io.Writer) int {
 		if err != nil {
 			return 1
 		}
-		fmt.Fprintf(w, "---\n%s", string(data))
+		_, err = fmt.Fprintf(w, "---\n%s", string(data))
+		if err != nil {
+			klog.Warningf("failed writing manifests to the console: %v\n", err)
+		}
```
ffromani, line 76: ditto

test/e2e/must-gather/must_gather_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"k8s.io/klog"
```
ffromani, line 23: misplaced import

tools/watchnrtattr/watchnrtattr.go
```diff
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"k8s.io/klog/v2"
```
ffromani, line 21: misplaced import

tools/verify-commit/verify-commit.go
```diff
@@ -94,4 +93,5 @@ func main() {
 			os.Exit(2)
 		}
 	}
+	fmt.Printf("done verifying commits from stdin\n")
```
ffromani, line 96: why this one?

tools/watchnrtattr/watchnrtattr.go
```diff
@@ -59,7 +60,8 @@ func main() {
 		nrtObjs := topologyv1alpha2.NodeResourceTopologyList{}
 		wa, err := cli.Watch(ctx, &nrtObjs)
 		if err != nil {
-			log.Fatalf("cannot watch NRT objects: %v", err)
+			klog.Errorf("cannot watch NRT objects: %v", err)
+			return
```
ffromani, line 64: why this one?

ffromani (commented):
mostly OK, with some inline questions

tools/nrtcacheck/nrtcacheck.go
```diff
@@ -244,8 +244,7 @@ func parseArgs(args ...string) (ProgArgs, error) {
 	flags.BoolVar(&pArgs.Version, "version", false, "Output version and exit")
 	flags.BoolVar(&pArgs.DumpNodes, "dump-nodes", false, "Force node PFP status dump")
 	flags.Usage = func() {
-		fmt.Fprintf(flags.Output(), "Usage: %s [options] [node0 [node1] ... [nodeN]]\noptions:\n", os.Args[0])
-		flags.PrintDefaults()
+		_, _ = fmt.Fprintf(flags.Output(), "Usage: %s [options] [node0 [node1] ... [nodeN]]\noptions:\n", os.Args[0])
```
yanirq, line 247: why was `flags.PrintDefaults()` removed ?

tools/verify-commit/verify-commit.go
shajmakh: because on fatal error the function will exit immediately and not execute deferred code. 
Maybe a better way is actually replacing `Fatal` with `klog.Errorf(); return;` 
it is common that in many unit tests we use `fatal` and in others just `t.Error`. I prefer to go with `logging an error and return (or continue in case of a loop) because on a fatal error the code will completely exit as soon as it hit it and the rest of the tests won't execute.

tools/watchnrtattr/watchnrtattr.go
shajmakh: same as above, in case of fatal error the program will exit without executing deferred code

tools/nrtcacheck/nrtcacheck.go
shajmakh: good catch, this is by mistake

tools/verify-commit/verify-commit.go
ffromani: >because on fatal error the function will exit immediately and not execute deferred code.

ffromani: I prefer to move to `return X` and drop the explicit `os.Exit` calls rather than moving away from `defer`. Could you please try replacing `os.Exit()`s with `return`s?

test/deviceplugin/pkg/numacell/cmd/numacell.go
shajmakh: I thought since this is printed out to a returned value then it must be important to check. I skipped fmt.Fprintf for this linter

tools/verify-commit/verify-commit.go
shajmakh: yes indeed it is due to exit ultimately, so by logging a fatal, it prints out the error to a writer and then exits. 
this is main program, we can't return codes in main() (the proposal was eventually rejected). I will test this with only return, if that turns out to be fine we can mention the error code in the fmt.FprintF.

ffromani: ah right, `main` doesn't return int. Sigh.
feel free to move all the logic into another function and call it from `main`. Please add a comment referencing/summarizing this convo.

tools/verify-commit/verify-commit.go
```diff
@@ -66,6 +66,22 @@ func expectedDCOCoauthorTag(commit GitCommit) string {
 }
 
 func main() {
+	/*
```
yanirq, line 69: please align the comment's text

.golangci.yml
```diff
@@ -88,4 +89,12 @@ linters-settings:
...
\ No newline at end of file
+      - commentFormatting
+  stylecheck:
+    checks:
+      # A function's error value should be its last return value.
+      # https://staticcheck.dev/docs/checks/#ST1008
+      - ST1008
+      # Importing the same package multiple times.
+      # https://staticcheck.dev/docs/checks/#ST1019
+      - ST1019
```
ffromani, line 100: ok, these will be removed in future PRs. Especially the second is concerning.

ffromani (commented):
/approve

formatting fix pending and we can merge. Thanks!

.golangci.yml
shajmakh: why removed? we enable them here

ffromani: sorry I misread them to be in the "exclude" part. These are actually enabled, so we're good.

PR #1191, opened 2025-02-12
e2e: don't print scheduler logs in tests

shajmakh:
Even in case of failure, printing the scheduling logs is excessive, and it is anyway stored in the artifacts of the d/s CI.


ffromani:
it is probably too much to print all the logs on failure, yes.

If we find ourselves in the future in the position on which we want to add back code like that, we should stop and think if and how we can solve the general issue.

shajmakh:
> it is probably too much to print all the logs on failure, yes.
> 
> If we find ourselves in the future in the position on which we want to add back code like that, we should stop and think if and how we can solve the general issue.

I agree. Scheduler logs can be beneficial but I believe we should be selective of what to print in the test. Perhaps filtered logs are better here, after all even if the full logs were printed out we'll anyway be looking at specific logs such as for test's pod logs or test time scoped.

PR #1192, opened 2025-02-13
serial tests fixes 

shajmakh:
please see commit messages for details.


ffromani (commented):
/approve
/lgtm

this is a wakeup call for us to review (later, in a separate PRs) our helper to see how many of them are actual helper and how many of them should be rewritten from scratch. I can think of at least two helpers that should be inlined on top of my head.

shajmakh:
/hold 
for another run of the affected tests

PR #1193, opened 2025-02-16
SCC: Added annotation to deployment and daemonset pods template

rbaturov:
Since OCP 4.18  that all workloads of platform namespaces (openshift-*) explicitly pin the SCCs they need. https://docs.openshift.com/container-platform/4.17/authentication/managing-security-context-constraints.html#security-context-constraints-requiring_configuring-internal-oauth https://github.com/openshift/origin/blob/master/pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go


bundle/manifests/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -496,6 +496,7 @@ spec:
           template:
             metadata:
               annotations:
+                openshift.io/required-scc: anyuid
```
ffromani, line 499: seems ok

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -745,6 +745,10 @@ func daemonsetUpdater(poolName string, gdm *rtestate.GeneratedDesiredManifest) e
 		return err
 	}
 	if gdm.ClusterPlatform != platform.Kubernetes {
+		if gdm.DaemonSet.Spec.Template.Annotations == nil {
+			gdm.DaemonSet.Spec.Template.Annotations = make(map[string]string)
+		}
+		gdm.DaemonSet.Spec.Template.Annotations["openshift.io/required-scc"] = "resource-topology-exporter"
```
ffromani, line 751: ok in principle but should be probably done in the deployer libraries

rbaturov: But isn't the deployer an u/s project that won't be used on OCP specifically?

ffromani: Deployer has provisions to deploy on OCP platforms already as it is a legit supported platform alongside kubernetes

ffromani: example: https://github.com/k8stopologyawareschedwg/deployer/tree/main/pkg/assets/selinux/policy

rbaturov: Is this [daemonset](https://github.com/rbaturov/deployer/blob/main/pkg/manifests/yaml/rte/daemonset.yaml#L10) yaml a valid place to put that annotation?

ffromani: I'd prefer to have it in https://github.com/k8stopologyawareschedwg/deployer/blob/main/pkg/objectupdate/rte/rte.go#L73

bundle/manifests/numaresources-operator.clusterserviceversion.yaml
Tal-or, line 499: Is there a reason for picking this specific SCC?

Tal-or (commented):
LGTM 
I just wonder why did we go with `anyuid` and not `restricted` for example? 
Is there a best practice approach somewhere?

rbaturov:
@ffromani Ready for your review

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -746,10 +746,18 @@ func daemonsetUpdater(poolName string, gdm *rtestate.GeneratedDesiredManifest) e
 	}
 	if gdm.ClusterPlatform != platform.Kubernetes {
 		if gdm.IsCustomPolicyEnabled && gdm.ClusterPlatform == platform.OpenShift {
-			k8swgrteupdate.SecurityContext(gdm.DaemonSet, selinux.RTEContextTypeLegacy)
+			k8swgrteupdate.SecurityContextWithOpts(gdm.DaemonSet,
+				k8swgrteupdate.SecurityContextOptions{
+					SELinuxContextType:  selinux.RTEContextTypeLegacy,
+					SecurityContextName: rteupdate.MainContainerName,
```
ffromani, line 752: this is wrong, the SCCName is not the same thing as the container name. It happens to be the same value but we must use a different constant (or determine it dynamically)

rbaturov: I have added a constant name for this matter.

ffromani (commented):
/approve

deferring to other reviewers the LGTM

for the future: please use active form in commit messages: "add annotation to..."

ffromani:
do we still need this change?

rbaturov:
> do we still need this change?

For now let's close it. 
This change was needed to pass a test in the unfinished work for NROP upgrade lane.
When I'll have some more capacity we can reconsider pursuing that effort.

PR #1194, opened 2025-02-18
performance: minimize update calls

Tal-or:
Many of the `Update` calls done by the operator
are changing nothing.

This is happening because the API server defaults some of the fields
while our code leave them empty.

This result in a diff in the comparison function and leads to redundant calls.

This package purpose it to set the default values in our objects similar to what
the API server is doing.

The code is based on upstream code that is
not in staging so we can't import it unfortunately.

Below benchmarks shows the benefit of those changes:

**Simple patch against `numaresources` CR (before change):**
```
oc patch numaresourcesoperators.nodetopology.openshift.io numaresourcesoperator \
  --type='json' -p '[{"op": "add", "path": "/spec/nodeGroups/0/annotations", "value": {"key3":"value6"}}]'
numaresourcesoperator.nodetopology.openshift.io/numaresourcesoperator patched
```
```
./kubectl-dev_tool audit -f audit.log --after 2025-02-18T17:10:00+02:00 -otop --by resource --verb=update --verb=create  --user=system:serviceaccount:openshift-numaresources:numaresources-controller-manager  --resource=*.* --resource=-leases.* 
count: 7, first: 2025-02-18T17:12:26+02:00, last: 2025-02-18T17:12:26+02:00, duration: 59.093ms
2x                   security.openshift.io/v1/securitycontextconstraints
1x                   apiextensions.k8s.io/v1/customresourcedefinitions
1x                   nodetopology.openshift.io/v1/numaresourcesoperators
1x                   v1/services
1x                   apps/daemonsets
1x                   v1/serviceaccounts
```

**Simple patch against `numaresources` CR (after change):**
```
oc patch numaresourcesoperators.nodetopology.openshift.io numaresourcesoperator \
  --type='json' -p '[{"op": "add", "path": "/spec/nodeGroups/0/annotations", "value": {"key3":"value6"}}]'
numaresourcesoperator.nodetopology.openshift.io/numaresourcesoperator patched
```
```
./kubectl-dev_tool audit -f audit.log --after 2025-02-18T15:30:00+02:00 -otop --by resource --verb=update --verb=create  --user=system:serviceaccount:openshift-numaresources:numaresources-controller-manager  --resource=*.* --resource=-leases.* 
count: 1, first: 2025-02-18T17:02:34+02:00, last: 2025-02-18T17:02:34+02:00, duration: 0s
1x                   nodetopology.openshift.io/v1/numaresourcesoperators
```

**Realistic benchmark against serial suite (before change):**
```
E2E_NROP_INSTALL_SKIP_KC=true ./bin/e2e-nrop-serial.test  --ginkgo.skip='reboot_required|schedrst' --ginkgo.label-filter='!(feature: containsAny { ngpoolname, nganns })' --ginkgo.label-filter='tier0||tier1' --ginkgo.dry-run

Ran 32 of 120 Specs in 2895.370 seconds
SUCCESS! -- 32 Passed | 0 Failed | 0 Pending | 88 Skipped
```
```
./kubectl-dev_tool audit -f audit.log --after 2025-02-18T14:10:00+02:00 -otop --by resource --verb=update --verb=create --user=system:serviceaccount:openshift-numaresources:numaresources-controller-manager --resource=*.* --resource=-leases.* 
count: 113, first: 2025-02-18T14:40:20+02:00, last: 2025-02-18T15:07:09+02:00, duration: 26m48.655819s
32x                  security.openshift.io/v1/securitycontextconstraints
16x                  apps/daemonsets
16x                  v1/serviceaccounts
16x                  apiextensions.k8s.io/v1/customresourcedefinitions
16x                  nodetopology.openshift.io/v1/numaresourcesoperators
16x                  v1/services
1x                   v1/events
```

**Realistic benchmark against serial suite (after change):**

```
E2E_NROP_INSTALL_SKIP_KC=true ./bin/e2e-nrop-serial.test  --ginkgo.skip='reboot_required|schedrst' --ginkgo.label-filter='!(feature: containsAny { ngpoolname, nganns })' --ginkgo.label-filter='tier0||tier1' --ginkgo.dry-run

Ran 32 of 120 Specs in 2901.802 seconds
SUCCESS! -- 32 Passed | 0 Failed | 0 Pending | 88 Skipped
```

```
./kubectl-dev_tool audit -f audit.log --after 2025-02-18T16:00:00+02:00 -otop --by resource --verb=update --verb=create --user=system:serviceaccount:openshift-numaresources:numaresources-controller-manager --resource=*.* --resource=-leases.* 

// return empty. no redundant call at all
```

**NOTE:** we must add a CI lane that tracks the amount of update call and failed if we're exceeding a certain limit.
This is needed because atm we don't have a better way to track changes in the set-default  logic of the API server.


pkg/apply/apply.go
```diff
@@ -68,15 +68,21 @@ func ApplyObject(ctx context.Context, cli k8sclient.Client, objState objectstate
 		return objState.Desired, true, nil
 	}
 
+	// set default to the desired object to avoid the comparison from failing
+	// on default values which are set by the API server
+	desiredWithDefault := objState.Default(objState.Existing, objState.Desired)
+
 	// Merge the desired object with what actually exists
-	merged, err := objState.Merge(objState.Existing, objState.Desired)
+	merged, err := objState.Merge(objState.Existing, desiredWithDefault)
```
ffromani, line 76: another option could be to make our merge functions smarter. We should only copy the desired fields in the existing object. This allows us to avoid defaults (and the need to depend on defaulter) and was the original intent of the `merge` package, which was however lost in the initial implementation.

We can take the chance to rethink the `Compare` step. Can we do `Merge+Compare` in one single go? Can the merging step infer if a field we care about changed during the merge?

ffromani (commented):
Impressive PR. My only concern is to depend on external defaulter package we need to cut and paste here. Can we avoid it with smarter merge functions?

ffromani:
some historical context.
The person who wrote the merge code (cough cough) implemented the bare minimum merge code back in time. That was not very smart and admittedly pretty raw. The thought was like "we will optimize later" and later, of course, never came.

I'm not against using `defaulter`. This is actually a good solution and another very valid approach which complements smart and precise merge functions. The problem is having easy access to that package.

pkg/objectstate/setdefault/setdefault.go
```diff
@@ -0,0 +1,310 @@
+package setdefault
```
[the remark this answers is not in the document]
Tal-or: The functions are located as part of the API:
https://github.com/kubernetes/kubernetes/blob/e5976909c6fb129228a67515e0f86336a53884f0/pkg/apis/core/v1/zz_generated.defaults.go

pkg/apply/apply.go
Tal-or: Defining a list of fields we care about using a black/white list is possible however it has its own cons:
1. We need to define the fields we care about which is not that easy because we don't know what will happen in case that a field "we didn't care about" is misconfigured, could it break anything?
2. We need to write a custom compare/merge functions that will be smart enough to skip and focus on the fields according to our needs, which is basically similar to setting the defaults and goes through everything.
3. What should we do if user is tampering with fields we didn't care about and messing things up? we'll never be able to fix it automatically, a manual intervention will be needed.

go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/openshift-kni/numaresources-operator
 
-go 1.22.0
-
-toolchain go1.22.8
+go 1.23.0
```
ffromani, line 3: bumping deps is fine. Bumping the golang version is fine (but we need to check the builder images as well). Bumping them both in the same commit is not fine

go.mod
```diff
@@ -144,9 +147,15 @@ replace (
...
 	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.31.2
 	k8s.io/metrics => k8s.io/metrics v0.31.2
 	k8s.io/mount-utils => k8s.io/mount-utils v0.31.2
 	k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.31.2
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.31.2
 )
+
+replace (
+	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20241107164952-923091dd2b1a // release-4.18
+	sigs.k8s.io/structured-merge-diff/v4 => sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // remove when bumping k8s.io/apimachinery => v0.32.0
```
ffromani, line 160: why is this needed?

pkg/objectstate/defaulter/defaulter.go
```diff
@@ -0,0 +1,309 @@
+package defaulter
```
ffromani, line 1: missing copyright boilerplate (`hack/boilerplate.go.txt`)

pkg/objectstate/defaulter/defaulter.go
```diff
@@ -0,0 +1,309 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	securityv1 "github.com/openshift/api/security/v1"
+	"github.com/openshift/apiserver-library-go/pkg/securitycontextconstraints/sccdefaults"
+)
+
+func Service(original, mutated client.Object) client.Object {
+	service := mutated.(*corev1.Service)
```
ffromani, line 17: do we need a check or do we want to panic if the object is not what we expect? applies to everything else below

Makefile
```diff
@@ -571,3 +571,9 @@ mv $(1) $(1)-$(3) ;\
 } ;\
 ln -sf $(1)-$(3) $(1)
 endef
+
+
+.PHONY: defaults
+defaults:
+	@echo "generating defaults"
+	hack/generate-defaults.sh
```
ffromani, line 579: which files this script will generate?

pkg/objectstate/merge/merge.go
```diff
@@ -148,3 +161,20 @@ func preserveServiceAccountPullSecrets(original, mutated *corev1.ServiceAccount)
 
 	mutated.Secrets = original.Secrets
 }
+
+// preserve the IP configuration from the original object since
+// those are assigned by external operator (not ours)
+func preserveIPConfigurations(original, mutated *corev1.ServiceSpec) {
```
ffromani, line 167: makes sense. I guess we will need to play catchup to find cases like this in the future?
nit: please make the comment docstring compliant:
```
// preserveIPConfigurations keep the configuration from the original object
// because it is owned by an external operator (not ours)
```

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"github.com/openshift-kni/numaresources-operator/pkg/objectstate/defaulter"
```
ffromani, line 21: misplaced import

Makefile
```diff
@@ -144,7 +144,7 @@ test-must-gather-e2e: build-must-gather-e2e
 
 .PHONY: lint
 lint: update-buildinfo golangci-lint ## Run golangci-lint linter
-	$(GOLANGCI_LINT) --verbose run --print-resources-usage --exclude-dirs test/internal/k8simported
+	$(GOLANGCI_LINT) --verbose run --print-resources-usage --exclude-dirs 'test/internal/k8simported|pkg/objectstate/defaulter'
```
ffromani, line 147: perhaps we can isolate the autogenerated code in a subpackage of `defaulter` (or into another package even?)
so we can minimize the code we have to blindly trust

ffromani (commented):
The more I review, the more I like the PR. But it is an important and relatively invasive change, so it has to be considered carefully.
Questions and suggestions inline

go.mod
Tal-or: keep API compatibility, we have other dependencies which cannot use the latest version

Makefile
Tal-or: everything under the defaulter package but the `defaulter.go` file.

I think I'll move all the generated code into a separate dir inside `defaulter`

pkg/objectstate/defaulter/defaulter.go
Tal-or: This code pulled directly from u/s we cannot change it if we want to keep it aligned with u/s

Tal-or: > This code pulled directly from u/s we cannot change it if we want to keep it aligned with u/s

Sorry my bad it's not.

so basically we can, but the object is given by our code, meaning it's not provided from the API server or any other producer. 
We can log an error and return doing nothing. is that good enough?

go.mod
Tal-or: I saw it already bumped on separate PR

pkg/objectstate/defaulter/defaulter.go
Tal-or: Fixed

ffromani:
> @ffromani PR is reviewable again

unfortunately needs a rebase. Generally speaking I want this PR merged and even possibly backported to 4.18

go.mod
```diff
@@ -36,6 +36,12 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require (
+	github.com/openshift/apiserver-library-go v0.0.0-20250502092109-8d341e94467d
+	k8s.io/apiserver v0.32.3
+	k8s.io/kubernetes v1.33.0
+)
```
ffromani, line 43: why a separate group?

go.mod
```diff
@@ -145,3 +155,8 @@ replace (
 	k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.32.3
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.32.3
 )
+
+replace (
+	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20241107164952-923091dd2b1a
+	k8s.io/kubernetes => k8s.io/kubernetes v1.32.2
```
ffromani, line 161: this is probably the biggest open issue about this PR. Depending again on `k8s.io/kubernetes` is something I'm extremely reluctant to because it brings a world of pain for rebases/updates. We spent a fair bit of energy to remove this dep so even though this PR is compelling, is still a debatable proposition.

ffromani:
reviewing this PR is high on my priority. This is very good work. Conceptually I want this in, let's see how we can address the implementation details.

go.mod
Tal-or: > this is probably the biggest open issue about this PR. Depending again on `k8s.io/kubernetes` is something I'm extremely reluctant to because it brings a world of pain for rebases/updates. We spent a fair bit of energy to remove this dep so even though this PR is compelling, is still a debatable proposition.

We were discussing about that in the initial iterations of this PR.
Depending on `k8s.io/kubernetes` is the price that we pay for updating the default functions automatically.

go.mod
Tal-or: Mistake probably. will fix in next iteration

go.mod
ffromani: correct. This is the main/only friction point. I need to invest time here. Will review everything else in between.

ffromani:
I spent most of the morning reviewing this PR and trying to remove the `k8s.io/kubernetes` dependency. It's quite hard.
At the same time, reintroducing the dep on `k8s.io/kubernetes` is something we can't do. Despite how good this is PR, the ROI is still not high enough.
What we can do to move forward is to extract the few changes which don't depend on the `k8s.io/kubernetes` (directly or indirectly) and start merging those. I think 1371710b (plus: ```
diff --git a/pkg/objectstate/merge/merge.go b/pkg/objectstate/merge/merge.go
index 27c8119b..8ccf4df6 100644
--- a/pkg/objectstate/merge/merge.go
+++ b/pkg/objectstate/merge/merge.go
@@ -21,7 +21,7 @@ import (
 
        appsv1 "k8s.io/api/apps/v1"
        corev1 "k8s.io/api/core/v1"
-       v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+       apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
        "k8s.io/apimachinery/pkg/util/sets"
 
        "sigs.k8s.io/controller-runtime/pkg/client"
@@ -62,8 +62,8 @@ func StatusForUpdate(current client.Object, updated client.Object) (client.Objec
                updated.(*appsv1.DaemonSet).Status = currentTyped.Status
        case *corev1.Service:
                updated.(*corev1.Service).Status = currentTyped.Status
-       case *v1.CustomResourceDefinition:
-               updated.(*v1.CustomResourceDefinition).Status = currentTyped.Status
+       case *apiextv1.CustomResourceDefinition:
+               updated.(*apiextv1.CustomResourceDefinition).Status = currentTyped.Status
        default:
                return updated, nil
        }
```
and 1c64a334 are easy to go in. Everything else needs more thought

Tal-or:
> I spent most of the morning reviewing this PR and trying to remove the `k8s.io/kubernetes` dependency. It's quite hard. At the same time, reintroducing the dep on `k8s.io/kubernetes` is something we can't do. Despite how good this is PR, the ROI is still not high enough. What we can do to move forward is to extract the few changes which don't depend on the `k8s.io/kubernetes` (directly or indirectly) and start merging those. I think 1371710b (plus:
> 
> ```
> diff --git a/pkg/objectstate/merge/merge.go b/pkg/objectstate/merge/merge.go
> index 27c8119b..8ccf4df6 100644
> --- a/pkg/objectstate/merge/merge.go
> +++ b/pkg/objectstate/merge/merge.go
> @@ -21,7 +21,7 @@ import (
>  
>         appsv1 "k8s.io/api/apps/v1"
>         corev1 "k8s.io/api/core/v1"
> -       v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
> +       apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
>         "k8s.io/apimachinery/pkg/util/sets"
>  
>         "sigs.k8s.io/controller-runtime/pkg/client"
> @@ -62,8 +62,8 @@ func StatusForUpdate(current client.Object, updated client.Object) (client.Objec
>                 updated.(*appsv1.DaemonSet).Status = currentTyped.Status
>         case *corev1.Service:
>                 updated.(*corev1.Service).Status = currentTyped.Status
> -       case *v1.CustomResourceDefinition:
> -               updated.(*v1.CustomResourceDefinition).Status = currentTyped.Status
> +       case *apiextv1.CustomResourceDefinition:
> +               updated.(*apiextv1.CustomResourceDefinition).Status = currentTyped.Status
>         default:
>                 return updated, nil
>         }
> ```
> 
> and 1c64a334 are easy to go in. Everything else needs more thought

Thank you for looking at this, I can't find the commits with the hash mentioned above.

Without having a way to keep the defaults up-to-date, there's not much benefit in this PR, because sooner or later it will break.

Another option is to write an e2e test that will catch the breakage at runtime which will force us to fix the logic.

ffromani:
https://github.com/openshift-kni/numaresources-operator/pull/1194/commits/528cad64c4822601654c38815edbf276ce7bfac3 and https://github.com/openshift-kni/numaresources-operator/pull/1194/commits/57425ccd5e31ffb8fbb20ead0a27091de895a1f3 we can merge quickly

ffromani:
I spent few hours looking at the generator script.
The problem is the script is out of necessity complex and does many transformations to the the original source code to create our packages. I see why it is like this. I can't think of a better way. Still, this is too complex and fragile to be maintained. We can extract some more bits from this PR (https://github.com/openshift-kni/numaresources-operator/pull/1313) but I think the generator script per se cannot move forward, so the `defaulter` idea can't progress.

However.
Thinking hard and (relatively) long about how to achieve the same result without the defaulter, there is a relatively contained possible way reorganizing how we do the merge step.

The merge step is how we compute the delta between the desired state and the current state, which is what we use as payload of our `Update` calls.
The existing state is what we got from the cluster (a glorified set of `Get` calls hidden in `FromClient` calls).
The desired state is created by deployer libraries:
1. starting from zeroed objects
2. enforcing the values we want, optionally mutated by the config

Thinking about it, we already have our delta, but implicit and buried in the deployer libs.
If the deployer libs:
1. make all the changes they need to manifests programmatically, instead of relying on loaded data (implementation detail)
2. can take a random object as input to mutate, vs implicitly and always use the loaded data as starting point

then we can
1. use the existing data as base
2. apply mutations from deployer libs to reach the desired state
3. use this data (plus optimization we have already maybe) as payload of `Update` calls.

this makes our `merge` package useless, replaced by functionality from deployer libs, but should be doable with targeted and contained changes in our controller logic.

PR #1195, opened 2025-02-19
serial tests fixes

shajmakh:
please see commits for more details.

### DIFF
--- a/internal/devices/devices.go
+++ b/internal/devices/devices.go
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package devices
+
+const (
+	DevType1EnvVar = "E2E_NROP_DEVICE_TYPE_1"
+	DevType2EnvVar = "E2E_NROP_DEVICE_TYPE_2"
+	DevType3EnvVar = "E2E_NROP_DEVICE_TYPE_3"
+)
--- a/internal/noderesourcetopology/equality.go
+++ b/internal/noderesourcetopology/equality.go
@@ -18,12 +18,16 @@ package noderesourcetopology
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	nrtv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/klog"
+
+	"github.com/openshift-kni/numaresources-operator/internal/devices"
 )
 
 func EqualZones(zonesA, zonesB nrtv1alpha2.ZoneList, isRebootTest bool) (bool, error) {
@@ -71,7 +75,16 @@ func EqualResourceInfos(resInfosA, resInfosB nrtv1alpha2.ResourceInfoList, isReb
 
 func EqualResourceInfo(resInfoA, resInfoB nrtv1alpha2.ResourceInfo, isRebootTest bool) (bool, error) {
 	if isRebootTest && strings.Compare(resInfoA.Name, string(corev1.ResourceMemory)) == 0 {
-		return EqualResourceInfoWithDeviation(resInfoA, resInfoB)
+		return EqualMemoryWithDeviation(resInfoA, resInfoB)
+	}
+
+	// TODO we almost never use real devices but in case we do we still pass the names using same env variables as for
+	// sample devices; need a way to distinguish between real vs sample devices used in the ci
+	if isRebootTest && resourceIsDevice(resInfoA.Name) {
+		// TODO: Feb 2025: NROP serial is using sample-device plugin which has a known issue that after operations like node reboot or taints update,
+		// quantity of the device resources is messed up, so this is a workaround for reboot tests should be removed once
+		// https://issues.redhat.com/browse/CNF-12824 is resolved. Note that this doesn't happen with real devices like sriov.
+		return EqualDevicesWithDeviation(resInfoA, resInfoB)
 	}
 
 	if resInfoA.Name != resInfoB.Name {
@@ -89,7 +102,58 @@ func EqualResourceInfo(resInfoA, resInfoB nrtv1alpha2.ResourceInfo, isRebootTest
 	return true, nil
 }
 
-func EqualResourceInfoWithDeviation(resInfoA, resInfoB nrtv1alpha2.ResourceInfo) (bool, error) {
+func EqualDevicesWithDeviation(a, b nrtv1alpha2.ResourceInfo) (bool, error) {
+	if a.Name != b.Name {
+		return false, fmt.Errorf("mismatched resource name initial=%q vs updated=%q", a.Name, b.Name)
+	}
+	if a.Capacity.Cmp(b.Capacity) != 0 {
+		klog.Warningf("mismatched resource Capacity initial=%v vs updated=%v", ResourceInfoToString(a), ResourceInfoToString(b))
+	}
+	if a.Allocatable.Cmp(b.Allocatable) != 0 {
+		klog.Warningf("mismatched resource Allocatable initial=%v vs updated=%v", ResourceInfoToString(a), ResourceInfoToString(b))
+		// we shouldn't tolerate different consumptions ratios though
+		ac := a.Capacity
+		bc := b.Capacity
+		ac.Sub(a.Allocatable)
+		bc.Sub(b.Allocatable)
+		if ac.Cmp(bc) != 0 {
+			return false, fmt.Errorf("mismatched resource consumption; expected %v==%v=true", ac.String(), bc.String())
+		}
+	}
+	if a.Available.Cmp(b.Available) != 0 {
+		klog.Warningf("mismatched resource Available initial=%v vs updated=%v", ResourceInfoToString(a), ResourceInfoToString(b))
+		// we shouldn't tolerate different consumptions ratios though
+		ac := a.Capacity
+		bc := b.Capacity
+		ac.Sub(a.Available)
+		bc.Sub(b.Available)
+		if ac.Cmp(bc) != 0 {
+			return false, fmt.Errorf("mismatched resource consumption; expected %v==%v=true", ac.String(), bc.String())
+		}
+	}
+	return true, nil
+}
+
+func resourceIsDevice(resName string) bool {
+	var name string
+	var ok bool
+
+	name, ok = os.LookupEnv(devices.DevType1EnvVar)
+	if ok && resName == name {
+		return true
+	}
+	name, ok = os.LookupEnv(devices.DevType2EnvVar)
+	if ok && resName == name {
+		return true
+	}
+	name, ok = os.LookupEnv(devices.DevType3EnvVar)
+	if ok && resName == name {
+		return true
+	}
+	return false
+}
+
+func EqualMemoryWithDeviation(resInfoA, resInfoB nrtv1alpha2.ResourceInfo) (bool, error) {
 	/*
 		tests that involve reboot show a difference in the memory capacity on numa nodes, e.g memory capacity
 		 is reallocated across numa nodes upon startup. This is a a kernel behavior which we have no control
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -235,10 +235,10 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 				Expect(err).ToNot(HaveOccurred())
 
 				//this will trigger node reboot as the NROP settings will be reapplied to the unlabelled node, so new node is added under the old mcp hence the MachineCount update type
-				waitForMcpUpdate(fxt.Client, context.TODO(), MachineCount, initialMcpInfo)
+				waitForMcpUpdate(fxt.Client, context.TODO(), MachineCount, time.Now().String(), newMcpInfo)
 			}()
 
-			waitForMcpUpdate(fxt.Client, context.TODO(), MachineConfig, newMcpInfo)
+			waitForMcpUpdate(fxt.Client, context.TODO(), MachineConfig, time.Now().String(), newMcpInfo)
 
 			By(fmt.Sprintf("modifying the NUMAResourcesOperator nodeGroups field to match new mcp: %q labels %q", mcp.Name, mcp.Labels))
 			Eventually(func(g Gomega) {
@@ -274,12 +274,13 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 				By("waiting for mcps to start updating")
 				// this will trigger mcp update only for the initial mcps because the mcp-test nodes are still labeled
 				// with the old labels, so worker mcp will switch back to the NROP mc
-				waitForMcpUpdate(fxt.Client, context.TODO(), MachineConfig, initialMcpInfo)
+
+				waitForMcpUpdate(fxt.Client, context.TODO(), MachineConfig, time.Now().String(), initialMcpInfo)
 			}() //end of defer
 
 			By("waiting for the mcps to update")
 			// on old mcp because the ds will no longer include the worker node that is not labeled with mcp-test, so returning to MC without NROP settings
-			waitForMcpUpdate(fxt.Client, context.TODO(), MachineConfig, initialMcpInfo)
+			waitForMcpUpdate(fxt.Client, context.TODO(), MachineConfig, time.Now().String(), initialMcpInfo)
 
 			By(fmt.Sprintf("Verify RTE daemonsets have the updated node selector matching to the new mcp %q", mcp.Name))
 			Eventually(func() (bool, error) {
@@ -578,12 +579,12 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 						)).ToNot(HaveOccurred())
 					}
 					By("waiting for mcp to update")
-					waitForMcpUpdate(fxt.Client, context.TODO(), MachineConfig, mcpsInfo...)
+					waitForMcpUpdate(fxt.Client, context.TODO(), MachineConfig, time.Now().String(), mcpsInfo...)
 				}
 			}()
 
 			By("waiting for mcp to update")
-			waitForMcpUpdate(fxt.Client, context.TODO(), MachineConfig, mcpsInfo...)
+			waitForMcpUpdate(fxt.Client, context.TODO(), MachineConfig, time.Now().String(), mcpsInfo...)
 
 			By("checking that NUMAResourcesOperator's ConfigMap has changed")
 			cmList := &corev1.ConfigMapList{}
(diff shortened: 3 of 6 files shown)


shajmakh:
proposal for rewriting the test is found here: https://github.com/openshift-kni/numaresources-operator/pull/1196

test/e2e/serial/tests/configuration.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/openshift/cluster-node-tuning-operator/pkg/performanceprofile/controller/performanceprofile/resources"
```
shajmakh: changes are not needed in this commit, should be cleanedup

internal/noderesourcetopology/equality.go
```diff
@@ -89,6 +102,59 @@ func EqualResourceInfo(resInfoA, resInfoB nrtv1alpha2.ResourceInfo, isRebootTest
...
+		return true
+	}
+	name, ok = os.LookupEnv(devices.DevType3EnvVar)
+	if ok && resName == name {
+		return true
+	}
+	return false
+}
+
+// TODO rename to EqualMemoryWithDeviation because it is dedicated for memory resource
```
shajmakh, line 157: need a cleanup

internal/devices/devices.go
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package devices
+
+const (
+	DevType1EnvVar = "E2E_NROP_DEVICE_TYPE_1"
+	DevType2EnvVar = "E2E_NROP_DEVICE_TYPE_2"
+	DevType3EnvVar = "E2E_NROP_DEVICE_TYPE_3"
+)
```
ffromani, line 23: what's the benefit of moving here the constants?

internal/devices/devices.go
```diff
@@ -0,0 +1,23 @@
+/*
```
ffromani, line 1: the commit message is nice but there's no need to be so formal. The previous informal style was fine (and actually I liked it a bit more).

internal/noderesourcetopology/equality.go
```diff
@@ -89,6 +102,59 @@ func EqualResourceInfo(resInfoA, resInfoB nrtv1alpha2.ResourceInfo, isRebootTest
 	return true, nil
 }
 
+func EqualDevicesWithWarning(a, b nrtv1alpha2.ResourceInfo) (bool, error) {
```
ffromani, line 105: no need to use "WithWarning", code is free to call the logger

internal/noderesourcetopology/equality.go
```diff
@@ -74,6 +78,15 @@ func EqualResourceInfo(resInfoA, resInfoB nrtv1alpha2.ResourceInfo, isRebootTest
 		return EqualResourceInfoWithDeviation(resInfoA, resInfoB)
 	}
 
+	// TODO we almost never use real devices but in case we do we still pass the names using same env variables as for
+	// sample devices; need a way to distinguish between real vs sample devices used in the ci
+	if isRebootTest && resourceIsDevice(resInfoA.Name) {
+		// TODO: Feb 2025: NROP serial is using sample-device plugin which has a known issue that after operations like node reboot or taints update,
+		// quantity of the device reaources is messed up, so this is a workaround for reboot tests should be removed once
```
ffromani, line 85: typo: resources

test/e2e/serial/tests/configuration.go
```diff
@@ -98,6 +97,10 @@ type mcpInfo struct {
 	sampleNode    corev1.Node
 }
 
+func (i mcpInfo) ToString() string {
+	return fmt.Sprintf("name %q; config %q; sample node %q\n", i.mcpObj.Name, i.initialConfig, i.sampleNode.Name)
```
ffromani, line 104: Stringer functions, either `String()` or our custom `ToString()` lesser format should never add the trailing "\n"

test/e2e/serial/tests/configuration.go
```diff
@@ -144,18 +144,18 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			// nrts > 1 is checked in the setup
+			targetedNodeName, _ := nodesNameSet.PopAny()
+			targetedNode := &corev1.Node{}
+			err = fxt.Client.Get(context.TODO(), client.ObjectKey{Name: targetedNodeName}, targetedNode)
+			Expect(err).ToNot(HaveOccurred(), "failed to pull node %s object", targetedNodeName)
 
 			// we need to save a node that will still be associated to the initial mcps so later we conduct the MachineConfig check on it
-			initialMcpSampleNode := workers[len(workers)-targetIdx-1]
+			initialMCPNodeName, _ := nodesNameSet.PopAny()
+			initialMCPNode := &corev1.Node{}
```
ffromani, line 159: why a pointer? can be a local variable

test/e2e/serial/tests/configuration.go
```diff
@@ -144,18 +144,18 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
-				e2efixture.Skipf(fxt, "the test requires at least 2 worker nodes, found %d", len(workers))
-			}
-
-			targetIdx, ok := e2efixture.PickNodeIndex(workers)
-			Expect(ok).To(BeTrue())
-			targetedNode := workers[targetIdx]
+			nodesNameSet := e2enrt.AccumulateNames(nrts)
+			// nrts > 1 is checked in the setup
+			targetedNodeName, _ := nodesNameSet.PopAny()
+			targetedNode := &corev1.Node{}
```
ffromani, line 153: why a pointer? can be a local variable

test/e2e/serial/tests/tolerations.go
```diff
@@ -841,11 +841,18 @@ func sriovToleration() corev1.Toleration {
 }
 
 func waitForMcpUpdate(cli client.Client, ctx context.Context, mcpsInfo []mcpInfo, updateType MCPUpdateType) {
```
ffromani, line 843: this is a good idea but I'm puzzled by the implementation. We don't have `GinkgoHelper()` calls anymore, so failures should point to the specific test, should not they?

test/e2e/serial/tests/tolerations.go
```diff
@@ -841,11 +841,18 @@ func sriovToleration() corev1.Toleration {
...
+	waitForMcpUpdateWithID(cli, ctx, mcpsInfo, updateType, time.Now().String())
```
ffromani, line 844: unnecessary, internal functions can (and should) change API freely

test/internal/deploy/deploy.go
```diff
@@ -155,9 +155,9 @@ func WaitForMCPsCondition(cli client.Client, ctx context.Context, mcps []*machin
 			defer GinkgoRecover()
 			ts := time.Now()
 			err := wait.With(cli).Interval(interval).Timeout(timeout).ForMachineConfigPoolCondition(ctx, mcp, condition)
-			klog.Infof("MCP %q condition=%s err=%v after %v", mcp.Name, condition, err, time.Since(ts))
+			klog.InfoS("wait.ForMachineConfigPoolCondition result", "MCP", mcp.Name, "condition", condition, "err", err, "after", time.Since(ts))
 			return err
 		})
 	}
-	return eg.Wait()
+	return eg.Wait() // will block the flow until goroutines are finished or return on first error
```
ffromani, line 162: yes, this is well documented so this comment is redundant

test/e2e/serial/tests/configuration.go
```diff
@@ -1510,3 +1523,19 @@ func getLabelRoleWorker() string {
 func getLabelRoleMCPTest() string {
 	return fmt.Sprintf("%s/%s", depnodes.LabelRole, roleMCPTest)
 }
+
+func isCustomPolicySupportEnabled(nro *nropv1.NUMAResourcesOperator) bool {
+	GinkgoHelper()
```
ffromani, line 1524: so here we do want to call `GinkgoHelper`?

ffromani (commented):
mixed bag. Some commits are clear improvements, some other raises some question.
I can see the reason for all the changes, but some implementations can use some polishing, probably.

internal/devices/devices.go
shajmakh: if they were kept in `test/` we cannot use them in the `internal/` directory because of cycle imports because `test/` directory already imports `internal/`. to solve this we move the common consts in internal and use them in `test/` and internal helpers without issue. I believe it would help to have this change in separate commit. I'll split that.

internal/devices/devices.go
shajmakh: ack, good to know this feedback, I'll proceed with the informal style with the needed details :)

internal/devices/devices.go
ffromani: I've grown less and less entusiast of constants-only packages (and yes, I'm guilty too) but this makes sense.

test/e2e/serial/tests/configuration.go
```diff
@@ -205,10 +205,14 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
 				Expect(err).ToNot(HaveOccurred())
 
 			}()
-			//so far 0 machine count for mcp-test -> no nodes -> no updates status
+
+			//so far 0 machine count for mcp-test -> no nodes -> no updates status -> empty status
+			var updatedNewMcp *machineconfigv1.MachineConfigPool
+			Expect(fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(mcp), updatedNewMcp)).To(Succeed())
```
Tal-or: `updatedNewMcp` is `nil`, it should be passed as empty initialized struct.
```
updatedNewMcp := &machineconfigv1.MachineConfigPool{}
```

test/e2e/serial/tests/configuration.go
Tal-or: I presume it's because it passed to the client which accept a pointer.
it can be pass to the client using `&` but I saw it's a common usage to declare as a pointer.

test/e2e/serial/tests/configuration.go
```diff
@@ -144,18 +144,18 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
-			}
-
-			targetIdx, ok := e2efixture.PickNodeIndex(workers)
-			Expect(ok).To(BeTrue())
-			targetedNode := workers[targetIdx]
+			nodesNameSet := e2enrt.AccumulateNames(nrts)
+			// nrts > 1 is checked in the setup
+			targetedNodeName, _ := nodesNameSet.PopAny()
+			targetedNode := &corev1.Node{}
+			err = fxt.Client.Get(context.TODO(), client.ObjectKey{Name: targetedNodeName}, targetedNode)
```
Tal-or, line 154:
```
fxt.Client.Get(context.TODO(), client.ObjectKey{Name: targetedNodeName}, targetedNode).Should(Succeed(), "failed to pull node %s object", targetedNodeName)
```

test/e2e/serial/tests/configuration.go
```diff
@@ -144,18 +144,18 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			Expect(err).ToNot(HaveOccurred(), "failed to pull node %s object", targetedNodeName)
 
 			// we need to save a node that will still be associated to the initial mcps so later we conduct the MachineConfig check on it
-			initialMcpSampleNode := workers[len(workers)-targetIdx-1]
+			initialMCPNodeName, _ := nodesNameSet.PopAny()
+			initialMCPNode := &corev1.Node{}
+			err = fxt.Client.Get(context.TODO(), client.ObjectKey{Name: initialMCPNodeName}, initialMCPNode)
```
Tal-or, line 160:
```
fxt.Client.Get(context.TODO(), client.ObjectKey{Name: initialMCPNodeName}, initialMCPNode).Should(Succeeed(), "failed to pull node %s object", initialMCPNodeName)
```

test/e2e/serial/tests/configuration.go
```diff
@@ -1510,3 +1523,19 @@ func getLabelRoleWorker() string {
 func getLabelRoleMCPTest() string {
 	return fmt.Sprintf("%s/%s", depnodes.LabelRole, roleMCPTest)
 }
+
+func isCustomPolicySupportEnabled(nro *nropv1.NUMAResourcesOperator) bool {
+	GinkgoHelper()
+
+	const minCustomSupportingVString = "4.18"
```
Tal-or, line 1526: This commit is going to be merge to 4.19 and (probably) backport to 4.18, so why there's a need to check if this code is running against a version older than 4.18

Tal-or (commented):
Looks like a lot of good improvements, good work

test/e2e/serial/tests/configuration.go
```diff
@@ -98,6 +97,10 @@ type mcpInfo struct {
 	sampleNode    corev1.Node
 }
 
+func (i mcpInfo) ToString() string {
```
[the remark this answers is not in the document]
shajmakh: Thanks for your comment. I see the ToString() methods are indeed with pointers, while String() is not:
https://pkg.go.dev/fmt#Stringer
is this the common convention?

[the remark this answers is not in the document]
ffromani: In pretty much all the cases I prefer the extra clarity over the possible speedup given by passing a pointer. In our codebase we almost always use the values except in
```
api/v1/numaresourcesoperator_types.go:229:func (ngc *NodeGroupConfig) ToString() string {
api/v1/numaresourcesoperator_types.go:253:func (ng *NodeGroup) ToString() string {
```
which, if memory serves, was a combination of laziness (the API types embed by pointer) and historical accident.

My take is that value is always preferred over pointer unless clearly proven beneficial performance-wise

test/e2e/serial/tests/configuration.go
ffromani: the golang escape analysis is probably doing a good job here, but I'd prefer in general to bias towards local variables.

test/e2e/serial/tests/configuration.go
shajmakh: I prefered the 2 lines version specially in cases that the one line version turns out to be long (I have to scroll to the side to see the error message)

test/e2e/serial/tests/configuration.go
shajmakh: right, this was fixed in later commit but should be working starting from this one, fixed. Thanks!

test/e2e/serial/tests/configuration.go
ffromani: lack in consistency in our codebase is a (minor) concern, but better to be addressed by a sweeping PR. Both versions are fine to me.

test/e2e/serial/tests/configuration.go
shajmakh: local var should be fine here as well, we'll just likely see more of the &<var> around

test/e2e/serial/tests/configuration.go
shajmakh: ack

test/e2e/serial/tests/tolerations.go
shajmakh: without the GinkgoHelper() call in a "helper" function, one will be able to see the exact failing line in the nested function and not the line of where this function is being called. 
IOW if inside a spec there is a call for a helper function in line 5, and that helper function has GinkgoHelper(); if the function fails on one of its assertions say line 91 (let's assume it's in the same file), the report will print out that the spec has failed in line 5. Whereas if GinkgoHelper() was to called, the report will print out that the spec has failed in line 91.

test/e2e/serial/tests/configuration.go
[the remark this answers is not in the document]
Tal-or: The problems begin when u need a receiver function to modify the value that its receiver points to.
Then you end up with mix of value and pointer receiver which is a bad practice.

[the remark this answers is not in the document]
ffromani: This is very true (see: https://go.dev/doc/faq#methods_on_values_or_pointers).
I guess the key decision factor is if we have a pure value object, or a object which never is mutated through methods. If we have at least a mutating method all the receivers should be pointers (and changing as that for consistency is fine and welcome).

My brain is wired in a weird way such as I'm not particularly bothered by mixed methods (some pointer receivers, some not) because I interpret these cases as applications of principle of least privilege, so it looks quite legit to me. And so I may have missed these occurrences. We should go and fix them (in different PRs obviously).

In general I don't think we should default to pointer receivers. It is something that can be done on a per-type basis. Of course related types should preferably be consistent.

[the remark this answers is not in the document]
ffromani: Uhm, digging more there is a good point about cases on which copies are dangerous:
https://go.dev/wiki/CodeReviewComments#receiver-type

I probably need to reverse my default bias, even though for different reasons than "performance!"

test/e2e/serial/tests/configuration.go
shajmakh: because we use and maintain same test image for all versions, so this code will also run on < 4.18 versions, and we do not backport changes done to the serial suite

Tal-or: What test image? 
I thought that we are checking out the code according to the cluster version and then run the appropriate tests.

Tal-or: It would be problematic to run tests adjusted for 4.19 on 4.17 because there are significant changes between the two. for example we not longer have MachineConfig

test/e2e/serial/tests/configuration.go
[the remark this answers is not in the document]
ffromani: more like because of `Finally, when in doubt, use a pointer receiver.`

test/e2e/serial/tests/configuration.go
shajmakh: the main purpose of GinkgoHelper() is to avoid using offset() for location calculation of the running code. It is especially helpful to be called on functions that are highly reused or are called several times in one spec. In that case it would be more clear to debug failures to get the location of the failing call rather than the failing line within the helper function. 
Considering this helper function is short and well defined I didn't mind to add the GinkgoHelper() call here. If this causes confusion we can remove it and address all that whenever the function is highly requested, then we can propose a cleanup PR to add this call if needed.

ffromani: I'm not completely sure I follow the logic, but this specific application makes sense to me, so let's move on

ffromani:
please rebase before you resubmit

test/e2e/serial/tests/configuration.go
ffromani: thing is: the problem I'm seeing (and which prompted the previous conversation?) was related to bad helpers which had too many expectations inside them and whose failure was hard to debug. This is still kinda related to the helper length, but the key discriminatror should be number and likeness of failure of expectations inside the helper rather than the helper length itself. Anyway, overall makes sense to me so we can discuss nuances later on.

test/e2e/serial/tests/configuration.go
Tal-or: but if you're checking the code to release-4.17, this piece of code won't be shown there.
What am I missing?

shajmakh: @Tal-or this was implemented u/s eventually to and one time d/s as it turned out the best approach to avoid maintenance load as much as possible, check this for the details: 
https://github.com/openshift-kni/numaresources-operator/tree/main/doc/features

ffromani: `minCustomSupportingVString` is hardcoded indeed, but `configuration.PlatVersion` is autodetected from the cluster the suite is running against

shajmakh: > but if you're checking the branch to release-4.17, this piece of code won't be shown there.
> What am I missing?

This piece of code will run on all releases and is shipped in our latest quay tests' image. The tl;dr is that the operator controller pod for every release exposes which features are supported per the running version; d/s we scan the serial tests for all the tests that are tagged with the supported feature, that way we run only supported tests. Active features per branch can be found here: 
https://github.com/openshift-kni/numaresources-operator/blob/main/internal/api/features/_topics.json
@Tal-or let me know if you have additional concerns.

test/e2e/serial/tests/configuration.go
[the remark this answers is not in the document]
shajmakh: This is a good and interesting discussion, thank you both for the references and for your input. I believe mcpInfo needs also a cleanup but will have this in another followup PR. 
/hold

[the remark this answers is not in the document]
ffromani: yeah we went a bit overboard. I think what you did is fine and we don't need any further change in this PR or in the short future.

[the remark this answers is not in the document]
shajmakh: Good to learn about this and it does makes sense eventually. From what I learned, considering the mcpInfo struct is small, using a receiver value is not costly. And most importantly ToString doesn't mutate any of the struct values even the pointer var fr the MCP object (we don't normalize the object and set defaults). Also No other methods are defined for this type because initially we implemented it just to have an easier way to pass several related parameters to the waitForUpdate function. I think accurences of ToString() and String() that use a pointer receiver are there due to consistency if the type have other methods with pointer receivers. 

> For types such as basic types, slices, and small structs, a value receiver is very cheap so unless the semantics of the method requires a pointer, a value receiver is efficient and clear.
For this case I'll keep a value receiver considering the above. 

I only missed checking the pointer of the mcp object is not nil so pushing again with fixed version of the body.

PR #1196, opened 2025-02-26
serial: rewrite test 47674

shajmakh:
The previous version was fragile and maintained several times already.
This fact means that the test should be rewritten and once the known bugs
are fixed, new tests will be added to cover the e2e full scenario.

The test goal is to verify that adjustments to NROP CR are done as
expected and reflected in the RTE daemonsets. Previously, this was done
via creating a new MCP for the test and associating to it one of the
worker nodes that were part of the initial NROP node groups. This was
causing a reboot when running on a cluster that is:
a. < 4.18 or
b. >= 4.18 and has the custom RTE selinux policy enabled

In the new version of the test, we keep checking the same functionality
while completely avoiding reboot. That is acheived by creating the new MCP,
keeping it empty (no nodes are associated to it) and appending it as a
second node group's selector. Then checks on the daemonsets are
performed and the only cleanup would be deleting the test mcp.

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -44,7 +44,6 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	depnodes "github.com/k8stopologyawareschedwg/deployer/pkg/clientutil/nodes"
-	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
 	nrtv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -54,7 +53,6 @@ import (
 
 	nropv1 "github.com/openshift-kni/numaresources-operator/api/v1"
 	"github.com/openshift-kni/numaresources-operator/api/v1/helper/namespacedname"
-	inthelper "github.com/openshift-kni/numaresources-operator/internal/api/annotations/helper"
 	nropmcp "github.com/openshift-kni/numaresources-operator/internal/machineconfigpools"
 	intnrt "github.com/openshift-kni/numaresources-operator/internal/noderesourcetopology"
 	intobjs "github.com/openshift-kni/numaresources-operator/internal/objects"
@@ -68,9 +66,7 @@ import (
 	"github.com/openshift-kni/numaresources-operator/pkg/validation"
 	rteconfig "github.com/openshift-kni/numaresources-operator/rte/pkg/config"
 	"github.com/openshift-kni/numaresources-operator/test/e2e/label"
-	e2eclient "github.com/openshift-kni/numaresources-operator/test/internal/clients"
 	"github.com/openshift-kni/numaresources-operator/test/internal/configuration"
-	"github.com/openshift-kni/numaresources-operator/test/internal/deploy"
 	e2efixture "github.com/openshift-kni/numaresources-operator/test/internal/fixture"
 	"github.com/openshift-kni/numaresources-operator/test/internal/images"
 	e2enrt "github.com/openshift-kni/numaresources-operator/test/internal/noderesourcetopologies"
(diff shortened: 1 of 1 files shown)


test/e2e/serial/tests/configuration.go
```diff
@@ -356,6 +360,166 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
 
 		})
 
+		It("[test_id:47674-V2][images][tier2] should be able to modify the configurable values under the NUMAResourcesOperator CR", Label("images", "tier2"), func(ctx context.Context) {
+			fxt.IsRebootTest = true
+			nroOperObj := &nropv1.NUMAResourcesOperator{}
```
ffromani, line 365: prefer local variables (not pointers) when feasible

test/e2e/serial/tests/configuration.go
```diff
@@ -356,6 +360,166 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			nroKey := objects.NROObjectKey()
+			Expect(fxt.Client.Get(ctx, nroKey, nroOperObj)).ToNot(HaveOccurred())
```
ffromani, line 367: better: 
```
Expect(fxt.Client.Get(ctx, nroKey, nroOperObj)).To(Succeed())
```

test/e2e/serial/tests/configuration.go
```diff
@@ -356,6 +360,166 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			initialNroOperObj := nroOperObj.DeepCopy()
+
+			mcps := &mcov1.MachineConfigPoolList{}
+			Expect(fxt.Client.List(ctx, mcps)).ToNot(HaveOccurred())
```
ffromani, line 371: ditto

test/e2e/serial/tests/configuration.go
```diff
@@ -356,6 +360,166 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+
+			if len(mcps.Items) < 2 {
+				e2efixture.Skip(fxt, "the test requires at least 2 MCPs")
+			}
+
+			nropMcps, err := nropmcp.GetListByNodeGroupsV1(context.TODO(), e2eclient.Client, nroOperObj.Spec.NodeGroups)
```
ffromani, line 377: why not using the top-level `ctx`?

test/e2e/serial/tests/configuration.go
```diff
@@ -356,6 +360,166 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+				}
+			}
+
+			customPolicySupportEnabled := isCustomPolicySupportEnabled(nroOperObj)
+
+			By(fmt.Sprintf("modifying the NUMAResourcesOperator nodeGroups field to match different mcp: %q labels %q", testMcp.Name, testMcp.Labels))
+			Eventually(func(g Gomega) {
+				// we need that for the current ResourceVersion
+				err := fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(initialNroOperObj), nroOperObj)
+				g.Expect(err).ToNot(HaveOccurred())
```
ffromani, line 398:
```
g.Expect(fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(initialNroOperObj), nroOperObj).To(Succeed())
```
?

test/e2e/serial/tests/configuration.go
```diff
@@ -356,6 +360,166 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+
+				// as filtered at the beginning of the test, it supports single node group
+				nroOperObj.Spec.NodeGroups[0].MachineConfigPoolSelector.MatchLabels = testMcp.Labels
+				err = fxt.Client.Update(context.TODO(), nroOperObj)
+				g.Expect(err).ToNot(HaveOccurred())
```
ffromani, line 403: likewise

test/e2e/serial/tests/configuration.go
```diff
@@ -356,6 +360,166 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+				By("CLEANUP: reverting the changes under the NUMAResourcesOperator object")
+				nroOperObj := &nropv1.NUMAResourcesOperator{}
+				Eventually(func(g Gomega) {
+					// we need that for the current ResourceVersion
+					err := fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(initialNroOperObj), nroOperObj)
+					g.Expect(err).ToNot(HaveOccurred())
+
+					nroOperObj.Spec = initialNroOperObj.Spec
+					err = fxt.Client.Update(context.TODO(), nroOperObj)
+					g.Expect(err).ToNot(HaveOccurred())
```
ffromani, line 416: likewise

test/e2e/serial/tests/configuration.go
```diff
@@ -356,6 +360,166 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+				}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(Succeed())
+
+				if customPolicySupportEnabled {
+					By(fmt.Sprintf("waiting for mcp %q to start updating", initialMcp.Name))
+					Expect(deploy.WaitForMCPsCondition(fxt.Client, context.TODO(), []*machineconfigv1.MachineConfigPool{initialMcp}, machineconfigv1.MachineConfigPoolUpdating)).To(Succeed())
+					By(fmt.Sprintf("waiting for mcp %q to complete update", initialMcp.Name))
+					Expect(deploy.WaitForMCPsCondition(fxt.Client, context.TODO(), []*machineconfigv1.MachineConfigPool{initialMcp}, machineconfigv1.MachineConfigPoolUpdated)).To(Succeed())
```
ffromani, line 423: this proven to be fragile, it seems better to wait for the MCP to become Updated, but with a different `ResourceVersion`.

test/e2e/serial/tests/configuration.go
```diff
@@ -356,6 +360,166 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+				}
+			}() //end of defer
+
+			if customPolicySupportEnabled {
+				By(fmt.Sprintf("waiting for mcp %q to start updating", initialMcp.Name))
+				Expect(deploy.WaitForMCPsCondition(fxt.Client, context.TODO(), []*machineconfigv1.MachineConfigPool{initialMcp}, machineconfigv1.MachineConfigPoolUpdating)).To(Succeed())
+				By(fmt.Sprintf("waiting for mcp %q to complete update", initialMcp.Name))
+				Expect(deploy.WaitForMCPsCondition(fxt.Client, context.TODO(), []*machineconfigv1.MachineConfigPool{initialMcp}, machineconfigv1.MachineConfigPoolUpdated)).To(Succeed())
```
ffromani, line 431: ditto

test/e2e/serial/tests/configuration.go
```diff
@@ -356,6 +360,166 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			}
+
+			By(fmt.Sprintf("Verify RTE daemonsets have the updated node selector matching to the new mcp %q", testMcp.Name))
+			Eventually(func() (bool, error) {
```
ffromani, line 435: do we ever return `err != nil`?

ffromani (commented):
I'm generally in favor of this rewrite. Partial review.

test/e2e/serial/tests/configuration.go
ffromani: indeed: AFAIK the safest way to wait is to get BOTH `resourceVersion` > current and state == Updated

shajmakh:
/hold
the bug turned out to be an MCO one, and reproduces also on 4.18. 
To overcome this we need to find MCP config that wouldn't trigger a reboot

test/e2e/serial/tests/configuration.go
```diff
@@ -381,7 +382,102 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
 				}
 				return true, nil
 			}).WithTimeout(5*time.Minute).WithPolling(10*time.Second).Should(BeTrue(), "failed to update RTE container with LogLevel %q", operatorv1.Trace)
+		})
+
+		It("[replaces-47674][images][tier2] should be able to modify the configurable values under the NUMAResourcesOperator CR", Label("images", "tier2"), func(ctx context.Context) {
```
Tal-or: Although #1170 has not been merged, we can drop the obsolete tags.

test/e2e/serial/tests/configuration.go
```diff
@@ -381,7 +382,102 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+				err := wait.With(fxt.Client).
+					Interval(configuration.MachineConfigPoolUpdateInterval).
+					Timeout(configuration.MachineConfigPoolUpdateTimeout).
+					ForMachineConfigPoolDeleted(context.TODO(), testMCP)
+				Expect(err).ToNot(HaveOccurred())
+			}()
+			// keep the mcp with zero machine count intentionally to avoid reboots
+
+			testNG := nropv1.NodeGroup{
+				MachineConfigPoolSelector: &metav1.LabelSelector{
```
Tal-or, line 177: Can we use `poolName` instead? since this is the new API we would like to adopt and it's common for both OCP and HCP

Tal-or (commented):
If the new MCP's nodeSelector is not associate with any of the nodes, then why do we need it in the test?

Is the fact that we're adding it under the nodeGroup is enough to make NROP to create new RTE pods?

test/e2e/serial/tests/configuration.go
shajmakh: this is reinnovation for an old test that is designed to run on all versions, if we use poolName we need to label this with `ngpoolname` to run it only on 4.18 and avoid it failing on <4.18. Using poolName will cause to skip running the test on < 4.18 because we use one version of tests image

test/e2e/serial/tests/configuration.go
shajmakh: can you specify what tags are obsolete here please?

Tal-or: `[images][tier2]`

test/e2e/serial/tests/configuration.go
ffromani: ok, so we want to use the MCP to make sure this test can run against older version of NROP. I'm fine with that.

test/e2e/serial/tests/configuration.go
```diff
@@ -143,245 +140,100 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
-			waitForMcpUpdate(fxt.Client, context.TODO(), MachineConfig, time.Now().String(), newMcpInfo)
 
-			By(fmt.Sprintf("modifying the NUMAResourcesOperator nodeGroups field to match new mcp: %q labels %q", mcp.Name, mcp.Labels))
+			By(fmt.Sprintf("modifying the NUMAResourcesOperator nodeGroups field to include new group: %q labels %q", testMCP.Name, testMCP.Labels))
+			newLogLevel := operatorv1.Trace
 			Eventually(func(g Gomega) {
 				// we need that for the current ResourceVersion
-				err := fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(initialNroOperObj), nroOperObj)
-				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(fxt.Client.Get(ctx, client.ObjectKeyFromObject(initialNroOperObj), &nroOperObj)).To(Succeed())
```
ffromani, line 186: it would be better to use a local variable to perform the update, but this also works

test/e2e/serial/tests/configuration.go
Tal-or: Yea, probably no other choice

ffromani (commented):
/approve

I like the new rewrite more than the old code

ffromani (commented):
/lgtm

test/e2e/serial/tests/configuration.go
shajmakh: yeah this didn't show any issue and the idea was to reuse the vars as much as relevant

ffromani: reusing vars is almost always a bad idea though

shajmakh: I agree, in this case though the reuse is completely overriding the old value so no risk in using old data or something like that

PR #1138, opened 2025-01-01
NO-JIRA: govulncheck ci

Tal-or:
Add a new CI lane for checking go code vulnerabilities.

Signed-off-by: Talor Itzhak <titzhak@redhat.com>

PR #1200, opened 2025-02-28
deps: bump x/oauth2

ffromani:
bump package dependency. Prompted by https://pkg.go.dev/vuln/GO-2025-3488 but we are not affected by this CVE (per scan and inspection: there's no `jws.go` in our vendor tree)

Bump the dep just in case, for defence in depth.


ffromani:
> Thanks for this update. do you think adding the vulnerability check as an optional CI lane might be a good idea to identify the CVEs as early as possible?

it's already there: https://github.com/openshift-kni/numaresources-operator/pull/1138

PR #1202, opened 2025-02-28
[release-0.17] deps bump: glog, x/oauth2

ffromani:
bump deps to proactive remove possible vulnerabilities. Note in both cases our production code is NOT AFFECTED ( we either don't even include the code in our vendor tree or not use in production code, but in test code).

Still, bump deps to be proactive and make the risk exactly zero

glog   -> Vulnerability GO-2025-3372
oauth2 -> Vulnerability GO-2025-3488


swatisehgal (commented):
/lgtm

PR #1206, opened 2025-03-03
[release-4.15] deps: bump x/oauth2 to 0.26

ffromani:
This was prompted by GO-2025-3488 but we are NOT AFFECTED by that CVE because we don't use, thus we don't vendor, the relevant code at all.

Still, we bump the dep to the last version which is compatible with the golang version we use in this branch, to be proactive.


swatisehgal (commented):
/lgtm

PR #1211, opened 2025-03-04
bump deps: glog, grpc, x/oauth2

ffromani:
bump deps to proactive remove possible vulnerabilities. Note our production code is NOT AFFECTED by none of the listed vulnerabilities, because we do not use in production code.
Some code (glog, grpc) is used by a tool we have to support testing - so not meant to be anywhere near production sites; the other dep is used as package, but we don't physically pull the affected file pertaining to a functionality we don't need.

glog	-> 1.2.4	related to: GO-2025-3372
oauth2	-> 0.26		related to: GO-2025-3488
gprc	-> 1.56.3	related to: GO-2023-2153

Still, we bump the deps to the last version which is compatible with the golang version we use in this branch, to be proactive.


swatisehgal (commented):
/lgtm

PR #1213, opened 2025-03-06
api: allow to skip reconciliation of NRT CRD 

ffromani:
Add custom annotation to stop reconciling the NRT API.
This annotation is semi-public and will be used strictly on a case-by-case way if there are multiple operators which
claim ownership of the NRT CRD.

With this annotation, NROP goes above and beyond and trusts blindly whatever NRT CRD is found in the cluster instead of owning the API.

Stop owning the NRT CRD greatly weakens the stability of the NUMA-aware scheduling and it is dangerous and definitely NOT recommended. But can be necessary in circumstances on which other operator claim ownership of the CRD, and we can't reconcile the differences.

### DIFF
--- a/internal/api/annotations/annotations.go
+++ b/internal/api/annotations/annotations.go
@@ -26,6 +26,9 @@ const (
 
 	PauseReconciliationAnnotation        = "config.numa-operator.openshift.io/pause-reconciliation"
 	PauseReconciliationAnnotationEnabled = "enabled"
+
+	NRTAPIDefinitionAnnotation = "config.numa-operator.openshift.io/nrt-api-definition"
+	NRTAPIFromCluster          = "cluster" // trust whatever it is already in the cluster, if at all
 )
 
 func IsCustomPolicyEnabled(annot map[string]string) bool {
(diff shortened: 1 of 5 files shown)


ffromani:
/hold

WIP

ffromani:
/hold cancel

ready for review

Tal-or (commented):
Codewise LGTM

I would understand this rational if this operator was a mandatory one.
But if the user want to manage some of the resources on its own, it should not install this operator at all, but helm/kustomize or any other solution that allow greater versatility wrt objects management.

ffromani:
> Codewise LGTM
> 
> I would understand this rational if this operator was a mandatory one. But if the user want to manage some of the resources on its own, it should not install this operator at all, but helm/kustomize or any other solution that allow greater versatility wrt objects management.

Yes, you are 100% right. This is why intended on our side and designed accordingly on our side. We go as far as installing the CRD conditionally, vs installing it unconditionally with OLM. Thing is, for this to work _all_ operators in the cluster need to be cooperative.

Tal-or:
> Thing is, for this to work _all_ operators in the cluster need to be cooperative.

What does it means?

Tal-or (approved):
Agreed to proceed on this direction after discussion offline.

PR #1215, opened 2025-03-07
[release-4.12] deps: x/crypto: downgrade to 0.33

ffromani:
Building x/crypto with golang 1.19.z yields
```
golang.org/x/crypto/pkcs12/internal/rc2: cannot compile Go 1.23 code
```

So move back to x/crypto 0.33


ffromani:
discovered locally. I wonder why CI didn't complain :\

PR #1216, opened 2025-03-10
OCPBUGS-51261:controller: track rte configmap state

Tal-or:
Kubelet controller is responsible to track `KubeletConfig` objects and
generate a `ConfigMap` data from them that will be served to RTE.

NROP controller is responsible to track the `ConfigMap` creation,
feed it to RTE and restart RTE in case the `ConfigMap` data has changed.

The latter is broken, and this PR is fixing it.

In addition the `ConfigMap` should be per `DaemonSet`/`NodeGroup`, while
previously it was common at the CR level which is wrong.

Because cluster might have multiple node groups which are configured
slightly different.

### DIFF
--- a/internal/controller/numaresourcesoperator_controller.go
+++ b/internal/controller/numaresourcesoperator_controller.go
@@ -58,7 +58,6 @@ import (
 	"github.com/openshift-kni/numaresources-operator/internal/dangling"
 	"github.com/openshift-kni/numaresources-operator/internal/relatedobjects"
 	"github.com/openshift-kni/numaresources-operator/pkg/apply"
-	"github.com/openshift-kni/numaresources-operator/pkg/hash"
 	"github.com/openshift-kni/numaresources-operator/pkg/images"
 	"github.com/openshift-kni/numaresources-operator/pkg/loglevel"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectnames"
@@ -535,15 +534,6 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 		return dsPoolPairs, err
 	}
 
-	// ConfigMap should be provided by the kubeletconfig reconciliation loop
-	if r.RTEManifests.Core.ConfigMap != nil {
-		cmHash, err := hash.ComputeCurrentConfigMap(ctx, r.Client, r.RTEManifests.Core.ConfigMap)
-		if err != nil {
-			return dsPoolPairs, err
-		}
-		rteupdate.DaemonSetHashAnnotation(r.RTEManifests.Core.DaemonSet, cmHash)
-	}
-
 	rteupdate.SecurityContextConstraint(r.RTEManifests.Core.SecurityContextConstraint, true) // force to legacy context
 	// SCC v2 needs no updates
 
@@ -753,6 +743,8 @@ func daemonsetUpdater(poolName string, gdm *rtestate.GeneratedDesiredManifest) e
 			klog.V(5).InfoS("DaemonSet update: selinux options", "container", manifests.ContainerNameRTE, "context", selinux.RTEContextType)
 		}
 	}
+	// it's possible that the hash will be empty if kubelet controller hasn't created a configmap
+	rteupdate.DaemonSetHashAnnotation(gdm.DaemonSet, gdm.RTEConfigHash)
 	return nil
 }
 
--- a/pkg/objectstate/rte/machineconfigpool.go
+++ b/pkg/objectstate/rte/machineconfigpool.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
@@ -28,6 +29,7 @@ import (
 	nropv1 "github.com/openshift-kni/numaresources-operator/api/v1"
 	nodegroupv1 "github.com/openshift-kni/numaresources-operator/api/v1/helper/nodegroup"
 	"github.com/openshift-kni/numaresources-operator/internal/api/annotations"
+	"github.com/openshift-kni/numaresources-operator/pkg/hash"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectnames"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectstate"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectstate/compare"
@@ -56,6 +58,12 @@ func (obj machineConfigPoolFinder) UpdateFromClient(ctx context.Context, cli cli
 		if dsm.daemonSetError = cli.Get(ctx, key, ds); dsm.daemonSetError == nil {
 			dsm.daemonSet = ds
 		}
+
+		cm := &corev1.ConfigMap{}
+		if err := cli.Get(ctx, key, cm); err == nil {
+			// we're storing the updated hash only in the case that kubelet controller created a configmap
+			dsm.rteConfigHash = hash.ConfigMapData(cm)
+		}
 		obj.em.daemonSets[generatedName] = dsm
 
 		mcName := objectnames.GetMachineConfigName(obj.instance.Name, mcp.Name)
@@ -76,12 +84,14 @@ func (obj machineConfigPoolFinder) FindState(mf Manifests, tree nodegroupv1.Tree
 	for _, mcp := range tree.MachineConfigPools {
 		var existingDs client.Object
 		var loadError error
+		var rteComputedConfigHash string
 
 		generatedName := objectnames.GetComponentName(obj.instance.Name, mcp.Name)
 		existingDaemonSet, ok := obj.em.daemonSets[generatedName]
 		if ok {
 			existingDs = existingDaemonSet.daemonSet
 			loadError = existingDaemonSet.daemonSetError
+			rteComputedConfigHash = existingDaemonSet.rteConfigHash
 		} else {
 			loadError = fmt.Errorf("failed to find daemon set %s/%s", mf.Core.DaemonSet.Namespace, mf.Core.DaemonSet.Name)
 		}
@@ -101,6 +111,7 @@ func (obj machineConfigPoolFinder) FindState(mf Manifests, tree nodegroupv1.Tree
 			MachineConfigPool:     mcp.DeepCopy(),
 			NodeGroup:             tree.NodeGroup.DeepCopy(),
 			DaemonSet:             desiredDaemonSet,
+			RTEConfigHash:         rteComputedConfigHash,
 			IsCustomPolicyEnabled: annotations.IsCustomPolicyEnabled(tree.NodeGroup.Annotations),
 		}
 
--- a/pkg/objectstate/rte/nodegroup.go
+++ b/pkg/objectstate/rte/nodegroup.go
@@ -21,11 +21,13 @@ import (
 	"fmt"
 
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	nropv1 "github.com/openshift-kni/numaresources-operator/api/v1"
 	nodegroupv1 "github.com/openshift-kni/numaresources-operator/api/v1/helper/nodegroup"
 	"github.com/openshift-kni/numaresources-operator/internal/api/annotations"
+	"github.com/openshift-kni/numaresources-operator/pkg/hash"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectnames"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectstate"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectstate/compare"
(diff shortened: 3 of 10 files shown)


ffromani:
conceptually LGTM. Please fix CI errors

pkg/objectstate/rte/rte.go
```diff
@@ -407,3 +416,20 @@ func getObject(ctx context.Context, cli client.Client, key client.ObjectKey, obj
 	*err = cli.Get(ctx, key, obj)
 	return *err == nil
 }
+
+func findConfigMapState(name, namespace string, configMaps map[string]configMapManifest) *corev1.ConfigMap {
+	var desiredConfigMap *corev1.ConfigMap
+	existingConfigMap, ok := configMaps[name]
+	errFormat := fmt.Sprintf("failed to find config map %s/%s", namespace, name)
```
ffromani, line 423: it seems we can log this in few flows, but all the flows should be mutually exclusive. So I'd avoid this and inline the formatting in the logging sites.

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -753,6 +744,11 @@ func daemonsetUpdater(poolName string, gdm *rtestate.GeneratedDesiredManifest) e
 			klog.V(5).InfoS("DaemonSet update: selinux options", "container", manifests.ContainerNameRTE, "context", selinux.RTEContextType)
 		}
 	}
+	if gdm.ConfigMap != nil {
```
ffromani, line 747: on which cases we do expect `gdm.ConfigMap` to be nil?

ffromani: ok, makes sense

pkg/objectstate/rte/nodegroup.go
```diff
@@ -80,11 +89,27 @@ func (obj nodeGroupFinder) FindState(mf Manifests, tree nodegroupv1.Tree) []obje
...
+	// a case where kubelet controller is not creating the ConfigMap, is undesired but yet possible.
+	// this is why the operator should log here
+	// and not bubble up the error that would interfere with its normal operation.
+	switch {
+	case !ok:
+		klog.ErrorS(errFormat, "")
+	case existingConfigMap.configMapError != nil:
+		klog.ErrorS(existingConfigMap.configMapError, errFormat.Error())
+	default:
+		desiredConfigMap = existingConfigMap.configMap
```
ffromani, line 104: do we need a DeepCopy here?

pkg/objectstate/rte/machineconfigpool.go
```diff
@@ -96,11 +105,27 @@ func (obj machineConfigPoolFinder) FindState(mf Manifests, tree nodegroupv1.Tree
 			updateError = fmt.Errorf("the machine config pool %q does not have node selector", mcp.Name)
 		}
 
+		var desiredConfigMap *corev1.ConfigMap
+		existingConfigMap, ok := obj.em.configMaps[generatedName]
+		errFormat := fmt.Errorf("failed to find config map %s/%s", mf.Core.DaemonSet.Namespace, generatedName)
```
ffromani, line 110: please inline in the calling site

pkg/objectstate/rte/nodegroup.go
```diff
@@ -80,11 +89,27 @@ func (obj nodeGroupFinder) FindState(mf Manifests, tree nodegroupv1.Tree) []obje
 		HyperShiftNodePoolLabel: poolName,
 	}
 
+	var desiredConfigMap *corev1.ConfigMap
+	existingConfigMap, ok := obj.em.configMaps[generatedName]
+	errFormat := fmt.Errorf("failed to find config map %s/%s", mf.Core.DaemonSet.Namespace, generatedName)
```
ffromani, line 94: please inline in the calling site

ffromani:
this work should probably be backported down to 4.16

pkg/objectstate/rte/nodegroup.go
Tal-or: No. We're not mutating the config map.
All the controller is doing is exercise a hash function on the data

Tal-or:
> so `ConfigMap` is already an exception in the context of this controller. We don't own it, we don't reconcile it, we "just" need to read it. So, idea: why don't we read it _and_ store its hash in the internal data structures vs storing the object and computing when needed? this
> 
> 1. will further reinforce the exceptional treatment
> 2. will remove any ambiguity about ownerhip, mutation or not

Sure this is an option. I used the above approach simply because it was aligned with the existing architecture

ffromani:
> > so `ConfigMap` is already an exception in the context of this controller. We don't own it, we don't reconcile it, we "just" need to read it. So, idea: why don't we read it _and_ store its hash in the internal data structures vs storing the object and computing when needed? this
> > 
> > 1. will further reinforce the exceptional treatment
> > 2. will remove any ambiguity about ownerhip, mutation or not
> 
> Sure this is an option. I used the above approach simply because it was aligned with the existing architecture

thanks, looks better IMO and almost ready to go (gonna review shortly).
My general rule is: code which behaves the same (or kinda the same) should look the same (or kinda the same). Code which behaves different can and should look different. So in this case looking different is a feature IMO.

pkg/objectstate/rte/nodegroup.go
```diff
@@ -74,6 +82,8 @@ func (obj nodeGroupFinder) FindState(mf Manifests, tree nodegroupv1.Tree) []obje
 
 	desiredDaemonSet := mf.Core.DaemonSet.DeepCopy()
 	desiredDaemonSet.Name = generatedName
+	// it's possible that the hash will be empty if kubelet controller hasn't created a configmap
+	rteupdate.DaemonSetHashAnnotation(desiredDaemonSet, existingDaemonSet.rteComputedConfigHash)
```
ffromani, line 86: I'm thinking aloud. Should we set it to "" (empty string) or should we set it only if `!= ""`? What are the pros and cons in both scenarios?

ffromani: do we log when we change annotation? we probably should (V>=4)

Tal-or: if kubeletconfig has been completely deleted from the cluster, we should be able to signal that correctly.
setting the annotation value to `""` gives us what we need.
It will cause RTE pods to restart and they'll reflect the right behavior as in the case of no kubeletconfig configuration at all.

ffromani: good point, it's worthy to keep it. Please do *NOT* resubmit just to add logs

pkg/objectstate/rte/machineconfigpool.go
```diff
@@ -21,17 +21,20 @@ import (
...
 
 	nropv1 "github.com/openshift-kni/numaresources-operator/api/v1"
 	nodegroupv1 "github.com/openshift-kni/numaresources-operator/api/v1/helper/nodegroup"
 	"github.com/openshift-kni/numaresources-operator/internal/api/annotations"
+	"github.com/openshift-kni/numaresources-operator/pkg/hash"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectnames"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectstate"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectstate/compare"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectstate/merge"
+	rteupdate "github.com/openshift-kni/numaresources-operator/pkg/objectupdate/rte"
```
ffromani, line 37: ok, this is the only remaining item (extra logging is a nice bonus but not required).
Conceptually, `objectstate` sits below in the stack with respect to `objectupdate`:

```
controller
objectupdate
objectstate
client
```
IOW, objectstate can't import objectupdate, but objectupdate can call objectstate (controller can and do import them directly).
We can bypass layers from up to down, but we should never break layering from the bottom to the up

ffromani (commented):
ok, last comment inline, but is worthy. If you add the logs (V>=4) we were talking about it's a nice bonus.

pkg/objectstate/rte/machineconfigpool.go
Tal-or: And that's why I preferred passing the configmap, because it is more align with current flow. Anyway I'll fix that, I'll find a way to update the ds later in the code

ffromani: I may be wrong but  I think there should be a reasonnably easy way to not break the layering and yet only carry to hash around

ffromani (commented):
the main problem here is we push updates in the state extraction code which was a bad idea in the first place back in time `daemonsetUpdater`. It should have been cleaner since the beginning.

Now's probably late to clean the board - we should wait 4.20 for it, if ever, so let's try to have a more targeted fix.

ffromani (commented):
something like this should fit more the existing (very debatable, see #1217) layout:
```
diff --git a/internal/controller/numaresourcesoperator_controller.go b/internal/controller/numaresourcesoperator_controller.go
index a10433bc..c44e8c21 100644
--- a/internal/controller/numaresourcesoperator_controller.go
+++ b/internal/controller/numaresourcesoperator_controller.go
@@ -758,6 +758,9 @@ func daemonsetUpdater(poolName string, gdm *rtestate.GeneratedDesiredManifest) e
 			klog.V(5).InfoS("DaemonSet update: selinux options", "container", manifests.ContainerNameRTE, "context", selinux.RTEContextType)
 		}
 	}
+
+	// set hash here
+
 	return nil
 }
 
diff --git a/pkg/objectstate/rte/machineconfigpool.go b/pkg/objectstate/rte/machineconfigpool.go
index 618319db..fa0a9140 100644
--- a/pkg/objectstate/rte/machineconfigpool.go
+++ b/pkg/objectstate/rte/machineconfigpool.go
@@ -75,12 +75,14 @@ func (obj machineConfigPoolFinder) FindState(mf Manifests, tree nodegroupv1.Tree
 	var ret []objectstate.ObjectState
 	for _, mcp := range tree.MachineConfigPools {
 		var existingDs client.Object
+		var rteConfigHash string
 		var loadError error
 
 		generatedName := objectnames.GetComponentName(obj.instance.Name, mcp.Name)
 		existingDaemonSet, ok := obj.em.daemonSets[generatedName]
 		if ok {
 			existingDs = existingDaemonSet.daemonSet
+			rteConfigHash = existingDaemonSet.rteComputedConfigHash
 			loadError = existingDaemonSet.daemonSetError
 		} else {
 			loadError = fmt.Errorf("failed to find daemon set %s/%s", mf.Core.DaemonSet.Namespace, mf.Core.DaemonSet.Name)
@@ -101,6 +103,7 @@ func (obj machineConfigPoolFinder) FindState(mf Manifests, tree nodegroupv1.Tree
 			MachineConfigPool:     mcp.DeepCopy(),
 			NodeGroup:             tree.NodeGroup.DeepCopy(),
 			DaemonSet:             desiredDaemonSet,
+			RTEComputedConfigHash: rteConfigHash,
 			IsCustomPolicyEnabled: annotations.IsCustomPolicyEnabled(tree.NodeGroup.Annotations),
 		}
 
diff --git a/pkg/objectstate/rte/nodegroup.go b/pkg/objectstate/rte/nodegroup.go
index 902c8e6d..c4adfff6 100644
--- a/pkg/objectstate/rte/nodegroup.go
+++ b/pkg/objectstate/rte/nodegroup.go
@@ -59,6 +59,7 @@ func (obj nodeGroupFinder) UpdateFromClient(ctx context.Context, cli client.Clie
 func (obj nodeGroupFinder) FindState(mf Manifests, tree nodegroupv1.Tree) []objectstate.ObjectState {
 	var ret []objectstate.ObjectState
 	var existingDs client.Object
+	var rteConfigHash string
 	var loadError error
 
 	poolName := *tree.NodeGroup.PoolName
@@ -67,6 +68,7 @@ func (obj nodeGroupFinder) FindState(mf Manifests, tree nodegroupv1.Tree) []obje
 	existingDaemonSet, ok := obj.em.daemonSets[generatedName]
 	if ok {
 		existingDs = existingDaemonSet.daemonSet
+		rteConfigHash = existingDaemonSet.rteComputedConfigHash
 		loadError = existingDaemonSet.daemonSetError
 	} else {
 		loadError = fmt.Errorf("failed to find daemon set %s/%s", mf.Core.DaemonSet.Namespace, mf.Core.DaemonSet.Name)
@@ -85,6 +87,7 @@ func (obj nodeGroupFinder) FindState(mf Manifests, tree nodegroupv1.Tree) []obje
 		MachineConfigPool:     nil,
 		NodeGroup:             tree.NodeGroup.DeepCopy(),
 		DaemonSet:             desiredDaemonSet,
+		RTEComputedConfigHash: rteConfigHash,
 		IsCustomPolicyEnabled: annotations.IsCustomPolicyEnabled(tree.NodeGroup.Annotations),
 	}
 
diff --git a/pkg/objectstate/rte/rte.go b/pkg/objectstate/rte/rte.go
index c13a4f5f..43df352e 100644
--- a/pkg/objectstate/rte/rte.go
+++ b/pkg/objectstate/rte/rte.go
@@ -54,8 +54,9 @@ type rteHelper interface {
 }
 
 type daemonSetManifest struct {
-	daemonSet      *appsv1.DaemonSet
-	daemonSetError error
+	daemonSet             *appsv1.DaemonSet
+	daemonSetError        error
+	rteComputedConfigHash string
 }
 
 type machineConfigManifest struct {
@@ -227,6 +228,7 @@ type GeneratedDesiredManifest struct {
 	NodeGroup         *nropv1.NodeGroup
 	// generated manifests
 	DaemonSet             *appsv1.DaemonSet
+	RTEComputedConfigHash string
 	IsCustomPolicyEnabled bool
 }
 ```

ffromani (commented):
/approve
/lgtm

thanks!

PR #1223, opened 2025-03-11
[release-4.12] deps: move to openshift fork of x/oauth2 and x/crypto

ffromani:
consume the forked branch with only the targeted fix for oauth2 over v0.26.0
crypto over v0.33.0


swatisehgal (approved):
/lgtm

PR #1224, opened 2025-03-12
cleanup: remove hash computation from shceduler's deployment

Tal-or:
In contrary to nrop controller where the configmap is generated by a different controller, on the sheduler side we're fully control the configmap which is static and not configurable, so we don't need to keep the configmap hash in the deployment.

We do change the API of one of the functions, but as of now, we're not aware of anyone using it but this flow.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -471,7 +471,7 @@ func renderSchedulerManifests(schedManifests schedmanifests.Manifests, imageSpec
 	// empty string is fine. Will be handled as "disabled".
 	// We only care about setting the environ variable to declare it exists,
 	// the best setting is "present, but disabled" vs "missing, thus implicitly disabled"
-	schedupdate.DeploymentConfigMapSettings(mf.Deployment, schedManifests.ConfigMap.Name, hash.ConfigMapData(schedManifests.ConfigMap))
+	schedupdate.DeploymentConfigMapSettings(mf.Deployment, schedManifests.ConfigMap.Name)
 	return mf, nil
 }
 
--- a/internal/controller/numaresourcesscheduler_controller.go
+++ b/internal/controller/numaresourcesscheduler_controller.go
@@ -44,7 +44,6 @@ import (
 	"github.com/openshift-kni/numaresources-operator/internal/api/annotations"
 	"github.com/openshift-kni/numaresources-operator/internal/relatedobjects"
 	"github.com/openshift-kni/numaresources-operator/pkg/apply"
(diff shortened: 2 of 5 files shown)


ffromani (commented):
you're right this code may very well be redundant but I can't recall if we have a test which ensure that the scheduler config changes are correctly picked up by the scheduler. IOW, before to remove this code (which I'd like) we need to make sure we don't have other bugs hidden by this code.

Tal-or:
> you're right this code may very well be redundant but I can't recall if we have a test which ensure that the scheduler config changes are correctly picked up by the scheduler. IOW, before to remove this code (which I'd like) we need to make sure we don't have other bugs hidden by this code.

Please keep me honest here but isn't configmap for the scheduler is static? 
We have a static `yaml` file that we create as a config map and we cannot configure it by changing the CR

ffromani:
> > you're right this code may very well be redundant but I can't recall if we have a test which ensure that the scheduler config changes are correctly picked up by the scheduler. IOW, before to remove this code (which I'd like) we need to make sure we don't have other bugs hidden by this code.
> 
> Please keep me honest here but isn't configmap for the scheduler is static? We have a static `yaml` file that we create as a config map and we cannot configure it by changing the CR

you are correct. NROP owns the scheduler config map, noone else is supposed to change it.
Now: if someone edits the `numaresourcesscheduler` CR, changes should be reflected in the `configmap` (and I think this works fine btw). Now, is the scheduler expect to pick up the change automatically? As user, I'd totally expect no further changes on my side. If the scheduler pod needs to be restarted, so be it.
What we must not do is introduce a regression in the form of a hidden change of behavior. So:
1. if previously the change was automatically picked by the scheduler **and this is the behavior we want**, it must keep working like that
2. if previously the change was automatically picked by the scheduler **and that was wrong** (was it? why?), then we must clearly document the change of behavior and release note it

ffromani:
> > NROP owns the scheduler config map, noone else is supposed to change it.
> > Now: if someone edits the numaresourcesscheduler CR, changes should be reflected in the configmap (and I think this works fine btw).
> 
> That's the part I'm not sure about and that's why I'm asking, is changing `numaresourcesscheduler` actually changes the configmap? because when I did a sweep through the code I saw we're using the configmap manifest verbatim. IOW nothing in the CR is affecting the configmap data

Some parameters are taken from the CR, yes. PTAL to `configParamsFromSchedSpec`

Tal-or:
Changes from CR are reflected in the ConfigMap so we should keep the hash logic

PR #1229, opened 2025-03-16
verify-commit: allow dependabot signoff

Tal-or:
The `AuthorEmail` of the dependabot is different than the allowed pattern.

adding a separate verification for the dependabot to make the CI happy.


Tal-or:
This PR is to make the CI pass in: https://github.com/openshift-kni/numaresources-operator/pull/1226

ffromani (commented):
/approve
/lgtm

thanks, this is pretty much how I would have done it.

PR #1231, opened 2025-03-17
sched: ha: rbac: fix autodetection

ffromani:
HA autosizing need to know how many worker objects we have so we fix the RBAC permissions


Tal-or (approved):
So the logic was there but broken?

ffromani:
> So the logic was there but broken?

the autodetection was broken, yes. But the feature is not supported yet (and of course disabled by default) so no big deal

ffromani:
/retest

can it be that adding a RBAC rule make the hypershift lane to fail? it seems unlikely to me

PR #1233, opened 2025-03-17
tools: sortimports

Tal-or:
A tool to automate the sorting of imports for our files in the project.
The tool expect a one or more  absolute paths on which it will iterate recursevly
sorting the import according to the following scheme:
 1. Standard libraries
 2. "k8s.io" imports
 3. "sigs.k8s.io" imports
 4. Other third-party libraries
 5. "github.com/openshift" imports
 6. "github.com/openshift-kni" imports

The tool can also be given a list of directory to skip, vendor
directory for example.

### DIFF
--- a/internal/controller/numaresourcesscheduler_controller_test.go
+++ b/internal/controller/numaresourcesscheduler_controller_test.go
@@ -20,12 +20,7 @@ import (
 	"context"
 	"time"
 
-	"github.com/onsi/ginkgo/v2"
-	"github.com/onsi/gomega"
-	configv1 "github.com/openshift/api/config/v1"
-
 	"github.com/google/go-cmp/cmp"
-
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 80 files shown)


ffromani (commented):
yes please, this was on my wishlist since a while

let's sort the CI issues and we can merge

tools/sortimports/main.go
```diff
@@ -0,0 +1,257 @@
...
+	ExcludeDirs sets.Set[string]
+	Verbose     bool
+}
+
+// Sorts imports based on:
+// 1. Standard libraries
+// 2. "k8s.io" imports
+// 3. "sigs.k8s.io" imports
+// 4. Other third-party libraries
+// 5. "github.com/openshift" imports
```
shajmakh, line 58: it is probably not common to use packages from other openshift projects but maybe we can add another filter specific for the local nrop imports and place them at the bottom:
`github.com/openshift-kni/numaresources-operator`

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -39,15 +34,19 @@ import (
 
 	depmanifests "github.com/k8stopologyawareschedwg/deployer/pkg/manifests"
 	depobjupdate "github.com/k8stopologyawareschedwg/deployer/pkg/objectupdate"
+	"github.com/onsi/ginkgo/v2"
```
shajmakh, line 37: do we want testing frameworks imports grouped with other packages?

internal/controller/suite_test.go
```diff
@@ -22,15 +22,15 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-
-	nropv1 "github.com/openshift-kni/numaresources-operator/api/v1"
-	securityv1 "github.com/openshift/api/security/v1"
-	machineconfigv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
shajmakh, line 25: same here, onsi's imports are not seperated from k8s ones

tools/sortimports/main.go
```diff
@@ -0,0 +1,257 @@
...
+
+	for _, imp := range imports {
+		path := strings.Trim(imp.Path.Value, `"`)
+		switch {
+		case isStdLib(path):
+			stdLibs = append(stdLibs, imp)
+		case strings.HasPrefix(path, "k8s.io") || strings.HasPrefix(path, "sigs.k8s.io"):
+			k8sLibs = append(k8sLibs, imp)
+		case strings.HasPrefix(path, "github.com/openshift"):
+			openShiftLibs = append(openShiftLibs, imp)
```
shajmakh, line 71: IIUC openshift and opnshift-kni will be grouped (and sorted later) together. I think the local imports of numaresources-operator project would look nicer if split from other openshift packages

shajmakh (commented):
Good to have this indeed, thanks!
initial review

internal/controller/numaresourcesscheduler_controller_test.go
Tal-or: it's considered as a third party so it's fine

internal/controller/suite_test.go
Tal-or: they should I'll check why not

Tal-or:
Move to: https://github.com/Tal-or/tool-box due to lack of interest

PR #1234, opened 2025-03-18
CNF-15623:must-gather: add support for hypershift

Tal-or:
To support running  MG on hosted clusters, we should check whether we're
running on standard OpenShift or HyperShift
and collect the relevant data accordingly.

### DIFF
--- a/Dockerfile.must-gather
+++ b/Dockerfile.must-gather
@@ -2,4 +2,4 @@ FROM quay.io/openshift/origin-cli:4.19
 
 COPY must-gather/collection-scripts/* /usr/bin/
 
-ENTRYPOINT /usr/bin/gather
+ENTRYPOINT ["/usr/bin/gather"]
--- a/must-gather/collection-scripts/gather
+++ b/must-gather/collection-scripts/gather
@@ -9,8 +9,8 @@ version >> /must-gather/version
 resources+=(nodes machineconfigs machineconfigpools featuregates kubeletconfigs)
 
 # run the collection of resources using must-gather
-for resource in ${resources[@]}; do
-  /usr/bin/oc adm inspect --dest-dir must-gather --all-namespaces ${resource}
+for resource in "${resources[@]}"; do
+  /usr/bin/oc adm inspect --dest-dir must-gather --all-namespaces "${resource}"
 done
 
 # Collect NRO details
--- a/must-gather/collection-scripts/gather
+++ b/must-gather/collection-scripts/gather
@@ -1,12 +1,28 @@
 #!/bin/bash
 
+
+function is_hypershift() {
+  control_plane_topology=$(/usr/bin/oc get infrastructure cluster -o jsonpath='{.status.controlPlaneTopology}'
+  if [[ "${control_plane_topology}" == "External" ]]; then
+    return 0
+  fi
+  return 1
+}
+
 # generate /must-gather/version file
 . version
 echo "numaresources-operator/must-gather" > /must-gather/version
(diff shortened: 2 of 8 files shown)


Tal-or:
/hold 
e2e tests are missing

Tal-or:
/hold cancel

PR is reviewable

must-gather/collection-scripts/gather
```diff
@@ -1,12 +1,27 @@
 #!/bin/bash
 
+
+function is_hypershift() {
+  control_plane_topology=$(/usr/bin/oc get infrastructure cluster -ojson | jq .status.controlPlaneTopology)
```
ffromani, line 5: we can also use kubectl advanced formatting and very likely get rid of the `jq` part, but AFAIR `jq` is already installed, so good enough for now

must-gather/collection-scripts/gather
```diff
@@ -1,12 +1,27 @@
...
 . version
 echo "numaresources-operator/must-gather" > /must-gather/version
 version >> /must-gather/version
 
 # machine/node resources
-resources+=(nodes machineconfigs machineconfigpools featuregates kubeletconfigs)
+resources+=(nodes featuregates)
+
+if is_hypershift; then
+  resources+=(configmaps)
```
ffromani, line 21: all of them?

Dockerfile.must-gather
```diff
@@ -2,4 +2,4 @@ FROM quay.io/openshift/origin-cli:4.19
 
 COPY must-gather/collection-scripts/* /usr/bin/
 
-ENTRYPOINT /usr/bin/gather
+ENTRYPOINT ["/usr/bin/gather"]
```
ffromani, line 5: this is fine, but doesn't clarify why the exec form is better (I'm fine with both TBH)

ffromani (commented):
provisional LGTM, some questions inside

Dockerfile.must-gather
Tal-or: I added a link to the official docs that explains why

must-gather/collection-scripts/gather
ffromani: I mean, on which namespace?

Tal-or: fixed that. configmap will be collected only from `openshift-config-managed` namespace which suppose to contain the kubeletconfig configmap

Tal-or: yes. should be here:https://github.com/openshift-kni/numaresources-operator/pull/1234/files#diff-2257fba988165a6d7631558cbd11f1c7aafb71d06f02e10d2fd4e210519f7c4eR21

Dockerfile.must-gather
ffromani: Yes, I saw it, thanks. Could you please make it explicit which of the reasons apply to us?
1. just use the preferred form because both are valid for our use case (a bit light, but OK)
2. prevent the shell process to be run
3. avoid any possible issue with signal delivery
4. streamline `ENTRYPOINT` and `CMD` interaction and avoid future bugs

all the above?
something else?

Tal-or: `1` the default approach is that we should always follow best-practice unless we have a well defined reason why not to.

Please correct me if i'm wrong here

Tal-or:
/hold

I would like to add some additional tests

Tal-or:
/hold cancel

The tests are more robust now.

ffromani (commented):
/approve
/lgtm

PR #1235, opened 2025-03-20
e2e: serial: abuse ginkgo to report involved version

ffromani:
add a bogus (because it cannot fail) test to report both e2e serial suite version and cluster version.

### DIFF
--- a/test/e2e/serial/tests/version.go
+++ b/test/e2e/serial/tests/version.go
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package tests
+
+import (
+	"context"
+	"encoding/json"
+	"path/filepath"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	nropv1 "github.com/openshift-kni/numaresources-operator/api/v1"
+	"github.com/openshift-kni/numaresources-operator/pkg/version"
+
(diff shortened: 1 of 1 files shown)


test/e2e/serial/tests/version.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/openshift-kni/numaresources-operator/internal/api/buildinfo"
+	"github.com/openshift-kni/numaresources-operator/internal/remoteexec"
+
+	e2eclient "github.com/openshift-kni/numaresources-operator/test/internal/clients"
+	"github.com/openshift-kni/numaresources-operator/test/internal/deploy"
+	"github.com/openshift-kni/numaresources-operator/test/internal/objects"
+)
+
+var _ = Describe("[serial] numaresources version", Serial, Label("feature:config"), func() {
+	When("starting the configuration suite", func() {
```
shajmakh, line 39: `s/configuration/serial/`?

test/e2e/serial/tests/version.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/openshift-kni/numaresources-operator/internal/api/buildinfo"
+	"github.com/openshift-kni/numaresources-operator/internal/remoteexec"
+
+	e2eclient "github.com/openshift-kni/numaresources-operator/test/internal/clients"
+	"github.com/openshift-kni/numaresources-operator/test/internal/deploy"
+	"github.com/openshift-kni/numaresources-operator/test/internal/objects"
+)
+
+var _ = Describe("[serial] numaresources version", Serial, Label("feature:config"), func() {
```
shajmakh, line 38: is this guaranteed to run first this at the suite run? it is similar to other test files under `serial/tests`
if not, ca we make it run at suite level, the build info is useful at any run of the suite, and focusing on specific tests will skip running this one

shajmakh (commented):
Thanks! 
small comments below

test/e2e/serial/tests/version.go
ffromani: uhm, probably a complete rephrase is better

test/e2e/serial/tests/version.go
ffromani: this should run at random order. Moving at suite level you mean like in `BeforeSuite`?
I would prefer it to have in a regular `It` block at risk of being skipped if explicitely filtered, because this way it is captured in the test output logs in the general case

shajmakh: >this should run at random order. Moving at suite level you mean like in BeforeSuite?

BeforeSuite can be a one option as long as the configuration is all setup (last thing to run before any first spec), in that can it will be outputted every time the suite runs even if zero tests run.

PR #1242, opened 2025-04-01
[4.13] deps: Bump golang.org/x/net to v0.36.0

swatisehgal:
Proactive bump to resolve CVE: GO-2025-3503
More info: https://pkg.go.dev/vuln/GO-2025-3503

Some other deps bumped as well when we ran `go mod tidy`


vendor/modules.txt
```diff
@@ -355,8 +355,8 @@ go.uber.org/ratelimit
 golang.org/x/mod/internal/lazyregexp
 golang.org/x/mod/module
 golang.org/x/mod/semver
-# golang.org/x/net v0.33.0
-## explicit; go 1.18
+# golang.org/x/net v0.36.0
+## explicit; go 1.23
```
ffromani, line 359: uhm, I just spotted this.
Problem is: we build with golang 1.19
https://github.com/openshift-kni/numaresources-operator/blob/release-4.13/Dockerfile.tools#L5

swatisehgal: Thanks for point this. Since we don't want to bump golang versions other than the z versions and this is not a flagged CVE we can close this.

/close

PR #1246, opened 2025-04-07
[release-4.17] api: allow to skip reconciliation of NRT CRD

ffromani:
Add custom annotation to stop reconciling the NRT API. This annotation is semi-public and will be used strictly on a case-by-case way if there are multiple operators which claim ownership of the NRT CRD.

With this annotation, NROP goes above and beyond and trusts blindly whatever NRT CRD is found in the cluster instead of owning the API.

Stop owning the NRT CRD greatly weakens the stability of the NUMA-aware scheduling and it is dangerous and definitely NOT recommended. But can be necessary in circumstances on which other operator claim ownership of the CRD, and we can't reconcile the differences.

(cherry picked from commit bef76e3f887ed9c216cab9c2b1c38bb01bafed22)


ffromani:
/hold

tests need more work

swatisehgal (approved):
/lgtm

ffromani:
not worth the extra effort since there are also more fixes in other components

PR #1252, opened 2025-04-10
e2e: serial: Fix flaky NRO status updates

SargunNarula:
This PR addresses flaky tests by ensuring NRO status is checked only after controller reconciliation, using Eventually blocks for reliable updates.


ffromani (commented):
/lgtm
/approve

thanks!

PR #1253, opened 2025-04-10
CNF-18350: Enforce Network Policies 

rbaturov:
This PR introduces the baseline network policies that deny all traffic and selectively allow only the necessary communication paths identified above.

**Motivation:**

To improve the security posture of the cluster, we are introducing default-deny network policies for Day-2 operators starting with OCP 4.20. By default, Kubernetes allows unrestricted communication between pods, which may lead to unintended exposure. These network policies are designed to restrict all ingress and egress traffic by default and allow only explicitly permitted communication.

Implementation Overview:
As part of this initiative, we examined the network behavior of the following components:

* The operator pod

* The scheduler pod

* The RTE pods

Based on this analysis, we identified the following requirements:

* API Server Access: All three pod types require egress access to the Kubernetes API server. This access will be explicitly allowed in the network policies.

* Metrics Access: Network policies will also allow traffic to and from metrics endpoints to support observability.

* Webhook server - ingress traffic to the webhook service port will be permitted.

* Kubelet Probes: Liveness probes, readiness probes originate from the Kubelet. These communications are handled outside the scope of standard Kubernetes NetworkPolicies and are permitted by default.

Needs to be done:
- [ ] Wait for the OLM RFE to support network policies and rerun make bundle

### DIFF
--- a/config/networkpolicy/deny-all.yaml
+++ b/config/networkpolicy/deny-all.yaml
@@ -1,0 +1,24 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: operator-default-deny-all
+spec:
+  podSelector:
+    matchLabels:
+      app: numaresources-operator
+  policyTypes:
+  - Ingress
+  - Egress
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: rte-default-deny-all
+spec:
+  podSelector:
+    matchLabels:
+      name: resource-topology  
+  policyTypes:
+  - Ingress
+  - Egress
+
--- a/config/networkpolicy/egress-to-api-server.yaml
+++ b/config/networkpolicy/egress-to-api-server.yaml
@@ -1,0 +1,29 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: operator-egress-to-api-server
+spec:
+  podSelector:
+    matchLabels:
+      app: numaresources-operator 
+  egress:
+  - ports:
+    - protocol: TCP
+      port: 6443 
+  policyTypes:
+  - Egress
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: rte-egress-to-api-server
+spec:
+  podSelector:
+    matchLabels:
+      name: resource-topology  
+  egress:
+  - ports:
+    - protocol: TCP
+      port: 6443 
+  policyTypes:
+  - Egress
--- a/config/networkpolicy/kustomization.yaml
+++ b/config/networkpolicy/kustomization.yaml
@@ -1,0 +1,4 @@
+resources:
+- deny-all.yaml
+- egress-to-api-server.yaml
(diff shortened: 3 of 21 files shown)


config/networkpolicy/deny-all.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: operator-default-deny-all
```
jmencak, line 4: Is this really necessary? Doesn't creation of any NetworkPolicy that allows traffic actually create an implicit deny all rule?

config/networkpolicy/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+resources:
+- deny-all.yaml
+- egress-to-api-server.yaml
+- ingress-to-metrics.yaml
```
jmencak, line 4: Nit: EOL

config/networkpolicy/egress-to-api-server.yaml
```diff
@@ -0,0 +1,29 @@
...
+metadata:
+  name: operator-egress-to-api-server
+spec:
+  podSelector:
+    matchLabels:
+      app: numaresources-operator 
+  egress:
+  - ports:
+    - protocol: TCP
+      port: 6443
```
jmencak, line 12: Does this port work? I found I had to change this to 443 in NTO.

```
$ oc get po
NAME                                            READY   STATUS    RESTARTS      AGE
cluster-node-tuning-operator-7c664f7594-5b5mt   1/1     Running   6 (21m ago)   4d17h
tuned-bslv7                                     1/1     Running   1             4d17h
$ oc rsh cluster-node-tuning-operator-7c664f7594-5b5mt
sh-5.1$ env|grep KUBERN
KUBERNETES_SERVICE_PORT_HTTPS=443
KUBERNETES_SERVICE_PORT=443
KUBERNETES_PORT_443_TCP=tcp://172.30.0.1:443
KUBERNETES_PORT_443_TCP_PROTO=tcp
KUBERNETES_PORT_443_TCP_ADDR=172.30.0.1
KUBERNETES_SERVICE_HOST=172.30.0.1
KUBERNETES_PORT=tcp://172.30.0.1:443
KUBERNETES_PORT_443_TCP_PORT=443
```

config/networkpolicy/deny-all.yaml
rbaturov: I have to add podSelector per deployment/daemonset because this day2 operator can be deployed in an already created namespace (it doesn't necessarily use a separate namespace), and in this case, your suggested `NetworkPolicy` will deny all traffic for other pods in the namespace.

For NTO use case, which exclusively uses the "openshift-cluster-node-tuning-operator" your suggestion should be applied.

config/networkpolicy/egress-to-api-server.yaml
rbaturov: Yes it does work. I have tested it

ffromani:
/close

continued in https://github.com/openshift-kni/numaresources-operator/pull/1605

PR #1254, opened 2025-04-11
e2e: serial: more scheduler stall tests against burstable and guaranteed jobs

ffromani:
Add more scheduler stall tests to check that we correctly handle the interference against jobs which create guaranteed and burstable pods.

A "scheduler stall" is when the secondary scheduler can't recover nodes from "invalid topology data" state. We say a given node is "stalled". If all the nodes, or "too many" nodes are stalled, we call a scheduler stall.


shajmakh:
Thanks for the additional coverage, despite the known bugs and the failing test I think we should still merge. 
/lgtm

PR #1255, opened 2025-04-11
e2e: serial: Test for performance profile update and NRT reconciliation 

SargunNarula:
This PR implements the test for the performance profile update over the management cluster and verifies the NRT objects reconciliation with the updated changes on the hosted cluster.

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -35,6 +35,7 @@ import (
 	k8swait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
 
 	"github.com/google/go-cmp/cmp"
 	depnodes "github.com/k8stopologyawareschedwg/deployer/pkg/clientutil/nodes"
@@ -56,6 +57,7 @@ import (
 	"github.com/openshift-kni/numaresources-operator/pkg/validation"
 	rteconfig "github.com/openshift-kni/numaresources-operator/rte/pkg/config"
 	"github.com/openshift-kni/numaresources-operator/test/e2e/label"
+	e2eclient "github.com/openshift-kni/numaresources-operator/test/internal/clients"
 	"github.com/openshift-kni/numaresources-operator/test/internal/configuration"
 	e2efixture "github.com/openshift-kni/numaresources-operator/test/internal/fixture"
 	e2enrt "github.com/openshift-kni/numaresources-operator/test/internal/noderesourcetopologies"
@@ -63,6 +65,8 @@ import (
 	"github.com/openshift-kni/numaresources-operator/test/internal/objects"
 
 	serialconfig "github.com/openshift-kni/numaresources-operator/test/e2e/serial/config"
+	"github.com/openshift-kni/numaresources-operator/test/internal/hypershift"
+	"github.com/openshift-kni/numaresources-operator/test/internal/nodepools"
 )
 
 /*
@@ -82,6 +86,14 @@ type mcpInfo struct {
 	mcpObj        *machineconfigv1.MachineConfigPool
 	initialConfig string
 	sampleNode    corev1.Node
+}
+
+type Tuning struct {
+	Profile struct {
+		CPU struct {
+			TopologyManagerPolicy string `yaml:"topologyManagerPolicy"`
+		} `yaml:"cpu"`
+	} `yaml:"profile"`
 }
 
 func (i mcpInfo) ToString() string {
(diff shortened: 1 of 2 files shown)


test/e2e/serial/tests/configuration.go
```diff
@@ -532,6 +535,81 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
 			expectNRTConsumedResources(fxt, *nrtPreCreate, rl, testPod)
 		})
 
+		It("[test_id:80821] Verify Performance Profile updates on management cluster reflect in NRT resources", Label("reboot_required", label.HyperShift), func() {
```
shajmakh, line 538: if this is a slow test please add the label.Slow ginkgo label

test/e2e/serial/tests/configuration.go
```diff
@@ -532,6 +535,81 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+					break
+				}
+			}
+			Expect(initialTMPolicy).ToNot(BeEmpty(), "topologyManagerPolicy not found in initial NRT")
+			newTMPolicy := getNewTopologyManagerPolicyValue(initialTMPolicy)
+			Expect(initialTMPolicy).ToNot(Equal(newTMPolicy), "new TM policy should differ from the initial one")
+
+			By("retrieving the PerformanceProfile configmap from the management cluster")
+			hostedClusterName, err := hypershift.GetHostedClusterName()
+			Expect(err).To(Not(HaveOccurred()))
```
shajmakh, line 561: To(Not -> ToNot(

test/e2e/serial/tests/configuration.go
```diff
@@ -532,6 +535,81 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+					break
+				}
+			}
+			Expect(targetCM).ToNot(BeNil(), fmt.Sprintf("ConfigMap %q not found in namespace: clusters", tuningConfigName))
+
+			By("updating the ConfigMap with the new TM policy")
+			yamlData := targetCM.Data["tuning"]
+			Expect(yamlData).ToNot(BeEmpty(), "tuning section not found in ConfigMap")
+
+			modifiedYaml := strings.Replace(yamlData, initialTMPolicy, newTMPolicy, 1)
```
shajmakh, line 588: there could be more than one empty value "" in that yaml (although it shouldn't happen because TAS works with single numa policy but better use another way to do the update, or use full key:value pairs for update.

test/e2e/serial/tests/configuration.go
```diff
@@ -532,6 +535,81 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+				if attr.Name == "topologyManagerPolicy" {
+					initialTMPolicy = attr.Value
+					break
+				}
+			}
+			Expect(initialTMPolicy).ToNot(BeEmpty(), "topologyManagerPolicy not found in initial NRT")
+			newTMPolicy := getNewTopologyManagerPolicyValue(initialTMPolicy)
+			Expect(initialTMPolicy).ToNot(Equal(newTMPolicy), "new TM policy should differ from the initial one")
+
+			By("retrieving the PerformanceProfile configmap from the management cluster")
```
shajmakh, line 559: we need to do extra verification layer before, how do you know if the configmap is for the PP ? and whether NROP is using PP config or kubeletconfig?

shajmakh (commented):
Thanks! below is the initial review

test/e2e/serial/tests/configuration.go
SargunNarula: Added slow label. Thanks

test/e2e/serial/tests/configuration.go
SargunNarula: Resolved with latest commit.

test/e2e/serial/tests/configuration.go
SargunNarula: Resolved by replacing the string manipulation with YAML marshalling.

test/e2e/serial/tests/configuration.go
SargunNarula: Since the ConfigMap is retrieved from the `tuningConfig` field under the NodePool spec, it effectively serves as the source of the PerformanceProfile configuration. In the NROP-HCP environment, there is no separate KubeletConfig as far as I’m aware. On the hosted cluster, the KubeletConfig ConfigMap is generated by the HyperShift operator and is marked as `immutable: true`.

test/e2e/serial/tests/configuration.go
```diff
@@ -285,6 +297,90 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+
+			tuning.Profile.CPU.TopologyManagerPolicy = newTMPolicy
+
+			modifiedYamlBytes, err := yaml.Marshal(&tuning)
+			Expect(err).ToNot(HaveOccurred(), "failed to marshal modified tuning YAML")
+			modifiedYaml := string(modifiedYamlBytes)
+			Expect(modifiedYaml).ToNot(Equal(yamlData), "no changes applied to ConfigMap YAML")
+
+			targetCM.Data["tuning"] = modifiedYaml
+			err = e2eclient.MNGClient.Update(context.TODO(), targetCM)
```
shajmakh, line 363: If I'm correct the reboot should happen on the update, where are we waiting for reboot?

SargunNarula: Yes, the node reboots on a PerformanceProfile update, but we don’t wait for the reboot explicitly. Since NRT relies on the RTE DaemonSet, and RTE pods are recreated only after the node becomes Ready, waiting for the updated NRT implicitly covers the reboot. Please let me know if I’ve misunderstood anything here.

Tal-or (commented):
I didn't went fully into reviewing the code itself, but there's an hidden assumption here, which is that a performance profile is available for the test.

This is not always the case because one can create a kubeletconfig directly.
The only valid assumption is that we have a kubeleconfig available, not a performance profile.

test/e2e/serial/tests/configuration.go
shajmakh: so is performance-profile is a must in HCP? what is the user want to do kubeletchanges but not involve PP? can't this just be done via kubeletconfig like in OCP?

test/e2e/serial/tests/configuration.go
```diff
@@ -841,6 +1000,14 @@ func objRefListToStringList(objRefs []configv1.ObjectReference) []string {
 	return ret
 }
 
+func getNewTopologyManagerPolicyValue(oldTopologyManagerPolicyValue string) string {
```
Tal-or, line 1003: I would rename it `getDifferentTopologyManagerPolicyFrom(oldTopologyManagerPolicy string) string`

test/e2e/serial/tests/configuration.go
```diff
@@ -84,6 +88,14 @@ type mcpInfo struct {
 	sampleNode    corev1.Node
 }
 
+type Tuning struct {
```
Tal-or, line 91: Worth checking if we already importing hypershift APIs

Tal-or (commented):
LGTM overall.
Please rebase if still relevant.

test/e2e/serial/tests/configuration.go
```diff
@@ -286,6 +298,153 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
 			Expect(schedOK).To(BeTrue(), "pod %s/%s not scheduled with expected scheduler %s", updatedPod.Namespace, updatedPod.Name, serialconfig.SchedulerTestName)
 		})
 
+		It("[test_id:80821] Verify Performance Profile updates on management cluster reflect in NRT resources", Label("reboot_required", label.Slow, label.HyperShift), func() {
```
rshemtov13, line 301: Since we don't have the dependencies for NTO here, In the serial suite we've added a while ago a dynamicClient to get the performanceprofile. It's being used [here](https://github.com/openshift-kni/numaresources-operator/blob/main/test/e2e/serial/tests/configuration.go#L558)
It's possible this test can use that.

Tal-or:
@SargunNarula is this still relevant?

Tal-or:
/close

please reopen again if still relevant

PR #1258, opened 2025-04-15
deps: reduce MCO dependencies

ffromani:
In order to facilitate the transition to the MCO APIs after the merge in `openshift/api`, reduce where it is cheap our usage of the MCO types. Less dep surface, less porting effort.

No intended changes in behavior.


shajmakh:
Thanks for this. 
I think we should also update the imports to use the new 
github.com/openshift/api/tree/master/machineconfiguration/v1instead; the module is already required in go.mod and the API would be more up-to-date via the api project

shajmakh:
/lgtm
updating MCP api import is not blocking

PR #1263, opened 2025-04-21
OCPBUGS-51261: controller: RTE `ConfigMap` predicate

Tal-or:
A continuation of #1216 which turns out to solve the problem only partially.

The ConfigMap generated by Kubelet controller
should be watched by NUMAResourceoperator controller,
so it will be able to act upon a ConfigMap change.

### DIFF
--- a/internal/controller/kubeletconfig_controller.go
+++ b/internal/controller/kubeletconfig_controller.go
@@ -20,12 +20,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"slices"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
@@ -36,7 +38,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	mcov1 "github.com/openshift/api/machineconfiguration/v1"
 
@@ -55,9 +59,10 @@ const (
 )
 
 const (
-	HypershiftKubeletConfigConfigMapLabel = "hypershift.openshift.io/kubeletconfig-config"
-	HyperShiftNodePoolLabel               = "hypershift.openshift.io/nodePool"
-	HyperShiftConfigMapConfigKey          = "config"
+	HypershiftKubeletConfigConfigMapLabel     = "hypershift.openshift.io/kubeletconfig-config"
+	HyperShiftNodePoolLabel                   = "hypershift.openshift.io/nodePool"
+	HyperShiftConfigMapConfigKey              = "config"
+	HyperShiftKubeletConfigConfigMapNamespace = "openshift-config-managed"
 )
 
 // KubeletConfigReconciler reconciles a KubeletConfig object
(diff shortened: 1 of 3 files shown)


test/e2e/install/install_test.go
```diff
@@ -356,9 +360,80 @@ var _ = Describe("[Install] durability", Serial, func() {
...
+			Eventually(func() bool {
+				updatedNrtObjs := &nrtv1alpha2.NodeResourceTopologyList{}
+				err := e2eclient.Client.List(context.TODO(), updatedNrtObjs)
+				Expect(err).ToNot(HaveOccurred())
+
+				for i := range updatedNrtObjs.Items {
+					nrt := &updatedNrtObjs.Items[i]
+					// in this specific test deployment,
+					// the same configuration should apply to all NRT objects
+					matchingResult := checkTopologyManagerConfigMatching(nrt, &cfg)
```
rbaturov, line 393: I think a better name instead of matchingResult could be `mismatchErrors` or `misconfigurationErrors`, since the function only returns details when a mismatch is found. What do you think?
I would change it inside the `checkTopologyManagerConfigMatching` function as well.

Tal-or: Thanks. Fixed

swatisehgal (commented):
Thanks for the PR and also adding e2e test.

I understand that the e2e test you have added allows us to ensure that we have expected Topology Manager configuration and I think this is of value but I am failing to understand how it is proof of the fact that we have resolved the issue at hand (NROP controller not creating new CM on node groups updates). Can you please explain?

Can we add an e2e test for validating that the code changes proposed here resolve the main issue and that new CMs are created if we point to a different node group. Adding an e2e would also automate the verification process of the bug.

Tal-or:
> Thanks for the PR and also adding e2e test.
> 
> I understand that the e2e test you have added allows us to ensure that we have expected Topology Manager configuration and I think this is of value but I am failing to understand how it is proof of the fact that we have resolved the issue at hand (NROP controller not creating new CM on node groups updates). Can you please explain?
> 
> Can we add an e2e test for validating that the code changes proposed here resolve the main issue and that new CMs are created if we point to a different node group. Adding an e2e would also automate the verification process of the bug.

The bug describes a symptom, but not the root cause. 
The root cause was that NROP controller is not watching for CM changes that are done by the Kubelet controller. This is  why we've added the predicate for.

With that being said it might be useful to add a test that mimic the bug description exactly. We should add it in the serial suite though since it would be too complex as an u/s test.

swatisehgal:
> With that being said it might be useful toadd a test that mimic the bug description exactly. We should add it in the serial suite though since it would be too complex as an u/s test.

Sure, addition to the serial suite sounds reasonable. Do you prefer to do that in the same PR or a follow-up? I am fine either way.

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -672,6 +684,36 @@ func (r *NUMAResourcesOperatorReconciler) mcpToNUMAResourceOperator(ctx context.
 	return requests
 }
 
+func (r *NUMAResourcesOperatorReconciler) configMapToNUMAResourceOperator(ctx context.Context, cmObj client.Object) []reconcile.Request {
```
shajmakh, line 687: why do we need this if `configMapPredicates` is already filtering the config maps based on the NROP label?

Tal-or:
split the serial tests to https://github.com/openshift-kni/numaresources-operator/pull/1266 since we don't need to backport them

internal/controller/numaresourcesoperator_controller.go
Tal-or: `configMapToNUMAResourceOperator` is a function which converts the `ConfigMap` (A) events  to `NUMAResourceOperator` requests, because the `NUMAResourcesOperatorReconciler` expects to reconcile the NUMAResourceOperator CR and not a `ConfigMap` objects

The configMapPredicates is under filter layer that minimizes the `ConfigMap` events that gets into the `configMapToNUMAResourceOperator` function.

internal/controller/kubeletconfig_controller.go
```diff
@@ -155,9 +159,22 @@ func (r *KubeletConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	}
+	numaResourcesOperatorPredicate := predicate.Funcs{
+		UpdateFunc: func(e event.TypedUpdateEvent[client.Object]) bool {
+			nroOld := e.ObjectOld.(*nropv1.NUMAResourcesOperator)
+			nroNew := e.ObjectNew.(*nropv1.NUMAResourcesOperator)
+			oldNodeGroups := nroOld.Spec.NodeGroups
+			newNodeGroups := nroNew.Spec.NodeGroups
+			// if nodeGroups number has changed,
+			// it means that we should iterate and create/delete ConfigMap data for RTE pods
+			return len(oldNodeGroups) != len(newNodeGroups)
```
shajmakh, line 170: what about the case where the NROP CR was updated with different nodegroup but maintained same amount of entries in spec.nodegroup? in such case the condition would be false but in reality the groups changed

Tal-or: Right. let's do a simple compare instead

swatisehgal (commented):
LGTM from my side.

Will let @shajmakh add the label once she is happy enough.

internal/controller/kubeletconfig_controller.go
```diff
@@ -20,12 +20,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/equality"
```
shajmakh, line 23: nit: misplaced import

internal/controller/numaresourcesoperator_controller.go
shajmakh: ack I see how builder Watches is expecting an event handler and options, thanks for clarifying

shajmakh:
/lgtm
Thanks for handing this

PR #1268, opened 2025-04-25
internal: ensure total memory resources are equal on node level

shajmakh:
In
https://github.com/openshift-kni/numaresources-operator/commit/4f849a40fc669180f036b621f2cfb5c85d89f245 a deviation of the memory amount on NUMA level was added to allow more relaxed check for tests that involved reboot. This is correct but still need to verify that the total memory resources on node level is maintained after the reboot.
Add a verification step to verify that the total memory resources per compute node is same before and after, and add unit tests coverage.


internal/noderesourcetopology/equality.go
```diff
@@ -38,6 +38,22 @@ func EqualZones(zonesA, zonesB nrtv1alpha2.ZoneList, isRebootTest bool) (bool, e
 	zA := SortedZoneList(zonesA)
 	zB := SortedZoneList(zonesB)
 
+	if isRebootTest {
```
Tal-or, line 41: if we're aiming for checking at the node level, I think we can remove the `isRebootTest` flag, and use this checkas a sanity for all the tests

Tal-or (commented):
I would also add a link for the kernel docs/article that will backup our claim for not checking at the NUMA node level.

internal/noderesourcetopology/equality.go
shajmakh: good points, this is valid for any test

shajmakh:
> I would also add a link for the kernel docs/article that will backup our claim for not checking at the NUMA node level.

we do check numa level resources but with deviation as explained in this PR: https://github.com/openshift-kni/numaresources-operator/commit/4f849a40fc669180f036b621f2cfb5c85d89f245

Tal-or:
> > I would also add a link for the kernel docs/article that will backup our claim for not checking at the NUMA node level.
> 
> we do check numa level resources but with deviation as explained in this PR: [4f849a4](https://github.com/openshift-kni/numaresources-operator/commit/4f849a40fc669180f036b621f2cfb5c85d89f245)

So that means the `reboot` flag is still needed? 

What I was thinking is that if the check per numa node is not reliable after a reboot, then what do we gain here?
isn't checking at the node level is enough?

shajmakh:
> > > I would also add a link for the kernel docs/article that will backup our claim for not checking at the NUMA node level.
> > 
> > 
> > we do check numa level resources but with deviation as explained in this PR: [4f849a4](https://github.com/openshift-kni/numaresources-operator/commit/4f849a40fc669180f036b621f2cfb5c85d89f245)
> 
> So that means the `reboot` flag is still needed?
>

yes the flag is still needed when comparing the per numa resources amounts.

 
> What I was thinking is that if the check per numa node is not reliable after a reboot, then what do we gain here? isn't checking at the node level is enough?

In general, NUMA resources should be stable and same before and after system reboot, resources allocation per NUMA might change due to few reasons like BIOS configuration update or hardware settings updates but these are irrelevant to our case, so we expect to see same data reflected on NUMA zones before and after the reboot, this applies to all resources. However, we have observed cases in which the amounts may change such as memory distribution on NUMA nodes and sample devices distributions post reboot (https://github.com/openshift-kni/numaresources-operator/commit/fddb81f8a85abc9c04fc22301436a9da5740274b). In https://github.com/openshift-kni/numaresources-operator/commit/4f849a40fc669180f036b621f2cfb5c85d89f245 we added a relaxed check for memory only considering that the change in the memory resources on reboot is consistent on every occurrence and not significant in amount. In this PR we ensure that even with a small deviation being applied on memory resources on comparison, amounts should still be consistent on node level which matches the observations.

Tal-or:
> > > > I would also add a link for the kernel docs/article that will backup our claim for not checking at the NUMA node level.
> > > 
> > > 
> > > we do check numa level resources but with deviation as explained in this PR: [4f849a4](https://github.com/openshift-kni/numaresources-operator/commit/4f849a40fc669180f036b621f2cfb5c85d89f245)
> > 
> > 
> > So that means the `reboot` flag is still needed?
> 
> yes the flag is still needed when comparing the per numa resources amounts.
> 
> > What I was thinking is that if the check per numa node is not reliable after a reboot, then what do we gain here? isn't checking at the node level is enough?
> 
> In general, NUMA resources should be stable and same before and after system reboot, resources allocation per NUMA might change due to few reasons like BIOS configuration update or hardware settings updates but these are irrelevant to our case, so we expect to see same data reflected on NUMA zones before and after the reboot, this applies to all resources. However, we have observed cases in which the amounts may change such as memory distribution on NUMA nodes and sample devices distributions post reboot ([fddb81f](https://github.com/openshift-kni/numaresources-operator/commit/fddb81f8a85abc9c04fc22301436a9da5740274b)). In [4f849a4](https://github.com/openshift-kni/numaresources-operator/commit/4f849a40fc669180f036b621f2cfb5c85d89f245) we added a relaxed check for memory only considering that the change in the memory resources on reboot is consistent on every occurrence and not significant in amount. In this PR we ensure that even with a small deviation being applied on memory resources on comparison, amounts should still be consistent on node level which matches the observations.

I understand, but if the checks at the numa node level is not reliable, then why we need it? 
If we want to check that memory after the test is similar to the initial memory (before the test) then checking at the node level is enough

shajmakh:
> I understand, but if the checks at the numa node level is not reliable, then why we need it? If we want to check that memory after the test is similar to the initial memory (before the test) then checking at the node level is enough

Because of 2 reasons:
1. in tests requiring reboot, we set a specific deviation value that if exceeded the test should fail and something must have changed and we should look into that. As mentioned above, as long as the difference is consistent the test result should be pass. If something changes we want to detect and handle that. It is not enough to compare node-level memory even in reboot tests, but we want to ensure that the deviation value is not exceeded. 
2. In tests that don't require reboot, we still want to verify that the resource amount is the same before and after the test regardless of what type of resource the check is running on.

Tal-or:
> > I understand, but if the checks at the numa node level is not reliable, then why we need it? If we want to check that memory after the test is similar to the initial memory (before the test) then checking at the node level is enough
> 
> Because of 2 reasons:
> 
> 1. in tests requiring reboot, we set a specific deviation value that if exceeded the test should fail and something must have changed and we should look into that. As mentioned above, as long as the difference is consistent the test result should be pass. If something changes we want to detect and handle that. It is not enough to compare node-level memory even in reboot tests, but we want to ensure that the deviation value is not exceeded.
> 2. In tests that don't require reboot, we still want to verify that the resource amount is the same before and after the test regardless of what type of resource the check is running on.

Fair enough

Tal-or (approved):

PR #1269, opened 2025-04-25
cleanup: remove dangling ConfigMaps

shajmakh:
When node group in NRO CR is deleted the CM that corresponds to that CM is kept dangling. Similar to dangling DS and MCs, delete the dangling CMs during the NRO controller resync loop.


Tal-or (commented):
I can see the rational here but we're not suppose to delete the ConfigMap since it's managed by the Kubeletconfig controller and owned by the KubeletConfig CR:
https://github.com/openshift-kni/numaresources-operator/blob/597d9df3ffac7b084fefef8192f28c414c9f3087/internal/controller/kubeletconfig_controller.go#L214 

IOW, when KubeletConfig CR gets deleted, the ConfigMap will get deleted automatically by k8s

PR #1273, opened 2025-04-28
e2e: serial: Update MCPs on NRO CR deletion when customPolicy is set

rshemtov13:
As reboot triggered by NROP CR deletion and creation is not necessary for versions >= 4.18 since it causes tests 75854 and 75855 to fail unintentionally.
The changes here removes the wait for mcp updates and in addition ensures the before each function checks that the RTE pods are deleted properly.In addition to preserve the old behavior for versions <= 4.17 we're checking to see if the custom policysupport is enabled which then a waitForMcpUpdate is necessary.

### DIFF
--- a/internal/wait/pod.go
+++ b/internal/wait/pod.go
@@ -22,6 +22,8 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/sync/errgroup"
+
 	corev1 "k8s.io/api/core/v1"
 	k8swait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
@@ -108,4 +110,40 @@ func (wt Waiter) ForPodListAllRunning(ctx context.Context, pods []*corev1.Pod) (
 	}
 	wg.Wait()
 	return failed, updated
+}
+
+func (wt Waiter) ForPodListAllDeleted(ctx context.Context, pods []corev1.Pod) error {
+	if len(pods) == 0 {
+		klog.Infof("ForPodListAllDeleted called with an empty list of pods. Nothing to do.")
+		return nil
+	}
+
+	klog.Infof("Waiting for %d pod(s) to be deleted.", len(pods))
+
+	var eg errgroup.Group
+	for idx := range pods {
+		pod := pods[idx]
+		podKey := client.ObjectKeyFromObject(&pod).String()
+
+		eg.Go(func() error {
+			klog.Infof("Goroutine started: waiting for pod %s to be deleted.", podKey)
+			err := wt.ForPodDeleted(ctx, pod.Namespace, pod.Name)
+			if err != nil {
+				klog.Warningf("Failed to confirm deletion for pod %s: %v", podKey, err)
+			} else {
+				klog.Infof("Successfully confirmed deletion for pod %s.", podKey)
+			}
+			return err
+		})
+	}
+
+	err := eg.Wait()
+
+	if err != nil {
+		klog.Warningf("One or more pods were not deleted successfully: %v", err)
+	} else {
+		klog.Info("All pods were successfully deleted")
+	}
+
+	return err
 }
--- a/test/e2e/serial/tests/tolerations.go
+++ b/test/e2e/serial/tests/tolerations.go
@@ -26,8 +26,8 @@ import (
 	. "github.com/onsi/gomega"
 
 	"github.com/k8stopologyawareschedwg/deployer/pkg/clientutil/nodes"
-
 	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
+
 	inthelper "github.com/openshift-kni/numaresources-operator/internal/api/annotations/helper"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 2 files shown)


shajmakh (commented):
Thanks for checking this, but beware that these tests run also on OCP <4.18 so they will still require reboot on older versions. I suggest two things:
1. require changes in both u/s and d/s: detect when we should expect reboot on RTE updates and adjust the same test's code.
  1.1. to manage labels for same tests that on some versions it needs reboot and on others it doesn't, we could add another feature label pointing to the new selinux policy support in 4.18+ and adapt that change in d/s ci to make the exclusions. 
2. requires change in u/s code only: create 2 new feature labels one for old selinux policy and one for new, split the tests into 2 versions one for 4.18+ and one for <4.18 update topics.json for all the branches with the relevant supported label.

test/e2e/serial/tests/tolerations.go
```diff
@@ -536,18 +536,37 @@ var _ = Describe("[serial][disruptive][rtetols] numaresources RTE tolerations su
...
 
-					err = fxt.Client.Delete(ctx, &nroOperObj)
+					By("Get list of worker nodes")
+					workers, err = nodes.GetWorkers(fxt.DEnv())
 					Expect(err).ToNot(HaveOccurred())
 
-					waitForMcpUpdate(fxt.Client, ctx, MachineConfig, time.Now().String(), mcpsInfo...)
+					By("verifying RTE pods removal on all worker nodes")
+					maxRetries := 20
+					for step := range maxRetries {
```
SargunNarula, line 548: range does not work with Integers.

SargunNarula (commented):
Great work, one correction below:
Please update the the PR description and commit message as it contains wrong test_ids.

test/e2e/serial/tests/tolerations.go
rshemtov13: As of Go 1.22 ranging over integers is possible. Thanks for the wrong test_id's i'll change it.

rshemtov13:
The required changes currently apply only to version 4.18 and above. For versions below 4.18, these changes would break the tests, as deleting the NROP CR in those versions requires a reboot. I will update the code to handle both scenarios appropriately.

ffromani:
the linter failure is about (allegedly) no longer needed code

test/e2e/serial/tests/tolerations.go
```diff
@@ -531,24 +533,56 @@ var _ = Describe("[serial][disruptive][rtetols] numaresources RTE tolerations su
...
 
-					By("taint one worker node")
+					By("Get list of worker nodes")
+					var err error
 					workers, err = nodes.GetWorkers(fxt.DEnv())
 					Expect(err).ToNot(HaveOccurred())
 
+					By("verifying RTE pods removal on all worker nodes")
+					maxRetries := 10
+					for step := range maxRetries {
```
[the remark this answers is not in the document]
rshemtov13: Understood I'll move it before the CR deletion of NROP.

test/e2e/serial/tests/tolerations.go
```diff
@@ -574,15 +587,15 @@ var _ = Describe("[serial][disruptive][rtetols] numaresources RTE tolerations su
...
-
 						err = fxt.Client.Create(ctx, nropNewObj)
 						Expect(err).ToNot(HaveOccurred())
 
-						mcpsInfo, err := buildMCPsInfo(fxt.Client, ctx, *nropNewObj)
-						Expect(err).ToNot(HaveOccurred())
-						Expect(mcpsInfo).ToNot(BeEmpty())
-
-						waitForMcpUpdate(fxt.Client, ctx, MachineCount, time.Now().String(), mcpsInfo...)
+						if customPolicySupportEnabled {
```
shajmakh, line 593: I understand the goal of this is to wait for reboots only when needed, which means we also need to update the `					fxt.IsRebootTest = true` to false so expect no resources deviations

shajmakh: uless there is another event that would trigger a node reboot/mcp update which in that case we still need to have the flag enabled

rshemtov13: Right, in that case the test would be considered non reboot.
I'll add a check at the beginning of each ```It()``` to update the value of ```fxt.IsRebootTest```.
Another nit here is the ```Label("reboot_required")``` which is not correct for versions > 4.18.
I'll do a future PR to split the reboot_required label to two different labels, probably that would be the best solution for this.

internal/wait/pod.go
```diff
@@ -109,3 +109,49 @@ func (wt Waiter) ForPodListAllRunning(ctx context.Context, pods []*corev1.Pod) (
...
+
+			lock.Lock()
+			if err != nil {
+				klog.Warningf("Failed to confirm deletion for pod %s: %v", podKey, err)
+				runningPods = append(runningPods, currentPod)
+			} else {
+				klog.Infof("Successfully confirmed deletion for pod %s.", podKey)
+				deletedPods = append(deletedPods, currentPod)
+			}
+			lock.Unlock()
```
shajmakh, line 143: can you clarify why this block is needed beside reporting? I believe this is already checked in the next verification step

internal/wait/pod.go
```diff
@@ -109,3 +109,49 @@ func (wt Waiter) ForPodListAllRunning(ctx context.Context, pods []*corev1.Pod) (
...
+	}
+
+	wg.Wait()
+
+	klog.Infof("Finished waiting for pod deletions. DeletedPods: %d, RunningPods: %d.", len(deletedPods), len(runningPods))
+	if len(runningPods) > 0 {
+		for _, pod := range runningPods {
+			klog.Warningf("Pod %s/%s was NOT confirmed as deleted.", pod.Namespace, pod.Name)
+		}
+	}
```
shajmakh, line 154: I see the point of this, and it is a good addition, except that this can be simpler especially because the nexted ForPodDeleted already reports the undelete pod, and because there is no use of the returned pod lists apart from verifying they fulfill the intention of the function (e.g delete all pods)
instead of waitgroup you can use errgroup.group (https://github.com/openshift-kni/numaresources-operator/blob/main/test/internal/deploy/deploy.go#L150C9-L150C23) and take advantage of the returned error from ForPodDeleted. then at the end of the function return wg.Wait() and outside assert the error expecting it to be nil, fail otherwise. Please correct me if I am missing the point of this.

test/e2e/serial/tests/tolerations.go
```diff
@@ -27,6 +27,8 @@ import (
 
 	"github.com/k8stopologyawareschedwg/deployer/pkg/clientutil/nodes"
 
+	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
+	inthelper "github.com/openshift-kni/numaresources-operator/internal/api/annotations/helper"
```
shajmakh, line 31: misplaced imports

shajmakh (commented):
Thanks for addressing the discussed points. I have one comment apart from that it lgtm.

test/e2e/serial/tests/tolerations.go
rshemtov13: Right, will address.

internal/wait/pod.go
rshemtov13: I see what you mean, yes it would be better to return an error. I'll address that, thanks for noting this.

internal/wait/pod.go
```diff
@@ -109,3 +111,39 @@ func (wt Waiter) ForPodListAllRunning(ctx context.Context, pods []*corev1.Pod) (
...
+
+	err := eg.Wait()
+
+	if err != nil {
+		klog.Warningf("One or more pods were not deleted successfully: %v", err)
+	} else {
+		klog.Info("All pods were successfully deleted")
+	}
+
+	return eg.Wait()
```
shajmakh, line 148: just return `err`

shajmakh:
/lgtm
/approve
thanks for addressing the open points!

PR #1274, opened 2025-04-30
api: e2e: serial: initialize slice with correct length in CloneList

SargunNarula:
The CloneList function previously created a slice with length 0 and only set capacity, which led to a runtime panic when accessing ret[idx].


ffromani:
yes, this is my fault, thanks for catching up. Please let's add some unit tests to this code.

ffromani:
superceded by #1297

PR #1285, opened 2025-05-06
CNF-17778: CNF-17779: schedulercontroller: High availability by default

Tal-or:
Changing the topology-aware scheduler behavior to have HA by default.

Starting from OCP 4.20 the scheduler HA is enabled by default (previously it was not)
and replicas number is equal to the number of control-plane nodes.

In case the user wants to disable HA (retaining the old behavior)
he needs to explicitly set `spec.Replicas` to 1.

### DIFF
--- a/internal/controller/numaresourcesscheduler_controller.go
+++ b/internal/controller/numaresourcesscheduler_controller.go
@@ -62,7 +62,6 @@ type NUMAResourcesSchedulerReconciler struct {
 	Scheme             *runtime.Scheme
 	SchedulerManifests schedmanifests.Manifests
 	Namespace          string
-	AutodetectReplicas int
 }
 
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles,verbs=*
--- a/internal/controller/numaresourcesscheduler_controller.go
+++ b/internal/controller/numaresourcesscheduler_controller.go
@@ -27,8 +27,10 @@ import (
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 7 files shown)


internal/controller/numaresourcesscheduler_controller.go
```diff
@@ -186,6 +197,11 @@ func (r *NUMAResourcesSchedulerReconciler) syncNUMASchedulerResources(ctx contex
 
 	schedSpec := instance.Spec.Normalize()
 	cacheResyncPeriod := unpackAPIResyncPeriod(schedSpec.CacheResyncPeriod)
+	replicas, err := r.computeSchedulerReplicas(ctx, schedSpec)
+	if err != nil {
+		return nropv1.NUMAResourcesSchedulerStatus{}, fmt.Errorf("failed to compute scheduler replicas: %w", err)
```
ffromani, line 201: the benefit of the old approach was that the computation was done sporadically, so it was unlikely to fail.
I'm wondering if we should cache the value or if we should recompute every time. Ideally we should only recompute 1. when the user asks so or 2. when the cluster changes.

ffromani (commented):
I agree moving the autodetection into reconcile loop is probably better. I have some (minorish?) concerns about the reliability though. I need to think more about it

internal/controller/numaresourcesscheduler_controller.go
ffromani: In general listing all objects of some kind in every reconciliation step makes me uncomfortable (and yes, we do that already, and yes, I'm already uncomfortable).
In this case I expect us to reconcile if the object changed though, which should be still OK.
Don't we need to watch on nodes though? should we react if nodes are automatically added or deleted?

let's take a look about other operators (scheduler operator?) do. As much as I like this concept, I would avoid to be smarter than other standard operators.

Tal-or: I can look for an option to enter this flow only if the reconciliation triggered by event which is related to node objects.
But I don't think this is how operators are suppose to work

Tal-or: >last but not least: how we envision the autodection on hypershift? let's try hard to offer a consistent experience on hypershift and openshift

Yes, I will dedicate a separate PR for that.

ffromani: > I can look for an option to enter this flow only if the reconciliation triggered by event which is related to node objects.
> But I don't think this is how operators are suppose to work

it depends on how we define the autosizing feature. Let's first review how similar operator works.

Tal-or: > In general listing all objects of some kind in every reconciliation step makes me uncomfortable (and yes, we do that already, and yes, I'm already uncomfortable).

Since we're using the controllerruntime framework, we should not be worried about redundant `List` calls, because the controllerruntime client first reads from cache so it doesn't actually performs an API call to the server:
https://github.com/openshift-kni/numaresources-operator/blob/b72a1b849f681d7d3e95cc45889928363a38aa3a/vendor/sigs.k8s.io/controller-runtime/pkg/client/client.go#L361
We are setting the cache here:
https://github.com/openshift-kni/numaresources-operator/blob/860724c25dae93a091905e643c9170f760da7ba0/cmd/main.go#L253

ffromani: right, this is probably good enough in this case.

internal/controller/numaresourcesscheduler_controller.go
```diff
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
```
ffromani, line 22: misplaced import

internal/controller/numaresourcesscheduler_controller.go
```diff
@@ -171,13 +201,22 @@ func isDeploymentRunning(ctx context.Context, c client.Client, key nropv1.Namesp
...
+	if err := r.Client.List(ctx, nodeList, &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{
+			"node-role.kubernetes.io/control-plane": "",
+		}),
+	}); err != nil {
+		return schedSpec.Replicas, err
 	}
-	v := int32(r.AutodetectReplicas)
-	return &v
+	replicas := ptr.To(int32(len(nodeList.Items)))
```
ffromani, line 217: not needed, we can use a local variable and return its address. But not worth resubmitting just for this

internal/controller/numaresourcesscheduler_controller.go
```diff
@@ -171,13 +201,22 @@ func isDeploymentRunning(ctx context.Context, c client.Client, key nropv1.Namesp
...
+func (r *NUMAResourcesSchedulerReconciler) computeSchedulerReplicas(ctx context.Context, schedSpec nropv1.NUMAResourcesSchedulerSpec) (*int32, error) {
+	// do not autodetect if explicitly set by the user
+	if schedSpec.Replicas != nil {
+		return schedSpec.Replicas, nil
+	}
+	nodeList := &corev1.NodeList{}
+	if err := r.Client.List(ctx, nodeList, &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{
+			"node-role.kubernetes.io/control-plane": "",
+		}),
```
ffromani, line 213: good enough for now, let's clean up later

ffromani (commented):
/approve
/lgtm

thanks, nits inside are not worth a resubmit

PR #1296, opened 2025-05-11
Onboard NROP must gather to Konflux

rbaturov:
* Add initial tekton files for build pipelines.  
* Add rpm locking for must gather container (to be enabled with hermetic builds)
* Add build pipeline file containing metadata added to avoid code duplication across tekton files.
* Update tekton tasks images.
* Add renovate configuration and use pinned digest for dockerfiles


.konflux/must-gather/must-gather.konflux.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM brew.registry.redhat.io/rh-osbs/openshift-ose-must-gather-rhel9:v4.20@sha256:23a673af8dfe9e048c4a348a671d8c3b741e4a2fff37b54c0f825b9810940d99 as mgbuilder
+FROM registry.redhat.io/openshift4/ose-must-gather-rhel9@sha256:38c4e287bae87451ded2d57b8b87e82375f79d249e0901ed4f0f7aabf42ab4e2 as mgbuilder
```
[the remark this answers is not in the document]
rbaturov: https://catalog.redhat.com/software/containers/openshift4/ose-must-gather-rhel9/652809b17e435c863687fed2?container-tabs=gti

Thanks for catching this.
I took it from here. This is part of 4.18 release.
The tag is the latest.
Should we find one for 4.20?

.tekton/build-pipeline-must-gather.yaml
```diff
@@ -0,0 +1,580 @@
+apiVersion: tekton.dev/v1
```
yanirq, line 1: what is the diff between this file and the existing build-pipeline.yaml file ?  why the duplication ?

.tekton/build-pipeline.yaml
```diff
@@ -138,6 +138,8 @@ spec:
           value: $(params.output-image).git
         - name: ociArtifactExpiresAfter
           value: $(params.image-expires-after)
+        - name: dev-package-managers
```
yanirq, line 141: does this addition really belong here ?

rbaturov: Nice catch. It shouldn't be there.

.tekton/build-pipeline-must-gather.yaml
rbaturov: There are differences, it is not one to one. for example you added the generate-labels task. But there are more than that you added:

```
    - default: "true"
      description: Whether to append the platform string to the end of the build image tag
      name: image-append-platform
      type: string
      
    - default: "false"
      description: Enable dev-package-manager within prefetching task
      name: dev-package-managers
      type: string
```

We can possibly merge these, let's discuss it.

rbaturov: It has been decided to keep these two files separately and not try to merge them due to the changes.

.konflux/must-gather/must-gather.konflux.Dockerfile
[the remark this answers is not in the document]
yanirq: ok no , this is something we get only in GA (which at the time i am writing this its 4.18)
see if you can use the latest tag (or even without if it is pulling the latest) 
we need to remember to pin the version when we GA

PR #1297, opened 2025-05-12
api: e2e: serial: initialize slice with correct length in CloneList

ffromani:
The CloneList function previously created a slice with length 0 and only set capacity, which led to a runtime panic when accessing ret[idx].

replaces: #1274

### DIFF
--- a/api/v1/helper/nodegroup/nodegroup.go
+++ b/api/v1/helper/nodegroup/nodegroup.go
@@ -161,7 +161,7 @@ func flattenTrees(trees []Tree) []*mcov1.MachineConfigPool {
 }
 
 func CloneList(nodeGroups []nropv1.NodeGroup) []nropv1.NodeGroup {
-	ret := make([]nropv1.NodeGroup, 0, len(nodeGroups))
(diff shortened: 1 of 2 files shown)


Tal-or:
> /cherry-pick release-4.19

If this code is being used only by the serial test, we can backport it but don't have to.

ffromani:
> > /cherry-pick release-4.19
> 
> If this code is being used only by the serial test, we may backport it but don't have to.

correct, but it's a fix in a general package which we may end up use (and backport) later on, so I prefer to backport, especially considering the fix is simple and uninitrusive.

Tal-or:
> > > /cherry-pick release-4.19
> > 
> > 
> > If this code is being used only by the serial test, we may backport it but don't have to.
> 
> correct, but it's a fix in a general package which we may end up use (and backport) later on, so I prefer to backport, especially considering the fix is simple and uninitrusive.

Sure, I just wanted to emphasis the (forgotten) fact that fixes to the serial suite not require a backport.
In this case this is a fix for a general package used by the serial suite, so that's ok

shajmakh (approved):

PR #1306, opened 2025-05-18
 Activate production registry mapping for Konflux bundle

yanirq:


yanirq:
/hold

test conforma first - then perform final alignments.

PR #1309, opened 2025-05-20
internal: ds: fix condition of the wait loop

shajmakh:
The Eventually should return true if the condition is met, in the old version of the condition, when the condition is met it returns false (while should be the opposite), causing the loop context to exceed the timeout without the expected success.


ffromani (commented):
/approve
/lgtm

nice. How did we miss this so far?

shajmakh:
/unhold
7 tests affected and all ran and passed.

ffromani (commented):
/approve
/lgtm

nits are not worth a resubmit

test/e2e/serial/tests/workload_placement_no_nrt.go
```diff
@@ -192,11 +193,8 @@ func updateInfoRefreshPause(fxt *e2efixture.Fixture, newVal nropv1.InfoRefreshPa
 		err = fxt.Client.Get(context.TODO(), nroKey, updatedObj)
 		Expect(err).ToNot(HaveOccurred())
 
-		// TODO replace with updatedObj.Status.MachineConfigPools[0].Config.InfoRefreshPause when OCPBUGS-16058 is resolved.
-		// currently the functionality of the new flag works bt is matter of not reflecting it in resources status. also we
-		// wait for the respective ds to get ready with the new pods so we should be safe
-		if *updatedObj.Spec.NodeGroups[0].Config.InfoRefreshPause != newVal {
-			klog.Warningf("resource status is not updated yet: expected %q found %q", newVal, *updatedObj.Spec.NodeGroups[0].Config.InfoRefreshPause)
+		if *updatedObj.Status.MachineConfigPools[0].Config.InfoRefreshPause != newVal {
```
Tal-or, line 205: Will the status info present in `updatedObj.Status.MachineConfigPools[]` for the hypershift case as well?

shajmakh: addressed, thanks for pointing it out.

PR #1313, opened 2025-05-26
controller: minimize update calls (reduced)

ffromani:
Extract the remaining easy/uncontroversial changes from #1194 . The next optimization will require a different approach. See: https://github.com/openshift-kni/numaresources-operator/pull/1194#issuecomment-2948265049


Tal-or:
/lgtm
The rest of the work that were discussed is in progress

PR #1315, opened 2025-05-27
WIP: sched controller: InformerMode default to Shared

shajmakh:
TBD

### DIFF
--- a/api/v1/numaresourcesscheduler_normalize_test.go
+++ b/api/v1/numaresourcesscheduler_normalize_test.go
@@ -37,7 +37,7 @@ func TestNUMAResourcesSchedulerSpecNormalize(t *testing.T) {
 	cacheResyncPeriodCustom := 42 * time.Second
 	cacheResyncDebugCustom := CacheResyncDebugDisabled
 	cacheResyncDetectionCustom := CacheResyncDetectionAggressive
-	schedInformerCustom := SchedulerInformerShared
+	schedInformerCustom := SchedulerInformerDedicated
 	scoringStrategyCustom := ScoringStrategyParams{
 		Type:      MostAllocated,
 		Resources: []ResourceSpecParams{{Name: "cpu", Weight: 10}, {Name: "memory", Weight: 5}},
--- a/internal/controller/numaresourcesscheduler_controller_test.go
+++ b/internal/controller/numaresourcesscheduler_controller_test.go
@@ -412,7 +412,7 @@ var _ = ginkgo.Describe("Test NUMAResourcesScheduler Reconcile", func() {
 			_, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: key})
 			gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
-			expectCacheParams(reconciler.Client, depmanifests.CacheResyncAutodetect, depmanifests.ForeignPodsDetectOnlyExclusiveResources, depmanifests.CacheInformerDedicated)
+			expectCacheParams(reconciler.Client, depmanifests.CacheResyncAutodetect, depmanifests.ForeignPodsDetectOnlyExclusiveResources, depmanifests.CacheInformerShared)
 		})
 
 		ginkgo.It("should allow to set aggressive resync detection mode in configmap", func() {
@@ -427,14 +427,14 @@ var _ = ginkgo.Describe("Test NUMAResourcesScheduler Reconcile", func() {
 			_, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: key})
 			gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
-			expectCacheParams(reconciler.Client, depmanifests.CacheResyncAutodetect, depmanifests.ForeignPodsDetectAll, depmanifests.CacheInformerDedicated)
+			expectCacheParams(reconciler.Client, depmanifests.CacheResyncAutodetect, depmanifests.ForeignPodsDetectAll, depmanifests.CacheInformerShared)
 		})
 
-		ginkgo.It("should configure by default the informerMode to be Dedicated", func() {
+		ginkgo.It("should configure by default the informerMode to be Shared", func() {
 			key := client.ObjectKeyFromObject(nrs)
 			_, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: key})
 			gomega.Expect(err).ToNot(gomega.HaveOccurred())
-			expectCacheParams(reconciler.Client, depmanifests.CacheResyncAutodetect, depmanifests.CacheResyncOnlyExclusiveResources, depmanifests.CacheInformerDedicated)
+			expectCacheParams(reconciler.Client, depmanifests.CacheResyncAutodetect, depmanifests.CacheResyncOnlyExclusiveResources, depmanifests.CacheInformerShared)
 		})
 
 		ginkgo.It("should allow to change the informerMode to Shared", func() {
--- a/api/v1/numaresourcesscheduler_defaults.go
+++ b/api/v1/numaresourcesscheduler_defaults.go
@@ -25,7 +25,7 @@ import (
 const (
 	defaultCacheResyncPeriod    = 5 * time.Second
 	defaultCacheResyncDebug     = CacheResyncDebugDumpJSONFile
-	defaultSchedulerInformer    = SchedulerInformerDedicated
+	defaultSchedulerInformer    = SchedulerInformerShared
 	defaultCacheResyncDetection = CacheResyncDetectionRelaxed
 	defaultScoringStrategy      = LeastAllocated
 	defaultReplicas             = int32(1)
--- a/api/v1/numaresourcesscheduler_defaults_test.go
+++ b/api/v1/numaresourcesscheduler_defaults_test.go
@@ -37,7 +37,7 @@ func TestSetDefaults_NUMAResourcesSchedulerSpec(t *testing.T) {
 	cacheResyncPeriodCustom := 42 * time.Second
 	cacheResyncDebugCustom := CacheResyncDebugDisabled
 	cacheResyncDetectionCustom := CacheResyncDetectionAggressive
-	schedInformerCustom := SchedulerInformerShared
+	schedInformerCustom := SchedulerInformerDedicated
 	scoringStrategyCustom := ScoringStrategyParams{
 		Type:      MostAllocated,
 		Resources: []ResourceSpecParams{{Name: "cpu", Weight: 10}, {Name: "memory", Weight: 5}},
--- a/api/v1/numaresourcesscheduler_types.go
+++ b/api/v1/numaresourcesscheduler_types.go
@@ -38,10 +38,11 @@ const (
 type SchedulerInformerMode string
 
 const (
-	// SchedulerInformerDedicated makes the NodeResourceTopologyMatch plugin use the default framework informer.
+	// SchedulerInformerShared makes the NodeResourceTopologyMatch plugin use the default framework informer. Default.
 	SchedulerInformerShared SchedulerInformerMode = "Shared"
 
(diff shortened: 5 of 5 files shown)


internal/controller/numaresourcesscheduler_controller.go
```diff
@@ -224,7 +226,11 @@ func (r *NUMAResourcesSchedulerReconciler) syncNUMASchedulerResources(ctx contex
 	klog.V(4).Info("SchedulerSync start")
 	defer klog.V(4).Info("SchedulerSync stop")
 
-	schedSpec := instance.Spec.Normalize()
+	//intentionally define the feature gate here instead of pulling it because some verions may not have it supported at all
+	KubeletPodResourcesListUseActivePods := "KubeletPodResourcesListUseActivePods"
+	useActivePodsOnly := utilfeature.DefaultFeatureGate.Enabled(featuregate.Feature(KubeletPodResourcesListUseActivePods))
+
+	schedSpec := instance.Spec.Normalize(useActivePodsOnly)
```
ffromani, line 233: not really. This is well-intentioned but doesn't work.
The problem is now you import the packages from the version we depend on, and you consume that value. This means the default value hardcoded in the libraries we import and you use here will be the truth value. But this is not what we want. We want to change the default depending on what is actually on the cluster.

So what we should do, I think, is rather

- detect the full cluster version (X.Y.Z) and pass this information to the `numresourcesschedulereconcciler` object.
- keep the normalization stage as-is
- add a pre-normalization stage. IFF the user did NOT set a value, compute the value based on the cluster version
- call the API normalization (which is static and must not depend on anything else) as usual

ffromani: the only way is depending on the version once we have a fixed-version in the kubelet.

internal/controller/numaresourcesscheduler_controller.go
```diff
@@ -44,6 +44,8 @@ import (
 	k8swgmanifests "github.com/k8stopologyawareschedwg/deployer/pkg/manifests"
 	k8swgrbacupdate "github.com/k8stopologyawareschedwg/deployer/pkg/objectupdate/rbac"
 
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
```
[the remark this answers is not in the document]
ffromani: in addition: OCP doesn't usually expose kube FGs directly. While this can be an argument to use the `library-go` package (albeit the extra deps are not trivial), we should validate that we have visibility on that FG.

api/v1/numaresourcesscheduler_normalize.go
```diff
@@ -16,9 +16,39 @@
...
+	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
+	
+	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
+)
+
+const (
+	// ActivePodsResourcesSupportSince defines the OCP version which started to support the fixed kubelet
+	// in which the PodResourcesAPI lists the active pods by default
+	ActivePodsResourcesSupportSince = "4.20.999"
```
shajmakh, line 29: this must be adjusted once we know since which OCP version is the kubelet fix supported

internal/controller/numaresourcesscheduler_controller.go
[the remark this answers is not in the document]
shajmakh: clarified offline

internal/controller/numaresourcesscheduler_controller.go
shajmakh: clarified offline

api/v1/numaresourcesscheduler_normalize.go
ffromani: correct. This is also in implementation detail, so the constant should start with a lowecase letter.

api/v1/numaresourcesscheduler_normalize_test.go
```diff
@@ -314,3 +316,60 @@ func TestNUMAResourcesSchedulerSpecNormalize(t *testing.T) {
...
 }
+
+func TestNUMAResourcesSchedulerSpec_PreNormalize(t *testing.T) {
+	schedInformer := defaultSchedulerInformer
+
+	schedInformerCustom := SchedulerInformerShared
+
+	greaterVersion, _ := platform.ParseVersion("4.20.1000")
+	equalVersion, _ := platform.ParseVersion(ActivePodsResourcesSupportSince)
+	lessVersion, _ := platform.ParseVersion("4.20.0")
```
ffromani, line 327: here we check per Z stream, which is fine and obviously necessary. I wonder if we need to check per Y stream. I tend to say no. WDYT?

internal/controller/numaresourcesscheduler_controller.go
```diff
@@ -58,12 +59,24 @@ import (
 	"github.com/openshift-kni/numaresources-operator/pkg/status"
 )
 
+const (
+	// ActivePodsResourcesSupportSince defines the OCP version which started to support the fixed kubelet
+	// in which the PodResourcesAPI lists the active pods by default
+	activePodsResourcesSupportSince = "4.20.999"
```
shajmakh, line 65: This version value should be modified per Ystream branch (main is 4.20.z, release-4.19 will use 4.19.z where z is the first supported stream in this minor release.)

internal/controller/numaresourcesscheduler_controller.go
```diff
@@ -225,6 +238,10 @@ func (r *NUMAResourcesSchedulerReconciler) syncNUMASchedulerResources(ctx contex
 	klog.V(4).Info("SchedulerSync start")
 	defer klog.V(4).Info("SchedulerSync stop")
 
+	if r.PlatformInfo.Platform == platform.OpenShift {
+		preNormalize(&instance.Spec, r.PlatformInfo.Version)
```
ffromani, line 242: I would probably have renamed `preNormalize` in something like `platformNormalization` and pushed the `if` block inside the function, but it's largely personal taste. The current approach is not wrong or bad.

internal/controller/numaresourcesscheduler_controller.go
```diff
@@ -293,6 +310,22 @@ func (r *NUMAResourcesSchedulerReconciler) syncNUMASchedulerResources(ctx contex
 	return schedStatus, nil
 }
 
+func preNormalize(spec *nropv1.NUMAResourcesSchedulerSpec, version platform.Version) {
+	parsedVersion, _ := platform.ParseVersion(activePodsResourcesSupportSince)
+	ok, err := version.AtLeast(parsedVersion)
+	if err != nil {
+		klog.Infof("failed to compare version %v with %v, err %v", parsedVersion, version, err)
```
ffromani, line 317: so return here?

internal/controller/numaresourcesscheduler_controller.go
```diff
@@ -293,6 +310,22 @@ func (r *NUMAResourcesSchedulerReconciler) syncNUMASchedulerResources(ctx contex
...
+	}
+
+	if !ok {
+		return
+	}
+
+	if spec.SchedulerInformer == nil {
+		spec.SchedulerInformer = ptr.To(nropv1.SchedulerInformerShared)
+		klog.InfoS("SchedulerInformer default is overridden", "PlatformVersion", version.String(), "kubeletPodResourcesListACtivePodsByDefault", "enabled", "SchedulerInformer", &spec.SchedulerInformer)
```
ffromani, line 326: yes, we should log but probably at levels >= 4 and probably `kubeletPodResourcesListACtivePodsByDefault` should not be mentioned

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -631,6 +633,137 @@ var _ = ginkgo.Describe("Test NUMAResourcesScheduler Reconcile", func() {
...
+
+				expectCacheParams(reconciler.Client, depmanifests.CacheResyncAutodetect, depmanifests.CacheResyncOnlyExclusiveResources, string(informerMode))
+			})
+		})
+	})
+})
+
+var _ = ginkgo.Describe("Test scheduler spec PreNormalize", func() {
+	ginkgo.When("Spec.SchedulerInformer is not set by the user", func() {
+		ginkgo.It("should override default informer to Shared if kubelet is fixed - started version", func() {
```
ffromani, line 722: `started`?

internal/controller/numaresourcesscheduler_controller.go
shajmakh: good point, thanks, fixed

internal/controller/numaresourcesscheduler_controller.go
shajmakh: updated, looks cleaner when the condition is internal in the function. I have updated the condition to apply to hypershift too as I missed that from the start, and updated the tests accordingly. Thanks for raising this point.

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -631,6 +633,167 @@ var _ = ginkgo.Describe("Test NUMAResourcesScheduler Reconcile", func() {
...
+
+				reconciler.PlatformInfo = reconcilerPlatInfo
+
+				key := client.ObjectKeyFromObject(nrs)
+				_, err = reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: key})
+				gomega.Expect(err).ToNot(gomega.HaveOccurred())
+				expectedInformer := depmanifests.CacheInformerDedicated
+				if reconcilerPlatInfo.Platform == platform.OpenShift || reconcilerPlatInfo.Platform == platform.HyperShift {
+					expectedInformer = depmanifests.CacheInformerShared
+				}
```
ffromani, line 661: this should be a test parameter so it's trivial to verified. If we want to determine implicitly it should also be a function of version, which is basically rewriting the production code logic. Having it explicit bypasses this problem.

ffromani:
/hold

let's fix the tests before to merge

shajmakh:
the failure is legit and is about not found CRD yaml. will add a fix.

shajmakh:
needs https://github.com/openshift-kni/numaresources-operator/pull/1507 to fix CI E2E

ffromani (commented):
CI failures seems glitches
I see the test comment not addressed, is that intentional?
The commit message is great, but please be aware that if you reference just a commit, everyone (human or tooling) assumes it's a commit in the same repo; if it is a commit on a different repo, you need to fully qualify it (repo@commit or equivalent)

Besides that, great job, kudos!

internal/controller/numaresourcesscheduler_controller_test.go
shajmakh: I see your point and makes sense indeed, fixed by passing the expected value.

internal/controller/numaresourcesscheduler_controller.go
```diff
@@ -293,6 +308,27 @@ func (r *NUMAResourcesSchedulerReconciler) syncNUMASchedulerResources(ctx contex
...
+		return
+	}
+
+	if !ok {
+		return
+	}
+
+	if spec.SchedulerInformer == nil {
+		spec.SchedulerInformer = ptr.To(nropv1.SchedulerInformerShared)
+		klog.V(4).InfoS("SchedulerInformer default is overridden", "Platform", platInfo.Platform, "PlatformVersion", platInfo.Version.String(), "SchedulerInformer", &spec.SchedulerInformer)
```
ffromani, line 329: nit: always prefer active form:
```
klog.V(4).InfoS("SchedulerInformer: override default based on platform detection", "Platform", platInfo.Platform, "PlatformVersion", platInfo.Version.String(), "SchedulerInformer", &spec.SchedulerInformer)
```

ffromani (commented):
/approve
/lgtm

ffromani:
/test ci-e2e-install-hypershift

hypershift is acting today. Let's give it another go.

internal/controller/numaresourcesscheduler_controller.go
shajmakh: I intentially used past form to ensure that the overide actually happened and was successful, however this is not an immediate update so the active form would fit too

PR #1316, opened 2025-05-27
CNF-18077: rte: log: initialize controller-runtime

ffromani:
controller-runtime log package has an interesting deferring log mechanism which require initialization within a timeout. Otherwise a scary stacktrace is emitted and no logs are shown.
Make sure to initialize this log rerouting to the main klog logger.


shajmakh:
Thanks for catching and filing this. 
/lgtm

PR #1367, opened 2025-06-07
[release-4.16] Backport Konflux changes to 4.16

yanirq:


.konflux/operator/konflux.Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_9_golang_1.22@sha256:9febad53eab6f5b6d1f9a32c7b5f22afdebd6d7735e08f068e0dcbe2ffa6ca64 as builder
```
shajmakh, line 1: I'm uncertain if we want go 1.22 or 1.21 but I see openshift is using 1.21 while in midstream we used to consume 1.22 :/

yanirq: I kept the m/s convention.  any breaking changes from what we currently have needs to be examined separately from the konflux migration

PR #1458, opened 2025-06-18
[release-4.20]: Konflux: Catalog configuration for 4.20

yanirq:
- Introduce a fbc workflow (manual)
- Introduce a fbc workflow (automatic)
- Setup tekton fbc configuration
- Introduce a 'run-script' workflow
  - It runs 'konflux-build-catalog.sh' to render a catalog-template.in.yaml file
    That way we can check if access to 'registry.redhat.io' is authorized.
- Add mirroring file


shajmakh:
Discussed offline, this is part of the onboarding and we need it. It will however need couple of rebases and fixes according to that the conforms which as stated by Yanir will be addressed in upcoming PRs. 
/approve
/lgtm

PR #1465, opened 2025-06-19
e2e: change deprecated reboot labels to new reboot label

rshemtov13:
Changing the depricated reboot_required labels to use the new reboot label


ffromani:
/retitle e2e: change deprecated reboot labels to new reboot label

PR #1514, opened 2025-06-25
Changing SSL client key and cert to ENV var

rbaturov:


.konflux/must-gather/rpms.lock.yaml
```diff
@@ -18,5 +18,12 @@ arches:
     name: rsync
     evr: 3.2.3-19.el9
```
shajmakh, line 19: I'm missing what us the difference between `registry.redhat.io/rhel9-4-els/rhel-minimal` and `registry.redhat.io/rhel9-4-els/rhel-minimal:9.4` that made the lockfile generating miss the binaries? I think this also needs clarification in the commit message. Let's also clarify please how did you use the image while generating the lockfile, did you replace the one in the dockerfile or is it the one you ran the container with?

shajmakh: usually we need to align the final base image in the dockerfile with the one we use to generate the lockfile

rbaturov: @shajmakh Thanks for noticing. One of the images in the commit message was wrong.
I have changed the commit message, you can review it now.

shajmakh: Thanks Ronny it is more clear now and makes sense. 
I believe 4.20 should use 9.6 according to the OCP streams schedule. This fix is however is needed and upgrading to 9.6 can be done in a future PR.

PR #1527, opened 2025-06-27
lint: add gci linter to enforce import ordering

ffromani:
add and use the gci linter to enforce the import ordering.
The adopted ordering is not the nicest nor the best, but it's the closest to our existing unwritten standard and can be
now enforced automatically, which is more than good enough.

Subtrees intentionally excluded by linting:
- api: because autogeneration
- cmd: because kubebuilder scaffolding requirement

To prevent cascading failures, fix all our tree using the tool before we enable the linter. PTAL to individual commits to verify the commandline used.

GCI is: https://github.com/daixiang0/gci

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,7 +16,11 @@ issues:
   exclude-rules:
     - path: "api/*"
       linters:
+        - gci
         - lll
+    - path: "cmd/*"
+      linters:
+        - gci
     - path: "internal/*"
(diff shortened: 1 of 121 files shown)


ffromani:
/cc @Tal-or 

we talked about this in the not-so-distant past

shajmakh:
Thanks!
/lgtm

PR #1536, opened 2025-06-30
hack: add a script to reformat the imports

ffromani:
add a helper script which reads the sort ordering expectation from our .golangci.yml (source of truth) and automatically constructs the recommended command line to run to conform.

Depends on python3 and gci: https://github.com/daixiang0/gci

### DIFF
--- a/hack/sort-imports.py
+++ b/hack/sort-imports.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python
+
+import argparse
+import os
+import subprocess
+import sys
+import yaml
+
+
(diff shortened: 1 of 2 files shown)


hack/sort-imports.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python
```
shajmakh, line 1: why in python?

Tal-or (commented):
who's the caller?
How this script can be used?

Can we add a simple make target `sort-imports` that will do all the work for us.
We can even force the call to this make target by using github pre-commits

PR #1537, opened 2025-06-30
e2e:config: support old RTE config format

Tal-or:
rte configmap API has changed at 4.18
legacy version:
https://github.com/openshift-kni/numaresources-operator/blob/release-4.16/rte/pkg/config/config.go#L41 updated version (4.18+):
https://github.com/openshift-kni/numaresources-operator/blob/release-4.18/rte/pkg/config/config.go#L48

The test code is using the API structures from main (which is same as 4.18) but the same test code is being exercised on versions which are older than 4.18. The yaml.Umarshal call failed to parse the structure properly and that's why we're getting an empty string and the comparison failed.

In order to fix that we're adding a fallback method that will try to craft rte config from the legacy config.

In addition replace Unmarshal -> UnmarshalStrict to catch the error when data is the old format but the struct is the new format.

### DIFF
--- a/test/internal/configuration/configuration.go
+++ b/test/internal/configuration/configuration.go
@@ -23,11 +23,14 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 
 	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
 	nrtv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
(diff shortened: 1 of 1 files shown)


ffromani (commented):
/approve

thanks, very nice catch

Tal-or:
> Thanks Talor, have this been verified on both versions old and new?

It was tested only for the old version.
let me check if someone can help me test on new version, I don't have a setup atm

ffromani:
/test ci-e2e-install-hypershift

it's a glitch or a real issue?

rshemtov13:
@Tal-or I'll test on the new version in just a bit and update you

rshemtov13:
test passed after the changes on new and older versions

ffromani:
/test images
/test ci-index-tested-numaresources-operator-bundle
/test ci-unit

lovely timing.

ffromani:
succeeded before, taking the risk.

PR #1564, opened 2025-07-03
#1539 cleanup: konflux: mustgather: remove redundant lines and restore `tar`

abraham2512:
A follow-up cleanup of #1539, please see commits for details.


.konflux/must-gather/redhat.repo
```diff
@@ -1,28 +1,3 @@
-#
-# Certificate-Based Repositories
```
shajmakh, line 2: because the file is already part of the project, we need to provide why we update it now, you may mention in the commit message that this is a cleanup of the `.repo` file from irrelevant comments and unneeded RPM repo as no binary needs it in order to be installed

shajmakh:
/lgtm
/approve
Thanks!

Issue #1257, opened 2025-04-15
reintroduce removed tier2 e2e tests which depend on performance profile


ffromani:
reimplement the tests removed in https://github.com/openshift-kni/numaresources-operator/pull/1256 using dynamic client, e.g. without using the performance profile API package. No other change is expected.
We can use the unstructured package, e.g. https://github.com/k8stopologyawareschedwg/deployer/blob/main/pkg/objectupdate/sched/render.go

ffromani:
fixed by #1579

PR #1579, opened 2025-07-06
[AA] e2e: serial: reintroduce test 75354 using dymanic client

rshemtov13:
These changes where AI-Assisted (hense the AA tag)

Initial prompt:
```
I want to show you a test that was removed which needs to be reintroduced in a new way. I'll give you the code of the test that was removed here: <code-was-copied-here>.

This test was removed because we removed the dependencies for using the performance profile API, I want you to change the test file so that the changes that were made through using the performance profile API would be done with a dynamic client instead.
```
This PR serves to add test_id:75354 which is one of the missing tests from #1257 after it was removed in #1256.

More manual refinement of the test's code was needed after the model proposed some changes, and deeper explanations were required for the model to return any good results from the prompt.

Assisted-by: Cursor AI
Assisted-by-model: claude-4-sonnet

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -30,14 +31,20 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
 	k8swait "k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/klog/v2"
+	"k8s.io/kubelet/config/v1beta1"
 	"k8s.io/utils/ptr"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
 	configv1 "github.com/openshift/api/config/v1"
 	machineconfigv1 "github.com/openshift/api/machineconfiguration/v1"
@@ -52,11 +59,14 @@ import (
 	"github.com/openshift-kni/numaresources-operator/api/v1/helper/namespacedname"
 	"github.com/openshift-kni/numaresources-operator/api/v1/helper/nodegroup"
 	inthelper "github.com/openshift-kni/numaresources-operator/internal/api/annotations/helper"
+	nropmcp "github.com/openshift-kni/numaresources-operator/internal/machineconfigpools"
 	"github.com/openshift-kni/numaresources-operator/internal/nodegroups"
 	intnrt "github.com/openshift-kni/numaresources-operator/internal/noderesourcetopology"
 	intobjs "github.com/openshift-kni/numaresources-operator/internal/objects"
+	"github.com/openshift-kni/numaresources-operator/internal/podlist"
 	"github.com/openshift-kni/numaresources-operator/internal/relatedobjects"
 	"github.com/openshift-kni/numaresources-operator/internal/wait"
+	"github.com/openshift-kni/numaresources-operator/pkg/kubeletconfig"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectnames"
 	"github.com/openshift-kni/numaresources-operator/pkg/status"
 	"github.com/openshift-kni/numaresources-operator/pkg/validation"
@@ -66,6 +76,7 @@ import (
 	e2eclient "github.com/openshift-kni/numaresources-operator/test/internal/clients"
 	"github.com/openshift-kni/numaresources-operator/test/internal/configuration"
 	e2efixture "github.com/openshift-kni/numaresources-operator/test/internal/fixture"
+	"github.com/openshift-kni/numaresources-operator/test/internal/images"
 	e2enrt "github.com/openshift-kni/numaresources-operator/test/internal/noderesourcetopologies"
 	"github.com/openshift-kni/numaresources-operator/test/internal/nrosched"
 	"github.com/openshift-kni/numaresources-operator/test/internal/objects"
@@ -105,6 +116,7 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 	var fxt *e2efixture.Fixture
 	var nrtList nrtv1alpha2.NodeResourceTopologyList
 	var nrts []nrtv1alpha2.NodeResourceTopology
+	var dynamicClient dynamic.Interface
 
 	BeforeEach(func() {
 		Expect(serialconfig.Config).ToNot(BeNil())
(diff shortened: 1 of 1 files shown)


test/e2e/serial/tests/configuration.go
```diff
@@ -114,6 +126,12 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
 		fxt, err = e2efixture.Setup("e2e-test-configuration", serialconfig.Config.NRTList)
 		Expect(err).ToNot(HaveOccurred(), "unable to setup test fixture")
 
+		cfg, err := config.GetConfig()
+		Expect(err).ToNot(HaveOccurred(), "unable to get kubeconfig")
+
+		dynamicClient, err = dynamic.NewForConfig(cfg)
+		Expect(err).ToNot(HaveOccurred(), "unable to create dynamic client")
```
shajmakh, line 133: if setting up the dynamic client fails, all configuration.go tests will fail. I think we should avoid that and instead we can move the setup into a function and invoke it only when needed. even not in the beginning of the new test, but rather at the beginning of updatePerformanceProfileFieldUnstructured because it is only used there. Then you can also change the args passed to that function accordingly.

test/e2e/serial/tests/configuration.go
```diff
@@ -1172,3 +1460,45 @@ func isCustomPolicySupportEnabled(nro *nropv1.NUMAResourcesOperator) bool {
...
+	}
+	profile, err := dynamicClient.Resource(performanceProfileGVR).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	err = unstructured.SetNestedField(profile.Object, value, fieldPath...)
+	if err != nil {
+		return err
+	}
+	_, err = dynamicClient.Resource(performanceProfileGVR).Update(ctx, profile, metav1.UpdateOptions{})
```
shajmakh, line 1502: should this part be called within a loop or eventually like we do in structured client?

test/e2e/serial/tests/configuration.go
```diff
@@ -1172,3 +1460,45 @@ func isCustomPolicySupportEnabled(nro *nropv1.NUMAResourcesOperator) bool {
...
+		Version:  "v2",
+		Resource: "performanceprofiles",
+	}
+	profile, err := dynamicClient.Resource(performanceProfileGVR).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	err = unstructured.SetNestedField(profile.Object, value, fieldPath...)
+	if err != nil {
+		return err
```
shajmakh, line 1500: let's rephrase the error and append the original one within a new more clear one pointing to the set operation that failed.

test/e2e/serial/tests/configuration.go
```diff
@@ -1172,3 +1460,45 @@ func isCustomPolicySupportEnabled(nro *nropv1.NUMAResourcesOperator) bool {
...
+// It takes the field path as variadic string arguments and updates the field with the provided value.
+func updatePerformanceProfileFieldUnstructured(ctx context.Context, dynamicClient dynamic.Interface, name string, value interface{}, fieldPath ...string) error {
+	performanceProfileGVR := schema.GroupVersionResource{
+		Group:    "performance.openshift.io",
+		Version:  "v2",
+		Resource: "performanceprofiles",
+	}
+	profile, err := dynamicClient.Resource(performanceProfileGVR).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return err
```
shajmakh, line 1496: let's generate a new error here please

test/e2e/serial/tests/configuration.go
```diff
@@ -1172,3 +1460,45 @@ func isCustomPolicySupportEnabled(nro *nropv1.NUMAResourcesOperator) bool {
...
+	}
+	err = unstructured.SetNestedField(profile.Object, value, fieldPath...)
+	if err != nil {
+		return err
+	}
+	_, err = dynamicClient.Resource(performanceProfileGVR).Update(ctx, profile, metav1.UpdateOptions{})
+	return err
```
shajmakh, line 1503: same here

shajmakh (commented):
Thanks for reviving this. 
Overall I see that the interaction with dynamic client is simple given the use case we are using it for in this context. I left some comments related to handling errors as I learned that this is especially important since this kind of client is usually provides complex errors that are hard to parse.
I also saw that the old PR removing this test also removed another one that had PP involved, do you plan to add the rest in this PR?

rshemtov13:
@shajmakh Thanks for the review comments I'll address them. Regarding the second test that was removed, I wanted to keep the usage of using AA code to just one test and create another PR for it independently.
WDYT, should I implement them both in this PR?

test/e2e/serial/tests/configuration.go
rshemtov13: Yes, correct let me change that to be a part of the function.

test/e2e/serial/tests/configuration.go
rshemtov13: You're right, it should rather be wrapped in an eventually, let me modify it accordingly.

shajmakh:
> @shajmakh Thanks for the review comments I'll address them. Regarding the second test that was removed, I wanted to keep the usage of using AA code to just one test and create another PR for it independently. WDYT, should I implement them both in this PR?

Given that the tests ran enough times to ensure they are stable, I don't see any harm in reviving both of them in this PR.

test/e2e/serial/tests/configuration.go
```diff
@@ -1485,20 +1480,49 @@ func createTAEDeployment(fxt *e2efixture.Fixture, ctx context.Context, name, sch
...
+		profile, err := dynamicClient.Resource(performanceProfileGVR).Get(ctx, name, metav1.GetOptions{})
+		Expect(err).ToNot(HaveOccurred(), "Failed to fetch PerformanceProfile using the dynamic client: %v", err)
+
+		err = unstructured.SetNestedField(profile.Object, value, fieldPath...)
+		Expect(err).ToNot(HaveOccurred(), "Failed to set field %v in PerformanceProfile %s: %v", fieldPath, name, err)
+
+		_, err = dynamicClient.Resource(performanceProfileGVR).Update(ctx, profile, metav1.UpdateOptions{})
+		Expect(err).ToNot(HaveOccurred())
+	}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(Succeed())
+	return nil
```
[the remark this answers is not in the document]
rshemtov13: I agree with your view, I think I'll just go with the approach of changing it to void and changing the invoked instances accordingly.

test/e2e/serial/tests/configuration.go
```diff
@@ -1172,3 +1456,81 @@ func isCustomPolicySupportEnabled(nro *nropv1.NUMAResourcesOperator) bool {
...
+		_, err = dynamicClient.Resource(performanceProfileGVR).Update(ctx, profile, metav1.UpdateOptions{})
+		Expect(err).ToNot(HaveOccurred())
+	}).WithTimeout(10 * time.Minute).WithPolling(30 * time.Second).Should(Succeed())
+	return
+}
+
+// newDynamicClient creates and returns a new dynamic client for Kubernetes API operations.
+// This helper function should be called once per test to initialize the dynamic client.
+func newDynamicClient() dynamic.Interface {
+	cfg, err := config.GetConfig()
```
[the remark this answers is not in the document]
rshemtov13: Sure will add.

test/e2e/serial/tests/configuration.go
```diff
@@ -452,6 +463,279 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			// a deployment that is expected to fail due to TopologyAffinityError it will otherwise fail for failed scheduling.
+			// We expect that failure to occur only when the bug is present.
+			// Changes are made through the kubeletconfig directly or through performance profile.
+			By("verify owner reference of kubeletconfig")
+			Expect(numTargetedKCOwnerReferences).To(BeNumerically("<", 2)) // so 0 or 1
+
+			if initialTopologyManagerPolicy != v1beta1.NoneTopologyManagerPolicy {
+				if numTargetedKCOwnerReferences == 0 {
+					By("modifying Topology Manager Policy to 'none' under kubeletconfig")
+					Eventually(func(g Gomega) {
```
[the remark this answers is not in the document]
rshemtov13: Actually the go linter is shouting when I removed it. I'll add it back and use g.Expect on all the Expects there, like around the codebase.

[golangci-lint: test/e2e/serial/tests/configuration.go#L760](https://github.com/openshift-kni/numaresources-operator/pull/1579/files#annotation_38082037065)
ginkgo-linter: Success matcher only support a single error value, or function with Gomega as its first parameter (ginkgolinter)

Tal-or (approved):
/lgtm
/approve

PR #1595, opened 2025-07-09
[release-4.19] konflux: must-gather: restore `tar` rpm 

abraham2512:
This commit includes the tar source rpm in the lockfile which was missed. This is needed for nrop must-gather to
function correctly.


shajmakh:
For future reference, the package rpm was removed in https://github.com/openshift-kni/numaresources-operator/commit/0e12e234ffee231f27e08d317aa99ca30e8f9519 due to a bug in the lockfile generator tool. 
/lgtm
/approve
Thanks!

PR #1598, opened 2025-07-09
[release-4.18] controller: test: verify RTE has a critical priority class 

shajmakh:
bump deployer and verify RTE has a critical priority class. see commits for details.


go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/openshift-kni/numaresources-operator
 
-go 1.22.0
+go 1.23.0
 
-toolchain go1.22.8
+toolchain go1.23.5
```
ffromani, line 5: this should not change

shajmakh: The bump for the go and go toolchain was automatic when I ran `go mod tidy`, which adapts to the minimum Go version required by the dependencies. Investigating this further, it doesn't relate to the deployer version but actually to this bump:
 https://github.com/openshift-kni/numaresources-operator/commit/a553b30b580b6b789b3e3d6f63edd154adf9012a
 where auth2 was bumped from 0.21 (used go 1.18 https://github.com/golang/oauth2/blob/v0.21.0/go.mod) to 0.27 (uses go1.23 https://github.com/golang/oauth2/blob/v0.27.0/go.mod), namely because of https://github.com/golang/oauth2/commit/3f78298beea38fb76a3fbca33e3056f4b7eb5502. 
 If we want to proceed with 1.22 for 4.18 we need to replace oauth2 version to 0.26 at most while preserving goal of   https://github.com/openshift-kni/numaresources-operator/commit/a553b30b580b6b789b3e3d6f63edd154adf9012a.

shajmakh: I see 4.17 bumped to 0.26 https://github.com/openshift-kni/numaresources-operator/pull/1202/files so perhaps that is a good balance between Go version and the intent of the bump.

ffromani: very nice investigation, thanks. Yes, let's downgrade `oauth2` to 0.26

ffromani:
> /hold we need this [370eeb7](https://github.com/openshift-kni/numaresources-operator/commit/370eeb75bb40812dd065c38134de98dd7dd75b39)

OK, but why now? because of the deployer bump?

shajmakh:
> > /hold we need this [370eeb7](https://github.com/openshift-kni/numaresources-operator/commit/370eeb75bb40812dd065c38134de98dd7dd75b39)
> 
> OK, but why now? because of the deployer bump? [EDIT] And because of this [k8stopologyawareschedwg/deployer@e314dab](https://github.com/k8stopologyawareschedwg/deployer/commit/e314dab4e0d6d1c8b41912229e5811e7eb3128fc) right?

Yes exactly. This was intended to be consumed in 4.18 but this PR is the first to bump to deployer version > 0.21.3.

ffromani:
/lgtm

failures tracked in https://github.com/openshift-kni/numaresources-operator/pull/1638 and https://github.com/openshift-kni/numaresources-operator/pull/1635

do we still need the hold?

PR #1628, opened 2025-07-15
AA: tests: switch to dot import for ginkgo and gomega

shajmakh:
AI-assisted modifications, please take a look at the commits for details.

### DIFF
--- a/internal/controller/numaresourcesscheduler_controller_test.go
+++ b/internal/controller/numaresourcesscheduler_controller_test.go
@@ -22,8 +22,6 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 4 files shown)


ffromani (commented):
seems straighforward, but why do we want to change the dot/non-dot import in the first place?

ffromani:
> > seems straightforward, but why do we want to change the dot/non-dot import in the first place?
> 
> Simply to be consistent with the rest of the test files where no conflicts are found, and allow direct access to the packages object. The intention was originally for `numaresourcesscheduler_controller_test.go`, where the testing packages are most commonly used with no conflicting methods by other packages, so there is no reason for the prefixes. Besides that it is the safest thing to try with AA as a start.

fair enough!

PR #1641, opened 2025-07-16
hack: remove `ginkgo.fail-fast` for non-install suites

shajmakh:
For suites that in which the tests are successive, failing fast e.i exit the suite of the first failure saves execution time knowing that the initial steps failed so there would be no point to proceed in checking the next expected results because they are likely to fail as well.

However, running non-install (and non-uninstall) suites would benefit from keeping the suite run alive even on one test failure because normally tests in such suites are independent, and a continuous run would provide more verbose picture about the overall state, and would also give hints and  narrow down the possible reasons, providing more efficient debugging experience.


ffromani:
/approve
/lgtm

do we need backports?

shajmakh:
> /approve /lgtm
> 
> do we need backports?

Thanks. For now I only see the need for when ci-e2e prow lane fails (e.g https://github.com/openshift/release/pull/66829) which is uncommon in older branches. It would be a proactive backport and won't hurt.

PR #1681, opened 2025-07-21
CNF-17780:numaresourcesscheduler: support HA for HyperShift platform

Tal-or:
Add support for scheduler high availability when running on HyperShift platform. On HyperShift, scheduler replicas are deployed on worker nodes, as opposed to standard OpenShift where replicas run on control-plane nodes.

Implement intelligent replica management with the following logic:
- Use worker nodes for replica count calculation on HyperShift
- Use control-plane nodes for replica count calculation on OpenShift
- Prefer odd-sized replica counts (1 or 3) to avoid split-brain scenarios
- Cap maximum replicas at 3 to prevent resource overconsumption
- Return error when no target nodes are found for the platform

The automatic replica detection can be overridden by explicitly setting the replicas field in the NUMAResourcesScheduler custom resource.

Includes comprehensive unit tests covering all platform and node count scenarios.

### DIFF
--- a/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
@@ -62,7 +62,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2025-07-15T16:26:43Z"
+    createdAt: "2025-07-23T11:36:11Z"
     olm.skipRange: '>=4.19.0 <4.20.0'
     operatorframework.io/cluster-monitoring: "true"
     operators.operatorframework.io/builder: operator-sdk-v1.40.0
@@ -615,11 +615,7 @@ spec:
           resources:
           - networkpolicies
           verbs:
-          - create
-          - delete
-          - get
-          - list
-          - update
+          - '*'
         serviceAccountName: numaresources-controller-manager
     strategy: deployment
   installModes:
--- a/internal/controller/numaresourcesscheduler_controller.go
+++ b/internal/controller/numaresourcesscheduler_controller.go
@@ -63,6 +63,9 @@ const (
 	// ActivePodsResourcesSupportSince defines the OCP version which started to support the fixed kubelet
 	// in which the PodResourcesAPI lists the active pods by default
 	activePodsResourcesSupportSince = "4.20.999"
+
+	// MaxSchedulerReplicas is the maximum number of scheduler replicas that can be deployed
+	MaxSchedulerReplicas = 3
 )
 
 type PlatformInfo struct {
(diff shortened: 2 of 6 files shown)


ffromani (commented):
I think you need to resubmit to make CI happy. If so, PTAL to inline ~nits~ comments

internal/controller/numaresourcesscheduler_controller.go
```diff
@@ -221,16 +221,52 @@ func (r *NUMAResourcesSchedulerReconciler) computeSchedulerReplicas(ctx context.
...
+
+	// check if no nodes are found for the target role
+	if replicaCount == 0 {
+		return nil, fmt.Errorf("failed to compute scheduler replicas: no %s nodes found for platform %s", nodeRoleDescription, r.PlatformInfo.Platform)
+	}
+
+	// prefer odd-sized replica counts to avoid split-brain scenarios
+	// cap replicas at 3 for both OpenShift and HyperShift
+	if replicaCount > 3 {
+		replicaCount = 3
```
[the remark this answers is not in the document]
Tal-or: Done.
>super-nit: capping the replicas is a good idea and should be ideally done in its own commit. I won't complain in this case though.

Thanks I'll take that into consideration in similar future scenarios

bundle/manifests/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -615,11 +615,7 @@ spec:
           resources:
           - networkpolicies
           verbs:
-          - create
-          - delete
-          - get
-          - list
-          - update
+          - '*'
```
ffromani, line 618: it is still better to list all the verbs explicitly, but I don't mind enough now. Let's move on.

Tal-or:
> /hold
> 
> The predicates should be restricted to the operator namespace.

there's a limitation in the operatorSDK framework that doesn't allow setting predicates for an object without giving a permission to the operator at the cluster role level. (Yes, even if you want to watch networkpolicieis at a specific namespace).

pkg/objectstate/rte/rte.go
```diff
@@ -211,14 +211,21 @@ func (em *ExistingManifests) State(mf Manifests) []objectstate.ObjectState {
 			Error:    em.errs.Core.APIServerNetworkPolicy,
 			Desired:  mf.Core.APIServerNetworkPolicy.DeepCopy(),
 			Compare:  compare.Object,
-			Merge:    merge.ObjectForUpdate,
+			Merge:    merge.MetadataForUpdate,
```
Tal-or, line 214: Good enough, `merge.ObjectForUpdate` also merges status which networkpolicy doesn't have

ffromani:
/retest-required

the first one is a freebie

ffromani:
```
 INFO[2025-07-24T08:28:22Z] Step ci-e2e-ipi-conf-telemetry failed after 6s. 
INFO[2025-07-24T08:28:22Z] Step phase pre failed after 15s.  
```
hardly this PR's fault

PR #1682, opened 2025-07-21
konflux: catalog and image pinning alignments

yanirq:


.konflux/catalog/.gitignore
```diff
@@ -0,0 +1 @@
+numaresources-operator/catalog.yaml
```
shajmakh, line 1: I'm unsure if we should keep `.gitkeep` if we are explicit about the files to ignore in the `numaresources-operator` directory

PR #1687, opened 2025-07-22
WIP: dangling: delete unused NRTs

shajmakh:
RTE daemonsets are deleted but that isn't cleaning corresponding NRTs. Remove unused NRTS after scanning the relevant DSs.


ffromani:
> I would prefer to fix this setting the owner reference of NRT objects. We have facilities to bind NRT objects to the corresponding nodes. We need to check if we enable this facility, and if not why (likely efficiency reasons, we should not list/get nodes at each NRT update). In general, it is questionable that we return NRT objects when the corresponding DS goes away. But if we need to do this binding, I'd rather use OwnerReference rather than use explicit delete programmatically.

or not? https://github.com/openshift-kni/numaresources-operator/pull/1692

shajmakh:
> I would prefer to fix this setting the owner reference of NRT objects. We have facilities to bind NRT objects to the corresponding nodes. We need to check if we enable this facility, and if not why (likely efficiency reasons, we should not list/get nodes at each NRT update). In general, it is questionable that we return NRT objects when the corresponding DS goes away. But if we need to do this binding, I'd rather use OwnerReference rather than use explicit delete programmatically.

I agree that adding and owner references is the best way to manage related resources and garbage collecting. There are multiple ways to delete the dangling NRTs and I picked this one because with the current RTE duplication bug, adding the ownerReference might come with extra duplication issue of the NRT, e.i same node has multiple NRTs but each belong to different DS, while according to the intended implementation there should be 1:1 relationship between: node:nodePool, nodepool:DS, node:NRT. However the Fix for the RTE duplication is in progress, I will revisit this cleanup after the fix is approved.

ffromani:
all in all it seems that the best (as in "most practical") approach seems to do explicit cleanup, so I'll review again with this in mind

shajmakh:
after clarifying the way forward with the overlapping node sets of target MCPs, we see that the best way to manage this is using the Kubernetes built in capabilities by using ownerReference under the NRTs that would point to the owner DS. That way deleting the owner object (dangling DS) triggers cascading deletion of dependents (i.e. NRTs).

PR #1694, opened 2025-07-24
controller: add `PodAntiAffinity` to RTE DS

shajmakh:
controller: add PodAntiAffinity to RTE DS

We want to avoid having more than 1 RTE pod deployed on node when there
are overlapping nodeGroups specified in the NRO CR.

We cannot prevent the misconfiguration of the cluster node labels with
respect to MCO (e.i a node may match to more than one MCP's node
selector field) but we can mitigate having duplicated RTEs on the same
node.As a result, the order of RTE pods being created following
configuration modifications, like nodes labels, MCP node selectors or any
other adjustments that may relate to the specified nodeGroup(s), will
keep the pod pending starting the second one.

Important note: this change doesn't prevent running the RTE pod on nodes that do not
belong to the specified MCP but still match their nodeSelectors. The
only way to manage that is by not having overlapping node labels in the
first place. The operator depends on the MCO to group nodes of the
cluster into one configuration, but it should not duplicate how MCO
computes which node belongs to which MCP.

### DIFF
--- a/internal/controller/numaresourcesoperator_controller.go
+++ b/internal/controller/numaresourcesoperator_controller.go
@@ -524,8 +524,13 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
 		}
 	}
 
-	// using a slice of poolDaemonSet instead of a map because Go maps assignment order is not consistent and non-deterministic
 	dsPoolPairs := []poolDaemonSet{}
+	err = rteupdate.DaemonSetAffinitySettings(r.RTEManifests.Core.DaemonSet, r.RTEManifests.Core.DaemonSet.Spec.Template.Labels)
+	if err != nil {
+		klog.ErrorS(err, "failed to update RTE affinity settings")
+	}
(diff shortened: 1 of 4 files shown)


internal/controller/numaresourcesoperator_controller.go
```diff
@@ -524,6 +524,19 @@ func (r *NUMAResourcesOperatorReconciler) syncNUMAResourcesOperatorResources(ctx
...
+				{
+					LabelSelector: &metav1.LabelSelector{
+						MatchLabels: r.RTEManifests.Core.DaemonSet.Spec.Template.Labels,
+					},
+					TopologyKey: "kubernetes.io/hostname",
+				},
+			},
+		},
+	}
+
```
ffromani, line 539: this looks correct but should be a helper function in `pkg/objectupdate/rte`. Considered you added a controller test, unit tests are welcome but not needed.

If there is a suitable existing helper in `pkg/objectupdate/rte` feel free to add this code there.

PR #1734, opened 2025-07-31
controller: sched: report DedicatedInformerActive behavior

shajmakh:
We want to be able to see the informer mode of the scheduler as a preparation step for having the informer default change according to the kubelet version.
This commit enables reflecting the informer value (default or configured) as part of the status conditions which is less direct than reporting it in spec and is valid.


shajmakh:
issue during cluster setup
/retest

PR #1748, opened 2025-08-04
controller: sched: adapt that kubelet PodResourcesAPI fix has landed

shajmakh:
The fix has landed in OCP nightly 4.20, make sure this is reflected to consume the new default of schedulerInformer in case it is not set.

### DIFF
--- a/internal/controller/numaresourcesscheduler_controller.go
+++ b/internal/controller/numaresourcesscheduler_controller.go
@@ -60,17 +60,14 @@ import (
 )
 
 const (
-	// ActivePodsResourcesSupportSince defines the OCP version which started to support the fixed kubelet
-	// in which the PodResourcesAPI lists the active pods by default
-	activePodsResourcesSupportSince = "4.20.999"
-
 	// MaxSchedulerReplicas is the maximum number of scheduler replicas that can be deployed
 	MaxSchedulerReplicas = 3
 )
 
 type PlatformInfo struct {
-	Platform platform.Platform
-	Version  platform.Version
+	Platform                     platform.Platform
+	Version                      platform.Version
+	KubeletSupportsActivePodList bool
 }
 
 // NUMAResourcesSchedulerReconciler reconciles a NUMAResourcesScheduler object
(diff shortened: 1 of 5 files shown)


shajmakh:
/hold 
for next nightly

internal/platform/activepodresources/supportversion.go
```diff
@@ -0,0 +1,104 @@
+package activepodresources
```
ffromani, line 1: missing copyright boilerplate (`hack/boilerplate.go.txt`) which makes me realize we don't have a CI rule which ensure all (new) files have the boilerplate, so we should add such CI rule (obviously not related to this PR)

internal/platform/activepodresources/supportversion.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
+)
+
+// This file processes the known OCP platform versions, so far nightly, konflux-nightly, CI, dev-preview, and RC
+
+var (
+	// first build is always the nightly. Yet that doesn't necessarily mean that builds of other lanes
+	// (e.i CI, dev-preview) that produce the build after the specific nightly build date will certainly
+	// have the fix, thus we need to track them separately
+	StableSupportSince           = "4.20.999"
```
ffromani, line 16: we need to make logic on OCP versions, but this seems like a NROP version. Is this right? Ditto for all but `NightlySupportSince`. Is this what you meant in https://github.com/openshift-kni/numaresources-operator/pull/1748/commits/54a5afe71095cdc6dea855371cebac391f5e6923#r2253978141 ?

internal/platform/activepodresources/supportversion_test.go
```diff
@@ -0,0 +1,85 @@
+package activepodresources
```
ffromani, line 1: likewise, missing boilerplate

internal/controller/numaresourcesscheduler_controller.go
```diff
@@ -361,7 +358,13 @@ func platformNormalize(spec *nropv1.NUMAResourcesSchedulerSpec, platInfo Platfor
 		return
 	}
 
-	parsedVersion, _ := platform.ParseVersion(activePodsResourcesSupportSince)
+	leastSupported, err := activepodresources.GetLeastSupportedByBuildType(platInfo.Version)
```
ffromani, line 361: ideally this should be computed once at startup: the platform should not change while the operator is running (but can change across operator restarts obviously).

ffromani (commented):
Thanks, this seems like a good general direction

internal/controller/numaresourcesscheduler_controller.go
shajmakh: good point, I'll see where this goes better

internal/platform/activepodresources/supportversion.go
ffromani: yes, so these constants should be openshift/kubelet versions, and they don't use the `.999` versioning like NROP does

shajmakh:
issue during cluster setup
/retest

ffromani:
obsoleted by #1915

PR #1605, opened 2025-07-10
CNF-18350: Enforce Network Policies (UPDATED)

Tal-or:
Picking up the work from: https://github.com/openshift-kni/numaresources-operator/pull/1253

PR #1801, opened 2025-08-12
bundle: disable networkpolicies again

ffromani:
we're not ready to ship them in the bundle.
Note this commit does *NOT* remove the network policies installed by the operator for the operand, just the ones from the main manifest bundle.


Tal-or:
Is it fine having the NP for the operands installed?
 wouldn't it be easier to revert https://github.com/openshift-kni/numaresources-operator/pull/1605?

PR #1804, opened 2025-08-13
WIP: rbac: controllers: narrow down permissions

ffromani:
narrow down RBAC permissions in two ways:
1. (minor) reduce verb requirements; this is still ongoing and the changes in this area are just initial. We should avoid asterisk permission and always spell them out even if it's the full list.
2. (major) reduce scoping. Move from cluster-role to roles as much as possible. This requires also cache tunings to actually query the target namespace.

The PR is reviewable but should be merged post the 4.20 cutoff, and we need to careful test it works with different namespaces - I'm testing using the default `numaresources` namespace. The expectation is it does, but we need to verify both the component contracts and the actual behavior.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -36,14 +36,12 @@ import (
 	machineconfigv1 "github.com/openshift/api/machineconfiguration/v1"
 	securityv1 "github.com/openshift/api/security/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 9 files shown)


cmd/main.go
```diff
@@ -261,13 +259,10 @@ func main() {
 
 	klog.InfoS("metrics server", "enabled", params.enableMetrics, "addr", params.metricsAddr)
 
+	lm := controller.NewLocalManager(namespace)
```
Tal-or, line 262: Instead of introducing a cache implementation can't we simply set 
```
Cache: cache.Options{
			DefaultNamespaces: map[string]cache.Config{
				namespace:            {namespace-we-care-about},
			},
		},
```
?

ffromani: the reasons are long forgotten, but I remember this model back in time didn't work - we need to track objects across multiple namespaces. Anyway, this PR is lost. Better to restart from scratch, if ever.

ffromani:
context is long lost. Good riddance.

PR #1932, opened 2025-09-04
OCPBUGS-61355: [release-4.19] controller: sched: conditionally enable Shared informer by default 

shajmakh:
https://github.com/openshift/kubernetes/pull/2391 has landed in
openshift, which means that the behavior described is enabled by default and we want
the scheduler to adapt to this behavior by default iff the user didn't
explicitly set the informer mode in the CR.

Until now, the scheduler works in a dedicated informer mode which is meant to take into
account the pods in terminal state (such as pods that ran and completed)
in the PFP computation to ensure it matches the PFP computed by the RTE
and reported in NRT.

The intended behavior which the new kubelet behavior is about ignoring
such pods and accounting only for active pods, so if this behavior is
enabled in the RTE-NRT, while kept the default dedicated in the
scheduler there will be misalignment in the computed vs the expected PFP
from scheduler's POV vs NRT's POV and a scheduler stall will happen that
will never recover.

In this commit we adjust the informer default value to Shared (instead
of Dedicated) only if both below conditions are met:
1. the cluster version supports the fixed kubelet which is met if the
   cluster version is equal or greater to the known-to-be fixed OCP version.
2. the user didn't set the Spec.SchedulerInformer field in the NRS CR
This modification will enable the shared mode which in turn includes only
running pods (= active pods) in the PFP computation from POV allowing
ultimately PFP alignment with NRT.

(cherry picked from commit https://github.com/openshift-kni/numaresources-operator/commit/6a56840544a420c1393c39fc61e3580f563c1d1f)

The above is the core of this PR and it is followed by related enhancements as described in the different commits that were cherry-picked into this PR. The commits form a complete follow-up for the core commit and that is why I chose to put them in one PR. Please review each one to obtain more details about the changes.

Original PRs:
https://github.com/openshift-kni/numaresources-operator/pull/1315
https://github.com/openshift-kni/numaresources-operator/pull/1715
https://github.com/openshift-kni/numaresources-operator/pull/1734
https://github.com/openshift-kni/numaresources-operator/pull/1915

### Changes specific to 4.19

These involve modifications to 4.19 build numbers and which is derived from the fact that 4.19 is GA and the fix landed in a zstream. Thus, there is no dev-preview nor release candidate builds for already GA releases.

### DIFF
--- a/internal/platforminfo/platforminfo_test.go
+++ b/internal/platforminfo/platforminfo_test.go
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package platforminfo
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
+)
+
+func TestNewWithDiscover(t *testing.T) {
+	testcases := []struct {
+		description string
+		plat        platform.Platform
+		ver         platform.Version
+		expected    PlatformProperties
+	}{
+		{
+			description: "empty",
+		},
+		{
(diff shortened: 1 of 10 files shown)


internal/platforminfo/platforminfo_test.go
```diff
@@ -0,0 +1,85 @@
...
+			plat:        platform.OpenShift,
+			ver:         mustParseVersion("4.20.0"), // at time of writing
+			expected: PlatformProperties{
+				PodResourcesListFilterActivePods: true,
+			},
+		},
+		{
+			description: "next Z-stream fixed, must never regress", // at time of writing
+			plat:        platform.OpenShift,
+			ver:         mustParseVersion("4.20.1"),
```
ffromani, line 55: this should be 4.19.11 or so

ffromani (commented):
/approve

minorish thing

PR #1933, opened 2025-09-04
platforminfo: update first builds that has the kubelet fix

shajmakh:
Update the different types of build releases that had the kubelet PodResourceAPI fix first.

CI and Nightly confirmed using sippy: https://sippy.dptools.openshift.org/sippy-ng/pull_requests/Presubmits
Konflux nightly builds are meant for internal testing only and should be
ignored.


ffromani (commented):
/approve
/lgtm

trusting the research

PR #1977, opened 2025-09-09
e2e: ensure sched pod count is correct for different ocp versions

rshemtov13:
As of version 4.20, the scheduler deployment will consist of 3 pods instead of 1. Added logic to correctly identify the running version so test 75354 won't break from 4.20 onwards.


ffromani:
/approve
/lgtm

there's probably a better way, but let's unblock first

ffromani:
/lgtm

thanks!

rshemtov13:
@ffromani Thanks for the comments!

PR #1981, opened 2025-09-09
sched: status: reset the status conditions to defaults

shajmakh:
The problem with the current implementation is that if the status condition A is updated to `True` (like Progressing or Degraded condition) it will never be updated again if the reconcilation interation is not updating that exact condition. This creates accumulated status conditions as follows:

```
    conditions:
    - lastTransitionTime: "2025-09-09T10:24:02Z"
      message: ""
      reason: AsExpected
      status: "True"
      type: Available
    - lastTransitionTime: "2025-09-09T10:24:02Z"
      message: ""
      reason: Upgradeable
      status: "True"
      type: Upgradeable
    - lastTransitionTime: "2025-09-09T10:23:47Z"
      message: ""
      reason: AsExpected
      status: "True"
      type: Progressing
    - lastTransitionTime: "2025-09-09T10:23:47Z"
      message: 'could not apply (apps/v1, Kind=Deployment) openshift-numaresources/secondary-scheduler:
        could not update object (apps/v1, Kind=Deployment) openshift-numaresources/secondary-scheduler:
        Operation cannot be fulfilled on deployments.apps "secondary-scheduler": the
        object has been modified; please apply your changes to the latest version
        and try again'
      reason: InternalError
      status: "True"
      type: Degraded
    - lastTransitionTime: "2025-09-09T10:24:02Z"
      message: ""
      observedGeneration: 1
      reason: DedicatedInformerActive
      status: "False"
      type: DedicatedInformerActive
```

From which the user cannot know what condition is the true one. To fix this, in the phase where the scheduler status is being updated, make the status starts with default conditions instead of the old ones to ensure a reset of the old conditions and a fresh start.

This is a reporting issue and doesn't affect the scheduler functionality.

### DIFF
--- a/internal/controller/numaresourcesscheduler_controller.go
+++ b/internal/controller/numaresourcesscheduler_controller.go
@@ -304,7 +304,8 @@ func (r *NUMAResourcesSchedulerReconciler) syncNUMASchedulerResources(ctx contex
 	}
 
 	informerCondition := buildDedicatedInformerCondition(*instance, schedSpec)
-	schedStatus.Conditions = status.GetUpdatedSchedulerConditions(schedStatus.Conditions, informerCondition)
+	// intentionally override with nil conditions to trigger a reset of the old status conditions
+	schedStatus.Conditions = status.GetUpdatedSchedulerConditions(nil, informerCondition)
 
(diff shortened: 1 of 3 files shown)


ffromani (commented):
can we have controller tests to ensure we got this right (and we keep it right) this time?

PR #2002, opened 2025-09-11
tree: make main point to 4.21

shajmakh:
release-4.20 is branched out, ensure main branch points to 4.21. Konflux updates related to 4.21 bundle SHA will have a follow-up PR.


shajmakh:
the auto konflux PRs that update the catalog and bundle are creating a conflict, I'll drop the conflicting changes since they anyway need a follow-up PR to get updated with the expected true values.

ffromani:
@shajmakh are the failures expected? should we override the failing lanes?

PR #2009, opened 2025-09-12
OCPBUGS-61676: [release-4.17]: controller: sched: conditionally enable Shared informer by default 

shajmakh:
manual cherry-pick of #1979


shajmakh:
after taking a deeper look, the CI failure is on NRO never being reported because the RTEs didn't manage to be ready due to:
`main.go:118] failed to execute: failed to initialize ResourceMonitor: error while updating node allocatable: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /host-podresources/kubelet.sock: connect: permission denied"`

this isn't related to the scheduler conditions changes. investigating what causes this permission issue.

PR #2037, opened 2025-09-14
Konflux: adopt telco5g-konflux shared scripts

yanirq:
Rework makefile and d/s konflux mechanism to make use of telco5g-konflux shared library

- Add telco5g-konflux as a git submodule
- Switch bundle overlay to submodule script
- Catalog updates
- Makefile integration:
  - Wire Konflux targets to telco5g-konflux scripts
  - Use separate input/output for catalog template
  - Set bundle name suffix and production names for this repo
- Cleanup obsolete scripts now provided by submodule

Assisted-by: Cursor 1.5.11
AI Attribution: Hybrid , application with https://github.com/openshift-kni/lifecycle-agent/pull/1978 as base ,  GPT-5


.konflux/bundle/overlay/release.in.yaml
```diff
@@ -0,0 +1,15 @@
...
+  description: |
+      NUMA Resources Operator for OpenShift
+      Provides NUMA-aware scheduling and resources management for workloads.
+
+      You can find additional guidance in the upstream repository: https://github.com/openshift-kni/numaresources-operator
+  display_name: "NUMA Resources Operator"
+  manager_version: "numaresources-operator.v4.21.0"
+  # min_kube_version should match ocp
+  # https://access.redhat.com/solutions/4870701
+  min_kube_version: "1.32.0"
```
yanirq: @ffromani  up until now this was set to 1.23.0 ,  it makes sense to change it to something more up to date (although this was never blocking)
Do you see any risk here ?  cc @shajmakh

Makefile
```diff
@@ -595,53 +608,96 @@ endef
...
+	$(MAKE) konflux-validate-catalog
+
+.PHONY: konflux-filter-unused-redhat-repos
+konflux-filter-unused-redhat-repos: ## Filter unused repositories from redhat.repo in operator lock folder
+	@echo "Filtering unused repositories from operator runtime lock folder..."
+	$(MAKE) -C $(PROJECT_DIR)/telco5g-konflux/scripts/rpm-lock filter-unused-repos REPO_FILE=$(PROJECT_DIR)/.konflux/operator/redhat.repo
+	@echo "Filtering completed."
+
+.PHONY: konflux-compare-catalog
+konflux-compare-catalog: sync-git-submodules ## Compare generated catalog with upstream FBC image
```
fontivan: did you run the compare catalog target to verify the catalog generated from the template was the same?

yanirq: Both catalogs reference the same bundle SHA256 (ae16e8e5ded756b3d7cbe388429bab4b62b613e9872310e7c69b67edc1db9bab)
All other content (metadata, properties, CRD descriptions, annotations) is identical
The only difference is the image registry/namespace:
Upstream catalog is production (uses registry.redhat.io)
Generated catalog is non-production (uses quay.io)

PR #2048, opened 2025-09-16
scheduler:  regression fixes for 4.21

ffromani:
fix regression introduced in the 4.21 branching
along the way (unrelated) update the manifest examples to catch up with modern features.


shajmakh:
/lgtm
Thanks for catching and fixing!

PR #2204, opened 2025-09-30
e2e: Serial: Added test ids to the untagged serial tests

SargunNarula:
This PR assigns unique IDs to previously un-tagged tests in the serial suite.


ffromani (approved):
/approve
/lgtm

PR #2261, opened 2025-10-09
WIP: OCPBUGS-61756: ctrl: kubeletConfig: avoid updating ConfigMap on Paused MCP

shajmakh:
Having a paused MCP should prevent updating/creating the corresponding
config map for the specified node group. So far, the code wasn't
considering the case of paused MCPs, which leads to creating/updating the config map
a thing that caused a mismatch between the configuration in the config
map and the one reflected on the NRTs.
In this commit, we modify the kubeletconfig controller to skip on
updating RTE config maps that belong to paused MCPs.

### DIFF
--- a/internal/controller/kubeletconfig_controller.go
+++ b/internal/controller/kubeletconfig_controller.go
@@ -48,6 +48,7 @@ import (
 	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
 
 	nropv1 "github.com/openshift-kni/numaresources-operator/api/v1"
+	intkubeletconfig "github.com/openshift-kni/numaresources-operator/internal/kubeletconfig"
 	"github.com/openshift-kni/numaresources-operator/internal/machineconfigpools"
 	"github.com/openshift-kni/numaresources-operator/pkg/apply"
 	"github.com/openshift-kni/numaresources-operator/pkg/kubeletconfig"
@@ -57,7 +58,8 @@ import (
 )
 
 const (
-	kubeletConfigRetryPeriod = 30 * time.Second
+	kubeletConfigRetryPeriod           = 30 * time.Second
+	MachineConfigPoolPausedRetryPeriod = 2 * time.Minute
 )
 
 const (
@@ -90,6 +92,7 @@ type kubeletConfigHandler struct {
 type reconcileErrorHandler struct {
 	err           error
 	tolerateError bool
+	result        ctrl.Result
 }
 
 // Namespace Scoped
@@ -126,21 +129,22 @@ func (r *KubeletConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		var klErr *InvalidKubeletConfig
 		if errors.As(errHandler.err, &klErr) {
 			r.Recorder.Event(instance, "Normal", "ProcessSkip", "ignored kubelet config "+klErr.ObjectName)
-			return ctrl.Result{}, nil
+			return errHandler.result, nil
 		}
 		if errHandler.tolerateError {
 			r.Recorder.Event(instance, "Normal", "ProcessSkip", errHandler.err.Error())
-			return ctrl.Result{}, nil
+			return errHandler.result, nil
 		}
 
 		klog.ErrorS(errHandler.err, "failed to reconcile configmap", "controller", "kubeletconfig")
 
 		r.Recorder.Event(instance, "Warning", "ProcessFailed", "Failed to update RTE config from kubelet config "+req.NamespacedName.String())
-		return ctrl.Result{}, errHandler.err
+		return errHandler.result, errHandler.err
 	}
 
 	r.Recorder.Event(instance, "Normal", "ProcessOK", fmt.Sprintf("Updated RTE config %s/%s from kubelet config %s", cm.Namespace, cm.Name, req.NamespacedName.String()))
-	return ctrl.Result{}, nil
+	//return ctrl.Result{}, nil
+	return errHandler.result, nil
 }
 
 func (r *KubeletConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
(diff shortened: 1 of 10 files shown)


internal/reconcile/step.go
```diff
@@ -92,3 +101,35 @@ func StepFailed(err error) Step {
 		Error:         err,
 	}
 }
+
+func StepWarning(err error) Step {
```
Tal-or: lets add comments explaining the new steps purposes.

Tal-or (commented):
Good change overall.

In case the MCP is paused, the final update to the configmap will be triggered when the MCP changes back to unpause?

internal/reconcile/event.go
```diff
@@ -0,0 +1 @@
+package reconcile
```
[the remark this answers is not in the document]
shajmakh: correct, was thinking if we need a different wrapper other than the one we use for NRO controller, for now we can use the same `step` record

internal/reconcile/step.go
shajmakh: of course

yanirq:
also fixes https://issues.redhat.com/browse/OCPBUGS-52859

ffromani (commented):
initial review

internal/controller/kubeletconfig_controller.go
```diff
@@ -244,63 +249,106 @@ func (r *KubeletConfigReconciler) syncConfigMap(ctx context.Context, kubeletConf
...
 		if mcoKc.Spec.KubeletConfig == nil {
 			klog.InfoS("detected KubeletConfig with empty payload, ignoring", "name", kcKey.Name)
-			return nil, &InvalidKubeletConfig{ObjectName: kcKey.Name}
+			return nil, intreconcile.StepNormalSkip(fmt.Errorf("Invalid KubeletConfig %s", kcKey.Name))
+		}
+
+		if mcp.Spec.Paused {
+			klog.InfoS("detected paused MCP", "name", mcp.Name)
+			//if the CM exists -> just skip;
+			//if the CM does not exist -> create it based on the current active machineConfig
```
Tal-or: Maybe I'm not following but how such case is possible? is it the default `01-worker-kubelet` machineConfig?

internal/controller/kubeletconfig_controller.go
```diff
@@ -244,63 +249,106 @@ func (r *KubeletConfigReconciler) syncConfigMap(ctx context.Context, kubeletConf
...
+					}
+
+					decodeKc, err := intkubeletconfig.DecodeKubeletConfigurationFromData(dataInBytes)
+					if err != nil {
+						klog.ErrorS(err, "cannot decode the current KubeletConfig data from MachineConfig", "name", currentConfigName)
+						return nil, stepNormalSkipForPausedupdates("failed to decode the current KubeletConfig data from MachineConfig %s: %v", currentConfigName, err)
+					}
+
+					successStepWithRetry := intreconcile.StepNormalSucess(fmt.Sprintf("Created ConfigMap based on the current machineConfig for paused MCP %s", mcp.Name))
+					successStepWithRetry.Result = ctrl.Result{Requeue: true, RequeueAfter: MachineConfigPoolPausedRetryPeriod}
```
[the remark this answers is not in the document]
shajmakh: that sounds better, Thanks for highliting, will update this part

internal/controller/kubeletconfig_controller.go
```diff
@@ -87,6 +90,12 @@ type kubeletConfigHandler struct {
 	setCtrlRef func(owner, controlled metav1.Object, scheme *runtime.Scheme, opts ...controllerutil.OwnerReferenceOption) error
 }
 
+type reconcileErrorHandler struct {
```
ffromani: this is not a handler. a Handler manages something. This seems more a `reconcileResult`

internal/controller/kubeletconfig_controller.go
```diff
@@ -126,21 +129,22 @@ func (r *KubeletConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 		klog.ErrorS(errHandler.err, "failed to reconcile configmap", "controller", "kubeletconfig")
 
 		r.Recorder.Event(instance, "Warning", "ProcessFailed", "Failed to update RTE config from kubelet config "+req.NamespacedName.String())
-		return ctrl.Result{}, errHandler.err
+		return errHandler.result, errHandler.err
 	}
 
 	r.Recorder.Event(instance, "Normal", "ProcessOK", fmt.Sprintf("Updated RTE config %s/%s from kubelet config %s", cm.Namespace, cm.Name, req.NamespacedName.String()))
-	return ctrl.Result{}, nil
+	//return ctrl.Result{}, nil
```
Tal-or: leftovers I presume

internal/controller/kubeletconfig_controller.go
```diff
@@ -254,63 +261,128 @@ func (r *KubeletConfigReconciler) syncConfigMap(ctx context.Context, kubeletConf
...
+
+				return &kubeletConfigHandler{
+					ownerObject: decodeKc,
+					mcoKc:       decodeKc,
+					poolName:    mcp.Name,
+					setCtrlRef:  controllerutil.SetControllerReference,
+				}, reconcileErrorHandler{result: ctrl.Result{Requeue: true, RequeueAfter: MachineConfigPoolPausedRetryPeriod}}
+			}
+
+			klog.InfoS("MachineConfigPool of KubeletConfig %s is paused and configMap %s exists", kcKey.Name, existingCM.Name)
```
ffromani: please check verbosity to avoid log spam

internal/controller/kubeletconfig_controller_test.go
```diff
@@ -272,4 +272,171 @@ var _ = Describe("Test KubeletConfig Reconcile", func() {
...
+			event := <-fakeRecorder.Events
+			Expect(event).To(ContainSubstring("ProcessSkip"))
+			Expect(event).To(ContainSubstring("failed to find the current machineConfig"))
+
+			cm := &corev1.ConfigMap{}
+			key := client.ObjectKey{
+				Namespace: testNamespace,
+				Name:      objectnames.GetComponentName(nro.Name, poolName2),
+			}
+			Expect(reconciler.Client.Get(context.TODO(), key, cm)).To(HaveOccurred())
```
ffromani: and check IsNotFound perhaps?

ffromani (commented):
the direction seems fine, I have implementation comments.
The last commit would need another review round from me, need more time to grasp it.

Tal-or (commented):
Some comments inside.
One thing that bother me the most is the `tolerateError` concept.
if it's tolerable, I wouldn't consider it an error.

can't we simply log the error in place, and Requeue the request as we usually do?

PR #2292, opened 2025-10-15
rte: controller: tune daemonset rollout

ffromani:
The reason why have the current rollout settings is we never challenged them. There is no technical reason not to have MaxUnavailable up to 10% (or more), so we set it as such.

### DIFF
--- a/internal/controller/numaresourcesoperator_controller_test.go
+++ b/internal/controller/numaresourcesoperator_controller_test.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
+	"strings"
 	"time"
 
 	igntypes "github.com/coreos/ignition/v2/config/v3_2/types"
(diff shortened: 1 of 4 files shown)


internal/controller/numaresourcesoperator_controller_test.go
```diff
@@ -319,6 +321,19 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
...
+					dsKey := client.ObjectKey{
+						Name:      objectnames.GetComponentName(nro.Name, pn1),
+						Namespace: testNamespace,
+					}
+					var ds appsv1.DaemonSet
+					Expect(reconciler.Client.Get(context.TODO(), dsKey, &ds)).To(Succeed())
+
+					Expect(ds.Spec.UpdateStrategy.Type).To(Equal(appsv1.RollingUpdateDaemonSetStrategyType))
+					Expect(ds.Spec.UpdateStrategy.RollingUpdate).ToNot(BeNil())
+					Expect(ds.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable.String()).To(BeStringPercentageAtLeast(10))
```
shajmakh: de we still intend this to be 10? and considering the logic of `MaxUnavailable` shouldn't this be "at most" instead of "at least"?

ffromani: The intent here is to set (and check) we do set a `MaxUnavaialble` whose value is 10% or greater, do you think this is expressed clearly enough from the testcase?

PR #2322, opened 2025-10-21
e2e:install:hcp: poll for status

Tal-or:
The status about daemonset reported by NROP is missing. Adding `Eventually` to make sure it\'s not a case of racecondition where NROP takes long time to update/report the status.


ffromani (commented):
/approve
/lgtm

good approach in general, which we still miss from time to time

Tal-or:
> /approve /lgtm
> 
> good approach in general, which we still miss from time to time

Not sure it will resolve the issue.
I didn't have ab handy environment to test locally

ffromani:
> Not sure it will resolve the issue.

fair enough, makes the code better anyway.

ffromani (commented):
/approve
/lgtm

makes the code better anyway

PR #2325, opened 2025-10-21
CNF-20022: status: nrosched: set ObservedGeneration in status conds

ffromani:
The NUMAResourcesScheduler object status condition reports ObservedGeneration only in the `DedicatedInformerActive` condition.
Let's expose it in all the relevant conditions, doing some minimal internal refactoring along the way.
Note that we can't test this change using controller tests due both the unique requirements of the operator and the way we (mis)use the controller tests; we should add checks in (existing?) e2e tests, but we postpone this to a later change.

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -243,67 +243,68 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 		})
 
 		It("[test_id:54916] should be able to modify the configurable values under the NUMAResourcesScheduler CR", Label(label.Tier2, "schedrst"), Label("feature:schedrst"), func() {
+			By("getting the initial scheduler object state")
 			initialNroSchedObj := &nropv1.NUMAResourcesScheduler{}
 			nroSchedKey := objects.NROSchedObjectKey()
-			err := fxt.Client.Get(context.TODO(), nroSchedKey, initialNroSchedObj)
+			ctx := context.TODO()
+			err := fxt.Client.Get(ctx, nroSchedKey, initialNroSchedObj)
 			Expect(err).ToNot(HaveOccurred(), "cannot get %q in the cluster", nroSchedKey.String())
 			nroSchedObj := initialNroSchedObj.DeepCopy()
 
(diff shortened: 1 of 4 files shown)


test/e2e/serial/tests/configuration.go
```diff
@@ -243,67 +243,69 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
-					err = fxt.Client.Update(context.TODO(), currentSchedObj)
-					g.Expect(err).ToNot(HaveOccurred())
+					g.Expect(fxt.Client.Update(rctx, currentSchedObj)).To(Succeed())
 				}).WithTimeout(5*time.Minute).WithPolling(10*time.Second).Should(Succeed(), "failed to revert changes the changes to the NRO scheduler object")
 
 				updatedSchedObj := &nropv1.NUMAResourcesScheduler{}
 				Eventually(func() string {
-					err = fxt.Client.Get(context.TODO(), client.ObjectKeyFromObject(initialNroSchedObj), updatedSchedObj)
-					Expect(err).ToNot(HaveOccurred())
+					Expect(fxt.Client.Get(rctx, client.ObjectKeyFromObject(initialNroSchedObj), updatedSchedObj)).To(Succeed())
```
shajmakh: same here, probably never failed on this because Get is unlikly to fail. I'm seeing a couple of examples in the suite that do this mistake, we should fix in a cleanup PR, I expect that to stabilize the suite more.

ffromani: indeed, fixed all the occurrences in this test.

test/e2e/serial/tests/configuration.go
```diff
@@ -243,67 +243,69 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
 					return updatedSchedObj.Status.SchedulerName
 				}).WithTimeout(time.Minute).WithPolling(time.Second*15).Should(Equal(initialNroSchedObj.Status.SchedulerName), "failed to revert the schedulerName field,expected %q but found %q", initialNroSchedObj.Status.SchedulerName, updatedSchedObj.Status.SchedulerName)
+			}(context.TODO())
```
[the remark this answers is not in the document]
ffromani: because we seen failure on cleanup when the context using in test was cancelled. Truth to be told, I'm still not sure about what's the best way to handle cleanup.

PR #2722, opened 2025-12-17
release-4.20: status: set ObservedGeneration in all status conds

ffromani:
The NUMAResourcesScheduler (sic: 4.20 note: we actually change NUMAResourcesOperator) object status condition never reports ObservedGeneration in any status conditions, unlike the scheduler object.

Likewise the scheduler change, let's expose it in all the relevant conditions, doing some minimal internal refactoring along the way.

Note that we can't test this change using controller tests due both the unique requirements of the operator and the way we (mis)use the controller tests; we should add checks in (existing?) e2e tests, but we postpone this to a later change.

Note: this is a release-4.20-specific backport of
commit d223bdaeb2703a8d860017f416d8ed1aadf59521

### DIFF
--- a/internal/api/features/_topics.json
+++ b/internal/api/features/_topics.json
@@ -20,6 +20,7 @@
     "ngpoolname",
     "nganns",
     "metrics",
-    "nrtcrdanns"
+    "nrtcrdanns",
+    "operobsgen"
   ]
 }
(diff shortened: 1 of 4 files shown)


ffromani:
/hold cancel

it seems we prefer this over https://github.com/openshift-kni/numaresources-operator/pull/2732
so we can move forward pending review and verification

Tal-or (approved):
/lgtm
/approve

thanks

Tal-or:
keep backport to 4.18?

ffromani:
> keep backport to 4.18?

yes, we can still move forward with that

PR #2732, opened 2025-12-18
conditional use of ObservedGeneration

ffromani:
in PR https://github.com/openshift-kni/numaresources-operator/pull/2431 we added reporting of `ObservedGeneration` in the `NUMAResourcesOperator.Status`. This is a good change and closes a known gap, but we can't just use the value unconditionally in e2e tests, because a long tail of old versions (pre https://github.com/openshift-kni/numaresources-operator/pull/2722) won't expose this value, which will create a storm of false negatives tests.

Ignoring the value till all the version supports it seems excessive; reverting the tests is wrong, reverting the change seems worse. The best way forward seems to be a more granular check, introduced in this PR.

This creates a challenge for future tests. The best option is probably to use ObservedGeneration only in the new tests and stop refactoring of existing codebase. Either way, we will need to be more careful in the future.

### DIFF
--- a/internal/api/features/types.go
+++ b/internal/api/features/types.go
@@ -23,7 +23,13 @@ import (
 )
 
 const (
-	Version = "v5.0.0"
+	// Version is the version of the API, and covers data layout.
+	// The version should be bumped when the API definition changes,
+	// and is not strictly tied to the operator version.
+	// The payload (e.g. active topics) may and will changes even
+	// within the same operator version; e.g. is totally legit
+	// to add more topics during a Z stream.
+	Version = "v1.0.0"
 )
 
 type Metadata struct {
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -1843,6 +1844,10 @@ func isNROOperSyncedAt(nropStatus *nropv1.NUMAResourcesOperatorStatus, condition
 	}
 	if cond.Status != metav1.ConditionTrue {
 		return errors.New("condition not true")
+	}
+	// TODO: until all the version reports ObservedGeneration
+	if serialconfig.Config != nil && !slices.Contains(serialconfig.Config.ClusterTopics.Active, "operobsgen") {
+		return nil
 	}
 	if cond.ObservedGeneration != gen {
 		return fmt.Errorf("condition at %d expected %d", cond.ObservedGeneration, gen)
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -235,25 +235,29 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 				g.Expect(isNROOperSyncedAt(&updatedOperObj.Status, status.ConditionAvailable, updatedOperObj.Generation)).To(Succeed())
 			}).WithTimeout(10*time.Minute).WithPolling(30*time.Second).Should(Succeed(), "failed to update RTE daemonset node selector")
 
-			dss, err := objects.GetDaemonSetsByNamespacedName(fxt.Client, ctx, updatedOperObj.Status.DaemonSets...)
-			Expect(err).ToNot(HaveOccurred())
+			// Eventually is not strictly needed but till all the supported version report ObservedGeneration in status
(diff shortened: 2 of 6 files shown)


test/e2e/serial/tests/configuration.go
```diff
@@ -1521,10 +1522,18 @@ func isNROOperAvailableAt(nropStatus *nropv1.NUMAResourcesOperatorStatus, gen in
...
+		return true
+	}
 	return cond.ObservedGeneration == gen
 }
 
 func isNROOperUpToDate(nrop *nropv1.NUMAResourcesOperator) bool {
+	// TODO: until all the version reports ObservedGeneration
+	if serialconfig.Config != nil && !slices.Contains(serialconfig.Config.ClusterTopics.Active, "operobsgen") {
+		return true
+	}
```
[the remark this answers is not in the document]
ffromani: not sure I got you right: we check the condition before, don't we?

internal/api/features/types.go
```diff
@@ -23,7 +23,13 @@ import (
...
 
 const (
-	Version = "v5.0.0"
+	// Version is the version of the API, and covers data layout.
+	// The version should be bumped when the API definition changes,
+	// and is not strictly tied to the operator version.
+	// The payload (e.g. active topics) may and will changes even
+	// within the same operator version; e.g. is totally legit
+	// to add more topics during a Z stream.
+	Version = "v1.0.0"
```
Tal-or, line 32: Will be really hard to track, or i'll have to be more blunt we'll simply forget to update this value.
Without an automated check it won't fly IMVHO, which I'm sure will also add more complexity.

Tal-or (commented):
This PR has good intentions but if I may, I would like to suggest different approach:
1. backport https://github.com/openshift-kni/numaresources-operator/pull/2722 all the way to 4.18 - the change is relatively small and scoped.
2. change  the check in the `isNROOperSyncedAt` function:
from:
```
if cond.ObservedGeneration != gen {
		return fmt.Errorf("condition at %d expected %d", cond.ObservedGeneration, gen)
	}
```
```
to:
if cond.ObservedGeneration != gen && cond.ObservedGeneration != 0 {
		return fmt.Errorf("condition at %d expected %d", cond.ObservedGeneration, gen)
	}
```
if `cond.ObservedGeneration == 0` it's likely because the operator is running the old code so don't fail the test on this part.

internal/api/features/types.go
ffromani: likely correct, but using the version as before is pointless. We can just remove the version at that point. Either it's dependable signal or should be ignored.

ffromani:
> This PR has good intentions but if I may, I would like to suggest different approach:
> 
>     1. backport #2722 all the way to 4.18 - the change is relatively small and scoped.
> 
>     2. change  the check in the `isNROOperSyncedAt` function:
>        from:
> 
> 
> ```
> if cond.ObservedGeneration != gen {
> 		return fmt.Errorf("condition at %d expected %d", cond.ObservedGeneration, gen)
> 	}
> ```
> 
> ```
> to:
> if cond.ObservedGeneration != gen && cond.ObservedGeneration != 0 {
> 		return fmt.Errorf("condition at %d expected %d", cond.ObservedGeneration, gen)
> 	}
> ```
> 
> if `cond.ObservedGeneration == 0` it's likely because the operator is running the old code so don't fail the test on this part.

It's an interesting alternative. I'm ok backporting https://github.com/openshift-kni/numaresources-operator/pull/2722 down to 4.18, but I'm not sure _that_ is the oldest supported version, and therefore makes this PR useless.

Tal-or:
> > if `cond.ObservedGeneration == 0` it's likely because the operator is running the old code so don't fail the test on this part.

 > It's an interesting alternative. I'm ok backporting #2722 down to 4.18, but I'm not sure _that_ is the oldest supported version, and therefore makes this PR useless.

yes but adding also suggestion `2` will do.

ffromani:
> > > if `cond.ObservedGeneration == 0` it's likely because the operator is running the old code so don't fail the test on this part.
> 
> > It's an interesting alternative. I'm ok backporting #2722 down to 4.18, but I'm not sure _that_ is the oldest supported version, and therefore makes this PR useless.
> 
> yes but adding also suggestion `2` will do.


True. But here's the value proposition of this PR: don't do guesswork with ambiguous valuye (0), but rather get an explicit signal.

ffromani:
in the end https://github.com/openshift-kni/numaresources-operator/pull/4063 made this unnecessary

ffromani:
/close

reason:

> in the end #4063 made this unnecessary

PR #2874, opened 2026-01-20
GH actions: run CI on all supported main branches

shajmakh:
Keep the CI running on all official supported branches to avoid merging content with less checks.
Given that the repo is public and the runners used in the workflows (`ubuntu-latest`) are standard github-hosted runners, we can freely use them in the actions of this repo. ref:
https://docs.github.com/en/billing/concepts/product-billing/github-actions#free-use-of-github-actions

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,8 +4,14 @@ on:
   pull_request:
     branches:
       - main
-      - "release-4.18"
+      - "release-4.21"
       - "release-4.20"
+      - "release-4.19"
+      - "release-4.18"
+      - "release-4.17"
+      - "release-4.16"
+      - "release-4.14"
+      - "release-4.12"
   push:
     branches: 
       - main
--- a/.github/workflows/e2e-local.yaml
+++ b/.github/workflows/e2e-local.yaml
@@ -4,8 +4,14 @@ on:
   pull_request:
     branches:
       - main
-      - "release-4.18"
+      - "release-4.21"
       - "release-4.20"
+      - "release-4.19"
+      - "release-4.18"
+      - "release-4.17"
+      - "release-4.16"
+      - "release-4.14"
+      - "release-4.12"
     tags:
       - "v[0-9]+.[0-9]+.[0-9]+"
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -4,8 +4,14 @@ on:
...
       - main
-      - "release-4.18"
+      - "release-4.21"
       - "release-4.20"
+      - "release-4.19"
+      - "release-4.18"
+      - "release-4.17"
+      - "release-4.16"
+      - "release-4.14"
+      - "releasee-4.12"
```
ffromani: and we need to backport this down to 4.12 :\

ffromani: adding the even branches is OK to me, even though the GH doc page is confusing. Do we have 2k minutes/month or infinite minutes/month?
is this accounted per-repo or per-org?

shajmakh: yes we need this change in all older branches. Thinking about it, I don't see a strong reason to list the branches here except for main. 
The use is free for open-source (hence public) repositories that use standard github-hosted runners, which applies to this workflow yaml and all others used in this repo. the runners avialable in this repo are: https://github.com/openshift-kni/numaresources-operator/actions/runners and the larger ones are not used in this repo. here are the standard runner images:
https://github.com/actions/runner-images/tree/main?tab=readme-ov-file#available-images

ffromani: > The use is free for open-source (hence public) repositories that use standard github-hosted runners, which applies to this workflow yaml and all others used in this repo. the runners avialable in this repo are: https://github.com/openshift-kni/numaresources-operator/actions/runners and the larger ones are not used in this repo. here are the standard runner images:
https://github.com/actions/runner-images/tree/main?tab=readme-ov-file#available-images

this is true, and yet, doesn't answer my question above :\

shajmakh: yes indeed this better be consumed less. Thanks for your input.

ffromani:
I'm fine merging a reduced version: https://github.com/openshift-kni/numaresources-operator/pull/3601

Please reopen/file a followup if needed

PR #2924, opened 2026-01-29
deps: bump rte and ghw deps

ffromani:
consume latest `rte` and `ghw` dep to narrow down our dependency tree.


ffromani:
obsolete

PR #3169, opened 2026-03-08
e2e: add schedulable CP and Compact cluster node group config tests

rshemtov13:
Add two new e2e serial tests for verifying NROP node group configuration on clusters with schedulable control-pllane nodes and compact cluster.

- test 83869 Configures two non-overlapping node groups (master + worker-cnf) on MNO with schedulable masters and verifies DSs, pods and NRTs are healthy. Cannot run on compact clusters because all nodes carry both master and worker roles.

- test 83870 Configures two overlapping node groups (master + worker) and verifies duplicate RTE pods are kept pending due to PodsAntiAffinity. Will run on both compact and MNO cluster with masters schedulable.

Also introduces ClusterType labels for filtering the test case runs when running on these specific cluster type environments.

### DIFF
--- a/test/e2e/label/label.go
+++ b/test/e2e/label/label.go
@@ -36,6 +36,24 @@ const (
 	Reboot Kind = "reboot_required"
 )
 
+// ClusterType is a label that specifies the cluster topology
+// required by the test
+type ClusterType = string
+
+const (
+	// MNO is a multi-node OpenShift cluster with dedicated worker nodes
+	// and unschedulable control-plane nodes.
+	MNO ClusterType = "mno"
+
+	// MNOMastersSchedulable is a multi-node OpenShift cluster with
+	// schedulable control-plane nodes and dedicated worker nodes.
+	MNOMastersSchedulable ClusterType = "mno-masters-schedulable"
+
+	// Compact is a cluster where control-plane and worker roles
+	// are on the same nodes, with control-plane scheduling enabled.
+	Compact ClusterType = "compact"
+)
+
 // Tier is a label to classify tests under specific grade/level
 // that should roughly describe the execution complexity, maintainer identity and processing criteria.
 type Tier = string
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -1356,6 +1356,249 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 			})
 		})
 
+		Context("schedulable control-plane node group configuration", func() {
+			var initialOperObj *nropv1.NUMAResourcesOperator
+			var nroKey client.ObjectKey
+
+			BeforeEach(func(ctx context.Context) {
+				initialOperObj = &nropv1.NUMAResourcesOperator{}
+				nroKey = objects.NROObjectKey()
+				Expect(fxt.Client.Get(ctx, nroKey, initialOperObj)).To(Succeed(), "cannot get %q in the cluster", nroKey.String())
+			})
+
+			It("[test_id:83869] should configure NRO CR with multiple non-overlapping node groups and check e2e results", Label(label.Tier2, label.OpenShift, label.MNOMastersSchedulable), func(ctx context.Context) {
+				clusterType := getClusterType(ctx, fxt.Client)
+				if !(clusterType == label.MNOMastersSchedulable) {
+					e2efixture.Skipf(fxt, "test requires %q cluster type, got %q", label.MNOMastersSchedulable, clusterType)
+				}
+
+				workerCNF := "worker-cnf"
+				workerCNFRoleLabel := fmt.Sprintf("%s/%s", depnodes.LabelRole, workerCNF)
+
+				dedicatedWorkers := getDedicatedWorkerNodes(ctx, fxt.Client)
+				Expect(dedicatedWorkers).ToNot(BeEmpty(), "no dedicated worker nodes found while running on cluster type %q", clusterType)
+
+				e2efixture.By("labeling one dedicated worker node with %q", workerCNFRoleLabel)
+				targetNode := &dedicatedWorkers[0]
+				targetNodeInitial := targetNode.DeepCopy()
+				targetNode.Labels[workerCNFRoleLabel] = ""
+				Expect(fxt.Client.Patch(ctx, targetNode, client.MergeFrom(targetNodeInitial))).To(Succeed())
+				klog.InfoS("labeled node", "name", targetNode.Name, "label", workerCNFRoleLabel)
+
+				e2efixture.By("creating %q MCP", workerCNF)
+				mcp := objects.TestMCP()
+				mcp.Name = workerCNF
+				mcp.Labels = map[string]string{"machineconfiguration.openshift.io/role": workerCNF}
+				mcp.Spec.MachineConfigSelector = &metav1.LabelSelector{
+					MatchExpressions: []metav1.LabelSelectorRequirement{
+						{
+							Key:      "machineconfiguration.openshift.io/role",
+							Operator: metav1.LabelSelectorOpIn,
+							Values:   []string{depnodes.RoleWorker, workerCNF},
+						},
+					},
+				}
+				mcp.Spec.NodeSelector = &metav1.LabelSelector{
+					MatchLabels: map[string]string{workerCNFRoleLabel: ""},
+				}
+				Expect(fxt.Client.Create(ctx, mcp)).To(Succeed())
+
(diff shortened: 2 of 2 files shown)


test/e2e/label/label.go
```diff
@@ -36,6 +36,20 @@ const (
...
 )
 
+// ClusterType is a label that specifies the cluster topology
+// required by the test
+type ClusterType = string
+
+const (
+	// Compact means the test requires a compact cluster where
+	// all nodes are control-planes with scheduling enabled
+	Compact ClusterType = "compact"
```
Tal-or: Is it possible to detect against what type of cluster the tests are running, and execute the relevant tests according to the cluster type?

test/e2e/serial/tests/configuration.go
```diff
@@ -1276,6 +1276,230 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			BeforeEach(func(ctx context.Context) {
+				initialOperObj = &nropv1.NUMAResourcesOperator{}
+				nroKey = objects.NROObjectKey()
+				Expect(fxt.Client.Get(ctx, nroKey, initialOperObj)).To(Succeed(), "cannot get %q in the cluster", nroKey.String())
+			})
+
+			It("[test_id:83869] should configure NRO CR with multiple non-overlapping node groups and check e2e results", Label(label.Tier2, label.OpenShift, label.MNOMastersSchedulable), func(ctx context.Context) {
+				workerCnfRole := "worker-cnf"
+				workerCnfRoleLabel := fmt.Sprintf("%s/%s", depnodes.LabelRole, workerCnfRole)
+				masterPoolName := "master"
```
Tal-or: Master is the legacy term https://kubernetes.io/docs/reference/glossary/?all=true#term-master 

we may consider using the term control-plane which we also have constant for under `depnodes.RoleControlPlane`

test/e2e/serial/tests/configuration.go
```diff
@@ -1276,6 +1276,230 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+						},
+					},
+				}
+				mcp.Spec.NodeSelector = &metav1.LabelSelector{
+					MatchLabels: map[string]string{workerCnfRoleLabel: ""},
+				}
+				Expect(fxt.Client.Create(ctx, mcp)).To(Succeed())
+
+				defer func() {
+					By(fmt.Sprintf("CLEANUP: removing %q label from node %q", workerCnfRoleLabel, targetNode.Name))
```
Tal-or: use `fxt.By` instead

test/e2e/serial/tests/configuration.go
```diff
@@ -1276,6 +1276,230 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+					nodeBase := targetNode.DeepCopy()
+					delete(targetNode.Labels, workerCnfRoleLabel)
+					Expect(fxt.Client.Patch(ctx, targetNode, client.MergeFrom(nodeBase))).To(Succeed())
+
+					By("waiting for worker MCP to stabilize after node returns to worker pool")
+					workerMCP := &machineconfigv1.MachineConfigPool{}
+					Expect(fxt.Client.Get(ctx, client.ObjectKey{Name: "worker"}, workerMCP)).To(Succeed())
+					_ = wait.With(fxt.Client).
```
Tal-or: missing error handling here

test/e2e/serial/tests/configuration.go
```diff
@@ -1276,6 +1276,230 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+
+				var updatedNRO nropv1.NUMAResourcesOperator
+				Eventually(func(g Gomega) {
+					g.Expect(fxt.Client.Get(ctx, nroKey, &updatedNRO)).To(Succeed())
+					updatedNRO.Spec.NodeGroups = newNodeGroups
+					g.Expect(fxt.Client.Update(ctx, &updatedNRO)).To(Succeed())
+				}).WithTimeout(10*time.Minute).WithPolling(30*time.Second).Should(Succeed(), "failed to update node groups")
+
+				defer func() {
+					By(fmt.Sprintf("CLEANUP: reverting NodeGroups in NUMAResourcesOperator object %q", initialOperObj.Name))
```
Tal-or: `fxt.By()`

test/e2e/serial/tests/configuration.go
```diff
@@ -1276,6 +1276,230 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+						g.Expect(cond).ToNot(BeNil(), "condition Available was not found: %+v", revertNRO.Status.Conditions)
+						g.Expect(cond.Status).To(Equal(metav1.ConditionTrue), "expected operator Available=True, got: %+v", revertNRO.Status.Conditions)
+					}).WithTimeout(5*time.Minute).WithPolling(5*time.Second).Should(Succeed(), "operator did not return to Available state in time")
+				}()
+
+				nroNamespace := initialOperObj.Status.DaemonSets[0].Namespace
+				masterDSKey := client.ObjectKey{Namespace: nroNamespace, Name: objectnames.GetComponentName(updatedNRO.Name, masterPoolName)}
+				workerDSKey := client.ObjectKey{Namespace: nroNamespace, Name: objectnames.GetComponentName(updatedNRO.Name, workerPoolName)}
+
+				By(fmt.Sprintf("waiting for both DaemonSets to be created (%q and %q)", masterDSKey.Name, workerDSKey.Name))
```
Tal-or, line 1573: `fxt.By()`

test/e2e/serial/tests/configuration.go
```diff
@@ -1276,6 +1276,230 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+
+				var updatedNRO nropv1.NUMAResourcesOperator
+				Eventually(func(g Gomega) {
+					g.Expect(fxt.Client.Get(ctx, nroKey, &updatedNRO)).To(Succeed())
+					updatedNRO.Spec.NodeGroups = newNodeGroups
+					g.Expect(fxt.Client.Update(ctx, &updatedNRO)).To(Succeed())
+				}).WithTimeout(10*time.Minute).WithPolling(30*time.Second).Should(Succeed(), "failed to update node groups")
+
+				defer func() {
+					By(fmt.Sprintf("reverting NodeGroups in NUMAResourcesOperator object %q", initialOperObj.Name))
```
Tal-or: `fxt.By()`

Tal-or (commented):
looking good, minor comments

test/e2e/label/label.go
rshemtov13: Yes, it's possible to detect it on runtime, but I think it's still a good option to keep this label just like we have for platform for hypershift.
I'll create a generic function for clustertype detection and skip the tests accordingly.

rshemtov13:
@Tal-or Addressed your review comments in the latest commit, will squash once review finished.

test/e2e/serial/tests/configuration.go
```diff
@@ -1841,6 +2084,89 @@ func updatePerformanceProfileFieldUnstructured(dynamicClient dynamic.Interface,
 	return
 }
 
+func getClusterType(ctx context.Context, cli client.Client) label.ClusterType {
```
Tal-or, line 2087: I think we can simplify this function:
you have 3 criteria  you care about:
1. `oc get schedulers.config.openshift.io cluster -o jsonpath="{.spec.mastersSchedulable}"` : `true`/`false`
2. `oc get infrastructures.config.openshift.io cluster -o jsonpath="{.status.controlPlaneTopology}"` : `SingleReplica`/`HighlyAvailable`
3. `len(allNodes)`

```
if SingleReplica && mastersSchedulable:
    return SNO

if HighlyAvailable && mastersSchedulable:
   if len(allNodes) == 3:
       label.Compact
   return label.MNOMastersSchedulable

return label.MNO
```

WDYT? unless there is a compat cluster with more than 3 nodes which breaks this assumption.

rshemtov13: Hmm actually yes I can implement it as such, since this is the only way we'll run the serial suite.
We also don't need the SNO check since it's not even an option for us or for NROP.

rshemtov13: addressed and squashed into all commits.

Tal-or (approved):
/lgtm
/apptove

Thank you for your patience, good work

Tal-or (approved):
/lgtm
/approve

PR #3363, opened 2026-03-26
e2e: add TLS profile modification test

shajmakh:
Add e2e test that verifies the deployment is adhering to the new TLS profile and that the scheduler pods are restarted with the new settings.

Assisted-by: Cursor v2.3.37
AI-Attribution: AIA HAb Hin R Cursor v2.3.37 model:claude-4.6-opus-high

PR #3491, opened 2026-04-07
E2E: add test to check RTE tls args

mrniranjan:
Add E2E test to check tls args passed to rte container.

AI Attribution: AIA Human-AI blend, New content, Human-initiated, Reviewed v1.0

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -18,9 +18,11 @@ package tests
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/url"
 	"reflect"
 	"sort"
 	"strconv"
@@ -28,6 +30,7 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+	"golang.org/x/sync/errgroup"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -52,6 +55,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	machineconfigv1 "github.com/openshift/api/machineconfiguration/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
+	ctrltls "github.com/openshift/controller-runtime-common/pkg/tls"
 
 	depnodes "github.com/k8stopologyawareschedwg/deployer/pkg/clientutil/nodes"
 	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
@@ -72,6 +76,7 @@ import (
 	"github.com/openshift-kni/numaresources-operator/internal/wait"
 	"github.com/openshift-kni/numaresources-operator/pkg/kubeletconfig"
 	"github.com/openshift-kni/numaresources-operator/pkg/objectnames"
+	objtls "github.com/openshift-kni/numaresources-operator/pkg/objectupdate/tls"
 	"github.com/openshift-kni/numaresources-operator/pkg/status"
 	"github.com/openshift-kni/numaresources-operator/pkg/validation"
 	rteconfig "github.com/openshift-kni/numaresources-operator/rte/pkg/config"
@@ -84,6 +89,7 @@ import (
 	e2enrt "github.com/openshift-kni/numaresources-operator/test/internal/noderesourcetopologies"
 	"github.com/openshift-kni/numaresources-operator/test/internal/nrosched"
 	"github.com/openshift-kni/numaresources-operator/test/internal/objects"
+	intls "github.com/openshift-kni/numaresources-operator/test/internal/tls"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -1814,6 +1820,101 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 			})
 		})
 	})
+
+	Context("[tlscompliance][rte] rte complies with TLS Profile modifications", Label(label.Tier2, "feature:tlscompliance"), func() {
+		const (
+			rteDaemonSetCheckTimeout  = 30 * time.Second
+			rteDaemonSetCheckInterval = 5 * time.Second
+			apiServerUpdateTimeout    = 10 * time.Minute
+			apiServerUpdateInterval   = 10 * time.Second
+		)
+		It("[test_id:88383] should update RTE daemonset TLS args to adhere to modified TLS profile", Label(label.Reboot, label.Slow), func(ctx context.Context) {
+			fxt.IsRebootTest = true
+			mcpUpdateTimeout := 40 * time.Minute
+			mcpUpdatePolling := 20 * time.Second
+
+			waitForMCPsAndVerifyNodes := func(ctx context.Context, reason string) {
+				e2efixture.By("waiting for all MCPs to get updated")
+				eg := errgroup.Group{}
+				allMCPs := &machineconfigv1.MachineConfigPoolList{}
+				Expect(e2eclient.Client.List(ctx, allMCPs)).To(Succeed())
+				for _, mcp := range allMCPs.Items {
+					eg.Go(func() error {
+						err := waitForMCPConditionWithRetry(ctx, &mcp, machineconfigv1.MachineConfigPoolUpdating, mcpUpdateTimeout, mcpUpdatePolling)
+						if err != nil {
+							klog.InfoS("failed to wait for MCP to start updating", "mcp", mcp.Name, "err", err)
+							return err
+						}
+						klog.InfoS("MCP started updating; waiting for it to complete", "mcp", mcp.Name)
+						return waitForMCPConditionWithRetry(ctx, &mcp, machineconfigv1.MachineConfigPoolUpdated, mcpUpdateTimeout, mcpUpdatePolling)
+					})
+				}
+				Expect(eg.Wait()).To(Succeed(), "failed while waiting for mcp update")
+
+				e2efixture.By("verifying per-node MachineConfig alignment %s", reason)
+				for i := range allMCPs.Items {
+					Expect(verifyNodesMatchMCPConfig(ctx, &allMCPs.Items[i])).To(Succeed(),
+						"node config mismatch for MCP %q", allMCPs.Items[i].Name)
+				}
+			}
+
+			e2efixture.By("determining a different TLS profile to apply")
+			apiServerObj := &configv1.APIServer{}
+			apiServerKey := client.ObjectKey{Name: ctrltls.APIServerName}
+			Expect(e2eclient.Client.Get(ctx, apiServerKey, apiServerObj)).To(Succeed(), "failed to get APIServer object")
+			originalTLSProfile := apiServerObj.Spec.TLSSecurityProfile
+			// OpenShift defaults to Intermediate when no profile is explicitly set.
+			originalTLSProfileType := configv1.TLSProfileIntermediateType
+			if originalTLSProfile != nil {
+				originalTLSProfileType = originalTLSProfile.Type
+			}
+
+			newProfile := &configv1.TLSSecurityProfile{
+				Type:   configv1.TLSProfileModernType,
+				Modern: &configv1.ModernTLSProfile{},
+			}
+
+			if originalTLSProfileType == newProfile.Type {
+				newProfile = &configv1.TLSSecurityProfile{
+					Type:         configv1.TLSProfileIntermediateType,
+					Intermediate: &configv1.IntermediateTLSProfile{},
+				}
+			}
+
+			e2efixture.By("updating APIServer TLS profile from %v to %v", originalTLSProfileType, newProfile.Type)
+			Eventually(func(g Gomega) {
+				updatedAPIServer := &configv1.APIServer{}
+				g.Expect(e2eclient.Client.Get(ctx, apiServerKey, updatedAPIServer)).To(Succeed(), "failed to get APIServer object")
+				updatedAPIServer.Spec = apiServerObj.Spec
+				updatedAPIServer.Spec.TLSSecurityProfile = newProfile
+				g.Expect(e2eclient.Client.Update(ctx, updatedAPIServer)).To(Succeed(), "failed to update APIServer TLS profile")
+			}).WithTimeout(apiServerUpdateTimeout).WithPolling(apiServerUpdateInterval).Should(Succeed())
+
+			DeferCleanup(func(ctx context.Context) {
+				e2efixture.By("reverting APIServer TLS profile from %v to %v", newProfile.Type, originalTLSProfileType)
+				Eventually(func(g Gomega) {
+					updatedAPIServer := &configv1.APIServer{}
+					g.Expect(e2eclient.Client.Get(ctx, apiServerKey, updatedAPIServer)).To(Succeed(), "failed to get APIServer object")
+					updatedAPIServer.Spec = apiServerObj.Spec
+					g.Expect(e2eclient.Client.Update(ctx, updatedAPIServer)).To(Succeed(), "failed to update APIServer TLS profile")
+				}).WithTimeout(apiServerUpdateTimeout).WithPolling(apiServerUpdateInterval).Should(Succeed())
+
+				waitForMCPsAndVerifyNodes(ctx, "after MCP rollback")
+			})
+
+			waitForMCPsAndVerifyNodes(ctx, "after MCP rollout")
+
+			newTLSProfileSpec := *configv1.TLSProfiles[newProfile.Type]
+			newTLSConfigFn, _ := ctrltls.NewTLSConfigFromProfile(newTLSProfileSpec)
+			newTLSCfg := &tls.Config{}
+			newTLSConfigFn(newTLSCfg)
+			expectedTLSSettings := objtls.NewSettings(newTLSCfg)
+			e2efixture.By("verifying RTE DaemonSet TLS flags match the updated profile: %v", expectedTLSSettings)
+			Eventually(func(g Gomega) {
+				g.Expect(intls.CheckRTEDaemonSetTLSFlags(ctx, e2eclient.Client, expectedTLSSettings)).To(Succeed())
+			}).WithTimeout(rteDaemonSetCheckTimeout).WithPolling(rteDaemonSetCheckInterval).Should(Succeed())
+		})
+	})
 })
 
 func isNROSchedSyncedAt(nrosStatus *nropv1.NUMAResourcesSchedulerStatus, conditionType string, gen int64) error {
@@ -2173,4 +2274,93 @@ func rteConfigFrom(data string) (*rteconfig.Config, error) {
 		return nil, err
 	}
 	return conf, nil
+}
+
+// waitForMCPConditionWithRetry wraps ForMachineConfigPoolCondition with automatic
+// retries for transient network errors (e.g. connection resets, HTTP/2 stream errors,
+// 504 gateway timeouts) that are common during control-plane churn.
+// Unlike the plain ForMachineConfigPoolCondition call, which treats any API error as
+// terminal, this function distinguishes transient transport-level failures from real
+// condition failures and retries only the former.
+// The overall wall-clock time is bounded by timeout; each retry consumes the remaining
+// budget rather than resetting it.
+func waitForMCPConditionWithRetry(ctx context.Context, mcp *machineconfigv1.MachineConfigPool, condType machineconfigv1.MachineConfigPoolConditionType, timeout, interval time.Duration) error {
+	deadline := time.Now().Add(timeout)
+	var lastErr error
+	for {
+		remaining := time.Until(deadline)
+		if remaining <= 0 {
+			if lastErr != nil {
+				return fmt.Errorf("timed out waiting for MCP %q condition %q: %w", mcp.Name, condType, lastErr)
+			}
+			return fmt.Errorf("timed out waiting for MCP %q condition %q", mcp.Name, condType)
+		}
+
+		err := wait.With(e2eclient.Client).Timeout(remaining).Interval(interval).ForMachineConfigPoolCondition(ctx, mcp, condType)
+		if err == nil {
+			return nil
+		}
+		if !isTransientMCPError(err) {
+			return err
+		}
+
+		lastErr = err
+		klog.InfoS("transient MCP wait error; retrying", "mcp", mcp.Name, "condition", condType, "err", err)
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(interval):
+		}
+	}
+}
+
+// isTransientMCPError returns true for transport-level errors (url.Error, HTTP/2
+// stream errors, 504 gateway timeouts) that are expected during MCP transitions
+// when the API server is temporarily overloaded or unreachable, and are safe to retry.
+func isTransientMCPError(err error) bool {
+	var urlErr *url.Error
+	if errors.As(err, &urlErr) {
+		return true
+	}
+	if apierrors.IsTimeout(err) || apierrors.IsServerTimeout(err) || apierrors.IsServiceUnavailable(err) {
+		return true
+	}
+	// API transport hiccups seen during control-plane churn may be wrapped.
+	return strings.Contains(err.Error(), "stream error") || strings.Contains(err.Error(), "INTERNAL_ERROR")
+}
+
+// verifyNodesMatchMCPConfig re-reads the given MCP, lists the nodes matching
+// its NodeSelector, and checks that every node's currentConfig annotation
+// matches the MCP's rendered configuration (status.configuration.name).
+// This provides per-node confidence that the rollout actually happened.
+func verifyNodesMatchMCPConfig(ctx context.Context, mcp *machineconfigv1.MachineConfigPool) error {
+	updatedMCP := &machineconfigv1.MachineConfigPool{}
+	if err := e2eclient.Client.Get(ctx, client.ObjectKeyFromObject(mcp), updatedMCP); err != nil {
+		return fmt.Errorf("failed to re-read MCP %q: %w", mcp.Name, err)
+	}
+	renderedConfig := updatedMCP.Status.Configuration.Name
+	if renderedConfig == "" {
+		return fmt.Errorf("MCP %q has empty status.configuration.name", mcp.Name)
+	}
+
+	if updatedMCP.Spec.NodeSelector == nil {
+		return fmt.Errorf("MCP %q has nil NodeSelector", mcp.Name)
+	}
+	nodeList := &corev1.NodeList{}
+	if err := e2eclient.Client.List(ctx, nodeList, &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(updatedMCP.Spec.NodeSelector.MatchLabels),
+	}); err != nil {
+		return fmt.Errorf("failed to list nodes for MCP %q: %w", mcp.Name, err)
+	}
+
+	for i := range nodeList.Items {
+		node := &nodeList.Items[i]
+		currentConfig := node.Annotations[wait.CurrentConfigNodeAnnotation]
+		if currentConfig != renderedConfig {
+			return fmt.Errorf("node %q currentConfig=%q does not match MCP %q rendered config=%q",
+				node.Name, currentConfig, mcp.Name, renderedConfig)
+		}
+	}
+	klog.InfoS("all nodes match MCP config", "mcp", mcp.Name, "config", renderedConfig, "nodeCount", len(nodeList.Items))
+	return nil
 }
--- a/test/internal/tls/tls.go
+++ b/test/internal/tls/tls.go
@@ -17,12 +17,14 @@
 package tls
 
 import (
+	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
 	"net"
 	"strings"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
@@ -32,7 +34,14 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	libgocrypto "github.com/openshift/library-go/pkg/crypto"
 
+	"github.com/k8stopologyawareschedwg/deployer/pkg/flagcodec"
+	k8swgobjupdate "github.com/k8stopologyawareschedwg/deployer/pkg/objectupdate"
+
+	nropv1 "github.com/openshift-kni/numaresources-operator/api/v1"
 	"github.com/openshift-kni/numaresources-operator/internal/remoteexec"
+	"github.com/openshift-kni/numaresources-operator/pkg/objectnames"
+	rteupdate "github.com/openshift-kni/numaresources-operator/pkg/objectupdate/rte"
+	objtls "github.com/openshift-kni/numaresources-operator/pkg/objectupdate/tls"
 )
 
 // ErrTLSHandshakeRejected is a base error for any TLS connection refusal.
@@ -190,4 +199,52 @@ func TLSVersionBelow(v uint16) (uint16, error) {
 	default:
 		return 0, fmt.Errorf("unknown TLS version %d", v)
 	}
+}
+
+func CheckRTEDaemonSetTLSFlags(ctx context.Context, cli client.Client, expected objtls.Settings) error {
+	nropObj := &nropv1.NUMAResourcesOperator{}
+	if err := cli.Get(ctx, client.ObjectKey{Name: objectnames.DefaultNUMAResourcesOperatorCrName}, nropObj); err != nil {
+		return fmt.Errorf("failed to get NUMAResourcesOperator: %w", err)
+	}
+	if len(nropObj.Status.NodeGroups) == 0 {
+		return fmt.Errorf("expect the numaresourcesoperator to have at least one NodeGroup in status")
+	}
+	for _, ng := range nropObj.Status.NodeGroups {
+		ds := &appsv1.DaemonSet{}
+		if err := cli.Get(ctx, client.ObjectKey{Namespace: ng.DaemonSet.Namespace, Name: ng.DaemonSet.Name}, ds); err != nil {
+			return fmt.Errorf("failed to get DaemonSet %s/%s: %w", ng.DaemonSet.Namespace, ng.DaemonSet.Name, err)
+		}
+		rteCnt := k8swgobjupdate.FindContainerByName(ds.Spec.Template.Spec.Containers, rteupdate.MainContainerName)
+		if rteCnt == nil {
+			return fmt.Errorf("main container not found daemonsetName=%q", ds.Name)
+		}
+		rteFlags := flagcodec.ParseArgvKeyValue(rteCnt.Args, flagcodec.WithFlagNormalization)
+		if err := matchTLSFlag(rteFlags, "--metrics-tls-min-version", expected.MinVersion, ds.Name); err != nil {
+			return err
+		}
+		// TLS 1.3 cipher suites are fixed by the Go crypto/tls implementation and cannot
+		// be configured, so expected.CipherSuites will be empty for the Modern profile.
+		if err := matchTLSFlag(rteFlags, "--metrics-tls-cipher-suites", expected.CipherSuites, ds.Name); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func matchTLSFlag(rteFlags *flagcodec.Flags, flagName, expected, dsName string) error {
+	val, found := rteFlags.GetFlag(flagName)
+	if expected == "" {
+		if found && val.Data != "" {
+			return fmt.Errorf("%s should be absent or empty for this TLS profile daemonsetName=%q got=%q", flagName, dsName, val.Data)
+		}
+		return nil
+	}
+	if !found || val.Data != expected {
+		got := ""
+		if found {
+			got = val.Data
+		}
+		return fmt.Errorf("%s mismatch daemonsetName=%q expected=%q got=%q", flagName, dsName, expected, got)
+	}
+	return nil
 }
--- a/test/internal/tls/tls_test.go
+++ b/test/internal/tls/tls_test.go
@@ -17,14 +17,34 @@
 package tls
 
 import (
+	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"strings"
 	"testing"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	configv1 "github.com/openshift/api/config/v1"
 	libgocrypto "github.com/openshift/library-go/pkg/crypto"
+
+	nropv1 "github.com/openshift-kni/numaresources-operator/api/v1"
+	"github.com/openshift-kni/numaresources-operator/pkg/objectnames"
+	rteupdate "github.com/openshift-kni/numaresources-operator/pkg/objectupdate/rte"
+	objtls "github.com/openshift-kni/numaresources-operator/pkg/objectupdate/tls"
 )
+
+func init() {
+	utilruntime.Must(nropv1.AddToScheme(scheme.Scheme))
+}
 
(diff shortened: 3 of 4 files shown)


ffromani (commented):
I'm torn of this tests belongs here or in `test/e2e/serial/tests/configuration.go`

mrniranjan:
> I'm torn of this tests belongs here or in `test/e2e/serial/tests/configuration.go`

i am fine either way, let me know if you prefer configuration.go

test/e2e/rte/rte_test.go
```diff
@@ -305,6 +312,52 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+			Expect(clients.Client.Get(ctx, client.ObjectKey{Name: objectnames.DefaultNUMAResourcesOperatorCrName}, nropObj)).To(Succeed())
+			Eventually(func() error {
+				if err := clients.Client.Get(ctx, client.ObjectKey{Name: objectnames.DefaultNUMAResourcesOperatorCrName}, nropObj); err != nil {
+					return fmt.Errorf("failed to get NUMAResourcesOperator: %w", err)
+				}
+				if len(nropObj.Status.NodeGroups) == 0 {
+					return fmt.Errorf("expect the numaresourcesoperator to have at least one NodeGroup in status")
+				}
+				for _, ng := range nropObj.Status.NodeGroups {
+					ds, err := clients.K8sClient.AppsV1().DaemonSets(ng.DaemonSet.Namespace).Get(ctx, ng.DaemonSet.Name, metav1.GetOptions{})
```
Tal-or: why we're mixing different clients?

Tal-or (commented):
overall LGTM

left minor comment and we need @ffromani to decide about https://github.com/openshift-kni/numaresources-operator/pull/3491#pullrequestreview-4075211671 

it can go in `configuration.go` IMO as well

ffromani:
> > I'm torn of this tests belongs here or in `test/e2e/serial/tests/configuration.go`
> 
> i am fine either way, let me know if you prefer configuration.go

We can keep as-is and move later. No change needed

test/e2e/rte/rte_test.go
ffromani: Indeed. Please use `clients.Client` every time possible, resort to `K8sClient` if the `clients.Client` is not up to the task. The best would be to use `fixture.Fixture` though.

test/e2e/rte/rte_test.go
```diff
@@ -51,18 +53,23 @@ import (
...
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
+var (
+	timeout  = 30 * time.Second
+	interval = 5 * time.Second
+)
+
```
ffromani: please avoid the global scope if we can. Move in the smallest possible scope.

test/e2e/rte/rte_test.go
```diff
@@ -305,6 +309,54 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+					}
+					ciphers, found := rteFlags.GetFlag("--metrics-tls-cipher-suites")
+					if !found || ciphers.Data != tlsSettings.CipherSuites {
+						return fmt.Errorf("TLS cipher suites mismatch daemonsetName=%q expected=%q got=%q",
+							ds.Name, tlsSettings.CipherSuites, ciphers.Data)
+					}
+				}
+				return nil
+			}).WithTimeout(timeout).WithPolling(interval).Should(Succeed())
+		})
```
shajmakh: as a start the test is good but not enough to ensure the RTEs are restarting with the new settings upon TLS profile updates

mrniranjan: yes working on it, i will add more tests in subsequent PR's

shajmakh: thank you, then we can move forward with this.
to avoid unneeded reboots, we can add this check to the same test that modifies the profile here:
https://github.com/openshift-kni/numaresources-operator/pull/3363/changes#top or later combine them depends on what get's merged first)

test/e2e/rte/rte_test.go
```diff
@@ -305,6 +309,54 @@ var _ = Describe("with a running cluster with all the components", func() {
 			}
 		}
 	})
+	Context("[tlscompliance][rte] rte complies with TLS Profile modifications", Label(label.Tier0, "feature:tlscompliance"), func() {
+		timeout := 30 * time.Second
+		interval := 5 * time.Second
```
ffromani: Variables MUST be initialized in a BeforeEach block

ffromani (commented):
A remaining inline comment, other than that LGTM

test/e2e/rte/rte_test.go
```diff
@@ -322,6 +332,208 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+			rteDaemonSetCheckInterval time.Duration
+			apiServerUpdateTimeout    time.Duration
+			apiServerUpdateInterval   time.Duration
+		)
+		BeforeEach(func() {
+			rteDaemonSetCheckTimeout = 30 * time.Second
+			rteDaemonSetCheckInterval = 5 * time.Second
+			apiServerUpdateTimeout = 10 * time.Minute
+			apiServerUpdateInterval = 10 * time.Second
+		})
```
shajmakh: these should never change, we can make them a const without having before each. simply add them at the beginning of Context function.

test/e2e/rte/rte_test.go
```diff
@@ -322,6 +332,208 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+				// Wait for all pools to complete rollback reconciliation.
+				klog.InfoS("Waiting for all MCPs to get updated")
+				// Use goroutines so all pools are monitored in parallel.
+				eg := errgroup.Group{}
+				// Container for current MCP list.
+				allMCPs := &mcov1.MachineConfigPoolList{}
+				// List all pools present in cluster.
+				Expect(clients.Client.List(ctx, allMCPs)).To(Succeed())
+				// Launch one waiter per pool.
+				for _, mcp := range allMCPs.Items {
```
shajmakh: you'll need to exclude the MCPs that has no machine counts, otherwise the wait on line 458 will fail for such MCPs if any existed which is expected.

test/e2e/rte/rte_test.go
```diff
@@ -322,6 +332,208 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+						if err != nil {
+							klog.InfoS("failed to wait for MCP to start updating", "mcp", mcp.Name, "err", err)
+							return err
+						}
+						klog.InfoS("MCP started updating; waiting for it to complete", "mcp", mcp.Name)
+						return waitForMCPConditionWithRetry(ctx, &mcp, mcov1.MachineConfigPoolUpdated, mcpUpdateTimeout, mcpUpdatePolling)
+					})
+				}
+				// Ensure all pool waiters completed successfully.
+				Expect(eg.Wait()).To(Succeed(), "failed to wait for all MCPs to start updating")
```
shajmakh: we can't have this if we know some nodes will get stuck until manual intervention. we need other workaround for the current behavior -which I still think it is a bug.

shajmakh: IOW this assertion will always fail. we need to handle the result of the messy MCP behavior that cannot be automatic wait. It isn't always about turning the nodes to schedulable again and that is it; because the stuck node will likely be degraded and the its machine config might not be updated, to fix that we need to edit the node's annotations manually and wait again for the MCP to report updated. If we don't do that we cannot guarantee that the target node is indeed adhering to the new TLS profile.

test/e2e/rte/rte_test.go
```diff
@@ -322,6 +332,208 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+						klog.InfoS("failed to wait for MCP to start updating", "mcp", mcp.Name, "err", err)
+						// Propagate error to errgroup.
+						return err
+					}
+					// Log that pool started updating.
+					klog.InfoS("MCP started updating; waiting for it to complete", "mcp", mcp.Name)
+					// Then wait until pool reports Updated condition.
+					return waitForMCPConditionWithRetry(ctx, &mcp, mcov1.MachineConfigPoolUpdated, mcpUpdateTimeout, mcpUpdatePolling)
+				})
+			}
```
shajmakh: same comments apply here

test/e2e/rte/rte_test.go
```diff
@@ -322,6 +332,208 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+			// Ensure all pool waiters completed successfully.
+			Expect(eg.Wait()).To(Succeed(), "failed to wait for all MCPs to start updating")
```
shajmakh: "failed to wait for all MCPs to start updating" -> failed while waiting for MCP update"

test/e2e/rte/rte_test.go
```diff
@@ -322,6 +332,208 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+					if err := matchTLSFlag(rteFlags, "--metrics-tls-cipher-suites", tlsSettings.CipherSuites, ds.Name); err != nil {
+						return err
+					}
+				}
+				return nil
+			}).WithTimeout(rteDaemonSetCheckTimeout).WithPolling(rteDaemonSetCheckInterval).Should(Succeed())
+		})
+
+		It("[test_id:88383] should update scheduler deployment TLS args to adhere to modified TLS profile", Label(label.Slow, label.Reboot), func(ctx context.Context) {
+			// Maximum time allowed for each MCP reconciliation cycle.
```
shajmakh: I believe we should have this test in a different PR because it is hitting an unsual MCP behavior.

test/e2e/rte/rte_test.go
```diff
@@ -368,3 +598,56 @@ func toJSON(obj interface{}) string {
...
+
+// waitForMCPConditionWithRetry wraps ForMachineConfigPoolCondition with automatic
+// retries for transient network errors (e.g. connection resets, HTTP/2 stream errors,
+// 504 gateway timeouts) that are common during control-plane churn.
+// Unlike the plain ForMachineConfigPoolCondition call, which treats any API error as
+// terminal, this function distinguishes transient transport-level failures from real
+// condition failures and retries only the former.
+// The overall wall-clock time is bounded by timeout; each retry consumes the remaining
+// budget rather than resetting it.
+func waitForMCPConditionWithRetry(ctx context.Context, mcp *mcov1.MachineConfigPool, condType mcov1.MachineConfigPoolConditionType, timeout, interval time.Duration) error {
```
shajmakh: nice idea to have this wrapper around ForMachineConfigPoolCondition, but we still cannot tell if the nodes hitting the transition error are actually adhering to the new profile

shajmakh (commented):
I'm not confident about the test that updates the TLS profile. I would prefer to have it in a different PR, considering the unusual behavior in MCP updates. 
I explained why in the comments below. 
How is the test doing when you run it multiple sequential times without manual intervention?

test/e2e/rte/rte_test.go
```diff
@@ -359,19 +375,163 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+			Eventually(func(g Gomega) {
+				updatedAPIServer := &configv1.APIServer{}
+				g.Expect(clients.Client.Get(ctx, apiServerKey, updatedAPIServer)).To(Succeed(), "failed to get APIServer object")
+				updatedAPIServer.Spec = apiServerObj.Spec
+				updatedAPIServer.Spec.TLSSecurityProfile = newProfile
+				g.Expect(clients.Client.Update(ctx, updatedAPIServer)).To(Succeed(), "failed to updated APIServer TLS profile")
+			}).WithTimeout(apiServerUpdateTimeout).WithPolling(apiServerUpdateInterval).Should(Succeed())
+
+			// Always restore original TLS profile after test execution.
+			defer func() {
```
swatisehgal: This is a good place to use `DeferCleanup()` instead of `defer`.

test/e2e/rte/rte_test.go
```diff
@@ -359,19 +375,163 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+						return fmt.Errorf("main container not found daemonsetName=%q", ds.Name)
+					}
+					rteFlags := flagcodec.ParseArgvKeyValue(rteCnt.Args, flagcodec.WithFlagNormalization)
+					if err := matchTLSFlag(rteFlags, "--metrics-tls-min-version", expectedTLSSettings.MinVersion, ds.Name); err != nil {
+						return err
+					}
+					// in the case of TLS1.3 the expectedTLSSettings.CipherSuites will be empty
+					// When using a minimum of TLS 1.3—which is mandated by the Modern profile—you generally do not pass or explicitly configure cipher suites
+					if err := matchTLSFlag(rteFlags, "--metrics-tls-cipher-suites", expectedTLSSettings.CipherSuites, ds.Name); err != nil {
+						return err
```
swatisehgal: We’re repeating the same DaemonSet TLS flag validation logic in both tests (fetch DS, find RTE container,  parse flags, validate TLS args). It might be worth extracting this into a small helper (something like validateRTETLSFlags(...)) to reduce duplication.

test/e2e/rte/rte_test.go
```diff
@@ -305,6 +312,53 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+			Eventually(func() bool {
+				rteDss, err := getOwnedDss(clients.K8sClient, nropObj.ObjectMeta)
+				if err != nil {
+					klog.ErrorS(err, "failed to get the owned DaemonSets")
+					return false
+				}
+				if len(rteDss) == 0 {
+					klog.InfoS("expect the numaresourcesoperator to own at least one DaemonSet")
+					return false
+				}
```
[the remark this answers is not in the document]
mrniranjan: fixed in the latest commit

test/e2e/rte/rte_test.go
```diff
@@ -305,6 +312,53 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+			tlsConfig, _ := ctrltls.NewTLSConfigFromProfile(tlsProfileSpec)
+			tlsCfg := &tls.Config{}
+			tlsConfig(tlsCfg)
+			tlsSettings := objtls.NewSettings(tlsCfg)
+			klog.InfoS("Initial TLS Settings", "tlsSettings", tlsSettings)
+			By("Getting the initial NRO operator object")
+			nropObj := &nropv1.NUMAResourcesOperator{}
+			err = clients.Client.Get(ctx, client.ObjectKey{Name: objectnames.DefaultNUMAResourcesOperatorCrName}, nropObj)
+			Expect(err).ToNot(HaveOccurred())
+			Eventually(func() bool {
```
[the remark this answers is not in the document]
mrniranjan: fixed in the latest commit

test/e2e/rte/rte_test.go
```diff
@@ -305,6 +312,53 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+					klog.InfoS("expect the numaresourcesoperator to own at least one DaemonSet")
+					return false
+				}
+				for _, ds := range rteDss {
+					rteCnt := k8swgobjupdate.FindContainerByName(ds.Spec.Template.Spec.Containers, rteupdate.MainContainerName)
+					if rteCnt == nil {
+						klog.InfoS("main container not found", "daemonsetName", ds.Name)
+						return false
+					}
+					matchFn := getFlagMapper(rteCnt)
```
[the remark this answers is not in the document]
mrniranjan: fixed in the latest commit

test/e2e/rte/rte_test.go
```diff
@@ -305,6 +312,53 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+			Expect(err).ToNot(HaveOccurred(), "Unable to get TLS Profile from APIServer")
+			tlsConfig, _ := ctrltls.NewTLSConfigFromProfile(tlsProfileSpec)
+			tlsCfg := &tls.Config{}
+			tlsConfig(tlsCfg)
+			tlsSettings := objtls.NewSettings(tlsCfg)
+			klog.InfoS("Initial TLS Settings", "tlsSettings", tlsSettings)
+			By("Getting the initial NRO operator object")
+			nropObj := &nropv1.NUMAResourcesOperator{}
+			err = clients.Client.Get(ctx, client.ObjectKey{Name: objectnames.DefaultNUMAResourcesOperatorCrName}, nropObj)
+			Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
mrniranjan: fixed in the latest commit

test/e2e/rte/rte_test.go
```diff
@@ -305,6 +312,53 @@ var _ = Describe("with a running cluster with all the components", func() {
 			}
 		}
 	})
+	Context("[tlscompliance][rte] rte complies with TLS Profile modifications", Label(label.Tier0, "feature:tlscompliance"), func() {
+		It("should have RTE DaemonSet args aligned with the cluster TLS profile", func() {
+			var ctx context.Context = context.Background()
```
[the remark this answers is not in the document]
mrniranjan: Fixed in latest commits

test/e2e/rte/rte_test.go
mrniranjan: fixed in latest commits

test/e2e/rte/rte_test.go
mrniranjan: Addressed in latest commits

test/e2e/rte/rte_test.go
mrniranjan: Fixed in latest commit

test/e2e/rte/rte_test.go
mrniranjan: Why will there be any MCP's with no machine counts ?

test/e2e/rte/rte_test.go
mrniranjan: Is the above comment is based on Stuck MCP behavior that we are seeing. Because based on my  testing i did not encounter the mcp getting stuck once i fixed the topology ( the number of worker/worker-cnf nodes)

test/e2e/rte/rte_test.go
mrniranjan: fixed in latest commit

test/e2e/rte/rte_test.go
mrniranjan: Addressed this in latest commit where i am double checking not only did mcp get updated but also the mcp has rendered the desired configuration

test/e2e/rte/rte_test.go
```diff
@@ -464,9 +453,14 @@ var _ = Describe("with a running cluster with all the components", func() {
...
-				// Ensure all pool waiters completed successfully.
 				Expect(eg.Wait()).To(Succeed(), "failed to wait for all MCPs to start updating")
-			}()
+
+				By("Verifying per-node MachineConfig alignment after MCP rollback")
+				for i := range allMCPs.Items {
+					Expect(verifyNodesMatchMCPConfig(ctx, &allMCPs.Items[i])).To(Succeed(),
+						"node config mismatch for MCP %q", allMCPs.Items[i].Name)
+				}
+			})
```
shajmakh: why do we need these? we can rely on MCO for this. if it reports the MCP field Updated=true then that is it we keep moving with the test. what am I missing?

mrniranjan: In your earlier review you mentioned below , to make sure the nodes are actually adhering to the new profile by  checking the machine config that got created by profile modification is the one we switched to. 

`nice idea to have this wrapper around ForMachineConfigPoolCondition, but we still cannot tell if the nodes hitting the transition error are actually adhering to the new profile`

test/e2e/rte/rte_test.go
shajmakh: you are right, they should not be in this test context, but its existence is still valid, especially in the serial suite where I initially ran the scheduler e2e check. you can ignore that here

test/e2e/rte/rte_test.go
shajmakh: Sorry for the confusion. I had my [old comment ](https://github.com/openshift-kni/numaresources-operator/pull/3491#discussion_r3136083870) written before I knew about the PDB behavior and I thought that was a bug in MCO. 
Regarding the additional current vs desired check, I believe this is an additional way to confirm that the previous check of updated MCPs is true, basically a duplicate verification. However, it's fine to keep it considering its only verification of values and not waiting.

test/e2e/rte/rte_test.go
mrniranjan: Addressed in latest commits

test/e2e/rte/rte_test.go
mrniranjan: Addressed in latest commit

shajmakh:
The code LGTM except that none of the tests in test/e2e/rte/rte-test.go run in prow lanes nor in d/s from what I recall: 
https://github.com/openshift-kni/numaresources-operator/blob/main/hack/run-test-e2e.sh
I think it is enough to remove the ginkgo focus part to enable them:
` --ginkgo.focus='\[release\]'`
@mrniranjan can you try that please?
the next question is whether we want to have reboot tests running u/s, which I think not. @ffromani to comment on this.

shajmakh:
> The code LGTM except that none of the tests in test/e2e/rte/rte-test.go run in prow lanes nor in d/s from what I recall: https://github.com/openshift-kni/numaresources-operator/blob/main/hack/run-test-e2e.sh I think it is enough to remove the ginkgo focus part to enable them: ` --ginkgo.focus='\[release\]'` @mrniranjan can you try that please? the next question is whether we want to have reboot tests running u/s, which I think not. @ffromani to comment on this.

As second thought I think we need to keep `release` and add the rest of the tests tags, something like this:
`--ginkgo.focus='\[release\]|\[config\]|\[podfingerprint\]|\[tlscompliance\]'`

mrniranjan:
> The code LGTM except that none of the tests in test/e2e/rte/rte-test.go run in prow lanes nor in d/s from what I recall: https://github.com/openshift-kni/numaresources-operator/blob/main/hack/run-test-e2e.sh I think it is enough to remove the ginkgo focus part to enable them: ` --ginkgo.focus='\[release\]'` @mrniranjan can you try that please? the next question is whether we want to have reboot tests running u/s, which I think not. @ffromani to comment on this.

There  is 1 reboot and 1 non reboot, may be we can run 1 non-reboot test ? Also this will fail in ci because it needs https://github.com/openshift-kni/numaresources-operator/pull/3363 (the scheme changes specifically https://github.com/openshift-kni/numaresources-operator/pull/3363/changes#diff-e01e9ac51e985392a2296606e495a26d19fdad1f25ab3fb1355560d6abbda86b )

test/e2e/rte/rte_test.go
```diff
@@ -322,6 +333,151 @@ var _ = Describe("with a running cluster with all the components", func() {
...
+			tlsConfig(tlsCfg)
+			tlsSettings := objtls.NewSettings(tlsCfg)
+			klog.InfoS("Initial TLS Settings", "tlsSettings", tlsSettings)
+			By("Verifying RTE DaemonSet TLS flags match the cluster TLS profile")
+			Eventually(func() error {
+				return checkRTEDaemonSetTLSFlags(ctx, tlsSettings)
+			}).WithTimeout(rteDaemonSetCheckTimeout).WithPolling(rteDaemonSetCheckInterval).Should(Succeed())
+		})
+
+		It("[test_id:88383] should update scheduler deployment TLS args to adhere to modified TLS profile", Label(label.Slow, label.Reboot), func(ctx context.Context) {
```
shajmakh: "update scheduler deployment" -> should be RTE daemonset

mrniranjan: addressed in latest commit

hack/run-test-e2e.sh
```diff
@@ -36,7 +36,7 @@ export E2E_TOPOLOGY_MANAGER_SCOPE="${E2E_TOPOLOGY_MANAGER_SCOPE}"
 # --flake-attempts: rerun the test if it fails
 # -requireSuite: fail if tests are not executed because of missing suite
 echo "TM config policy=[${E2E_TOPOLOGY_MANAGER_POLICY}] scope=[${E2E_TOPOLOGY_MANAGER_SCOPE}]"
-${BIN_DIR}/e2e-nrop-rte.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.flake-attempts=2 --ginkgo.junit-report=${REPORT_DIR}/e2e.xml --ginkgo.skip='\[Disruptive\]|\[StateDirectories\]|\[NodeRefresh\]|\[local\]' --ginkgo.focus='\[release\]'
+${BIN_DIR}/e2e-nrop-rte.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.flake-attempts=2 --ginkgo.junit-report=${REPORT_DIR}/e2e.xml --ginkgo.skip='\[Disruptive\]|\[StateDirectories\]|\[NodeRefresh\]|\[local\]' --ginkgo.focus='\[release\]|\[config\]|\[podfingerprint\]|\[tlscompliance\]' --ginkgo.label-filter='!reboot_required'
```
shajmakh: we want the focus part indeed, but the skip part means for tls e2e that the test will never run in u/s . and in d/s we do not have a lane to run non-serial suites. either we  remove `--ginkgo.label-filter='!reboot_required'` or we move the reboot-required tests to serial/configuration.go

mrniranjan: i will try to see the !reboot_required and check

test/e2e/rte/rte_test.go
```diff
@@ -322,6 +333,151 @@ var _ = Describe("with a running cluster with all the components", func() {
...
 	})
+	Context("[tlscompliance][rte] rte complies with TLS Profile modifications", Label(label.Tier0, "feature:tlscompliance"), func() {
+		const (
+			rteDaemonSetCheckTimeout  = 30 * time.Second
+			rteDaemonSetCheckInterval = 5 * time.Second
+			apiServerUpdateTimeout    = 10 * time.Minute
+			apiServerUpdateInterval   = 10 * time.Second
+		)
+		It("[test_id:88380] should have RTE DaemonSet args aligned with the cluster TLS profile", func(ctx context.Context) {
+			By("Getting initial OCP TLS profile")
```
shajmakh: let's group all the config-check tests of TLS to test/e2e/tls. so 88380 goes in test/e2e/tls and let's have the reboot test live in serial/configuration with right ginkgo labels.
Next we need to enable these tests. It's on me since I added the suite but missed to actually consume in CI. 
Regarding the RTE failures we need another PR to handle those separately; they address a different issue. wdyt?

test/e2e/serial/tests/configuration.go
```diff
@@ -1571,6 +1577,138 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			By("Updating TLS profile to a new settings")
+			klog.InfoS("determining a different TLS profile to apply")
+			apiServerObj := &configv1.APIServer{}
+			apiServerKey := client.ObjectKey{Name: ctrltls.APIServerName}
+			Expect(e2eclient.Client.Get(ctx, apiServerKey, apiServerObj)).To(Succeed(), "failed to get APIServer object")
+			originalTLSProfile := apiServerObj.Spec.TLSSecurityProfile
+			originalTLSProfileType := configv1.TLSProfileIntermediateType
+			// If profile exists, use its actual type.
+			if originalTLSProfile != nil {
+				// Record the original type for later restoration.
```
shajmakh: comments can be removed unless it's explaining a non-obvious fact about why we are doing what we are doing

test/e2e/serial/tests/configuration.go
```diff
@@ -1571,6 +1577,138 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			// If cluster already uses Modern, switch to Intermediate instead.
+			if originalTLSProfileType == newProfile.Type {
+				newProfile = &configv1.TLSSecurityProfile{
+					Type:         configv1.TLSProfileIntermediateType,
+					Intermediate: &configv1.IntermediateTLSProfile{},
+				}
+			}
+
+			klog.InfoS("switching TLS profile", "from", originalTLSProfileType, "to", newProfile.Type)
+			By(fmt.Sprintf("updating APIServer TLS profile to %v", newProfile.Type))
```
shajmakh: let's use e2eFixture.By() here and in all the same places

test/e2e/serial/tests/configuration.go
```diff
@@ -1571,6 +1577,138 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			Eventually(func(g Gomega) {
+				updatedAPIServer := &configv1.APIServer{}
```
shajmakh, line 1886: can be defined outside the eventually so that it gets created once and be reused elsewhere in other eventually's

test/e2e/serial/tests/configuration.go
```diff
@@ -1571,6 +1577,138 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+				g.Expect(e2eclient.Client.Get(ctx, apiServerKey, updatedAPIServer)).To(Succeed(), "failed to get APIServer object")
+				updatedAPIServer.Spec = apiServerObj.Spec
+				updatedAPIServer.Spec.TLSSecurityProfile = newProfile
+				g.Expect(e2eclient.Client.Update(ctx, updatedAPIServer)).To(Succeed(), "failed to updated APIServer TLS profile")
```
shajmakh: nit: to updated -> to update

test/e2e/serial/tests/configuration.go
```diff
@@ -1571,6 +1577,138 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+					g.Expect(e2eclient.Client.Get(ctx, apiServerKey, updatedAPIServer)).To(Succeed(), "failed to get APIServer object")
+					updatedAPIServer.Spec = apiServerObj.Spec
+					g.Expect(e2eclient.Client.Update(ctx, updatedAPIServer)).To(Succeed(), "failed to updated APIServer TLS profile")
+				}).WithTimeout(apiServerUpdateTimeout).WithPolling(apiServerUpdateInterval).Should(Succeed())
+
+				klog.InfoS("Waiting for all MCPs to get updated")
+				eg := errgroup.Group{}
+				allMCPs := &machineconfigv1.MachineConfigPoolList{}
+				Expect(e2eclient.Client.List(ctx, allMCPs)).To(Succeed())
+				for _, mcp := range allMCPs.Items {
```
shajmakh: @coderabbitai the `mcp` var is being used in a goroutine, is this version of go allow that safely, or do we need to use a shadow copy like this :
```
mcp:=mcp
```

test/e2e/serial/tests/configuration.go
```diff
@@ -1571,6 +1577,138 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+				})
+			}
+			Expect(eg.Wait()).To(Succeed(), "failed while waiting for mcp update")
+
+			// We not only want to verify that the MCPs are updated, but also that the nodes are updated with the new config.
+			By("Verifying per-node MachineConfig alignment after MCP rollout")
+			for i := range allMCPs.Items {
+				Expect(verifyNodesMatchMCPConfig(ctx, &allMCPs.Items[i])).To(Succeed(),
+					"node config mismatch for MCP %q", allMCPs.Items[i].Name)
+			}
```
shajmakh: this can be defined as a function at the beginning of the test then called when needed, it will make the code cleaner from long duplicates

test/e2e/serial/tests/configuration.go
```diff
@@ -1571,6 +1577,138 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+
+			newTLSProfileSpec := *configv1.TLSProfiles[newProfile.Type]
+			newTLSConfigFn, _ := ctrltls.NewTLSConfigFromProfile(newTLSProfileSpec)
+			newTLSCfg := &tls.Config{}
+			newTLSConfigFn(newTLSCfg)
+			expectedTLSSettings := objtls.NewSettings(newTLSCfg)
+			klog.InfoS("expected TLS settings after update", "tlsSettings", expectedTLSSettings)
+			By("Verifying RTE DaemonSet TLS flags match the updated TLS profile")
+			Eventually(func() error {
+				return intls.CheckRTEDaemonSetTLSFlags(ctx, expectedTLSSettings)
```
shajmakh: remove the returned error and use `g Gomega ` to assert the check succeeds inside the eventually function

test/internal/tls/tls.go
```diff
@@ -191,3 +197,51 @@ func TLSVersionBelow(v uint16) (uint16, error) {
 		return 0, fmt.Errorf("unknown TLS version %d", v)
 	}
 }
+
+func CheckRTEDaemonSetTLSFlags(ctx context.Context, expected objtls.Settings) error {
```
shajmakh: can we add some unit tests with fake client please?

hack/run-test-e2e.sh
```diff
@@ -36,7 +36,7 @@ export E2E_TOPOLOGY_MANAGER_SCOPE="${E2E_TOPOLOGY_MANAGER_SCOPE}"
 # --flake-attempts: rerun the test if it fails
 # -requireSuite: fail if tests are not executed because of missing suite
 echo "TM config policy=[${E2E_TOPOLOGY_MANAGER_POLICY}] scope=[${E2E_TOPOLOGY_MANAGER_SCOPE}]"
-${BIN_DIR}/e2e-nrop-rte.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.flake-attempts=2 --ginkgo.junit-report=${REPORT_DIR}/e2e.xml --ginkgo.skip='\[Disruptive\]|\[StateDirectories\]|\[NodeRefresh\]|\[local\]' --ginkgo.focus='\[release\]'
+${BIN_DIR}/e2e-nrop-rte.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.flake-attempts=2 --ginkgo.junit-report=${REPORT_DIR}/e2e.xml --ginkgo.skip='\[Disruptive\]|\[StateDirectories\]|\[NodeRefresh\]|\[local\]' --ginkgo.focus='\[release\]|\[config\]|\[podfingerprint\]|\[tlscompliance\]'
```
shajmakh: I don't think this should be kept, particularly not in this PR, wdyt?

shajmakh (commented):
Thank you for addressing the comments. 
LGTM overall, I left some comments for further improvement. It is good to have commemts and it is most useful when they are explaining something that is not easy to understand by reading the code, so I believe comments that are not doing that ca be removed. 
Regarding the commits, you are planning to squash them so that we'll have a well-written focused commit message right?

test/e2e/serial/tests/configuration.go
mrniranjan: we can't move this outside of Eventually because on retry it can hold the stale ResourceVersion from the firtAttempt ?

hack/run-test-e2e.sh
mrniranjan: I removed it in latest commit

test/e2e/serial/tests/configuration.go
mrniranjan: Resolved in latest commit

test/e2e/serial/tests/configuration.go
mrniranjan: Fixed in latest commit

test/e2e/serial/tests/configuration.go
mrniranjan: Fixed in latest commit

test/e2e/serial/tests/configuration.go
mrniranjan: fixed in latest commit

test/e2e/serial/tests/configuration.go
mrniranjan: fixed in latest commit

test/internal/tls/tls.go
mrniranjan: @shajmakh fixed in latest commit. Can you review this

test/internal/tls/tls_test.go
```diff
@@ -199,3 +219,223 @@ func TestTLSVersionBelow(t *testing.T) {
...
+				),
+			},
+			settings: objtls.Settings{
+				MinVersion:   "VersionTLS12",
+				CipherSuites: "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
+			},
+			expectedError: "mismatch",
+		},
+	}
+	t.Run("CheckRTEDaemonSetTLSFlags Tests", func(t *testing.T) {
```
shajmakh: This line can be removed, just clarify in the name of the function above that it is for CheckRTEDaemonSetTLSFlags, so it'd be `TestCheckRTEDaemonSetTLSFlags`

test/internal/tls/tls_test.go
```diff
@@ -199,3 +219,223 @@ func TestTLSVersionBelow(t *testing.T) {
...
+func newRTEDaemonSetWithMainContainerArgs(namespace, name string, args ...string) *appsv1.DaemonSet {
+	return newRTEDaemonSet(namespace, name, corev1.Container{
+		Name: rteupdate.MainContainerName,
+		Args: args,
+	})
+}
+
+func TestGetRTEDaemonSet(t *testing.T) {
+	utilruntime.Must(nropv1.AddToScheme(testscheme))
+	utilruntime.Must(appsv1.AddToScheme(testscheme))
```
shajmakh: you can use `k8s.io/client-go/kubernetes/scheme` instead which is already populated with the standard k8s API types from client-go.

test/internal/tls/tls_test.go
```diff
@@ -199,3 +219,223 @@ func TestTLSVersionBelow(t *testing.T) {
...
+}
+
+func newRTEDaemonSetWithMainContainerArgs(namespace, name string, args ...string) *appsv1.DaemonSet {
+	return newRTEDaemonSet(namespace, name, corev1.Container{
+		Name: rteupdate.MainContainerName,
+		Args: args,
+	})
+}
+
+func TestGetRTEDaemonSet(t *testing.T) {
```
shajmakh: let's please rename this to match the function we are testing (CheckRTEDaemonSetTLSFlags)

test/internal/tls/tls_test.go
```diff
@@ -199,3 +219,223 @@ func TestTLSVersionBelow(t *testing.T) {
...
+	utilruntime.Must(nropv1.AddToScheme(testscheme))
+	utilruntime.Must(appsv1.AddToScheme(testscheme))
+
+	ctx := context.Background()
+
+	testCases := []struct {
+		name          string
+		makeNROP      func() *nropv1.NUMAResourcesOperator
```
shajmakh: this is used only one time per test, and it does not reduce duplication, so we can make it an NROP object instead of a function

test/internal/tls/tls_test.go
```diff
@@ -199,3 +219,223 @@ func TestTLSVersionBelow(t *testing.T) {
...
+		daemonSets    []client.Object
+		settings      objtls.Settings
```
shajmakh: can you rename this to expectedTLSSettings please ?

test/internal/tls/tls_test.go
```diff
@@ -199,3 +219,223 @@ func TestTLSVersionBelow(t *testing.T) {
...
+					"--metrics-tls-cipher-suites=TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA256",
+				),
+			},
+			settings: objtls.Settings{
+				MinVersion:   "VersionTLS12",
+				CipherSuites: "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
+			},
+			expectedError: "mismatch",
+		},
+	}
```
shajmakh, line 438: can we add a TLS1.3 scenario (positive and negative) with valid expectations regarding ciphers

test/internal/tls/tls_test.go
mrniranjan: fixed in latest commit

test/internal/tls/tls_test.go
mrniranjan: fixed in latest commit

test/internal/tls/tls_test.go
mrniranjan: fixed in latest commit

test/internal/tls/tls_test.go
mrniranjan: fixed in latest commit

test/internal/tls/tls_test.go
mrniranjan: fixed in latest commit

test/internal/tls/tls_test.go
mrniranjan: fixed in latest commit

shajmakh:
thank you for addressing all comments!
/lgtm
/approve
needs https://github.com/openshift-kni/numaresources-operator/pull/4193 to fix the failing lane

shajmakh:
@niranjan can you please squash the commits and update the commit message?

shajmakh:
rebased to get the govulnercheck fix, no content change
/lgtm
/approve

shajmakh:
@mrniranjan I just noticed the co-authored by Cursor, I was told we shouldn't use that but rather the AIA attribution , which provide more details arount the AI contribution to the code. If it's possible to change that it's would be great.

PR #3612, opened 2026-04-14
e2e: robustness fixes related to retry loops

ffromani:
make sure the update block get+modify+update are all wrapped inot a retry (Eventually) block, to avoid
false negatives caused by kube's optimistic concurrency.

In addition, instead of handling errors, logging them and return bool in the Eventually's closures, let the error be
returned and bubble up in the Eventually block, letting it handle by ginkgo.

### DIFF
--- a/test/e2e/rte/rte_test.go
+++ b/test/e2e/rte/rte_test.go
@@ -19,6 +19,7 @@ package rte
 import (
 	"context"
 	"encoding/json"
+	"errors"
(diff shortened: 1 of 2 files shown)


shajmakh (commented):
Thanks, looks better than before. using the gomega interface within eventually will make it look even cleaner so you don't need to return any value at all for these cases;  this also works though. 
/lgtm

PR #3639, opened 2026-04-15
e2e: serial: dump scheduler log on test failures

ffromani:
When a serial scheduler-related test fail, it is useful to see also the scheduler side.
Add utilities to dump the most recent X seconds of test logs, because we don't have primitives to just find the logs pertaining a pod scheduling attempt.

Use the utilties in the workload_unschedulable tests.

### DIFF
--- a/test/e2e/serial/tests/workload_unschedulable.go
+++ b/test/e2e/serial/tests/workload_unschedulable.go
@@ -25,6 +25,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 	corev1qos "k8s.io/kubectl/pkg/util/qos"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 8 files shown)


test/e2e/serial/tests/workload_unschedulable.go
```diff
@@ -982,6 +999,24 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload unsched
 	})
 })
 
+// educated guess. "far enough" in the past but without dumping too much. Subjected to further review.
+const schedLogWindow = 30 * time.Second
```
shajmakh, line 1003: the idea about dumping the logs is great, the only problem I see is that there is a chance that the logs may not actually be relevant for our debugging purposes. The best would likely be to get the logs based on the pod identity. I'm aware though that needs some level of design so for now we can experiment with how good 30 seconds is, and if we need more we can extend or fall back to must-gather logs.

shajmakh:
/lgtm
Thanks.

test/e2e/serial/tests/workload_unschedulable.go
ffromani: Indeed. It is possible we only add noise because we shoot in the dark. To be really effective we would need the old logtracer concept implemented, or at least some sever-side log filtering.

ffromani:
/hold

further refinement incoming

ffromani:
/hold cancel

last update for this PR (except review comments), thanks for your patience!

shajmakh:
Thanks for enabling additional option. LGTM.
/lgtm

PR #3601, opened 2026-04-14
GH actions: run CI on all supported main branches

ffromani:
Keep the CI running on all official supported branches to avoid merging content with less checks.
Given that the repo is public and the runners used in the workflows (`ubuntu-latest`) are standard github-hosted runners, we can freely use them in the actions of this repo. ref:
https://docs.github.com/en/billing/concepts/product-billing/github-actions#free-use-of-github-actions

Replaces #2874

PR #3642, opened 2026-04-16
e2e: test: internal: make sure to retry on conflicts

ffromani:
all the utilities in the testing call tree (and for kube clients in general) should wrap get+modify+update on a retry loop to handle optmistic concurrency conflicts.

Make sure to do so for all the code called by install_test. We use RetryOnConflict because utility code should not call Eventually directly - and yes, there is plenty of dirty code in the codebase which we should fix eventually (no pun intended).

AI-attribution: AIA PAI SeCe Hin R claude-4.6-opus-1M v1.0

### DIFF
--- a/test/internal/nodepools/nodepools.go
+++ b/test/internal/nodepools/nodepools.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/util/retry"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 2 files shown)


shajmakh:
/lgtm
Good modification. was there any test failing on not having this improvement?

ffromani:
We're seeing flakes lately like https://github.com/openshift-kni/numaresources-operator/pull/3601#issuecomment-4244672106 which prompted this investigation and the (AI-assisted) fix. Can't claim this is a conclusive fix, but should help.

PR #3687, opened 2026-04-20
wait: drop DaemonSets vs nodeGroups check from NRO wait helper

SargunNarula:
`ForDaemonsetInNUMAResourcesOperatorStatus` now only waits until `status.daemonSets` is non-empty.
Cross-field consistency belongs in e2e.

rte: assert daemonSets matches nodeGroups only when `status.nodeGroups`
is non-empty (backward-compatible status when it is empty).

Assisted-by: Claude

### DIFF
--- a/internal/wait/numaresources.go
+++ b/internal/wait/numaresources.go
@@ -18,13 +18,11 @@ package wait
 
 import (
 	"context"
-	"reflect"
 
 	k8swait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 
 	nropv1 "github.com/openshift-kni/numaresources-operator/api/v1"
-	testobjs "github.com/openshift-kni/numaresources-operator/internal/objects"
 )
 
 func (wt Waiter) ForNUMAResourcesOperatorDeleted(ctx context.Context, nrop *nropv1.NUMAResourcesOperator) error {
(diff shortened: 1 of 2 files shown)


internal/wait/numaresources.go
```diff
@@ -66,8 +66,10 @@ func (wt Waiter) ForDaemonsetInNUMAResourcesOperatorStatus(ctx context.Context,
 		}
 
 		dssFromNodeGroupStatus := testobjs.GetDaemonSetListFromNodeGroupStatuses(updatedNRO.Status.NodeGroups)
-		if !reflect.DeepEqual(updatedNRO.Status.DaemonSets, dssFromNodeGroupStatus) {
-			klog.Warningf("daemonset list mismatch: from NodeGroupStatus'es:\n%+v\n from operator status:\n%+v\n", dssFromNodeGroupStatus, updatedNRO.Status.NodeGroups)
+		// NodeGroups may be unset on clusters/operators that only populate the legacy DaemonSets
+		// field; in that case there is nothing to reconcile against.
+		if len(updatedNRO.Status.NodeGroups) > 0 && !reflect.DeepEqual(updatedNRO.Status.DaemonSets, dssFromNodeGroupStatus) {
```
shajmakh: let's return early if this is met and keep the old version past line 68:
```
if len(updatedNRO.Status.NodeGroups) > 0{
return true, nil
}
```

internal/wait/numaresources.go
```diff
@@ -66,8 +66,10 @@ func (wt Waiter) ForDaemonsetInNUMAResourcesOperatorStatus(ctx context.Context,
...
+			klog.Warningf("daemonset list mismatch: from NodeGroupStatus'es:\n%+v\n from operator status.DaemonSets:\n%+v\n", dssFromNodeGroupStatus, updatedNRO.Status.DaemonSets)
 			return false, nil
 		}
```
ffromani: we can just remove this self-consistency check. It doesn't really belong into a wait helper. This consistency check is already performed in e2e tests (rte_test.go and serial/tests/configuration.go). We should rather further enhance the tests there, if needed.

SargunNarula: @ffromani Thanks for guiding, Removed the check in latest commit.

internal/wait/numaresources.go
SargunNarula: Updated with removal of consistency check as decided over slack.

ffromani (commented):
/approve
/lgtm

thanks!

PR #3752, opened 2026-04-23
cleanup: remove unused `IsCustomPolicyEnabled` variable

Tal-or:
During the work of fixing https://redhat.atlassian.net/browse/OCPBUGS-84226 it was found that `IsCustomPolicyEnabled` variable never used (only the function did).

Deleted it on separate PR because we want to minimize backport changes.

### DIFF
--- a/pkg/objectstate/rte/machineconfigpool.go
+++ b/pkg/objectstate/rte/machineconfigpool.go
@@ -92,13 +92,12 @@ func (obj machineConfigPoolFinder) FindState(mf Manifests, tree nodegroupv1.Tree
 		}
 
 		gdm := GeneratedDesiredManifest{
-			ClusterPlatform:       obj.em.plat,
-			MachineConfigPool:     mcp.DeepCopy(),
-			NodeGroup:             tree.NodeGroup.DeepCopy(),
-			DaemonSet:             desiredDaemonSet,
-			RTEConfigHash:         rteConfigHash,
-			IsCustomPolicyEnabled: annotations.IsCustomPolicyEnabled(tree.NodeGroup.Annotations),
-			SecOpts:               mf.securityContextOptions(annotations.IsCustomPolicyEnabled(tree.NodeGroup.Annotations)),
+			ClusterPlatform:   obj.em.plat,
+			MachineConfigPool: mcp.DeepCopy(),
+			NodeGroup:         tree.NodeGroup.DeepCopy(),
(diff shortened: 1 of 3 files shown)


ffromani (commented):
The original intent was to do the reverse: compute the value once, store and use the same result for the full reconcile step.

I'm not terribly attached to this semantic, but we should be consistent. IIRC all the other annotations are checked once and stored, is that the case? If so we should not add an exception to the general rule.

Tal-or:
>The original intent was to do the reverse: compute the value once, store and use the same result for the full reconcile step.

As a general rule that's correct, but in the context of where the value was stored we were not reading it nor using the function, so it's completely redundant.

ffromani:
yep, but the main point is following the general pattern. What we usually do? compute once and use the stored value or, like we do in this PR, always compute the value when needed? Let's have a consistent codebase whatever approach we take. IOW, let's not introduce an exception for this flow unless there is a crystal clear and strong reason to (is there?)

Tal-or:
> yep, but the main point is following the general pattern. What we usually do? compute once and use the stored value or, like we do in this PR, always compute the value when needed? Let's have a consistent codebase whatever approach we take. IOW, let's not introduce an exception for this flow unless there is a crystal clear and strong reason to (is there?)

I'm fully agree with you on that. 
But again the change here is not replacing a stored value with one which is being compute every single time. 

I'm merely deleting the stored value because it's not being used at all.

ffromani:
yep, but the main point is following the general pattern. What we usually do? compute once and use the stored value or, like we do in this PR, always compute the value when needed? Let's have a consistent codebase whatever approach we take. IOW, let's not introduce an exception for this flow unless there is a crystal clear and strong reason to (is there?)

> > yep, but the main point is following the general pattern. What we usually do? compute once and use the stored value or, like we do in this PR, always compute the value when needed? Let's have a consistent codebase whatever approach we take. IOW, let's not introduce an exception for this flow unless there is a crystal clear and strong reason to (is there?)
> 
> I'm fully agree with you on that. But again the change here is not replacing a stored value with one which is being compute every single time.
> 
> I'm merely deleting the stored value because it's not being used at all.

fair enough. Then I suggest to make a sweeping PR to make the codebase consistent. This would have lower priority.

ffromani:
/lgtm

I stand corrected, this variable is never used and unnecessary. We will re-evaluate possible optimization (don't compute the value from annotations every time, store and reuse) later on.

PR #3801, opened 2026-04-26
OCPBUGS-84226: objectstate/rte: per-pool MachineConfig state with paused MCP awareness

Tal-or:
Each MachineConfigObjectState carries its own WaitForUpdated function,
pool name, and paused flag. The controller builds a per-pool wait map
and skips paused MCPs during sync.

Co-Authored-By: Francesco Romani <fromani@redhat.com>
Co-Authored-By: Shereen Haj <shajmakh@redhat.com>

### DIFF
--- a/internal/controller/numaresourcesoperator_controller.go
+++ b/internal/controller/numaresourcesoperator_controller.go
@@ -237,7 +237,7 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResourceAPI(ctx context.Conte
 func (r *NUMAResourcesOperatorReconciler) reconcileResourceMachineConfig(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, trees []nodegroupv1.Tree) intreconcile.Step {
 	// we need to sync machine configs first and wait for the MachineConfigPool updates
 	// before checking additional components for updates
-	mcpUpdatedFunc, err := r.syncMachineConfigs(ctx, instance, existing, trees)
+	waitByPool, err := r.syncMachineConfigs(ctx, instance, existing, trees)
 	if err != nil {
 		r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedMCSync", "Failed to set up machine configuration for worker nodes: %v", err)
 		err = fmt.Errorf("failed to sync machine configs: %w", err)
@@ -247,7 +247,7 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResourceMachineConfig(ctx con
 
 	// MCO needs to update the SELinux context removal and other stuff, and need to trigger a reboot.
 	// It can take a while.
-	mcpStatuses, mcpNamePending := syncMachineConfigPoolsStatuses(instance.Name, trees, r.ForwardMCPConds, mcpUpdatedFunc)
+	mcpStatuses, mcpNamePending := syncMachineConfigPoolsStatuses(instance.Name, trees, r.ForwardMCPConds, waitByPool)
 	instance.Status.MachineConfigPools = mcpStatuses
 
 	if mcpNamePending != "" {
@@ -391,7 +391,7 @@ func (r *NUMAResourcesOperatorReconciler) syncNodeResourceTopologyAPI(ctx contex
 	return (updatedCount == len(objStates)), err
 }
 
-func (r *NUMAResourcesOperatorReconciler) syncMachineConfigs(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, trees []nodegroupv1.Tree) (rtestate.MCPWaitForUpdatedFunc, error) {
+func (r *NUMAResourcesOperatorReconciler) syncMachineConfigs(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, trees []nodegroupv1.Tree) (map[string]rtestate.MCPWaitForUpdatedFunc, error) {
 	klog.V(4).InfoS("Machine Config Sync start", "trees", len(trees))
 	defer klog.V(4).Info("Machine Config Sync stop")
 
@@ -401,8 +401,12 @@ func (r *NUMAResourcesOperatorReconciler) syncMachineConfigs(ctx context.Context
 	// In case of operator upgrade from 4.1X → 4.18, it's necessary to remove the old MachineConfig,
 	// unless an emergency annotation is provided which forces the operator to use custom policy
 
-	objStates, waitFunc := existing.MachineConfigsState(r.RTEManifests)
-	for _, objState := range objStates {
+	mcObjStates := existing.MachineConfigsState(r.RTEManifests)
+	waitByPool := make(map[string]rtestate.MCPWaitForUpdatedFunc, len(mcObjStates))
+	for _, mcObjState := range mcObjStates {
+		waitByPool[mcObjState.PoolName] = mcObjState.WaitForUpdated
+
+		objState := mcObjState.ObjectState
 		klog.InfoS("objState", "desired", objState.Desired, "existing", objState.Existing, "createOrUpdate", objState.IsCreateOrUpdate())
 		if objState.IsCreateOrUpdate() {
 			if err2 := controllerutil.SetControllerReference(instance, objState.Desired, r.Scheme); err2 != nil {
@@ -421,19 +425,24 @@ func (r *NUMAResourcesOperatorReconciler) syncMachineConfigs(ctx context.Context
 			break
 		}
 	}
-	return waitFunc, err
+	return waitByPool, err
 }
 
-func syncMachineConfigPoolsStatuses(instanceName string, trees []nodegroupv1.Tree, forwardMCPConds bool, updatedFunc rtestate.MCPWaitForUpdatedFunc) ([]nropv1.MachineConfigPool, string) {
-	klog.V(4).InfoS("Machine Config Status Sync start", "trees", len(trees))
-	defer klog.V(4).Info("Machine Config Status Sync stop")
+func syncMachineConfigPoolsStatuses(instanceName string, trees []nodegroupv1.Tree, forwardMCPConds bool, waitByPool map[string]rtestate.MCPWaitForUpdatedFunc) ([]nropv1.MachineConfigPool, string) {
+	klog.V(4).InfoS("Machine Config Pools Status Sync start", "trees", len(trees))
+	defer klog.V(4).Info("Machine Config Pools Status Sync stop")
 
 	mcpStatuses := []nropv1.MachineConfigPool{}
 	for _, tree := range trees {
 		for _, mcp := range tree.MachineConfigPools {
 			mcpStatuses = append(mcpStatuses, extractMCPStatus(mcp, forwardMCPConds))
 
-			isUpdated := updatedFunc(instanceName, mcp)
+			waitFunc, ok := waitByPool[mcp.Name]
+			if !ok {
+				continue
+			}
+
+			isUpdated := waitFunc(instanceName, mcp)
 			klog.V(5).InfoS("Machine Config Pool state", "name", mcp.Name, "instance", instanceName, "updated", isUpdated)
 
 			if !isUpdated {
@@ -654,27 +663,33 @@ func (r *NUMAResourcesOperatorReconciler) mcpToNUMAResourceOperator(ctx context.
 		nro := &nros.Items[i]
 		mcpLabels := labels.Set(mcp.Labels)
 		for _, nodeGroup := range nro.Spec.NodeGroups {
-			if nodeGroup.MachineConfigPoolSelector == nil {
-				continue
-			}
-
-			nodeGroupSelector, err := metav1.LabelSelectorAsSelector(nodeGroup.MachineConfigPoolSelector)
-			if err != nil {
-				klog.Errorf("failed to parse the selector %v", mcp.Spec.NodeSelector)
-				return nil
-			}
-
-			if nodeGroupSelector.Matches(mcpLabels) {
+			if nodeGroupMatchesMCP(nodeGroup, mcp.Name, mcpLabels) {
 				requests = append(requests, reconcile.Request{
 					NamespacedName: client.ObjectKey{
 						Name: nro.Name,
 					},
 				})
+				break
 			}
 		}
 	}
 
 	return requests
+}
+
+func nodeGroupMatchesMCP(nodeGroup nropv1.NodeGroup, mcpName string, mcpLabels labels.Set) bool {
+	if nodeGroup.PoolName != nil {
+		return mcpName == *nodeGroup.PoolName
+	}
+	if nodeGroup.MachineConfigPoolSelector == nil {
+		return false
+	}
+	selector, err := metav1.LabelSelectorAsSelector(nodeGroup.MachineConfigPoolSelector)
+	if err != nil {
+		klog.Errorf("failed to parse the selector %v", nodeGroup.MachineConfigPoolSelector)
+		return false
+	}
+	return selector.Matches(mcpLabels)
 }
 
 func (r *NUMAResourcesOperatorReconciler) configMapToNUMAResourceOperator(ctx context.Context, cmObj client.Object) []reconcile.Request {
--- a/internal/controller/numaresourcesoperator_controller.go
+++ b/internal/controller/numaresourcesoperator_controller.go
@@ -31,6 +31,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 
@@ -237,7 +238,7 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResourceAPI(ctx context.Conte
 func (r *NUMAResourcesOperatorReconciler) reconcileResourceMachineConfig(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, trees []nodegroupv1.Tree) intreconcile.Step {
 	// we need to sync machine configs first and wait for the MachineConfigPool updates
 	// before checking additional components for updates
-	waitByPool, err := r.syncMachineConfigs(ctx, instance, existing, trees)
+	waitByPool, _, err := r.syncMachineConfigs(ctx, instance, existing, trees)
 	if err != nil {
 		r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedMCSync", "Failed to set up machine configuration for worker nodes: %v", err)
 		err = fmt.Errorf("failed to sync machine configs: %w", err)
@@ -294,7 +295,9 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
 	existing := rtestate.FromClient(ctx, r.Client, r.Platform, r.RTEManifests, instance, trees, r.Namespace)
 
 	if r.Platform == platform.OpenShift {
-		if step := r.reconcileResourceMachineConfig(ctx, instance, existing, trees); step.EarlyStop() {
+		var step intreconcile.Step
+		step = r.reconcileResourceMachineConfig(ctx, instance, existing, trees)
+		if step.EarlyStop() {
 			return step
 		}
 	}
@@ -391,7 +394,7 @@ func (r *NUMAResourcesOperatorReconciler) syncNodeResourceTopologyAPI(ctx contex
 	return (updatedCount == len(objStates)), err
 }
 
-func (r *NUMAResourcesOperatorReconciler) syncMachineConfigs(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, trees []nodegroupv1.Tree) (map[string]rtestate.MCPWaitForUpdatedFunc, error) {
+func (r *NUMAResourcesOperatorReconciler) syncMachineConfigs(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, trees []nodegroupv1.Tree) (map[string]rtestate.MCPWaitForUpdatedFunc, sets.Set[string], error) {
 	klog.V(4).InfoS("Machine Config Sync start", "trees", len(trees))
 	defer klog.V(4).Info("Machine Config Sync stop")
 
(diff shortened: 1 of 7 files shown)


pkg/objectstate/rte/machineconfigpool.go
```diff
@@ -134,8 +135,10 @@ func (em *ExistingManifests) MachineConfigsState(mf Manifests) ([]objectstate.Ob
 	if mf.Core.MachineConfig == nil {
 		return ret, nullMachineConfigPoolUpdated
 	}
-	enabledMCCount := 0
+	waitForMachineConfigPresenceByPoolName := make(map[string]bool)
```
ffromani: so this is basically a set of MCPs we need to wait for, isn't it?

pkg/objectstate/rte/machineconfigpool.go
```diff
@@ -161,6 +164,8 @@ func (em *ExistingManifests) MachineConfigsState(mf Manifests) ([]objectstate.Ob
 						Desired:  nil,
 					},
 				)
+				// wait until our machine config is no longer in the pool (see mcpWaitDispatchByPool)
+				waitForMachineConfigPresenceByPoolName[mcp.Name] = false
```
ffromani: What's the difference between adding a MCP to the map with `false` value and just not adding it to the map?

pkg/objectstate/rte/machineconfigpool.go
```diff
@@ -178,15 +183,54 @@ func (em *ExistingManifests) MachineConfigsState(mf Manifests) ([]objectstate.Ob
...
+	return ret, mcpWaitDispatchByPool(waitForMachineConfigPresenceByPoolName)
+}
+
+func formatMap(m map[string]bool) string {
+	flatPairs := make([]string, 0, len(m))
+	for k, v := range m {
+		flatPairs = append(flatPairs, fmt.Sprintf("[mcpName=%s, waitForMachineConfigPresence=%t]", k, v))
+	}
+	return strings.Join(flatPairs, ", ")
+}
```
ffromani: if we can use a set, this function devolves into a `strings.Join` we can do inline in the call site.

pkg/objectstate/rte/machineconfigpool.go
```diff
@@ -178,15 +183,54 @@ func (em *ExistingManifests) MachineConfigsState(mf Manifests) ([]objectstate.Ob
...
+	}
+	return mcp.Spec.Paused
+}
+
+// mcpWaitDispatchByPool returns a single wait function that, per MachineConfig pool name, waits either for
+// our machine config to be present (and pool updated) or for it to be removed from the pool, depending on
+// how that pool was handled in MachineConfigsState. A paused spec.paused pool is treated as ready (do not
+// block the reconcile loop) because the MCO will not roll in that case.
+// Pools not in the map were skipped in MachineConfigsState and are treated as ready.
+func mcpWaitDispatchByPool(waitForPresenceByMCPName map[string]bool) MCPWaitForUpdatedFunc {
```
ffromani: this works but using a closure is bordering too smart. Have we explored adding the wait (or changing how we wait) in the `ObjectState`?

internal/controller/numaresourcesoperator_controller_test.go
```diff
@@ -2281,8 +2281,10 @@ func checkSELinuxPolicyProcessing(ctx context.Context, nro *nropv1.NUMAResources
 	ensureMCPIsReady(mcp1, nro.Name)
 	Expect(reconciler.Client.Update(ctx, mcp1)).To(Succeed())
 
+	// Node group 2 has no custom SELinux policy: wait logic expects our MachineConfig to be *absent* from
+	// the pool, so do not put it in Configuration.Source (unlike the pool that has custom policy).
 	Expect(reconciler.Client.Get(ctx, client.ObjectKeyFromObject(mcp2), mcp2)).To(Succeed())
-	ensureMCPIsReady(mcp2, nro.Name)
+	ensureMCPIsReadyAfterMCDeletion(mcp2)
```
ffromani, line 2393: do we have coverage of the old behavior (uniform application?)
adding test coverage is ofc fine, let's make sure we do it in a purely additive way.
IOW, we need flows on which multiple MCPs have consistent config. and flows on which we have mixed configs.

pkg/objectstate/rte/machineconfigpool_test.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+	mf := Manifests{Core: rtemanifests.Manifests{MachineConfig: baseMachineConfigTemplate()}}
+	st, wait := em.MachineConfigsState(mf)
+	if len(st) != 1 {
+		t.Fatalf("paused mcp: expected one object state (object state is independent of spec.paused), got %d", len(st))
+	}
+	if st[0].Desired == nil {
+		t.Fatalf("with custom policy, expected non-nil desired MachineConfig")
+	}
+	// MCO is paused: wait should not keep reconciling in the "wait for pool" sense (see mcpWaitDispatchByPool).
```
ffromani: this comment is confusing. What are we expecting to do here?

pkg/objectstate/rte/machineconfigpool_test.go
```diff
@@ -0,0 +1,333 @@
+/*
+ * Copyright 2026 Red Hat, Inc.
```
ffromani: I'm not sure of the value of these tests, need to review them carefully. Do we have a e2e or at least a controller test which maps to the user flow we are fixing?

ffromani (commented):
initial review

pkg/objectstate/rte/machineconfigpool.go
Tal-or: Not exactly. the string contains the mcp name, the boolean value suggest whether for the given key (MCP) we expect a MachineConfig to be present (the MachineConfig that apply the custom SELinux policy)

pkg/objectstate/rte/machineconfigpool.go
Tal-or: We expect all MCPs to be present even the paused ones.

ffromani (commented):
Thanks for the update. Please do not use Co-Authored-By claude, use the AI-Attribution tag: https://aiattribution.github.io/create-attribution

internal/controller/numaresourcesoperator_controller_test.go
Tal-or: added.

test/e2e/serial/tests/configuration.go
```diff
@@ -254,6 +255,83 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			Expect(fxt.Client.Get(ctx, nroKey, &initialNroOperObj)).To(Succeed())
+
+			testMCP := objects.TestMCP()
+			e2efixture.By("creating new MCP: %q with zero machine count", testMCP.Name)
+			testMCP.Labels = map[string]string{"machineconfiguration.openshift.io/role": roleMCPTest}
+			testMCP.Spec.MachineConfigSelector = &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"machineconfiguration.openshift.io/role": depnodes.RoleWorker,
+				},
+			}
```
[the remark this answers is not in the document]
Tal-or: I'm not sure I'm following the explanation here

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -395,7 +391,7 @@ func (r *NUMAResourcesOperatorReconciler) syncNodeResourceTopologyAPI(ctx contex
 	return (updatedCount == len(objStates)), err
 }
 
-func (r *NUMAResourcesOperatorReconciler) syncMachineConfigs(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, trees []nodegroupv1.Tree) (rtestate.MCPWaitForUpdatedFunc, error) {
+func (r *NUMAResourcesOperatorReconciler) syncMachineConfigs(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, trees []nodegroupv1.Tree) (map[string]rtestate.MCPWaitForUpdatedFunc, []string, error) {
```
ffromani: we have 2 problems here:
1. the commit message summarizes what we do, adds no context on the why and why it's useful
2. we coalesce in 1 commit the per-pool wait and the pause tracking. Should be 2 commits, especailly since we just discard the paused MCP data in the current commit.

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -658,6 +670,18 @@ func (r *NUMAResourcesOperatorReconciler) mcpToNUMAResourceOperator(ctx context.
...
+				if mcp.Name == *nodeGroup.PoolName {
+					requests = append(requests, reconcile.Request{
+						NamespacedName: client.ObjectKey{
+							Name: nro.Name,
+						},
+					})
+					break
+				}
+				continue
+			}
```
ffromani: this is hard to read and to verify. Please move in its own helper function to streamline the flow. Something like
```
func nodeGroupMatchesMCP(nodeGroup nropv1.NodeGroup, mcpName string, mcpLabels labels.Set) bool {
       if nodeGroup.PoolName != nil {
               return mcpName == *nodeGroup.PoolName
       }
       if nodeGroup.MachineConfigPoolSelector == nil {
               return false
       }
       selector, err := metav1.LabelSelectorAsSelector(nodeGroup.MachineConfigPoolSelector)
       if err != nil {
               klog.Errorf("failed to parse the selector %v", nodeGroup.MachineConfigPoolSelector)
               return false
       }
       return selector.Matches(mcpLabels)
}
```
(please verify this code)

pkg/objectstate/rte/machineconfigpool.go
```diff
@@ -129,14 +129,29 @@ func MatchMachineConfigPoolCondition(conditions []machineconfigv1.MachineConfigP
...
+
+func (em *ExistingManifests) MachineConfigsState(mf Manifests) []MachineConfigObjectState {
+	var ret []MachineConfigObjectState
 	if mf.Core.MachineConfig == nil {
-		return ret, nullMachineConfigPoolUpdated
+		return ret
 	}
-	enabledMCCount := 0
 	for _, tree := range em.trees {
+		isCustomPolicy := annotations.IsCustomPolicyEnabled(tree.NodeGroup.Annotations)
```
ffromani, line 147: We should use computed data. No bother now, we can and will clean up later

pkg/objectstate/rte/machineconfigpool.go
```diff
@@ -129,14 +129,29 @@ func MatchMachineConfigPoolCondition(conditions []machineconfigv1.MachineConfigP
...
 		for _, mcp := range tree.MachineConfigPools {
+			if mcp.Spec.Paused {
+				ret = append(ret, MachineConfigObjectState{
```
ffromani: this is inconsistent though. We don't actually report the full object state for paused MCPs?

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -241,7 +237,7 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResourceAPI(ctx context.Conte
 func (r *NUMAResourcesOperatorReconciler) reconcileResourceMachineConfig(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, trees []nodegroupv1.Tree) intreconcile.Step {
 	// we need to sync machine configs first and wait for the MachineConfigPool updates
 	// before checking additional components for updates
-	mcpUpdatedFunc, err := r.syncMachineConfigs(ctx, instance, existing, trees)
+	waitByPool, _, err := r.syncMachineConfigs(ctx, instance, existing, trees)
```
ffromani: we're light on tests. Unit tests and controller tests which ensure the correctness of this change should be in this commit, not in followups. A e2e test (or tests) can be added in a later commit, but unit and integration tests totally are feasible and give us signal. E.g. 1 controller test with two pools, one with the custom sepolicy annotation, one vanilla. The reconciliation is expected to converge.

pkg/objectstate/rte/machineconfigpool.go
```diff
@@ -129,14 +129,29 @@ func MatchMachineConfigPoolCondition(conditions []machineconfigv1.MachineConfigP
 
 type MCPWaitForUpdatedFunc func(string, *machineconfigv1.MachineConfigPool) bool
 
-func (em *ExistingManifests) MachineConfigsState(mf Manifests) ([]objectstate.ObjectState, MCPWaitForUpdatedFunc) {
-	var ret []objectstate.ObjectState
+type MachineConfigObjectState struct {
+	objectstate.ObjectState
+	PoolName       string
+	Paused         bool
```
ffromani, line 136: Tracking paused MCPs is something we totally should do, but what's the value in embedding the flag in the wrapper type?

A paused MCP is a MCP we should ignore, so conceptually what we need is a set of strings (names).
This is actually reflecting on the current impl: we only track the name and the flag for paused MCPs, which is suggesting this modelization is questionable.

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
```
ffromani, line 23: like in the previous commit, the core logic seems sound and the need for these changes is real, but we should add unit and controller tests alongside the business logic changes.

internal/controller/numaresourcesoperator_controller_test.go
```diff
@@ -2140,15 +2140,96 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
...
 
 				mc2Key := client.ObjectKey{
 					Name: objectnames.GetMachineConfigName(nro.Name, mcp2.Name),
 				}
-				err = reconciler.Client.Get(context.TODO(), mc2Key, mc)
+				err = reconciler.Client.Get(ctx, mc2Key, mc)
 				Expect(apierrors.IsNotFound(err)).To(BeTrue(), "MachineConfig %s expected to be deleted; err=%v", mc2Key.Name, err)
 			})
+
+			It("should converge with mixed policy: one pool custom, one pool default", func(ctx context.Context) {
```
ffromani, line 2153: ok, so we have the required controller tests. Let's make sure to squash it in the same commit which introduces the business logic change

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -260,6 +260,8 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResourceMachineConfig(ctx con
 	}
 	instance.Status.MachineConfigPools = syncMachineConfigPoolNodeGroupConfigStatuses(instance.Status.MachineConfigPools, trees)
 
+	instance.Status.Conditions = updateMachineConfigPoolPausedCondition(instance.Status.Conditions, instance.Generation, pausedMCPNames)
+
```
ffromani: move this on line 256 above, right before the `if mcpNamePending != ""` block, so we correctly report the condition even if the operator status is Progressing; with the current code if the status is Progressing this condition is not reported, it is when Available (and Degraded IIRC?)

ffromani (commented):
The content is here and I'm convinced the changes are good. We probably just need to reorganize the commits

internal/controller/numaresourcesoperator_controller_test.go
```diff
@@ -2459,6 +2480,19 @@ func ensureMCPIsReady(mcp *machineconfigv1.MachineConfigPool, nroName string) {
 	}
 }
```
ffromani, line 2506: cheap tests to add:
```
diff --git a/internal/controller/numaresourcesoperator_controller_test.go b/internal/controller/numaresourcesoperator_controller_test.go
index ac10c136b..58e8d8993 100644
--- a/internal/controller/numaresourcesoperator_controller_test.go
+++ b/internal/controller/numaresourcesoperator_controller_test.go
@@ -2242,9 +2242,13 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
                                }
                                Expect(reconciler.Client.Get(ctx, mcp2DSKey, ds)).To(Succeed())
 
-                               By("Verify NRO status is Available")
+                               By("Verify NRO status is Available with no paused condition")
                                Expect(reconciler.Client.Get(ctx, client.ObjectKeyFromObject(nro), nro)).To(Succeed())
                                Expect(nro).To(BeInCondition(status.ConditionAvailable))
+
+                               pausedCond := getConditionByType(nro.Status.Conditions, status.ConditionMachineConfigPoolPaused)
+                               Expect(pausedCond).ToNot(BeNil())
+                               Expect(pausedCond.Status).To(Equal(metav1.ConditionFalse))
                        })
 
                        It("should converge when one pool is paused", func(ctx context.Context) {
@@ -2282,6 +2286,28 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
                                Expect(pausedCond).ToNot(BeNil())
                                Expect(pausedCond.Status).To(Equal(metav1.ConditionTrue))
                        })
+
+                       It("should report paused condition while Progressing", func(ctx context.Context) {
+                               // pool1 has custom policy -> MC created, needs MCO rollout
+                               // pool2 is paused -> skipped
+                               nro.Spec.NodeGroups[0].Annotations[annotations.SELinuxPolicyConfigAnnotation] = annotations.SELinuxPolicyCustom
+                               mcp2.Spec.Paused = true
+
+                               var err error
+                               reconciler, err = NewFakeNUMAResourcesOperatorReconciler(platform.OpenShift, defaultOCPVersion, nro, mcp1, mcp2)
+                               Expect(err).ToNot(HaveOccurred())
+
+                               key := client.ObjectKeyFromObject(nro)
+
+                               By("First reconcile: MC created for pool1, pool2 skipped (paused), waiting for MCO on pool1")
+                               Expect(reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: key})).To(CauseRequeue())
+
+                               By("Verify paused condition is set even while Progressing")
+                               Expect(reconciler.Client.Get(ctx, client.ObjectKeyFromObject(nro), nro)).To(Succeed())
+                               pausedCond := getConditionByType(nro.Status.Conditions, status.ConditionMachineConfigPoolPaused)
+                               Expect(pausedCond).ToNot(BeNil(), "paused condition should be set during Progressing phase")
+                               Expect(pausedCond.Status).To(Equal(metav1.ConditionTrue))
+                       })

```

pkg/status/status.go
```diff
@@ -183,6 +189,36 @@ func DefaultBaseConditions(timestamp time.Time) []metav1.Condition {
 	}
 }
 
+// NewNUMAResourcesOperatorConditions creates specific operator conditions on
+// top of DefaultBaseConditions.
+func NewNUMAResourcesOperatorConditions() []metav1.Condition {
```
ffromani, line 194: cheap test to add:
```
diff --git a/pkg/status/status_test.go b/pkg/status/status_test.go
index 6264aee11..85b4b4b17 100644
--- a/pkg/status/status_test.go
+++ b/pkg/status/status_test.go
@@ -282,6 +282,38 @@ func TestMessageFromError(t *testing.T) {
        }
 }
 
+func TestNewNUMAResourcesOperatorConditions(t *testing.T) {
+       conds := NewNUMAResourcesOperatorConditions()
+
+       baseTypes := []string{ConditionAvailable, ConditionUpgradeable, ConditionProgressing, ConditionDegraded}
+       for _, ct := range baseTypes {
+               found := false
+               for _, c := range conds {
+                       if c.Type == ct {
+                               found = true
+                               break
+                       }
+               }
+               if !found {
+                       t.Fatalf("missing base condition %q", ct)
+               }
+       }
+
+       var pausedCond *metav1.Condition
+       for i := range conds {
+               if conds[i].Type == ConditionMachineConfigPoolPaused {
+                       pausedCond = &conds[i]
+                       break
+               }
+       }
+       if pausedCond == nil {
+               t.Fatal("missing MachineConfigPoolPaused condition")
+       }
+       if pausedCond.Status != metav1.ConditionUnknown {
+               t.Fatalf("expected MachineConfigPoolPaused status %q, got %q", metav1.ConditionUnknown, pausedCond.Status)
+       }
+}
+
```

ffromani (commented):
I need a final pass to validate the tests, and that should be it.

pkg/objectstate/rte/machineconfigpool.go
Tal-or: We don't, because we're skipping applying their state anyway. 
once MCP  unpauses, it triggers the controller reconciliation loop which in turn reports the MC/MCP full state.

pkg/objectstate/rte/machineconfigpool.go
Tal-or: This is mostly to keep the API concise.

On higher level above the stack we're extracting this data and returning it.

pkg/objectstate/rte/machineconfigpool.go
Tal-or: ack

pkg/status/status.go
Tal-or: Added

internal/controller/numaresourcesoperator_controller_test.go
```diff
@@ -2221,9 +2225,35 @@ var _ = Describe("Test NUMAResourcesOperator Reconcile", func() {
 				}
 				Expect(reconciler.Client.Get(ctx, mcp2DSKey, ds)).To(Succeed())
 
-				By("Verify NRO status is Available")
+				By("Verify NRO status is Available with paused condition")
```
Tal-or, line 2228: Tests were added in this commit since this is the first commit where all the references can be found.

internal/controller/numaresourcesoperator_controller.go
Tal-or: Updated according to the comment. commits were splits as well.

internal/controller/numaresourcesoperator_controller.go
Tal-or: refactored. thanks.

pkg/objectstate/rte/machineconfigpool.go
```diff
@@ -132,6 +132,7 @@ type MCPWaitForUpdatedFunc func(string, *machineconfigv1.MachineConfigPool) bool
 type MachineConfigObjectState struct {
 	objectstate.ObjectState
 	PoolName       string
+	Paused         bool
```
ffromani, line 136: I still don't agree about this field here. We are conflating too many states in the same object. But I can turn a blind eye if this is the only missing piece. We can fix later.

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -293,11 +293,15 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
...
+	var pausedMCPNames []string
 	if r.Platform == platform.OpenShift {
-		if step := r.reconcileResourceMachineConfig(ctx, instance, existing, trees); step.EarlyStop() {
+		var step intreconcile.Step
+		pausedMCPNames, step = r.reconcileResourceMachineConfig(ctx, instance, existing, trees)
+		if step.EarlyStop() {
 			return step
 		}
 	}
+	_ = pausedMCPNames
```
ffromani: I don't get the intent here. If we don't need `pausedMCPNames`, and it seems we don't, let's just discard the return value.

ffromani (commented):
still need to fully process `machineconfigpool_test.go`. Nothing blocking so far.

ffromani:
LGTM the machineconfigpool_test.go can use some more polishing, but we can do that later.

please ping me when you want to merge, I'll add the proper label.

internal/controller/numaresourcesoperator_controller.go
Tal-or: old code. it was removed in later PR

pkg/objectstate/rte/machineconfigpool.go
Tal-or: since this is going to be backport the whole way down I can extract and return this value separately

ffromani: your call. It is non-blocking IMO. But yes, I'd like to have it cleaner.

Tal-or: Updated the code to return the paused MCPs in a separate set

ffromani:
/lgtm

good work here, thanks!

PR #3802, opened 2026-04-26
NO-JIRA: makefile: add version to golangci-lint make target

Tal-or:
current version printing was missing, add it.


ffromani (commented):
/approve
/lgtm

PR #3803, opened 2026-04-27
NO-JIRA:Dockerfile: disable RHEL repos when installing hwdata

Tal-or:
The UBI9 minimal base image ships with both RHEL subscription repos (rhel-9-for-x86_64-baseos-rpms, rhel-9-for-x86_64-appstream-rpms) and freely accessible UBI repos. When building outside of a subscribed RHEL host, microdnf fails with a 403 because the RHEL CDN requires valid entitlements.

The hwdata package is available in the UBI repos, so we can safely disable the rhel-* repos and still install it without a subscription. The subscription-aware Dockerfile.openshift, used for official builds in entitled environments, remains unchanged.


ffromani (commented):
/approve
/lgtm

why not?

PR #3830, opened 2026-04-28
add feature introspection topic for supported versions

rshemtov13:
Register the introspection feature in the active topics list so that the skip-list mechanism can gate the introspection tests on operator versions that predate the feature (< 4.21). So we need to backport this until 4.21 where it was first introduced.

### DIFF
--- a/internal/api/features/_topics.json
+++ b/internal/api/features/_topics.json
@@ -22,6 +22,7 @@
     "metrics",
     "nrtcrdanns",
     "netpols",
+    "introspection",
     "mgselinuxcollect",
     "tlscompliance"
   ]


ffromani:
Can't we get the same result with the exiting tools? if you get an empty topic list, then introspection is not supported

rshemtov13:
@ffromani Yes we can do it with existing tools, I can add the condition you suggested, although by convention we run the serial suite by excluding features using ginkgo filtering, and we only need to backport this to 4.21 so I thought it's simpler since no other change is necessary. Let me know what you think is best.

shajmakh:
not sure I understand why this is needed. do we have new tests for "introspection" feature of NROP?

rshemtov13:
@shajmakh No new tests indeed, just tests available from 4.21 and above

shajmakh:
/lgtm
/hold 
in case @ffromani has any comments

ffromani:
I'm still unsold to the core idea of this PR.

NROP supports introspection if it returns any topic. If it doesn't no introspection support. I'm struggling to see the benefit of adding this meta-topic.

Furthermore: for few years to come, there always be a old enough NROP version which don't have topics/introspection support, so how would we handle that case?

rshemtov13:
@ffromani Thanks for the feedback. Besides the current approach in this PR, we could also use a simple version gate (run only on 4.21+). Do you see any other option we should consider?

ffromani:
OK, I was wrong: the label `feature:introspection` sounded like it was about the topics/introspection system itself — which is what led me to think "this is a meta-topic, you can just check if topics exist." But it's actually a bucket label for a set of 4.21 features   (terminationMessagePolicy on operator and operand) that happen to be grouped under "introspection." In hindsight, bad label name.

rshemtov13:
manual backport for ```release-4-22``` branch #4069

PR #3849, opened 2026-04-30
[manual] [release-4.18] objectstate/rte: per-pool MachineConfig state with paused MCP awareness 

Tal-or:
Manual cherry-pick of https://github.com/openshift-kni/numaresources-operator/pull/3801

### DIFF
--- a/controllers/numaresourcesoperator_controller.go
+++ b/controllers/numaresourcesoperator_controller.go
@@ -31,6 +31,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 
@@ -247,7 +248,7 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResourceAPI(ctx context.Conte
 func (r *NUMAResourcesOperatorReconciler) reconcileResourceMachineConfig(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, trees []nodegroupv1.Tree) intreconcile.Step {
 	// we need to sync machine configs first and wait for the MachineConfigPool updates
 	// before checking additional components for updates
-	waitByPool, err := r.syncMachineConfigs(ctx, instance, existing, trees)
+	waitByPool, _, err := r.syncMachineConfigs(ctx, instance, existing, trees)
 	if err != nil {
 		r.Recorder.Eventf(instance, corev1.EventTypeWarning, "FailedMCSync", "Failed to set up machine configuration for worker nodes: %v", err)
 		err = fmt.Errorf("failed to sync machine configs: %w", err)
@@ -305,7 +306,8 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
 	existing := rtestate.FromClient(ctx, r.Client, r.Platform, r.RTEManifests, instance, trees, r.Namespace)
 
 	if r.Platform == platform.OpenShift {
-		if step := r.reconcileResourceMachineConfig(ctx, instance, existing, trees); step.EarlyStop() {
+		step := r.reconcileResourceMachineConfig(ctx, instance, existing, trees)
+		if step.EarlyStop() {
 			updateStatusConditionsIfNeeded(instance, step.ConditionInfo)
 			return step
 		}
@@ -408,7 +410,7 @@ func (r *NUMAResourcesOperatorReconciler) syncNodeResourceTopologyAPI(ctx contex
 	return (updatedCount == len(objStates)), err
 }
 
-func (r *NUMAResourcesOperatorReconciler) syncMachineConfigs(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, trees []nodegroupv1.Tree) (map[string]rtestate.MCPWaitForUpdatedFunc, error) {
+func (r *NUMAResourcesOperatorReconciler) syncMachineConfigs(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, trees []nodegroupv1.Tree) (map[string]rtestate.MCPWaitForUpdatedFunc, sets.Set[string], error) {
 	klog.V(4).InfoS("Machine Config Sync start", "trees", len(trees))
 	defer klog.V(4).Info("Machine Config Sync stop")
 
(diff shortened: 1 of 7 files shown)


Tal-or:
/hold

need to be tested manually and reviewed carefully

Tal-or:
@coderabbitai review

ffromani:
we need to be super-careful to not add fixes which are not also available on main branch. We should have main branch commit verbatim, commit(s) to adapt to the 4.18 branch and nothing else. Whatever improvements coderabbit finds, it's matter for `main` branch only.

Tal-or:
> we need to be super-careful to not add fixes which are not also available on main branch. We should have main branch commit verbatim, commit(s) to adapt to the 4.18 branch and nothing else. Whatever improvements coderabbit finds, it's matter for `main` branch only.

Sure. just wanted to be extra cautions to see we're not missing something super critical that might broke the code

Tal-or:
/hold cancel

Tested locally and behaves as expected.

test/e2e/serial/tests/configuration.go
```diff
@@ -355,6 +358,90 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
 
 		})
 
+		It("should converge with mixed SELinux policy across node groups", Label("tier2"), func(ctx context.Context) {
```
ffromani, line 361: there's no need to backport the e2e test because we use the most recent e2e suite against older branch. But since we have it we can keep it to make the backport trivial (or as trivial as possible)

test/e2e/serial/tests/configuration.go
```diff
@@ -355,6 +358,90 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
+			}(context.Background())
+
+			By("waiting for operator to converge with mixed policy")
+			updatedOperObj := &nropv1.NUMAResourcesOperator{}
+			Eventually(func(g Gomega) {
+				g.Expect(fxt.Client.Get(ctx, nroKey, updatedOperObj)).To(Succeed())
+				cond := status.FindCondition(updatedOperObj.Status.Conditions, status.ConditionAvailable)
+				g.Expect(cond).ToNot(BeNil(), "Available condition not found")
+				g.Expect(cond.Status).To(Equal(metav1.ConditionTrue), "operator not yet available")
+				g.Expect(cond.ObservedGeneration).To(Equal(updatedOperObj.Generation), "condition not yet updated for current generation")
```
ffromani, line 436: looks correct. For posterity: a careful backport of `isNROOperSyncedAt` would have been a viable option.

controllers/numaresourcesoperator_controller.go
```diff
@@ -29,6 +29,7 @@ import (
 	apiextensionv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
```
ffromani, line 32: attaching here because there's no hunk about `machineconfigpool_test.go`. Did you squash this fix in the original commit?

controllers/numaresourcesoperator_controller.go
```diff
@@ -212,6 +213,12 @@ func updateStatusConditionsIfNeeded(instance *nropv1.NUMAResourcesOperator, cond
 	klog.InfoS("updateStatus", "condition", cond.Type, "reason", cond.Reason, "message", cond.Message)
 	conditions, ok := status.UpdateConditions(instance.Status.Conditions, cond.Type, cond.Reason, cond.Message)
 	if ok {
+		// UpdateConditions replaces with base conditions only; preserve extras
```
ffromani, line 216: why we don't need this in `main` again?

controllers/numaresourcesoperator_controller.go
```diff
@@ -844,7 +851,7 @@ func updateMachineConfigPoolPausedCondition(conditions []metav1.Condition, gener
 		ObservedGeneration: generation,
 		LastTransitionTime: metav1.Now(),
 	}
-	conditions, _ = status.ComputeConditions(conditions, condition, time.Now())
+	apimeta.SetStatusCondition(&conditions, condition)
```
ffromani, line 854: if we bother to declare the fixes we need (and that's a good thing), let's pinpoint the reason why we need a fixup. Is this because 489a3175bc09b79e8d358b65ad086b17f33d515b ?

ffromani (commented):
looks reasonnable. Two main improvements besides inline comments.
1. please prefer `git cherry-pick -x` so the tooling add a reference about the main commit being backported, for easier backtracking
2. let's pick a side consistently: either fix each commit, but then add in the commit message what changed specifically to enable the backport, or cherrypick commit verbatim, accept each commit can break the build, and add a cumulative fixup commit so once the PR merges the build is clean again. Either work, we don't have a consolidated standard (yet). If you pick the second approach which seems already the direction taken, please use `cherry-pick -x`

Tal-or:
>let's pick a side consistently: either fix each commit, but then add in the commit message what changed specifically to enable the backport, or cherrypick commit verbatim, accept each commit can break the build, and add a cumulative fixup commit so once the PR merges the build is clean again. Either work, we don't have a consolidated standard (yet). If you pick the second approach which seems already the direction taken, please use cherry-pick -x

Coping the commit verbatim causes merge conflicts, so we'll end up with the first 4 commits containing unreadable annotations indicate about conflicts (`<<<<<<<<<<<<<<<<HEAD` and such)

you must have a minor degree of changes for ,at least, making the code syntax correctly 

EDIT: there are even files that were move into to different locations:
for example:

`internal/controller/numaresourcesoperator_controller.go` which doesn't exist on 4.18, the file is at `controllers/numaresourcesoperator_controller.go`

controllers/numaresourcesoperator_controller.go
Tal-or: in main we have `ComputeConditions` https://github.com/openshift-kni/numaresources-operator/blob/main/pkg/status/status.go#L98 which handle the extra conditions, while here it replaces and add only the base conditions.

so we need that for reporting the `MachineConfigPoolPaused` condition.

controllers/numaresourcesoperator_controller.go
Tal-or: Not sure I follow. the `machineconfigpool_test.go` was added in the previous commit. same as in main: 
https://github.com/openshift-kni/numaresources-operator/pull/3801/changes/bec6f8fbe87758c4c4ecda949c06d4462514d1cd

Tal-or:
The PR was open on feature branch from https://github.com/openshift-kni/numaresources-operator repo (and not from my own fork).

All branches here are protected from `push --force` so we must recreate this PR on my fork https://github.com/openshift-kni/numaresources-operator/pull/3910 and continue the conversation there.

I've already address the comments raised during this PR review.

controllers/numaresourcesoperator_controller.go
Tal-or: Done. see: https://github.com/openshift-kni/numaresources-operator/pull/3910/changes/45d43065f8fcf5332ac33886212faf25988d2e61#diff-a015b81c292215d1f79361f0a0297c9dcd6167b4a2aeaf248dcfbad60cdda000R854

test/e2e/serial/tests/configuration.go
Tal-or: done. see: https://github.com/openshift-kni/numaresources-operator/pull/3910/changes/03451dc3cc5253207f63f941107963e7c46b9893#diff-99dd14d7f4fade18212bf91dbe6165d7626335bf407a28732eb7d591d2252fd8R433

PR #3895, opened 2026-05-04
CNF-23039: add stable channel to bundle and catalog

yanirq:


ffromani (commented):
/approve
/lgtm

ok, let's fold back in the standard

PR #3956, opened 2026-05-08
must-gather: modified must-gather runner to enable cli arguments

SargunNarula:
Updated must-gather runner to enable cli arguments, to align its behavior 
with the serial runner so that, CI injected filters can be provided and applied.

### DIFF
--- a/hack/run-test-must-gather-e2e.sh
+++ b/hack/run-test-must-gather-e2e.sh
@@ -10,8 +10,85 @@ if ! which tput &> /dev/null 2>&1 || [[ $(tput -T$TERM colors) -lt 8 ]]; then
   NO_COLOR="-ginkgo.no-color"
 fi
 
+FOCUS="-ginkgo.focus='\\[must-gather\\]'"
+SKIP=""
+LABEL_FILTER=""
+VERBOSE=""
+REPORT_FILE="${REPORT_DIR}/must-gather.xml"
+
+# Keep argument handling compatible with run-e2e-nrop-serial.sh so
+# Jenkins-generated scripts can reuse the same options for must-gather.
+while [[ $# -gt 0 ]]; do
+	case "$1" in
+		--no-color)
+			NO_COLOR="-ginkgo.no-color"
+			shift
+			;;
+		--dry-run)
+			DRY_RUN="true"
+			shift
+			;;
+		--focus)
+			FOCUS="-ginkgo.focus='$2'"
+			shift
+			shift
+			;;
+		--skip)
+			SKIP="-ginkgo.skip='$2'"
+			shift
+			shift
+			;;
+		--label-filter)
+			LABEL_FILTER="-ginkgo.label-filter='$2'"
+			shift
+			shift
+			;;
+		--report-file)
+			REPORT_FILE="$2"
+			shift
+			shift
+			;;
+		--verbose)
+			VERBOSE="${VERBOSE} -ginkgo.v"
+			shift
+			;;
+		--debug)
+			VERBOSE="${VERBOSE} -ginkgo.vv -test.v"
+			shift
+			;;
+		--help)
+			echo "usage: $0 [options]"
+			echo "[options] are always '--opt val' and never '--opt=val'"
+			echo "available options:"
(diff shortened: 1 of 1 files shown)


hack/run-test-must-gather-e2e.sh
```diff
@@ -10,8 +10,85 @@ if ! which tput &> /dev/null 2>&1 || [[ $(tput -T$TERM colors) -lt 8 ]]; then
...
+done
+
 setupreport
 
 # Run must-gather test suite
 echo "Running NRO must-gather test suite"
-${BIN_DIR}/e2e-nrop-must-gather.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.junit-report=${REPORT_DIR}/must-gather.xml --ginkgo.focus='\[must-gather\]'
+runcmd ${BIN_DIR}/e2e-nrop-must-gather.test \
+	--ginkgo.v \
+	--ginkgo.timeout=5h \
```
rshemtov13: This timeout seems a bit too high for just running the must-gather tests. Let's lower it to a more reasonable value possibly max 1hr (and it's also pretty long)?
Also if it gets stuck for some reason the serial suite will get stack as well.

rshemtov13:
@SargunNarula Thanks for this PR, This fixes our downstream CI for the ```--label-filter``` option, which was missing for the must-gather tests since we didn't need it until now.

Apart from that small nit, lgtm

hack/run-test-must-gather-e2e.sh
SargunNarula: Updated the timeout to 20m, which is enough IMO, in the latest commit. Thanks.

hack/run-test-must-gather-e2e.sh
```diff
@@ -10,8 +10,85 @@ if ! which tput &> /dev/null 2>&1 || [[ $(tput -T$TERM colors) -lt 8 ]]; then
...
+done
+
 setupreport
 
 # Run must-gather test suite
 echo "Running NRO must-gather test suite"
-${BIN_DIR}/e2e-nrop-must-gather.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.junit-report=${REPORT_DIR}/must-gather.xml --ginkgo.focus='\[must-gather\]'
+runcmd ${BIN_DIR}/e2e-nrop-must-gather.test \
+	--ginkgo.v \
+	--ginkgo.timeout=20m \
```
ffromani, line 88: is this change down to 5h intentional?

hack/run-test-must-gather-e2e.sh
```diff
@@ -10,8 +10,85 @@ if ! which tput &> /dev/null 2>&1 || [[ $(tput -T$TERM colors) -lt 8 ]]; then
...
+			;;
+		--skip)
+			SKIP="-ginkgo.skip='$2'"
+			shift
+			shift
+			;;
+		--label-filter)
+			LABEL_FILTER="-ginkgo.label-filter='$2'"
+			shift
+			shift
```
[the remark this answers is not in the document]
SargunNarula: I believe this fix should be a part of refactor exercise among other runners as well and with sync to CI changes.
wdyt @ffromani @shajmakh  ?

ffromani (commented):
LGTM but some changes seems to have sneaked in? let's make sure these are intentional

hack/run-test-must-gather-e2e.sh
SargunNarula: I lower the timeout value to 20m, keeping in mind the usual MG test runtime. 5h was mistakenly set and got updated now.

hack/run-test-must-gather-e2e.sh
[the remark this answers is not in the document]
ffromani: yes, we should fix later alongside the other runners

hack/run-test-must-gather-e2e.sh
ffromani: ok, I guess we can start with this value

ffromani (commented):
/approve
/lgtm

PR #3989, opened 2026-05-11
mcp: selinux: extend annotation to stop reconciling

ffromani:
Add another value to the custom selinux policy annotation: if this value is set, the operand (RTE DaemonSet) will use the builtin policy and SCC v2, but the relevant MachineConfig will *stop being reconciled*.

Operators can remove the annotation anytime to trigger the reconciliation, causing a MC deletion and therefore a reboot.

This way, we decouple the SELinux upgrade from the MachineConfig removal, which helps (or should help) in the upgrade flow.


ffromani:
/cc @Tal-or 
/hold

tiny change, but not sure yet we want this

PR #3992, opened 2026-05-11
controllers: requeue on status update failure

ffromani:
make sure to not lose status updates: if we fail
to update the status subresource, we need to requeue the operation.

### DIFF
--- a/internal/controller/numaresourcesoperator_controller.go
+++ b/internal/controller/numaresourcesoperator_controller.go
@@ -192,6 +192,7 @@ func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctr
 
 	if err := r.updateStatus(ctx, initialStatus, instance); err != nil {
 		klog.InfoS("Failed to update numaresources-operator status", "error", err)
+		return ctrl.Result{RequeueAfter: numaResourcesRetryPeriod}, nil
 	}
 
 	return step.Result, step.Error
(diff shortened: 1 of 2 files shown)


internal/controller/numaresourcesoperator_controller.go
```diff
@@ -192,6 +192,7 @@ func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctr
 
 	if err := r.updateStatus(ctx, initialStatus, instance); err != nil {
 		klog.InfoS("Failed to update numaresources-operator status", "error", err)
+		return ctrl.Result{RequeueAfter: numaResourcesRetryPeriod}, nil
 	}
```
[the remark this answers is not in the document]
ffromani: check controller-runtime API.I can't both set RequeueAfter and return an error. Can I?

shajmakh:
/lgtm
I doubt the update status would fail because it is only a single controller that updates the resources so there will rarely (or never) be any conflict.

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -192,6 +192,7 @@ func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctr
 
 	if err := r.updateStatus(ctx, initialStatus, instance); err != nil {
 		klog.InfoS("Failed to update numaresources-operator status", "error", err)
+		return ctrl.Result{RequeueAfter: numaResourcesRetryPeriod}, nil
```
shajmakh, line 195: @coderabbitai at which cases may the status update fail during the reconciliation? any way to reduce that?

ffromani: switching to Patch is interesting

ffromani:
/hold

let's explore the patch approach

internal/controller/numaresourcesoperator_controller.go
shajmakh: @coderabbitai what ways can we use to distinguish if the status update is stuck in a loop of requeueing requests?

ffromani:
replaced by #4083

ffromani:
/close

#4083 merged

PR #3995, opened 2026-05-12
CNF-23818: machineconfigpools: avoid RTE pods duplication when single node matching multiple MCPs

Tal-or:
**Problem Statement**
When a node's labels match multiple `MachineConfigPools` (e.g. a custom pool that extends the worker role), the current `DaemonSet`'s `NodeSelector` — which copies the MCP's own nodeSelector — causes multiple RTE `DaemonSets` to target the same node, resulting in duplicate pods.

**Root Cause**
The MCP nodeSelectors are inherently overlapping: for example, a node with both `node-role.kubernetes.io/worker` and `node-role.kubernetes.io/workerB` labels matches the worker MCP and the workerB MCP simultaneously. Using those selectors directly as `DaemonSet`'s `NodeSelectors` guarantees 
collisions.

**Suggested Approach**
Place a single, NRO-owned label on each node that encodes its primary pool. The `DaemonSet`'s `NodeSelector` then targets this label instead of the MCP's nodeSelector, making placement 1:1 by construction.

We reuse MCO's own priority logic (master > custom > worker) to determine each node's primary pool, so the behavior is consistent with how MCO itself decides under what MCP the node belongs.

NOTE: 
This is not relevant for HyperShift since on HyperShift we're setting the `DaemonSet`'s `NodeSelector` as the `NodePool` name which are exclusive for each node, meaning a node can't have more than a single `NodePool` label. 

AIA Human-AI blend, New content, Human-initiated, Reviewed, Claude Opus 4.6 v1.0

### DIFF
--- a/internal/api/labels/labels.go
+++ b/internal/api/labels/labels.go
@@ -1,0 +1,29 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package labels
+
+const (
+	// NodePrimaryPool is set on each node managed by NRO to indicate
+	// its primary pool (MCP on OpenShift, node-pool on HyperShift).
+	// Using this exclusive label as DaemonSet NodeSelector prevents duplicate
+	// pods when a node's labels match multiple pools.
+	NodePrimaryPool = "numa-operator.openshift.io/primary-pool"
+
+	// NodeFinalizer is set on the NUMAResourcesOperator CR to ensure
+	// node labels are cleaned up when the CR is deleted.
+	NodeFinalizer = "numa-operator.openshift.io/node-labels"
+)
--- a/internal/controller/numaresourcesoperator_controller.go
+++ b/internal/controller/numaresourcesoperator_controller.go
@@ -57,7 +57,9 @@ import (
 	"github.com/openshift-kni/numaresources-operator/api/v1/helper/namespacedname"
 	nodegroupv1 "github.com/openshift-kni/numaresources-operator/api/v1/helper/nodegroup"
 	"github.com/openshift-kni/numaresources-operator/internal/api/annotations"
+	nrolabels "github.com/openshift-kni/numaresources-operator/internal/api/labels"
 	"github.com/openshift-kni/numaresources-operator/internal/dangling"
+	"github.com/openshift-kni/numaresources-operator/internal/nodes"
 	intreconcile "github.com/openshift-kni/numaresources-operator/internal/reconcile"
 	"github.com/openshift-kni/numaresources-operator/internal/relatedobjects"
 	"github.com/openshift-kni/numaresources-operator/pkg/apply"
@@ -119,8 +121,8 @@ type NUMAResourcesOperatorReconciler struct {
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings,verbs=*
 //+kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=*
 //+kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch
-//+kubebuilder:rbac:groups="",resources=nodes,verbs=list
-//+kubebuilder:rbac:groups=nodetopology.openshift.io,resources=numaresourcesoperators,verbs=get;list;watch
+//+kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch;patch
+//+kubebuilder:rbac:groups=nodetopology.openshift.io,resources=numaresourcesoperators,verbs=get;list;watch;update
 //+kubebuilder:rbac:groups=nodetopology.openshift.io,resources=numaresourcesoperators/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=nodetopology.openshift.io,resources=numaresourcesoperators/finalizers,verbs=update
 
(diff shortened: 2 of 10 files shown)


pkg/machineconfigpools/pools.go
```diff
@@ -0,0 +1,160 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/klog/v2"
+
+	mcov1 "github.com/openshift/api/machineconfiguration/v1"
+)
+
+const (
+	MachineConfigPoolMaster = "master"
```
ffromani, line 31: `master` or `control-plane`? or both?

internal/controller/numaresourcesoperator_controller_test.go
```diff
@@ -74,7 +74,31 @@ const (
...
+				Labels: mcp.Spec.NodeSelector.MatchLabels,
+			},
+		}
+		initObjects = append(initObjects, node)
+	}
+	return initObjects
+}
+
 func NewFakeNUMAResourcesOperatorReconciler(plat platform.Platform, platVersion platform.Version, initObjects ...runtime.Object) (*NUMAResourcesOperatorReconciler, error) {
+	initObjects = appendFakeNodesForMCPs(initObjects)
```
ffromani: so all the tests all the time get fake extra nodes which we filter?
I can see the reason, but this seems a bit coarse-grained

ffromani (commented):
skimmed over it, it seems a good idea worth looking into

internal/controller/numaresourcesoperator_controller_test.go
Tal-or: super cheap since this is all a mock and on local memory only, but i'll try to look into alternatives.

ffromani: the problem here is not the expensiveness of the test, but the fact that we lose the code path which exercises the expected clean state, aka the happy path. This is the main/only concern here.

Tal-or: IIUC you're concerned about the need to change the tests themselves to make the CI happy?  

We actually made the requirements stricter - no more dummy MCPs without node association. 
This is also closer to real life/production scenario

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -834,17 +835,58 @@ func updateMachineConfigPoolPausedCondition(conditions []metav1.Condition, gener
...
+				klog.V(4).InfoS("Error checking nodes for pool, keeping pool", "pool", mcp.Name, "error", err)
+				filteredMCPs = append(filteredMCPs, mcp)
+				continue
+			}
+			if has {
+				filteredMCPs = append(filteredMCPs, mcp)
+			} else {
+				klog.V(2).InfoS("Skipping MCP with no primary nodes", "pool", mcp.Name)
+			}
+		}
```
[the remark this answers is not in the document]
Tal-or: then the filtering is done for all trees together while we need the MCP per tree

Tal-or:
/hold

This requires more thinking, the existing approach is not good enough

pkg/machineconfigpools/pools.go
Tal-or: Here, we are referring to the MCPs coming out-of-the-box after OCP cluster installation, so it's `master` and `worker`. `control-plane` is irrelevant in this specific case.

Tal-or:
CI E2E - for tools depends on #4311

Tal-or:
/hold cancel

PR is reviewable again

Tal-or:
@coderabbitai review

pkg/machineconfigpools/pools.go
```diff
@@ -0,0 +1,167 @@
...
+	MachineConfigPoolWorker = "worker"
+)
+
+// ListPools categorizes the given pools into master, worker, and custom pools
+// based on which pools' node selectors match the given node's labels.
+// Adapted from MCO: https://github.com/openshift/machine-config-operator/blob/99cb8a46e6a31b2b72d6a8371c6cd4ee45393263/pkg/helpers/helpers.go#L155
+func ListPools(node *corev1.Node, pools []mcov1.MachineConfigPool) (master *mcov1.MachineConfigPool, worker *mcov1.MachineConfigPool, custom []*mcov1.MachineConfigPool, err error) {
```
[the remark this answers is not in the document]
Tal-or: We don't care about Windows use case. all OCP nodes are RHCOS which is a Linux distro

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -629,12 +655,29 @@ func (r *NUMAResourcesOperatorReconciler) SetupWithManager(mgr ctrl.Manager) err
...
+	nodePredicates := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if !validateUpdateEvent(&e) {
+				return false
+			}
+			_, oldHas := e.ObjectOld.GetLabels()[nrolabels.NodePrimaryPool]
+			_, newHas := e.ObjectNew.GetLabels()[nrolabels.NodePrimaryPool]
+			if oldHas || newHas {
+				return true
+			}
```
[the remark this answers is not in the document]
Tal-or: so more corrected approach is adding the `!apiequality.Semantic.DeepEqual(e.ObjectOld.GetLabels(), e.ObjectNew.GetLabels())`
in the `if oldHas || newHas`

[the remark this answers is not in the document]
Tal-or: but the predicate was originally introduce by this PR, so we're are the one setting the scope.

Tal-or:
@ffromani PTAL

internal/api/labels/labels.go
```diff
@@ -0,0 +1,29 @@
...
+const (
+	// NodePrimaryPool is set on each node managed by NRO to indicate
+	// its primary pool (MCP on OpenShift, node-pool on HyperShift).
+	// Using this exclusive label as DaemonSet NodeSelector prevents duplicate
+	// pods when a node's labels match multiple pools.
+	NodePrimaryPool = "numa-operator.openshift.io/primary-pool"
+
+	// NodeFinalizer is set on the NUMAResourcesOperator CR to ensure
+	// node labels are cleaned up when the CR is deleted.
+	NodeFinalizer = "numa-operator.openshift.io/node-labels"
```
ffromani, line 28: we need to review which domain we should use. Let's look for other examples in the codebase. Surely the domain should not be `openshift.io` but `openshift-kni.io`

ffromani (commented):
*before* the RBAC narrow down, this would have been easier to merge. Now we need a stronger justification to add more RBAC permissions.

ffromani:
Can we fix at RTE level? (I can take care) The main requisite is to avoid adding more RBAC permissions

Tal-or (commented):
> _before_ the RBAC narrow down, this would have been easier to merge. Now we need a stronger justification to add more RBAC permissions.

I understand that we should minimize RBAC permissions but where do we cross the line between what's needed and what's redundant? This change is for a bug fix (although minor) after all.

> Can we fix at RTE level? (I can take care) The main requisite is to avoid adding more RBAC permissions

This is the daemonset that populate the extra pod according to the label. The daemonset configuration is managed by NROP.

ffromani:
> > Can we fix at RTE level? (I can take care) The main requisite is to avoid adding more RBAC permissions
> 
> This is the daemonset that populate the extra pod according to the label. The daemonset configuration is managed by NROP.

Indeed. But the alternative would be to detect duplication from RTEs and halt the duplicated copy. https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/394 tries to fix a similar problem. Would the issue be fixable at RTE level?

Tal-or:
> > > Can we fix at RTE level? (I can take care) The main requisite is to avoid adding more RBAC permissions
> > 
> > 
> > This is the daemonset that populate the extra pod according to the label. The daemonset configuration is managed by NROP.
> 
> Indeed. But the alternative would be to detect duplication from RTEs and halt the duplicated copy. [k8stopologyawareschedwg/resource-topology-exporter#394](https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/394) tries to fix a similar problem. Would the issue be fixable at RTE level?

It's possible, but RTE level fix introduce this whole concept of node lease which is a bit more complex than a simple additional label at the node.
More over there's still more than a single RTE instance (where's only one is active) while the current PR won't allow more than a single instance, which is more accurate and clean from both UX and API perspective IMO.

PR #3998, opened 2026-05-12
CNF-23239: controller: nrop: per-tree processing and progress

ffromani:
implement "vertical split" vs current "horizontal split"
the split is about how the reconciliation makes progress. In the model we had till now, retroactively called "horizontal split", each NodeGroup must make progress on each reconciliation stage before any of these can progress to the next stage. IOW, the slowest NodeGroup blocks all the other NodeGroups at the stage it is stuck at.

With the approach proposed here, "vertical split", each NodeGroup advances independently on the reconciliation step, and only the *global reported status* is set pulled to the slowest state; IOW, with N nodegroups, N-1 can be Available and settled and 1 still Progressing; the global state will still be Progressing, but changes to the other NodeGroups can happen independently.

To simplify the implementation, a single goroutine handles all the NodeGroup reconciliation, so factually a NodeGroup can still influence the others, but this (bar more serious bugs we should fix anyway) can only create reconciliation delays, a far better stance than the issues caused by the horizontal split.

## backportability

This PR is broken down in commits which are intentionally shaped to make the backport down to the distant 4.18 as low risk as possible.
Compatibility layers are added and then removed in the context of the same PR explicitly and only in order
to enable gradual testing, not because their own merit.

## review notice

Adding and removing code in the same PR makes the review harder, even though the final tree state is identical.
It may be useful to review the final tree state (e.g. git diff <base>..HEAD) in addition to individual commits, since the end result is identical to a cleaner, main-only series.
The most independent commits are intentionally pushed towards the beginning of the series to enable this review mode as much as possible.

## LLM-assisted backport effort summary

| Branch | Effort | Notes |
|--------|--------|-------|
| 4.22 | minimal | Essentially clean cherry-picks |
| 4.21 | minimal | Essentially clean cherry-picks (after PR #4025 lands) |
| 4.20 | very low | Minor controller conflicts (after PR #3801 backport lands) |
| 4.19 | low | Minor conflicts; `rte.go` diverges more (after PR #3801 backport lands) |
| 4.18 | high | Path renames, no `MachineConfigsState`, `machineconfigpool.go` heavily diverged (see below) |

### 4.18 effort breakdown

| Commits | Effort | Nature |
|---------|--------|--------|
| 1 (Name() helper) | very low | Path adjustment: `api/v1/helper/` → `api/numaresourcesoperator/v1/helper/` |
| 2-3 (conditioninfo, step helpers) | minimal | `step.go` identical; `conditioninfo.go` nearly identical |
| 4 (move helper around) | low | Path adjustment: `internal/controller/` → `controllers/` |
| 5 (rte + compat wrappers) | hard/very hard | Key refactoring; `machineconfigpool.go` is 134 lines (vs 244 on main), no `MachineConfigObjectState` type, no `MachineConfigsState` method; `rte.go` is 436 lines (vs 379) with 288 diff lines |
| 6 (signature refactoring) | low | Path adjustment only |
| 7 (add per-tree functions) | mid/hard | New code, but depends on compat wrappers from commit 5 being adapted first |
| 8 (wire per-tree functions) | hard | Controller rewrite; building blocks from 5-7 must be proven first |
| 9-10 (narrow helpers, rename) | very low | Mechanical signature changes |
| 11 (remove compat wrappers) | mid | Adaptation for different `machineconfigpool.go` baseline |
| 12 (vertical processing) | mid | Controller change; same pattern as main but different baseline |

### 4.21/4.20/4.19 prerequisite

These branches require PR #3801 ("OCPBUGS-84226: per-pool MachineConfig state
with paused MCP awareness") to be backported first. PR #4025 covers 4.21;
4.20 and 4.19 need equivalent backports. Once that prerequisite lands,
these branches have the same `MachineConfigObjectState` type and
`MachineConfigsState` signature as main, making the cherry-picks
straightforward.

## LLM-assisted risk assessment

The proposed split should significantly reduce backport risk compared to
a simpler, monolithic approach:

1. **Commits 1-3** (Name helper, UpdateMessage, step helpers) are trivial
   cherry-picks on all branches. Done and verified quickly.
2. **Commit 4** (move helper) is a pure reorder, low risk.
3. **Commit 5** (rte compat wrappers) is the key risk point — it refactors
   the objectstate layer but adds compat wrappers so the controller continues
   working unchanged. Existing controller tests validate the compat wrappers
   without any controller changes.
4. **Commit 6** (signature refactoring) is mechanical prep work.
5. **Commit 7** (add per-tree functions) adds dead code. If it compiles,
   it is correct.
6. **Commit 8** (wire per-tree functions) is the core behavioral change.
   By this point all building blocks (1-7) are proven and verified.
7. **Commits 9-10** (narrow helpers, rename) are mechanical signature
   narrowing, safe after commit 8 removed the old callers.
8. **Commit 11** (remove compat wrappers) is mechanical cleanup.
9. **Commit 12** (vertical processing) is the final behavioral change,
   switching from horizontal to vertical orchestration.

The compat wrappers in commit 5 are the key risk reduction: they allow
validating the rte.go refactoring on any branch (including 4.18, the
hardest target) without touching the controller at all. Once that is
proven, the controller changes build on a solid foundation.

### DIFF
--- a/internal/controller/numaresourcesoperator_controller.go
+++ b/internal/controller/numaresourcesoperator_controller.go
@@ -377,7 +377,7 @@ func (r *NUMAResourcesOperatorReconciler) syncMachineConfig(ctx context.Context,
 	// In case of operator upgrade from 4.1X → 4.18, it's necessary to remove the old MachineConfig,
 	// unless an emergency annotation is provided which forces the operator to use custom policy
 
-	mcObjStates, pausedMCPNames := existing.MachineConfigsStateForTree(r.RTEManifests, tree)
+	mcObjStates, pausedMCPNames := existing.MachineConfigsState(r.RTEManifests, tree)
 	waitByPool := make(map[string]rtestate.MCPWaitForUpdatedFunc, len(mcObjStates))
 	for _, mcObjState := range mcObjStates {
 		waitByPool[mcObjState.PoolName] = mcObjState.WaitForUpdated
--- a/internal/controller/numaresourcesoperator_controller.go
+++ b/internal/controller/numaresourcesoperator_controller.go
@@ -403,56 +403,52 @@ func (r *NUMAResourcesOperatorReconciler) syncMachineConfigs(ctx context.Context
 	return waitByPool, pausedMCPNames, err
 }
 
-func syncMachineConfigPoolsStatuses(instanceName string, forwardMCPConds bool, waitByPool map[string]rtestate.MCPWaitForUpdatedFunc, trees ...nodegroupv1.Tree) ([]nropv1.MachineConfigPool, string) {
-	klog.V(4).InfoS("Machine Config Pools Status Sync start", "trees", len(trees))
+func syncMachineConfigPoolsStatuses(instanceName string, forwardMCPConds bool, waitByPool map[string]rtestate.MCPWaitForUpdatedFunc, tree nodegroupv1.Tree) ([]nropv1.MachineConfigPool, string) {
+	klog.V(4).InfoS("Machine Config Pools Status Sync start", "tree", tree.Name())
 	defer klog.V(4).Info("Machine Config Pools Status Sync stop")
 
 	mcpStatuses := []nropv1.MachineConfigPool{}
-	for _, tree := range trees {
-		for _, mcp := range tree.MachineConfigPools {
-			mcpStatuses = append(mcpStatuses, extractMCPStatus(mcp, forwardMCPConds))
+	for _, mcp := range tree.MachineConfigPools {
+		mcpStatuses = append(mcpStatuses, extractMCPStatus(mcp, forwardMCPConds))
 
-			waitFunc, ok := waitByPool[mcp.Name]
-			if !ok {
-				continue
-			}
+		waitFunc, ok := waitByPool[mcp.Name]
+		if !ok {
+			continue
+		}
 
-			isUpdated := waitFunc(instanceName, mcp)
-			klog.V(5).InfoS("Machine Config Pool state", "name", mcp.Name, "instance", instanceName, "updated", isUpdated)
+		isUpdated := waitFunc(instanceName, mcp)
+		klog.V(5).InfoS("Machine Config Pool state", "name", mcp.Name, "instance", instanceName, "updated", isUpdated)
 
-			if !isUpdated {
-				return mcpStatuses, mcp.Name
-			}
+		if !isUpdated {
+			return mcpStatuses, mcp.Name
 		}
 	}
 	return mcpStatuses, ""
 }
 
-func syncMachineConfigPoolNodeGroupConfigStatuses(mcpStatuses []nropv1.MachineConfigPool, trees ...nodegroupv1.Tree) []nropv1.MachineConfigPool {
-	klog.V(4).InfoS("Machine Config Pool Node Group Status Sync start", "mcpStatuses", len(mcpStatuses), "trees", len(trees))
+func syncMachineConfigPoolNodeGroupConfigStatuses(mcpStatuses []nropv1.MachineConfigPool, tree nodegroupv1.Tree) []nropv1.MachineConfigPool {
+	klog.V(4).InfoS("Machine Config Pool Node Group Status Sync start", "mcpStatuses", len(mcpStatuses), "tree", tree.Name())
 	defer klog.V(4).Info("Machine Config Pool Node Group Status Sync stop")
 
 	updatedMcpStatuses := []nropv1.MachineConfigPool{}
-	for _, tree := range trees {
-		klog.V(5).InfoS("Machine Config Pool Node Group tree update", "mcps", len(tree.MachineConfigPools))
+	klog.V(5).InfoS("Machine Config Pool Node Group tree update", "mcps", len(tree.MachineConfigPools))
 
-		for _, mcp := range tree.MachineConfigPools {
-			mcpStatus := getMachineConfigPoolStatusByName(mcpStatuses, mcp.Name)
+	for _, mcp := range tree.MachineConfigPools {
+		mcpStatus := getMachineConfigPoolStatusByName(mcpStatuses, mcp.Name)
 
-			var confSource string
-			if tree.NodeGroup != nil && tree.NodeGroup.Config != nil {
-				confSource = "spec"
-				mcpStatus.Config = tree.NodeGroup.Config.DeepCopy()
-			} else {
-				confSource = "default"
-				ngc := nropv1.DefaultNodeGroupConfig()
-				mcpStatus.Config = &ngc
-			}
-
-			klog.V(6).InfoS("Machine Config Pool Node Group updated status config", "mcp", mcp.Name, "source", confSource, "data", mcpStatus.Config.ToString())
-
-			updatedMcpStatuses = append(updatedMcpStatuses, mcpStatus)
+		var confSource string
+		if tree.NodeGroup != nil && tree.NodeGroup.Config != nil {
+			confSource = "spec"
+			mcpStatus.Config = tree.NodeGroup.Config.DeepCopy()
+		} else {
+			confSource = "default"
+			ngc := nropv1.DefaultNodeGroupConfig()
+			mcpStatus.Config = &ngc
 		}
+
+		klog.V(6).InfoS("Machine Config Pool Node Group updated status config", "mcp", mcp.Name, "source", confSource, "data", mcpStatus.Config.ToString())
+
+		updatedMcpStatuses = append(updatedMcpStatuses, mcpStatus)
 	}
 	return updatedMcpStatuses
 }
--- a/internal/controller/numaresourcesoperator_controller.go
+++ b/internal/controller/numaresourcesoperator_controller.go
@@ -249,98 +249,63 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResourceAPI(ctx context.Conte
 	return intreconcile.StepSuccess()
 }
 
-func (r *NUMAResourcesOperatorReconciler) reconcileResourceMachineConfig(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, trees []nodegroupv1.Tree) intreconcile.Step {
-	// we need to sync machine configs first and wait for the MachineConfigPool updates
-	// before checking additional components for updates
-	waitByPool, pausedMCPNames, err := r.syncMachineConfigs(ctx, instance, existing, trees...)
-	if err != nil {
-		err = fmt.Errorf("failed to sync machine configs: %w", err)
-		return intreconcile.StepFailed(err)
-	}
-
-	// MCO needs to update the SELinux context removal and other stuff, and need to trigger a reboot.
-	// It can take a while.
-	mcpStatuses, mcpNamePending := syncMachineConfigPoolsStatuses(instance.Name, r.ForwardMCPConds, waitByPool, trees...)
-	instance.Status.MachineConfigPools = mcpStatuses
-	instance.Status.Conditions = updateMachineConfigPoolPausedCondition(instance.Status.Conditions, instance.Generation, pausedMCPNames)
-
-	if mcpNamePending != "" {
-		// the Machine Config Pool still did not apply the machine config, wait for one minute
-		return intreconcile.StepOngoing(numaResourcesRetryPeriod).WithReason("MachineConfigPoolIsUpdating").WithMessage(mcpNamePending + " is updating")
-	}
-	instance.Status.MachineConfigPools = syncMachineConfigPoolNodeGroupConfigStatuses(instance.Status.MachineConfigPools, trees...)
-
-	return intreconcile.StepSuccess()
-}
-
-func (r *NUMAResourcesOperatorReconciler) reconcileResourceDaemonSet(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, trees []nodegroupv1.Tree) ([]poolDaemonSet, intreconcile.Step) {
-	daemonSetsInfoPerPool, err := r.syncNUMAResourcesOperatorResources(ctx, instance, existing, trees)
-	if err != nil {
-		err = fmt.Errorf("FailedRTESync: %w", err)
-		return nil, intreconcile.StepFailed(err)
-	}
-
-	if len(daemonSetsInfoPerPool) == 0 {
-		return nil, intreconcile.StepSuccess()
-	}
-
-	dssWithReadyStatus, dsNamePending, err := r.syncDaemonSetsStatuses(ctx, r.Client, daemonSetsInfoPerPool)
-	instance.Status.DaemonSets = dssWithReadyStatus
-	instance.Status.RelatedObjects = relatedobjects.ResourceTopologyExporter(r.Namespace, dssWithReadyStatus)
-	if err != nil {
-		return nil, intreconcile.StepFailed(err)
-	}
-	if dsNamePending != "" {
-		return nil, intreconcile.StepOngoing(5 * time.Second).WithReason("DaemonSetIsUpdating").WithMessage(dsNamePending + " is updating")
-	}
-
-	return daemonSetsInfoPerPool, intreconcile.StepSuccess()
-}
-
 func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context, instance *nropv1.NUMAResourcesOperator, trees []nodegroupv1.Tree) intreconcile.Step {
 	if step := r.reconcileResourceAPI(ctx, instance, trees); step.EarlyStop() {
 		return step
 	}
 
-	existing := rtestate.FromClient(ctx, r.Client, r.Platform, r.RTEManifests, instance, trees, r.Namespace)
+	existing := rtestate.FromClientTreeAgnostic(ctx, r.Client, r.Platform, r.RTEManifests, instance, r.Namespace)
+	if err := r.setupTreeAgnosticManifests(ctx, instance); err != nil {
+		return intreconcile.StepFailed(fmt.Errorf("FailedSharedResourceSync: %w", err))
+	}
+	if err := r.applyObjects(ctx, instance, existing.TreeAgnostic(r.RTEManifests)); err != nil {
+		return intreconcile.StepFailed(fmt.Errorf("FailedSharedResourceSync: %w", err))
+	}
 
+	err := dangling.DeleteUnusedDaemonSets(r.Client, ctx, instance, trees)
+	if err != nil {
+		klog.ErrorS(err, "failed to deleted unused daemonsets")
+	}
 	if r.Platform == platform.OpenShift {
-		if step := r.reconcileResourceMachineConfig(ctx, instance, existing, trees); step.EarlyStop() {
-			return step
-		}
-	}
-
-	dsPerPool, step := r.reconcileResourceDaemonSet(ctx, instance, existing, trees)
-	if step.EarlyStop() {
-		return step
-	}
-
-	// all fields of NodeGroupStatus are required so publish the status only when all daemonset and MCPs are updated which
-	// is a certain thing if we got to this point otherwise the function would have returned already
-	instance.Status.NodeGroups = syncNodeGroupsStatus(instance, dsPerPool)
-
-	return intreconcile.StepSuccess()
-}
-
-func (r *NUMAResourcesOperatorReconciler) syncDaemonSetsStatuses(ctx context.Context, rd client.Reader, daemonSetsInfo []poolDaemonSet) ([]nropv1.NamespacedName, string, error) {
-	dssWithReadyStatus := []nropv1.NamespacedName{}
-	for _, dsInfo := range daemonSetsInfo {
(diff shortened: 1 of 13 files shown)


internal/controller/numaresourcesoperator_controller.go
```diff
@@ -511,49 +603,55 @@ func getMachineConfigPoolStatusByName(mcpStatuses []nropv1.MachineConfigPool, na
...
+		}
+		_, _, err = apply.ApplyObject(ctx, r.Client, objState)
+		if err != nil {
+			return fmt.Errorf("failed to apply (%s) %s/%s: %w", objState.Desired.GetObjectKind().GroupVersionKind(), objState.Desired.GetNamespace(), objState.Desired.GetName(), err)
 		}
 	}
+	return nil
+}
 
+func (r *NUMAResourcesOperatorReconciler) prepareAndApplySharedResources(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests) error {
```
Tal-or: The name suggest shared but it's not clear.
I would add comment explaining that shared are objects that don't depend on any specific tree.

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -243,82 +244,173 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResourceAPI(ctx context.Conte
...
 
-	dsPerPool, step := r.reconcileResourceDaemonSet(ctx, instance, existing, trees)
-	if step.EarlyStop() {
-		return step
+	var results []treeReconcileResult
+	for _, tree := range trees {
+		treeExisting := existing.ForTree(ctx, r.Client, tree)
+
+		if r.Platform == platform.OpenShift {
+			mcResult := r.reconcileTreeMachineConfig(ctx, instance, treeExisting, tree)
```
Tal-or: We need to return early here if the MCP is paused so it won't update the DS

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -243,82 +244,173 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResourceAPI(ctx context.Conte
...
-		// the Machine Config Pool still did not apply the machine config, wait for one minute
-		return intreconcile.StepOngoing(numaResourcesRetryPeriod).WithReason("MachineConfigPoolIsUpdating").WithMessage(mcpNamePending + " is updating")
+		result.step = intreconcile.StepOngoing(numaResourcesRetryPeriod).WithReason("MachineConfigPoolIsUpdating").WithMessage(mcpNamePending + " is updating")
+		return result
 	}
-	instance.Status.MachineConfigPools = syncMachineConfigPoolNodeGroupConfigStatuses(instance.Status.MachineConfigPools, trees)
+	result.mcpStatuses = syncMachineConfigPoolNodeGroupConfigStatuses(result.mcpStatuses, singleTreeSlice)
 
-	return intreconcile.StepSuccess()
+	result.step = intreconcile.StepSuccess()
```
Tal-or: if we have pausedMCPNames, we should not set this `StepSuccess()` but `StepOngoing()` so controller wont continue to update DS for this specific tree

Tal-or (commented):
I know it's still WIP but already adding my comments so they won't slip.

internal/controller/numaresourcesoperator_controller.go
ffromani: addressed in the last update

internal/controller/numaresourcesoperator_controller.go
ffromani: StepOngoing() wants to requeue the request though. I'm not 100% sure this is the right approach.

internal/controller/numaresourcesoperator_controller.go
ffromani: makes sense, done

internal/controller/numaresourcesoperator_controller.go
Tal-or: `StepSuccess` continues to update the DS, and doing that while the MCP paused, might causes to misalignment with security context/ SELinux policy.

Requeue the request for later is the lesser evil option IMO. 
Another option is not continue and not requeue, a new request will come in anyway once the MCP moved to `paused:  false` because we're watching it. 
Not sure if we have a step that does that though.

ffromani: we can probably use `StepOngoing` with 0 value; this should report the right condition without causing a requeue

pkg/objectstate/rte/machineconfigpool.go
```diff
@@ -137,65 +137,73 @@ type MachineConfigObjectState struct {
 }
 
 func (em *ExistingManifests) MachineConfigsState(mf Manifests) ([]MachineConfigObjectState, sets.Set[string]) {
+	var allRet []MachineConfigObjectState
```
Tal-or: Since this area is going through a major refactoring anyway, it would be nice to make the names of variables and functions more accurate - it's the `MachineConfigPool*` that we're checking/updating and not `MachineConfig*`
as we all aware those are related though different objects.

pkg/objectstate/rte/rte.go
```diff
@@ -366,6 +371,27 @@ func FromClient(ctx context.Context, cli client.Client, plat platform.Platform,
 	return &ret
 }
 
+func (em *ExistingManifests) PerTree(ctx context.Context, cli client.Client, tree nodegroupv1.Tree) *ExistingManifests {
+	ret := &ExistingManifests{
+		existing:       em.existing,
+		errs:           em.errs,
+		daemonSets:     make(map[string]daemonSetManifest),
+		machineConfigs: make(map[string]machineConfigManifest),
```
Tal-or, line 359: If that's `PerTree` how come more than a single DamonSet and/or MachineConfig?

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -837,3 +838,191 @@ func getMachineConfigPoolStatusByName(mcpStatuses []nropv1.MachineConfigPool, na
 	}
 	return nropv1.MachineConfigPool{Name: name}
 }
+
+type treeReconcileResult struct {
```
Tal-or: slice of DSs and MCPs implies that this is for all trees so should be `treesReconcileResult` in plural or returning a slice of `treeReconcileResult` one per tree

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -330,39 +326,39 @@ func (r *NUMAResourcesOperatorReconciler) reconcileResource(ctx context.Context,
...
-	if r.Platform == platform.OpenShift {
-		for _, tree := range trees {
-			treeExisting := existing.PerTree(ctx, r.Client, tree)
-			mcResults = append(mcResults, r.reconcileResourceMachineConfig(ctx, instance, treeExisting, tree))
+		treeExisting := existing.PerTree(ctx, r.Client, tree)
+
+		mcResult := treeReconcileResult{}
+		if r.Platform == platform.OpenShift {
+			mcResult = r.reconcileResourceMachineConfig(ctx, instance, treeExisting, tree)
+			if mcResult.step.EarlyStop() /* || mcResult.pausedMCPNames.Len() > 0*/ {
```
Tal-or: forsaken code comment?

Tal-or (commented):
Usually it's helpful to break the PR into smaller commits, but here (especially in the few last commits) some of the newer commits are overriding/updating the logic of their previous ones. 

I know it was made to convey the steps clearly but it's bit confusing because now I see that I commented on lines of code that were deleted/moved in later commits.

Anyway lets address what we have right now and I'll have another round later on this week.

internal/controller/numaresourcesoperator_controller.go
ffromani: no, this is is a gap on my side. Let me have another cleanup pass. Restoring WIP for the time being.

ffromani:
/retitle WIP: [CNF-23239: controller: nrop: per-tree processing and progress](https://github.com/openshift-kni/numaresources-operator/pull/3998#top)

thanks for the review. I'll need to do more cleanup

pkg/objectstate/rte/machineconfigpool.go
ffromani: I agree about improving, thing is this function want to compute the existing state of MachineConfig objects, and compute the desired state of MachineConfig, which are indeed related to MachineConfigPools, but the pool is the scoping mechanism.

IOW, we want to reconcile MachineConfigs, and return ObjectState about these.

pkg/objectstate/rte/rte.go
ffromani: Mostly, because I'm reusing `ExistingManifests` which predates the split and had 1:N relationship because the horizontal partitioning.

We can probably simplify further as followup of this PR

internal/controller/numaresourcesoperator_controller.go
ffromani: this is meant to be a reconcile result of a single tree (therefore `perTreeReconcileResult` is probably better), but the problem here is we can't yet enforce the 1:1 mapping between trees and MCPs (https://github.com/openshift-kni/numaresources-operator/blob/main/internal/api/annotations/annotations.go#L25).
Well, there's an argument about doing that in 5.0 - we had 3 EUS version past 4.18 already, but then should be done as prerequisite of this PR, and would make backporting significantly more complex.
This is why I'm still carrying this forward and postponing to 5.2.

ffromani:
reviewable again

ffromani (commented):
I acknowledge the review difficulties. Not sure how to help further unless we want to completely pivot direction and make a clean-as-possible main-only, minimal PR. That would make for a trivial backport down to 4.21 or so. Past that, it will both become hard and touch branches deeper and deeper in stable state. Which alternatives do we have?

pkg/objectstate/rte/rte.go
```diff
@@ -366,6 +371,27 @@ func FromClient(ctx context.Context, cli client.Client, plat platform.Platform,
 	return &ret
 }
 
+func (em *ExistingManifests) PerTree(ctx context.Context, cli client.Client, tree nodegroupv1.Tree) *ExistingManifests {
```
shajmakh, line 354: this is never used; but if for any reason (why?) we want to keep it, can we call it PerTreeClone to illustrate that it clones the manifests?

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -857,3 +858,191 @@ func getMachineConfigPoolStatusByName(mcpStatuses []nropv1.MachineConfigPool, na
...
+	}
+
+	rteupdate.SecurityContextConstraint(r.RTEManifests.Core.SecurityContextConstraint, true) // force to legacy context
+
+	return nil
+}
+
+func (r *NUMAResourcesOperatorReconciler) applyObjects(ctx context.Context, instance *nropv1.NUMAResourcesOperator, objStates []objectstate.ObjectState) error {
+	klog.V(4).InfoS("Applying objects", "count", len(objStates))
+	defer klog.V(4).InfoS("Applyied objects", "count", len(objStates))
```
shajmakh: nit: Applied

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -857,3 +858,191 @@ func getMachineConfigPoolStatusByName(mcpStatuses []nropv1.MachineConfigPool, na
...
+
+func shouldReplaceStep(current, candidate intreconcile.Step) bool {
+	if current.Done() {
+		return true
+	}
+	if candidate.Failed() && !current.Failed() {
+		return true
+	}
+	if candidate.Ongoing() && current.Ongoing() {
+		return candidate.Result.RequeueAfter < current.Result.RequeueAfter
```
shajmakh: IIUC looks easy to miss a needed replace; what if the message/reason of the old Ongoing is different than the new one?

pkg/status/conditioninfo/conditioninfo.go
```diff
@@ -49,6 +49,18 @@ func (ci ConditionInfo) WithMessage(message string) ConditionInfo {
 	return ret
 }
 
+// UpdateMessage overrides the ConditionInfo message with the given value.
+// If a message is already set, the new value is prepended to it.
+func (ci ConditionInfo) UpdateMessage(message string) ConditionInfo {
+	ret := ci
```
Tal-or, line 55: what is the purpose of the additional assignment? why not assign to `ci` directly?

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -857,3 +715,213 @@ func IsRTEImageOverridable() bool {
...
+		Name: mcp.Name,
+	}
+	if !forwardMCPConds {
+		return mcpStatus
+	}
+	mcpStatus.Conditions = mcp.Status.Conditions
+	return mcpStatus
+}
+
+func getMachineConfigPoolStatusByName(mcpStatuses []nropv1.MachineConfigPool, name string) nropv1.MachineConfigPool {
```
Tal-or: The name of the function suggest that we're returning `mcp.Status` object but we're actually a MachineConfigPool

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -857,3 +715,213 @@ func IsRTEImageOverridable() bool {
...
+	mcpStatuses    []nropv1.MachineConfigPool
+	pausedMCPNames sets.Set[string]
+	step           intreconcile.Step
+}
+
+func (r *NUMAResourcesOperatorReconciler) reconcilePerTreeMachineConfig(ctx context.Context, instance *nropv1.NUMAResourcesOperator, existing *rtestate.ExistingManifests, tree nodegroupv1.Tree) perTreeResult {
+	result := perTreeResult{}
+
+	waitByPool, pausedMCPNames, err := r.syncMachineConfig(ctx, instance, existing, tree)
+	result.pausedMCPNames = pausedMCPNames
```
Tal-or: do we still want this assignment to happen when `err != nil`? (one line below)

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -857,3 +715,213 @@ func IsRTEImageOverridable() bool {
...
+	result.step = intreconcile.StepSuccess()
+	return result
+}
+
+func (r *NUMAResourcesOperatorReconciler) setupTreeAgnosticManifests(ctx context.Context, instance *nropv1.NUMAResourcesOperator) error {
+	rteupdate.DaemonSetRolloutSettings(r.RTEManifests.Core.DaemonSet)
+	err := rteupdate.DaemonSetAffinitySettings(r.RTEManifests.Core.DaemonSet, r.RTEManifests.Core.DaemonSet.Spec.Template.Labels)
+	if err != nil {
+		klog.ErrorS(err, "failed to update RTE affinity settings")
+	}
```
Tal-or:
```suggestion
	if err := rteupdate.DaemonSetAffinitySettings(r.RTEManifests.Core.DaemonSet, r.RTEManifests.Core.DaemonSet.Spec.Template.Labels); err != nil {
		klog.ErrorS(err, "failed to update RTE affinity settings")
	}
```

same for the others below, saves a few repetitive lines

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -857,3 +715,213 @@ func IsRTEImageOverridable() bool {
...
+
+	result.step = intreconcile.StepSuccess()
+	return result
+}
+
+func (r *NUMAResourcesOperatorReconciler) setupTreeAgnosticManifests(ctx context.Context, instance *nropv1.NUMAResourcesOperator) error {
+	rteupdate.DaemonSetRolloutSettings(r.RTEManifests.Core.DaemonSet)
+	err := rteupdate.DaemonSetAffinitySettings(r.RTEManifests.Core.DaemonSet, r.RTEManifests.Core.DaemonSet.Spec.Template.Labels)
+	if err != nil {
+		klog.ErrorS(err, "failed to update RTE affinity settings")
```
Tal-or: why here we're accepting to continue when `err != nil` and in the rest we don't

pkg/objectstate/rte/machineconfigpool.go
```diff
@@ -136,66 +136,66 @@ type MachineConfigObjectState struct {
...
+				Existing: existingMachineConfig.machineConfig,
+				Error:    existingMachineConfig.machineConfigError,
+				Desired:  desiredMachineConfig,
+				Compare:  compare.Object,
+				Merge:    merge.ObjectForUpdate,
+			},
+			PoolName:       mcp.Name,
+			WaitForUpdated: IsMachineConfigPoolUpdated,
+		})
 	}
```
Tal-or, line 199:
```suggestion
state := MachineConfigObjectState{
			ObjectState: objectstate.ObjectState{
				Existing: existingMachineConfig.machineConfig,
				Error:    existingMachineConfig.machineConfigError,
				// caution here: we want a *nil interface value*, not an *interface which points to nil*.
				// the latter would lead to apparently correct code leading to runtime panics. See:
				// https://trstringer.com/go-nil-interface-and-interface-with-nil-concrete-value/
				// (and many other docs like this)
				Desired: nil,
				Compare: compare.Object,
				Merge:   merge.ObjectForUpdate,
			},
			PoolName:       mcp.Name,
			WaitForUpdated: IsMachineConfigPoolUpdatedAfterDeletion,
		}

		if isCustomPolicy {
			mc := mf.Core.MachineConfig.DeepCopy()
			mc.Name = mcName
			mc.Labels = GetMachineConfigLabel(mcp)
			state.Desired = mc
			state.WaitForUpdated = IsMachineConfigPoolUpdated
		}

		ret = append(ret, state)
	}
```

shorter version, reusing what we can

pkg/status/conditioninfo/conditioninfo.go
Tal-or, line 55: why not using `ci` directly? what the additional assignment gives us?

Tal-or (commented):
looking good overall.
small comments from my side

pkg/objectstate/rte/rte.go
ffromani: `PerTree` will be used in later commits. The reason for this breakdown is explained (or I tried to) in the PR description. https://github.com/openshift-kni/numaresources-operator/pull/3998#issue-4427755021
I acknowledge this approach is unusual and creates friction. The challenge of a deep backport is a key concern and we will need to address it somehow, because it's gonna bite us hard later. I'm open to suggestions

Regarding the rename, not a fan of it but I can do it

internal/controller/numaresourcesoperator_controller.go
ffromani: right, will rename

internal/controller/numaresourcesoperator_controller.go
ffromani: worth moving below indeed

internal/controller/numaresourcesoperator_controller.go
ffromani: because we inherited code irregularity. I will fix in a followup
[EDIT] I stand corrected. The idea is (was?) to clone the receiver value without abusing the pass-by-value semantics. We can change, but it would be better to do in its own commit.

internal/controller/numaresourcesoperator_controller.go
ffromani: worth tightening the check indeed

pkg/status/conditioninfo/conditioninfo.go
ffromani: nothing actually. Old iteration left unchecked. Also, the name should be fixed because it is misleading

internal/controller/numaresourcesoperator_controller.go
ffromani: we use the idioms  freely and we never established a convention or guidelines. In this case I don't mind much but I also don't see this as a clear improvement. Applying anyway.

pkg/objectstate/rte/machineconfigpool.go
ffromani: the old code was intentionally explicit. Everything else equal, I'd like to preserve that.

internal/controller/numaresourcesoperator_controller.go
ffromani: closer inspection revealed a hidden coupling. Clairifying in the docs.

pkg/status/conditioninfo/conditioninfo.go
Tal-or: so we're leaving it as it is?

Tal-or: not a big of a deal anyway lets not resubmit for that

Tal-or (approved):
/approve
/lgtm

pkg/status/conditioninfo/conditioninfo.go
ffromani: for now, yes. I can post a followup cleanup once this merges

ffromani:
nice start
let's give main branch some testing time before we start the backport dance

Tal-or:
> nice start let's give main branch some testing time before we start the backport dance

Backport is going to be rough one

ffromani:
> > nice start let's give main branch some testing time before we start the backport dance
> 
> Backport is going to be rough one

Let's see and let's see if the weird strategy implemented here will pay dividends or not

PR #4018, opened 2026-05-13
CNF-21917: bump min_kube_version 1.32.0 to 1.35.0

yanirq:


.konflux/bundle/overlay/release.in.yaml
```diff
@@ -9,5 +9,5 @@ variables:
   manager_version: "numaresources-operator.v5.0.0"
   # min_kube_version should match ocp
   # https://access.redhat.com/solutions/4870701
-  min_kube_version: "1.32.0"
+  min_kube_version: "1.35.0"
```
ffromani, line 12: why? which feature do we _need_ from 1.35? does this mean we are silently depending on something is not available before?

yanirq: yes , in retrospective,  this PR is ahead of its time I agree. 
we can close it for now and realign when needed

ffromani:
/close

xref: https://github.com/openshift-kni/numaresources-operator/pull/4018#discussion_r3241505863

PR #4021, opened 2026-05-14
OCPBUGS-79418: ctrl: sched: add PodAntiAffinity to deployment

shajmakh:
The default functionality of the scheduler HA feature is to create scheduler
pod per node (control-plane in Hypershift and worker in Openshift).
This will create the pods regardless of the readiness of the nodes, and
doesn't necessarily place one pod on each node and this specially rises
when some of the hosting nodes are down. The scheduler pod that was
meant to be scheduled on that node will be running on another node which
already has the scheduler pod. The duplicate pod does not add much value while
 still consume node resources.

This commit adjusts the scheduler deployment to use PodAntiAffinity to
avoid creating the same app pod on the same host; IOW each possible host
will host only one scheduler pod.This applies only when the autodetection
of scheduler replicas is requested, meaning only when spec.replicas is
nil (unset).

However, having PodAntiAffinity is not enough specially for deployments running
on compact OpenShift clusters (control-plane nodes only, typically three) with high
availability and preferred NodeAffinity tied to control-plane nodes.
n such a case, the default RollingUpdate surge tries to add a new pod before terminating
an old one. With every node already hosting a scheduler pod, the extra pod
cannot schedule and the rollout stalls (new pods Pending).

To fix this, we need to adjust the rollout strategy on the deployment to first delete a pod then
create a new updated one,that's done useing RollingUpdate strategy with maxSurge 0 and maxUnavailable
1 so nodes free a slot before the replacement pod is scheduled.

### DIFF
--- a/internal/controller/numaresourcesscheduler_controller_test.go
+++ b/internal/controller/numaresourcesscheduler_controller_test.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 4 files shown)


shajmakh:
A cleaner version of https://github.com/openshift-kni/numaresources-operator/pull/3334
/cc @ffromani @Tal-or @swatisehgal

pkg/objectupdate/sched/sched.go
```diff
@@ -102,6 +106,52 @@ func DeploymentTLSSettings(dp *appsv1.Deployment, tlsSettings objtls.Settings) e
...
+// An explicit non-nil Replicas value means the user chose a replica count and keeps
+// full control (no required pod anti-affinity).
+func DeploymentAffinityWithStrategy(dp *appsv1.Deployment, spec nropv1.NUMAResourcesSchedulerSpec) error {
+	if spec.Replicas != nil {
+		if dp.Spec.Template.Spec.Affinity != nil {
+			dp.Spec.Template.Spec.Affinity.PodAntiAffinity = nil
+		}
+		dp.Spec.Strategy = appsv1.DeploymentStrategy{}
+		return nil
+	}
```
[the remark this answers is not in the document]
shajmakh: Incorrect commit message. addressing ..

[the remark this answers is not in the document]
shajmakh: the commit is actually correct. I updated the PR description. can you check now?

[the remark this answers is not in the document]
shajmakh: for the future, in case I want to ask you to review a specific part in the PR, is there any way to tag you in a comment?

shajmakh:
@ffromani Thank you for the extra live clarifications in https://github.com/openshift-kni/numaresources-operator/tree/sched-ha-podantiaff-redo. I addressed the comments as discussed and will be filing a different PR for e2e and a follow-up PR main-branch-only enhancements.

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -1031,6 +1032,143 @@ var _ = Describe("Test NUMAResourcesScheduler Reconcile", func() {
...
+		var (
+			schedDeployLabels  map[string]string
+			expectedPodAntiAff *corev1.PodAntiAffinity
+			expectedStrategy   appsv1.DeploymentStrategy
+		)
+
+		BeforeEach(func() {
+			schedDeployLabels = map[string]string{
+				"app": "secondary-scheduler",
+			}
```
ffromani: nit: unnecessary, used only once.

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -1031,6 +1032,143 @@ var _ = Describe("Test NUMAResourcesScheduler Reconcile", func() {
...
+
+		DescribeTable("should set PodAntiAffinity for scheduler Deployment according to replicas count", func(ctx context.Context, tc testCase) {
+			nrs := testobjs.NewNUMAResourcesScheduler("numaresourcesscheduler", "some/url:latest", testSchedulerName, 11*time.Second)
+			nrs.Spec.Replicas = tc.replicasCount
+			initObjects := []runtime.Object{nrs}
+			initObjects = append(initObjects, fakeNodes(3, 0)...)
+			reconciler, err := NewFakeNUMAResourcesSchedulerReconciler(initObjects...)
+			Expect(err).ToNot(HaveOccurred())
+			// Baseline node affinity comes from the scheduler deployment manifest; reconcile must preserve it.
+			expectedNodeAff := reconciler.SchedulerManifests.Deployment.Spec.Template.Spec.Affinity.NodeAffinity.DeepCopy()
```
ffromani: not critical: should be inlined as constant

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -1031,6 +1032,143 @@ var _ = Describe("Test NUMAResourcesScheduler Reconcile", func() {
...
+			// reconcile with non-nil replicas count
+			Eventually(func() bool {
+				Expect(reconciler.Client.Get(ctx, key, nrs)).ToNot(HaveOccurred())
+				nrs.Spec.Replicas = ptr.To(int32(2))
+				if err := reconciler.Client.Update(ctx, nrs); err != nil {
+					klog.Warningf("failed to update the scheduler object; err: %v", err)
+					return false
+				}
+				return true
+			}, timeout, interval).Should(BeTrue())
```
ffromani: nit: we should use WithPolling/WithTimeout. This removed the need for named variables/constants

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -1031,6 +1032,143 @@ var _ = Describe("Test NUMAResourcesScheduler Reconcile", func() {
...
+
+			dp := &appsv1.Deployment{}
+			Expect(reconciler.Client.Get(ctx, client.ObjectKey{Namespace: testNamespace, Name: "secondary-scheduler"}, dp)).To(Succeed())
+
+			podAntiAffinity := dp.Spec.Template.Spec.Affinity.PodAntiAffinity
+			Expect(podAntiAffinity).To(Equal(expectedPodAntiAff))
+			Expect(dp.Spec.Strategy).To(Equal(expectedStrategy))
+
+			// reconcile with non-nil replicas count
+			Eventually(func() bool {
```
ffromani: nit: better to return `error` so the assertion is self-evident

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -1031,6 +1032,143 @@ var _ = Describe("Test NUMAResourcesScheduler Reconcile", func() {
...
+			}, timeout, interval).Should(BeTrue())
+			_, err = reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: key})
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(reconciler.Client.Get(ctx, client.ObjectKey{Namespace: testNamespace, Name: "secondary-scheduler"}, dp)).To(Succeed())
+			Expect(dp.Spec.Template.Spec.Affinity.PodAntiAffinity).To(BeNil())
+			Expect(dp.Spec.Strategy).To(Equal(appsv1.DeploymentStrategy{}))
+
+			// reconcile with a different non-nil replicas count should keep the same result
+			Eventually(func() bool {
```
ffromani: nit:ditto

pkg/objectupdate/sched/sched.go
```diff
@@ -102,6 +106,52 @@ func DeploymentTLSSettings(dp *appsv1.Deployment, tlsSettings objtls.Settings) e
 	return nil
 }
 
+// DeploymentAffinityWithStrategy configures required pod anti-affinity on the scheduler
+// deployment only when the CR leaves replica count to autodetection (Replicas unset).
+// An explicit non-nil Replicas value means the user chose a replica count and keeps
+// full control (no required pod anti-affinity).
+func DeploymentAffinityWithStrategy(dp *appsv1.Deployment, spec nropv1.NUMAResourcesSchedulerSpec) error {
+	if spec.Replicas != nil {
```
ffromani, line 114: At glance this looked wrong, but thinking about it it's correct and it's actually consistent with what we do elsewhere. 0 replicas is a legit value to be set explicitly. That warrants for the default strategy.

pkg/objectupdate/sched/sched.go
```diff
@@ -102,6 +106,52 @@ func DeploymentTLSSettings(dp *appsv1.Deployment, tlsSettings objtls.Settings) e
...
+		if dp.Spec.Template.Spec.Affinity != nil {
+			dp.Spec.Template.Spec.Affinity.PodAntiAffinity = nil
+		}
+		dp.Spec.Strategy = appsv1.DeploymentStrategy{}
+		return nil
+	}
+
+	labels := dp.Spec.Template.Labels
```
ffromani, line 122: nit: shortcut used only once. Should inline

pkg/objectupdate/sched/sched_test.go
```diff
@@ -631,6 +635,125 @@ func TestDeploymentTLSSettingsRepeated(t *testing.T) {
...
+	}
+
+	if dp.Spec.Template.Spec.Affinity != nil && dp.Spec.Template.Spec.Affinity.PodAntiAffinity != nil {
+		t.Fatalf("Override failed: expected no podAntiAffinity but got %v", dp.Spec.Template.Spec.Affinity.PodAntiAffinity)
+	}
+	if dp.Spec.Strategy != (appsv1.DeploymentStrategy{}) {
+		t.Fatalf("expected strategy to be reset but got %v", dp.Spec.Strategy)
+	}
+}
+
```
ffromani, line 756: please add another controller test for the explicit -> autodetect transition:
start with Replicas=N (explicit), reconcile, then switch to Replicas=nil (autodetect) and verify PodAntiAffinity gets added.

ffromani (commented):
we need another test and we're good. For real this time.

pkg/objectupdate/sched/sched.go
shajmakh: meant to be used twice, thanks for catching.

internal/controller/numaresourcesscheduler_controller_test.go
shajmakh: we want to ensure the reconcile loop doesn't modify this, so we need t save a copy for the initial value because it might mutate

pkg/objectupdate/sched/sched_test.go
shajmakh: done, Thanks for catching.

ffromani:
/approve
/lgtm

thanks for the hard work here

ffromani:
/approve cancel

concurrent update, let me check again

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -1031,6 +1032,193 @@ var _ = Describe("Test NUMAResourcesScheduler Reconcile", func() {
...
+				Expect(affinity.PodAntiAffinity).To(Equal(expectedPodAntiAff))
+				Expect(dp.Spec.Strategy).To(Equal(expectedStrategy))
+			},
+				Entry("when replicas are not set, expected podAntiAffinity", testCase{expectPodAntiAffinity: true}),
+				Entry("when replicas are set and zero, no podAntiAffinity is set", testCase{replicasCount: ptr.To(int32(0)), expectPodAntiAffinity: false}),
+				Entry("when replicas are set and non-zero, no podAntiAffinity is set", testCase{replicasCount: ptr.To(int32(1)), expectPodAntiAffinity: false}),
+			)
+		})
+
+		Context("switch between replicas counts", func() {
```
ffromani, line 1129: Nit: good. Better:
```When("switching between replica counts" ...)```
produces better reading messages. Please **do not resubmit to fix this**

ffromani (commented):
/lgtm

thanks for the hard work

shajmakh:
/retitle OCPBUGS-79418: ctrl: sched: add PodAntiAffinity to deployment

PR #4028, opened 2026-05-14
e2e: sched: verify PodAntiAffinity and Rollout Strategy settings

shajmakh:
With https://github.com/openshift-kni/numaresources-operator/pull/4021
introducing the deployment PodAntiAffinity and the update to rollout
strategy to allow terminating a pod first before creating a new one,
a new condition was noticed on especially compact clusters (can be
reproduced on an MNO as well by marking worker-nodes as unschedulable)
that revealed that on a specific scenario of switching between
autodetection replicas mode to an explicit replicas count that's equal to
the autodetected count, the new version of the deployment will have the
default strategy which is 25% maxSurge and 25%maxUnavailable. The pods
of the new replicas will keep pending due to a stall rollout because of
the combination of the default strategy and PodAntiAffinityone.
By example it will be like this on a 3 control-plane cluster:
1. autodetect mode: deployment complete with 3 pods running
2. switch to explicit 3 replicas: because maxSurge allows one pod more
   it gets created, but because of the podAntiaffinity it keeps pending,
already using the maxUnavailable limit.
3. the result is a deadlock: that the deployment has 4 replicas (4 pods), 1 pending
   (new one part of the rollout); and the old 3 are running.

The solution is to allow at least to delete an old pod first then
continue deploying the new ones. That was the behavior set only when
podAntiAffinity is set. This commit extends that to be the default
behavior (set during the reconcile).

The PR includes implemented e2e that covers the podantiaffinity and the strategy's new expectations, plus an enhancement in the way the tests wait for an updated deployment to be complete.

### DIFF
--- a/internal/controller/numaresourcesscheduler_controller_test.go
+++ b/internal/controller/numaresourcesscheduler_controller_test.go
@@ -1033,10 +1033,75 @@ var _ = Describe("Test NUMAResourcesScheduler Reconcile", func() {
 		})
 	})
 
-	Context("when setting the scheduler PodAntiAffinity", func() {
+	When("setting the scheduler rollout strategy", func() {
+		var (
+			expectedStrategy appsv1.DeploymentStrategy
+			dpKey            client.ObjectKey
+		)
+
+		BeforeEach(func() {
+			expectedStrategy = appsv1.DeploymentStrategy{
+				Type: appsv1.RollingUpdateDeploymentStrategyType,
+				RollingUpdate: &appsv1.RollingUpdateDeployment{
+					MaxUnavailable: ptr.To(intstr.FromInt(1)),
+					MaxSurge:       ptr.To(intstr.FromInt(0)),
+				},
+			}
+			dpKey = client.ObjectKey{Namespace: testNamespace, Name: "secondary-scheduler"}
+
+		})
+
+		It("should set the rollout strategy with maxSurge set to 0 by default", func(ctx context.Context) {
+			nrs := testobjs.NewNUMAResourcesScheduler("numaresourcesscheduler", "some/url:latest", testSchedulerName, 11*time.Second)
+			initObjects := []runtime.Object{nrs}
+			initObjects = append(initObjects, fakeNodes(3, 0)...)
+			reconciler, err := NewFakeNUMAResourcesSchedulerReconciler(initObjects...)
+			Expect(err).ToNot(HaveOccurred())
+			key := client.ObjectKeyFromObject(nrs)
+			_, err = reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: key})
+			Expect(err).ToNot(HaveOccurred())
+
+			dp := &appsv1.Deployment{}
+			Expect(reconciler.Client.Get(ctx, dpKey, dp)).To(Succeed())
+			Expect(dp.Spec.Strategy).To(Equal(expectedStrategy))
+		})
+
+		It("should still set the rollout strategy with maxSurge set to 0 even if the replicas are set to non-zero", func(ctx context.Context) {
+			nrs := testobjs.NewNUMAResourcesScheduler("numaresourcesscheduler", "some/url:latest", testSchedulerName, 11*time.Second)
+			nrs.Spec.Replicas = ptr.To(int32(3))
+			initObjects := []runtime.Object{nrs}
+			initObjects = append(initObjects, fakeNodes(3, 0)...)
+			reconciler, err := NewFakeNUMAResourcesSchedulerReconciler(initObjects...)
+			Expect(err).ToNot(HaveOccurred())
+			key := client.ObjectKeyFromObject(nrs)
+			_, err = reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: key})
+			Expect(err).ToNot(HaveOccurred())
+
+			dp := &appsv1.Deployment{}
+			Expect(reconciler.Client.Get(ctx, dpKey, dp)).To(Succeed())
+			Expect(dp.Spec.Strategy).To(Equal(expectedStrategy))
+		})
+
+		It("should still set the rollout strategy with maxSurge set to 0 even if the replicas are set to zero", func(ctx context.Context) {
+			nrs := testobjs.NewNUMAResourcesScheduler("numaresourcesscheduler", "some/url:latest", testSchedulerName, 11*time.Second)
+			nrs.Spec.Replicas = ptr.To(int32(0))
+			initObjects := []runtime.Object{nrs}
+			initObjects = append(initObjects, fakeNodes(3, 0)...)
+			reconciler, err := NewFakeNUMAResourcesSchedulerReconciler(initObjects...)
+			Expect(err).ToNot(HaveOccurred())
+			key := client.ObjectKeyFromObject(nrs)
+			_, err = reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: key})
+			Expect(err).ToNot(HaveOccurred())
+
+			dp := &appsv1.Deployment{}
+			Expect(reconciler.Client.Get(ctx, dpKey, dp)).To(Succeed())
+			Expect(dp.Spec.Strategy).To(Equal(expectedStrategy))
+		})
+	})
+
+	When("setting the scheduler PodAntiAffinity", func() {
 		var (
 			expectedPodAntiAff *corev1.PodAntiAffinity
-			expectedStrategy   appsv1.DeploymentStrategy
 		)
 
 		BeforeEach(func() {
@@ -1050,13 +1115,6 @@ var _ = Describe("Test NUMAResourcesScheduler Reconcile", func() {
 						},
 						TopologyKey: "kubernetes.io/hostname",
 					},
-				},
-			}
-			expectedStrategy = appsv1.DeploymentStrategy{
-				Type: appsv1.RollingUpdateDeploymentStrategyType,
-				RollingUpdate: &appsv1.RollingUpdateDeployment{
-					MaxUnavailable: ptr.To(intstr.FromInt(1)),
-					MaxSurge:       ptr.To(intstr.FromInt(0)),
 				},
 			}
 		})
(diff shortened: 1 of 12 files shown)


internal/wait/deployment.go
```diff
@@ -25,6 +25,10 @@ import (
 )
 
 func (wt Waiter) ForDeploymentComplete(ctx context.Context, dp *appsv1.Deployment) (*appsv1.Deployment, error) {
+	return wt.ForDeploymentCompleteWithReplicas(ctx, dp, dp.Spec.Replicas)
+}
+
+func (wt Waiter) ForDeploymentCompleteWithReplicas(ctx context.Context, dp *appsv1.Deployment, expectedReplicas *int32) (*appsv1.Deployment, error) {
```
shajmakh, line 36: TODO: split this improvement into a different commit

shajmakh:
/test ci-e2e
infra issue

shajmakh:
compact cluster CI failure is real.
EDIT: needed a rebase. Let's see how it looks like after the CI is done.

yanirq:
@shajmakh is this really just an e2e change or touches also core payload ?

shajmakh:
> @shajmakh is this really just an e2e change or touches also core payload ?

@yanirq both. e2e has exposed a misconfiguration in the current approach. This PR includes a fix for that and the e2e that verifies it.

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -1033,10 +1033,75 @@ var _ = Describe("Test NUMAResourcesScheduler Reconcile", func() {
...
+
+		BeforeEach(func() {
+			expectedStrategy = appsv1.DeploymentStrategy{
+				Type: appsv1.RollingUpdateDeploymentStrategyType,
+				RollingUpdate: &appsv1.RollingUpdateDeployment{
+					MaxUnavailable: ptr.To(intstr.FromInt(1)),
+					MaxSurge:       ptr.To(intstr.FromInt(0)),
+				},
+			}
+			dpKey = client.ObjectKey{Namespace: testNamespace, Name: "secondary-scheduler"}
```
ffromani, line 1050: bit: this should be determined from the cluster state, not hardcoded

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -1033,10 +1033,75 @@ var _ = Describe("Test NUMAResourcesScheduler Reconcile", func() {
...
+			dp := &appsv1.Deployment{}
+			Expect(reconciler.Client.Get(ctx, dpKey, dp)).To(Succeed())
+			Expect(dp.Spec.Strategy).To(Equal(expectedStrategy))
+		})
+
+		It("should still set the rollout strategy with maxSurge set to 0 even if the replicas are set to non-zero", func(ctx context.Context) {
+			nrs := testobjs.NewNUMAResourcesScheduler("numaresourcesscheduler", "some/url:latest", testSchedulerName, 11*time.Second)
+			nrs.Spec.Replicas = ptr.To(int32(3))
+			initObjects := []runtime.Object{nrs}
+			initObjects = append(initObjects, fakeNodes(3, 0)...)
```
ffromani, line 1073: nit: in this case we can use
```
numNodes := 3
nrs.Spec.Replicas = ptr.To(int32(numNodes))
initObjects := []runtime.Object{nrs}
initObjects = append(initObjects, fakeNodes(numNodes, 0)...)
```

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -1033,10 +1033,75 @@ var _ = Describe("Test NUMAResourcesScheduler Reconcile", func() {
...
+			reconciler, err := NewFakeNUMAResourcesSchedulerReconciler(initObjects...)
+			Expect(err).ToNot(HaveOccurred())
+			key := client.ObjectKeyFromObject(nrs)
+			_, err = reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: key})
+			Expect(err).ToNot(HaveOccurred())
+
+			dp := &appsv1.Deployment{}
+			Expect(reconciler.Client.Get(ctx, dpKey, dp)).To(Succeed())
+			Expect(dp.Spec.Strategy).To(Equal(expectedStrategy))
```
ffromani, line 1098: nit: it seems we can make a tabular test? can probably be cleaned up later though

internal/wait/deployment.go
```diff
@@ -39,26 +44,110 @@ func (wt Waiter) ForDeploymentComplete(ctx context.Context, dp *appsv1.Deploymen
...
+			return false, nil
+		}
+
 		klog.Infof("deployment %s complete", key.String())
 		return true, nil
 	})
 	return updatedDp, err
 }
 
+func (wt Waiter) isDeploymentRolloutComplete(ctx context.Context, dp *appsv1.Deployment, expectedReplicas *int32) (bool, error) {
```
ffromani, line 67: why do we need to manually list items and can't we just trust and make logic on the deployment status data?

test/e2e/sched/sched_test.go
```diff
@@ -273,4 +275,146 @@ var _ = Describe("[Scheduler] imageReplacement", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			By("verifying new pods are created")
+			podsUIDsInitial = make([]types.UID, 0, len(podListInitial))
+			for _, pod := range podListInitial {
+				if pod.Status.Phase != corev1.PodRunning || pod.DeletionTimestamp != nil {
+					continue
+				}
+				podsUIDsInitial = append(podsUIDsInitial, pod.UID)
+			}
+			Expect(podsUIDsInitial).To(HaveLen(int(*autoDetectedReplicasCount)), "expected %d initial pods, got %d", *autoDetectedReplicasCount, len(podsUIDsInitial))
```
ffromani, line 346: this makes sense, but we're doublechecking the kubernetes deployment primitive does what it promises. Would it make sense to make logic on the deployment status?

test/e2e/sched/sched_test.go
```diff
@@ -273,4 +275,146 @@ var _ = Describe("[Scheduler] imageReplacement", func() {
...
+		DescribeTable("should set PodAntiAffinity on replicas autodetection (high-availability) and update it when replicas are set explicitly", func(ctx context.Context, replicasCount int) {
+			By("checking PodAntiAffinity is removed when replicas are set explicitly")
+			newCount := int32(replicasCount)
+			if replicasCount == -1 {
+				// intentionally challenge the condition by setting the replicas count to the autodetected count;
+				// this isolates the podAffinity/strategy check from the old deployment replica count (which results
+				//  the same number of pods on autodetection mode)
+				newCount = *autoDetectedReplicasCount
+			}
+			By(fmt.Sprintf("update NRS to use explicit %d replicas", newCount))
```
ffromani, line 358: please use `e2efixture.By()`

ffromani (commented):
I was thinking that we need the new strategy essentially when we are at capacity. I'm trying to think if we should tune the strategy for the case on which the user set replica count much lower than the control plane size, e.g. to 1. In that case we should NOT delete the old pod before we create the new one. Any chance to add this fix?
Ideally we should just use a single policy everytime, if we need to tune the strategy depending on the replica/control plane node count this is a smell.

the fix makes sense and so do the e2e tests, the main question is how do we wait for the deployment to complete - do we need to revalidate what the deployment controller is supposed to be doing?

everything else seems reasonnable.

shajmakh:
> the strategy for the case on which the user set replica count much lower than the control plane size, e.g. to 1. In that case we should NOT delete the old pod before we create the new one. Any chance to add this fix?

@ffromani anything can be added, no problem, I'm trying to think about the benefit of adding this though.
The problem is that having podantiaffinity existed on the old pods and a default rollout strategy is being applied will create a deadlock on small clusters 3 or below. exactly for the reason explained in the commit. 
Now for 1 controlplane node, there might be some more flexibility to assign the new pod on other node but only when we are sure that there are other nodes that are schedulable. And because are not aware of the consumer's environment and its node combination or health state, we cannot rely on that and we need to consider the worst scenario blindly. This was not the case before PodAntiAffinity because this exact affinity is the one creating the strictness combined with rollout strategy. new pods need to respect old pods' rules, which is this case the podantiaffinity on the old one. This is reflected under the new pending pod events:

```
  Warning  FailedScheduling  17m (x2 over 22m)  default-scheduler  0/3 nodes are available: 3 node(s) didn't satisfy existing pods anti-affinity rules. no new claims to deallocate, preemption: 0/3 nodes are available: 3 node(s) didn't satisfy existing pods anti-affinity rules.

```

notice the **"3 node(s) didn't satisfy existing pods anti-affinity rules."**. 
The fix for this deadlock can be done:
1. manually by simply deleting an old pod to allow room for the new pending one. 
2. same as (1) but automated by using rollout strategy with explicit 1 max unavailable (because 25% on 3 nodes rounds down to 0).
3. manually by clearing all the old replicasets or completley remove the deployment to trigger a recreate.
4. same as (3) but automated using Recrearte rollout strategy https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#recreate-deployment 

We want to avoid manual fixes so 2 and 4 are not considered. 3 is the cleanest option but it risks the availability of the scheduler and needs downtime. so 1 is the most reasonable despite that the rollout will take time to have the replicas all updated but it is guaranteed on all types of clusters.

shajmakh:
/hold 
Another deadlock scenario is discovered in a new e2e that uses the updated strategy. I need to reconsider the approach.

Tal-or (commented):
To make sure I follow:
This bug was revealed on compat clusters because they have only 3 nodes? and then when deployment tried to spin up the forth replica during rollout, it failed due to the PodAntiAffinity settings?

The part I don't understand from the PR description:
>The solution is to allow at least to delete an old pod first then
continue deploying the new ones. That was the behavior set only when
podAntiAffinity is set. This commit extends that to be the default
behavior (set during the reconcile).

Two questions here:
1. Why this is suppose to be the default behavior now if the issue occurs only when  podAntiAffinity is set?

2. Also it's mentioned that "first-delete-pod" behavior was already applied for when podAntiAffinity is set, so how the bug occurred on compact cluster?

PR #4069, opened 2026-05-18
[release-4.22] add feature introspection topic for supported versions

rshemtov13:
Register the introspection feature in the active topics list so that the skip list mechanism can gate the introspection tests on operator versions that predate the feature (< 4.21). So we need to backport this til 4.21 where it was first introduced.

(cherry picked from commit 32eb1f94578e775b38ac4f0037b1cc27f06b0043)


rshemtov13:
manual backport for release-4.21 #4082

PR #4083, opened 2026-05-19
controllers: rework status writes

ffromani:
we bundle two changes about how controllers write back their status to the apiserver:
- switch from Update() to Patch(): we can reuse more controller-runtime/client-go primitives and shed some equivalent but worse application code
- make sure to not lose status updates: if we fail to update the status subresource, we need to requeue the operation. Supercedes #3992


Tal-or (approved):
/approve
/lgtm

Looks really nice. we may consider (in future PR ofc) to switch to `Patch()` also for the resources themselves and not only the status.

Following up https://github.com/openshift-kni/numaresources-operator/pull/1502 which intended to minimize API server load and CPU churn

PR #4084, opened 2026-05-19
Revert "OCPBUGS-79418: ctrl: sched: add PodAntiAffinity to deployment"

shajmakh:
Reverts openshift-kni/numaresources-operator#4021

PR #4086, opened 2026-05-19
Revert "ctrl: sched: add PodAntiAffinity to deployment"

shajmakh:
This reverts commit b4e7fd1ab85aea2e8a57383f5e5882361eaba20e.

The approach presented in the merged commit causes a regression in a scenario where PodAntiAffinity with default deployment rollout strategy are conflicting.
The scenario is :
- Scheduler uses replicas autodetection mode
- manually update the replicas to a number that is >= autodetected count
- new pods will be pending forever


shajmakh:
/cc @ffromani @yanirq 
replaces https://github.com/openshift-kni/numaresources-operator/pull/4084

shajmakh:
closing for now as the main branch is not going to be released any soon. I'll handle the fix in a later PR.

PR #4177, opened 2026-05-25
OCPBUGS-79418: ctrl: sched: add topologySpreadConstraints to deployment

shajmakh:
To evenly spread the pods across the nodes in  a balanced way taking
into account the new replicaset pods while rollout.

previous discussions:
(reverted) https://github.com/openshift-kni/numaresources-operator/pull/4021
(cancelled due to regressions) https://github.com/openshift-kni/numaresources-operator/pull/4028


shajmakh:
requires https://github.com/openshift-kni/numaresources-operator/pull/4193

pkg/objectupdate/sched/sched.go
```diff
@@ -102,6 +103,29 @@ func DeploymentTLSSettings(dp *appsv1.Deployment, tlsSettings objtls.Settings) e
...
+			MaxSkew:           1,
+			TopologyKey:       "kubernetes.io/hostname",
+			WhenUnsatisfiable: corev1.DoNotSchedule,
+			// this is needed for safe rollouts to allow ignoring the old replicaset and calculates the spread
+			// purely based on the new replicaset, ensuring the final state is perfectly balanced without stalling.
+			MatchLabelKeys: []string{"pod-template-hash"},
+		},
+	}
+
+	klog.V(3).InfoS("scheduler deployment topology spread constraints", "constraints", dp.Spec.Template.Spec.TopologySpreadConstraints[0].String())
```
ffromani: NIT:
good. Better: since we know we will add just and only 1 spread constraint, this is one of the cases on which we can use a temp. var:
```
    schedConstr := corev1.TopologySpreadConstraint{
         {
			LabelSelector: &metav1.LabelSelector{
				MatchLabels: labels,
			},
			MaxSkew:           1,
			TopologyKey:       "kubernetes.io/hostname",
			WhenUnsatisfiable: corev1.DoNotSchedule,
			// this is needed for safe rollouts to allow ignoring the old replicaset and calculates the spread
			// purely based on the new replicaset, ensuring the final state is perfectly balanced without stalling.
			MatchLabelKeys: []string{"pod-template-hash"},
	}

    dp.Spec.Template.Spec.TopologySpreadConstraints = []corev1.TopologySpreadConstraint{schedConstr}
	klog.V(3).InfoS("scheduler deployment topology spread constraints", "constraints", schedConstr.String())
```

(note the example won't probably compile ad I sketched out live on the github UI)
please do not resubmit just for this

pkg/objectupdate/sched/sched_test.go
```diff
@@ -631,6 +631,38 @@ func TestDeploymentTLSSettingsRepeated(t *testing.T) {
...
+			LabelSelector: &metav1.LabelSelector{
+				MatchLabels: dp.Spec.Template.Labels,
+			},
+			MaxSkew:           1,
+			TopologyKey:       "kubernetes.io/hostname",
+			WhenUnsatisfiable: corev1.DoNotSchedule,
+			MatchLabelKeys:    []string{"pod-template-hash"},
+		},
+	}
+	if !reflect.DeepEqual(dp.Spec.Template.Spec.TopologySpreadConstraints, expectedConstraints) {
```
ffromani: nit: `cmp.Diff` is another good option here.

ffromani (commented):
Good work. Do we have a test which actually verify the happy path? 3 control-plane nodes, check we autodetect 3 replicas and each one lands on a different node.

shajmakh:
> Good work. Do we have a test which actually verify the happy path? 3 control-plane nodes, check we autodetect 3 replicas and each one lands on a different node.

we have that as a step in one of the tests:
https://github.com/openshift-kni/numaresources-operator/pull/4177/changes/b3da3928d049f9ead654b014d47f2e525b831373#diff-6e473fac5187d020f14665d506a6553280fc9db1866488805157ada1a93e82b0R479

pkg/objectupdate/sched/sched.go
shajmakh: Looks cleaner like that. Thanks for highlighting. I'm resubmitting anyway to address the AI comment

pkg/objectupdate/sched/sched.go
```diff
@@ -102,6 +103,28 @@ func DeploymentTLSSettings(dp *appsv1.Deployment, tlsSettings objtls.Settings) e
...
+	schedConstr := corev1.TopologySpreadConstraint{
+		LabelSelector: &metav1.LabelSelector{
+			MatchLabels: labels,
+		},
+		MaxSkew:           1,
+		TopologyKey:       "kubernetes.io/hostname",
+		WhenUnsatisfiable: corev1.DoNotSchedule,
+		// this is needed for safe rollouts to allow ignoring the old replicaset and calculates the spread
+		// purely based on the new replicaset, ensuring the final state is perfectly balanced without stalling.
+		MatchLabelKeys: []string{"pod-template-hash"},
```
Tal-or, line 123: is there a special meaning for the `pod-template-hash` string?

Tal-or (commented):
LGTM overall, left minor comment

pkg/objectupdate/sched/sched.go
shajmakh: yes, it's a label that's set by the deployment controller https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#pod-template-hash-label

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -1030,6 +1030,28 @@ var _ = Describe("Test NUMAResourcesScheduler Reconcile", func() {
 			Expect(dp.Spec.Template.Spec.Containers[0].Args).To(ContainElement("--tls-min-version=" + updatedSettings.MinVersion))
 			Expect(dp.Spec.Template.Spec.Containers[0].Args).To(ContainElement("--tls-cipher-suites=" + updatedSettings.CipherSuites))
 		})
+
+		It("should set the TopologySpreadConstraints in the deployment by default", func() {
+			_, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: nrsKey})
```
ffromani: non-blocking: use closures with context in new code:
```
		It("should set the TopologySpreadConstraints in the deployment by default", func(ctx context Context) {
			_, err := reconciler.Reconcile(context.TODO(), reconcile.Request{NamespacedName: nrsKey})
```

pkg/objectupdate/sched/sched_test.go
```diff
@@ -631,6 +633,43 @@ func TestDeploymentTLSSettingsRepeated(t *testing.T) {
...
+			},
+			MaxSkew:           1,
+			TopologyKey:       "kubernetes.io/hostname",
+			WhenUnsatisfiable: corev1.DoNotSchedule,
+			MatchLabelKeys:    []string{"pod-template-hash"},
+		},
+	}
+	if cmp.Diff(dp.Spec.Template.Spec.TopologySpreadConstraints, expectedConstraints) != "" {
+		t.Errorf("constraints mismatch\ngot:      %v\nexpected: %v", dp.Spec.Template.Spec.TopologySpreadConstraints, expectedConstraints)
+	}
```
ffromani: good. Better:
```
	if diff := cmp.Diff(dp.Spec.Template.Spec.TopologySpreadConstraints, expectedConstraints); diff != "" {
		t.Errorf("constraints mismatch\n%s", diff)
	}
```

test/e2e/sched/sched_test.go
```diff
@@ -273,4 +277,315 @@ var _ = Describe("[Scheduler] imageReplacement", func() {
...
+				},
+			}
+
+			By("verify spread constraints are set by default")
+			Expect(schedDp.Spec.Template.Spec.TopologySpreadConstraints).To(Equal(expectedTopologySpreadConstraints), "topology spread constraints mismatch")
+
+			autoDetectedReplicasCount = schedDp.Spec.Replicas
+			Expect(autoDetectedReplicasCount).ToNot(BeNil()) // must never happen
+		})
+
```
ffromani, line 372: let's add here a simple e2e happy path test and we're good to merge:
```
  It("should have each replica running on a different node", func(ctx context.Context) {
        pods, err := podlist.With(e2eclient.Client).ByDeployment(ctx, *schedDp)
        Expect(pods).To(HaveLen(int(*autoDetectedReplicasCount)))

        nodeNames := sets.New[string]()
        for _, pod := range pods {
                Expect(pod.Status.Phase).To(Equal(corev1.PodRunning))
                nodeNames.Insert(pod.Spec.NodeName)
        }
        Expect(nodeNames).To(HaveLen(len(pods)), "expected each replica on a different node")
  })
```

ffromani (commented):
let's add one more e2e test (suggested inline) and we can merge

pkg/objectupdate/sched/sched.go
Tal-or: then lets add this reference in the comment

shajmakh:
@ffromani @Tal-or Thank you for your reviews. I've addressed your comments and the CI flake is fixed now, please let me know if you have any concerns with the current version.

test/e2e/serial/tests/workload_placement.go
```diff
@@ -283,7 +283,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload placeme
 			dpReplicas, err = podlist.With(fxt.Client).ReplicaSetByDeployment(context.TODO(), *updatedDp)
 			Expect(err).ToNot(HaveOccurred())
 
-			updatedDp, err = wait.With(fxt.Client).Interval(10*time.Second).Timeout(time.Minute).ForDeploymentComplete(context.TODO(), dp)
+			updatedDp, err = wait.With(fxt.Client).Interval(10*time.Second).Timeout(time.Minute).ForDeploymentCompleteWithReplicas(context.TODO(), dp, *dp.Spec.Replicas)
```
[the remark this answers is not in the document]
shajmakh: we should keep the check against the old dp replicas which is what the test is expecting.

internal/wait/deployment.go
```diff
@@ -51,14 +51,28 @@ func (wt Waiter) ForDeploymentComplete(ctx context.Context, dp *appsv1.Deploymen
...
 
 func IsDeploymentComplete(dp *appsv1.Deployment, newStatus *appsv1.DeploymentStatus) bool {
-	return areDeploymentReplicasAvailable(newStatus, *(dp.Spec.Replicas)) &&
-		newStatus.ObservedGeneration >= dp.Generation
+	if newStatus.ObservedGeneration < dp.Generation {
+		klog.InfoS("generation is older than expected to indicate the deployment is complete", "old", dp.Generation, "new", newStatus.ObservedGeneration)
+		return false
+	}
+
+	return areDeploymentReplicasAvailable(newStatus, *(dp.Spec.Replicas))
```
ffromani: This is a bit heavy handed. Currently this is called only in test code, but the moment we pull in production code we will have log spam. There are options, but all require more work to fix. Can we simplify this commit or drop it entirely?

test/e2e/sched/sched_test.go
```diff
@@ -625,9 +625,11 @@ var _ = Describe("[Scheduler] scheduler object updates", func() {
 func infoUponDeploymentCompleteFailure(ctx context.Context, schedDp *appsv1.Deployment) {
 	klog.InfoS("deployment status", "status", schedDp.Status)
 	pods, err := podlist.With(e2eclient.Client).ByDeployment(ctx, *schedDp)
-	if err != nil {
+	if err == nil {
```
ffromani: this seems a logic change worth explaining (and keeping), not log improvements.

internal/wait/deployment.go
```diff
@@ -24,11 +24,8 @@ import (
 	"k8s.io/klog/v2"
 )
 
-func (wt Waiter) ForDeploymentComplete(ctx context.Context, dp *appsv1.Deployment) (*appsv1.Deployment, error) {
-	// This function waits for the readiness of the pods under the deployment. The best use of this check is for
-	// completely new deployments. If the deployment exists on the cluster and simply updated, this check is
-	// not enough to guarantee that the deployment is ready with the NEW replica, thus need to cover that by
-	// additional checks as the context requires
+// ForDeploymentCompleteWithReplicas waits for the deployment to be complete and have the specified number of replicas.
+func (wt Waiter) ForDeploymentCompleteWithReplicas(ctx context.Context, dp *appsv1.Deployment, newExpectedReplicas int32) (*appsv1.Deployment, error) {
```
ffromani, line 34: good addition. If most of the codebase actually uses the pattern
```
ForDeploymentCompleteWithReplicas(ctx, dp, *dp.Spec.Replicas)
```

and at glance it seems the case, it's worth keeping the old API `ForDeploymentComplete`

internal/wait/deployment.go
```diff
@@ -66,13 +63,13 @@ func areDeploymentReplicasAvailable(newStatus *appsv1.DeploymentStatus, replicas
 	return true
 }
 
-func IsDeploymentComplete(dp *appsv1.Deployment, newStatus *appsv1.DeploymentStatus) bool {
-	if newStatus.ObservedGeneration < dp.Generation {
-		klog.InfoS("generation is older than expected to indicate the deployment is complete", "old", dp.Generation, "new", newStatus.ObservedGeneration)
+func IsDeploymentComplete(oldGeneration int64, newStatus *appsv1.DeploymentStatus, expectedReplicas int32) bool {
```
ffromani, line 72: looking at the diff an explicit `oldGeneration` seems unnecessary. What's the motivation for it?

shajmakh: it is needed to ensure that the new deployment generation is updated and that the check is not checking against an old version of deployment

ffromani: If I'm not mistaken, we call the function always with dp.Generation, which would make the change hard to understand (different API, same end result). Is that the case?

internal/wait/deployment.go
shajmakh: I intentionally updated the codebase because ForDeploymentComplete without having the correct expected replicas count does not sound complete and can be confusing. Internally in the function, replicas check is a core one and in order to avoid using the incorrect function for deployment waiter I overrode the old API. This was actually an old bug as noted in the omitted comment; the new version is expected to handle all deployments now.

test/e2e/sched/sched_test.go
shajmakh: this was a mistake I should have pushed in the first commit introducing the helper function. will fix that.

internal/wait/deployment.go
ffromani: ok, so let's just rename `ForDeploymentCompleteWithReplicas` to `ForDeploymentComplete`

internal/wait/deployment.go
shajmakh: this was needed for debugging, it essentially describes why the functions fail which is why we have the last fixing [commit](https://github.com/openshift-kni/numaresources-operator/pull/4177/changes/ef1ff13fc0af69a9e3df1f1a8955316d7399a6ba) so this is very helpful details for troubleshooting the tests. 
If ever we decide to use that in production, we need to be aware of the old use which is initially meant for tests. we can anyway create a production version and improve it then as we see fit. 
Thinking about the backport, I think it's better to keep the old function `ForDeploymentComplete` to reduce any tests conflicts which will then simplify the commit a bit in terms of the adaptations across the codebase. would that sound good enough?

ffromani: discouraging production use could be as simple as move this code in a subtree of `test/` perhaps `test/internal/wait/...` but I won't do in this PR.

A way forward would be post a followup PR with first the trivial move and then adding the logs

internal/wait/deployment.go
shajmakh: in the old version we needed dp to get more than one field, however now that we know that the expected replicas count would not nessecarly be the same one as in `dp` , we pass that explicitly hence `expectedReplicas`. since we are left with one field needed from dp, then it's enough to pass only that specific field. the end result of the function has changed because expected replicas is no longer derived from  the "old" deployment.

ffromani: ok, makes sense

internal/wait/deployment.go
ffromani: actually let's move forward and let's revisit later

ffromani (commented):
/approve
/lgtm

we can defer the remaining comments to followups

shajmakh:
/hold
for updates to make the automated backport process smoother

internal/wait/deployment.go
shajmakh: in the latest push I restored ForDeploymentComplete to keep the diff as less as possible to reduce conflicts while backporting. I can address  changing the whole code base in a followup PR including https://github.com/openshift-kni/numaresources-operator/pull/4177#discussion_r3317478289

ffromani (commented):
/lgtm

PR #4227, opened 2026-05-27
OCPBUGS-12988: force degraded status if ExporterImage is set

ffromani:
ExporterImage is an escape hatch we had from day0 but which we never really leveraged. It made on the stable API and we can't deprecate it anytime soon, but we can at least narrow down the perimeter and make it obvious and loud it should not be used.
Force the status to Degraded if the ExporterImage is set by the user.


ffromani:
replaced by https://github.com/openshift-kni/numaresources-operator/pull/4592

PR #4327, opened 2026-06-03
fix u/s ci issues

shajmakh:
see details in commits.


ffromani:
/approve
/lgtm

the security failure is real but irrelevant real and we should tackle it (if at all) in a separate PR

shajmakh:
> /approve /lgtm
> 
> the security failure is real but irrelevant real and we should tackle it (if at all) in a separate PR

@ffromani thanks for checking this. The govulcheck is handled in https://github.com/openshift-kni/numaresources-operator/actions/runs/26875953209/job/79263122051?pr=4328.

ffromani:
merging https://github.com/openshift-kni/numaresources-operator/pull/4311

PR #4328, opened 2026-06-03
govulcheck: use aligned go versions

shajmakh:
Use aligned go version in setup and vulnerability check. ref:
https://github.com/golang/govulncheck-action/blob/master/action.yml


ffromani:
/approve
/lgtm

thanks!

PR #4338, opened 2026-06-09
docs: AI: consolidate agents instructions in AGENTS.md

ffromani:
instead of having custom or tool-specific directives, consolidate the recommendations in the generic `AGENTS.md`

Clarifications and content movement.


AGENTS.md
```diff
@@ -0,0 +1,25 @@
+# AGENTS.md
+
```
shajmakh, line 2: it would be helpful for history tracking to know why AGENTS.md would work instead of custom agent-specific file. maybe update the commit message to include https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/add-custom-instructions/add-repository-instructions#creating-custom-instructions?

AGENTS.md
```diff
@@ -0,0 +1,25 @@
...
+- Run `make help` to learn about Makefile targets
+- Always prefer Makefile targets over raw `go` commands: use `make test-unit` not `go test ./pkg/...`
+- To run controller tests, use `make test-controllers`
+
+## Dependencies
+
+- Always prefer standard library over third party packages.
+- Look first in the kubernetes libraries or kubernetes ecosystem libraries (e.g. sigs.k8s.io/...) before
+  pulling generic third party libraries.
```
shajmakh, line 13: I would guess that the agents are smart enough to infer that "reuse existing dependencies before pulling new ones, if the existing one can fulfill the requirement.

AGENTS.md
```diff
@@ -0,0 +1,25 @@
...
+- Always minimize external dependencies. If the dependency is small *and* well understood/common,
+  err towards reimplementing leveraging standard library or kubernetes or kubernetes-adjacent libraries
+
+## Development patterns
+
```
shajmakh, line 18: I personally usually ask something like this from the agent: 
"Concerning the current task, do not make assumptions if there is something that isn't clear enough; instead, ask the developer to confirm."
This will likely limit the agent's ability, but OTOH will keep the human involved more in the loop. wdyt?

shajmakh (commented):
LGTM, thanks for addressing. I left a couple of comments, let me know what we prefer.

AGENTS.md
ffromani: That was not always the case in my experience, even with the recent SOTA models I got access to.

AGENTS.md
ffromani: Seems like a good idea and matches the expectations, I can add this to the doc.

ffromani:
requires https://github.com/openshift-kni/numaresources-operator/pull/4351

AGENTS.md
shajmakh: I see, we can add that too in that case

PR #4351, opened 2026-06-12
chore: mco: fix CRD path

ffromani:
the MCO CRD we depended on was moved in openshift/api@b479107f81ea

Use the replacement accordingly.


ffromani:
if we need backports, we will need branch-specific backports - pinning the SHA version of the CRD for example

shajmakh:
Thanks for taking this on, LGTM. Just one clarification for future reference: SelfManagedHA is not the 1:1 replacement, but it is preferred over Hypershift_Default because the latter is still being ironed out, as per the referenced commit. 
I think CI will confirm this. 
/lgtm

PR #3335, opened 2026-03-24
manifests: add annotation to link must-gather image

ffromani:
Include the must gather (informative only ) annotation, so automation (and humans) have a way to learn about the soft dependency.

PR #4353, opened 2026-06-14
Revert "manifests: add annotation to link must-gather image"

yanirq:
Reverts #3335, which added the `operators.openshift.io/must-gather-image` annotation to the ClusterServiceVersion manifests.

Konflux validations on the bundle image will fail for non existing images. This can be reinstated once 5.0 is GA


shajmakh:
/lgtm
Thanks for handling, do we have any card to track this for 5.0 so it doesn't slip without updating?

PR #4446, opened 2026-06-21
[release-4.22] hack: use release-4.22 openshift/api branch for MCO CRD URLs

Tal-or:
The CRD URLs were hardcoded to fetch from the master branch of openshift/api which may have incompatible schema changes. Point to the matching release-4.22 branch instead.

It is possible to detect the branch from within the script, but I found it fragile.

Another option is passing the branch as an argument, but that involves with many changes around the callers in the code base and it does NOT guaranteed that the URL structure will not change again in the future.

### DIFF
--- a/hack/deploy-mco-crds.sh
+++ b/hack/deploy-mco-crds.sh
@@ -3,9 +3,9 @@
 source hack/common.sh
 
 # Specify the URL link to the machine config pool CRD
-CRD_MACHINE_CONFIG_POOL_URL="https://raw.githubusercontent.com/openshift/api/refs/heads/master/payload-manifests/crds/0000_80_machine-config_01_machineconfigpools-Default.crd.yaml"
+CRD_MACHINE_CONFIG_POOL_URL="https://raw.githubusercontent.com/openshift/api/refs/heads/release-4.22/payload-manifests/crds/0000_80_machine-config_01_machineconfigpools-Default.crd.yaml"
 # Specify the URL link to the kubeletconfig CRD
-CRD_KUBELET_CONFIG_URL="https://raw.githubusercontent.com/openshift/api/refs/heads/master/payload-manifests/crds/0000_80_machine-config_01_kubeletconfigs.crd.yaml"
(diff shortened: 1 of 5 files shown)


Tal-or:
Need to be backport all the way to 4.18
CC: @yanirq

hack/deploy-mco-crds.sh
```diff
@@ -3,9 +3,9 @@
 source hack/common.sh
 
 # Specify the URL link to the machine config pool CRD
-CRD_MACHINE_CONFIG_POOL_URL="https://raw.githubusercontent.com/openshift/api/refs/heads/master/payload-manifests/crds/0000_80_machine-config_01_machineconfigpools-Default.crd.yaml"
+CRD_MACHINE_CONFIG_POOL_URL="https://raw.githubusercontent.com/openshift/api/refs/heads/release-4.22/payload-manifests/crds/0000_80_machine-config_01_machineconfigpools-Default.crd.yaml"
```
yanirq, line 6: this was always master for all versions until now.  We should start pinning the versions ?

yanirq: If we do need to pin versions then we need to check the filename for each one

.github/workflows/govulncheck.yaml
```diff
@@ -26,7 +30,7 @@ jobs:
       - name: Set up golang
         uses: actions/setup-go@v5
         with:
-          go-version: 1.25.8
+          go-version: ${{ env.GOLANG_VERSION }}
```
ffromani: we should use the Y version, there's no real reason to pin to a specific Z version

hack/deploy-mco-crds.sh
ffromani: main branch should be floating, release branches should better be pinned to the corresponding release branch

.github/workflows/govulncheck.yaml
Tal-or: it was a test to see if changing the Golang version would help the test to pass

Tal-or:
depends on: https://github.com/openshift-kni/numaresources-operator/pull/4456

ffromani:
/approve
/lgtm

thanks

PR #4485, opened 2026-06-24
status: streamline reporting

ffromani:
our status update logic may determine there's no need to do a status update. This is totally expected and benign,
but wasn't conveyed properly in the log which sound unnecessarily scary. Fix the log to reflect reality, and make it visible only at debug log levels.

In addition, fix `Reason` on `Progressing` conditions. Failing to report reason will cause an apiserver reject on status update.


Tal-or (approved):
/lgtm
/approve

PR #4487, opened 2026-06-24
e2e: serial: tier list review

ffromani:
the idea is to keep the tier0 as small and focused as possible so it can used as highest ROI smoke test. The threshold is roughly around 24-25 testcase. Pushing down to 20 would be ideal, but also harder.


rshemtov13:
@ffromani Completely agree with your changes here. Let's see how this goes for sanity, and we'll check the tier0 success rate across all supported versions then we can adjust if needed.

PR #4568, opened 2026-06-30
api: fix PodsFingerprinting default doc in v1

SargunNarula:
PodsFingerprintingEnabled was documented as the default, but
`SetDefaults()` uses `EnabledExclusiveResources`. Update the const
comments to match `numaresourcesoperator_defaults.go`.


ffromani:
/approve
/lgtm

thanks

PR #4579, opened 2026-07-01
e2e: serial: add priority-based placement tests

shajmakh:
Add e2e tests to cover the NRT scheduler preemption support.

### DIFF
--- a/test/e2e/serial/tests/priority-based-placement.go
+++ b/test/e2e/serial/tests/priority-based-placement.go
@@ -1,0 +1,536 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package tests
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"golang.org/x/sync/errgroup"
+
+	corev1 "k8s.io/api/core/v1"
+	schedulingv1 "k8s.io/api/scheduling/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	nrtv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
+
+	intbaseload "github.com/openshift-kni/numaresources-operator/internal/baseload"
+	intnrt "github.com/openshift-kni/numaresources-operator/internal/noderesourcetopology"
+	"github.com/openshift-kni/numaresources-operator/internal/wait"
+	numacellapi "github.com/openshift-kni/numaresources-operator/test/deviceplugin/pkg/numacell/api"
+	"github.com/openshift-kni/numaresources-operator/test/e2e/label"
+	serialconfig "github.com/openshift-kni/numaresources-operator/test/e2e/serial/config"
+	e2efixture "github.com/openshift-kni/numaresources-operator/test/internal/fixture"
+	e2enrt "github.com/openshift-kni/numaresources-operator/test/internal/noderesourcetopologies"
+	"github.com/openshift-kni/numaresources-operator/test/internal/objects"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	systemNodeCriticalPriorityClassName = "system-node-critical"
+	customLowPriorityClassName          = "e2e-preemption-low"
+	customMediumPriorityClassName       = "e2e-preemption-medium"
+	customHighPriorityClassName         = "e2e-preemption-high"
+
+	pendingTimeout = 5 * time.Minute
+)
+
+var _ = Describe("[serial][disruptive][preemption] priority-based workload placement functionality", Serial, Label("disruptive", "scheduler"), Label("feature:preemption"), func() {
+	var fxt *e2efixture.Fixture
+	var nrtList nrtv1alpha2.NodeResourceTopologyList
+
+	BeforeEach(func(ctx context.Context) {
+		Expect(serialconfig.Config).ToNot(BeNil())
+		Expect(serialconfig.Config.Ready()).To(BeTrue(), "NUMA fixture initialization failed")
+
+		var err error
+		fxt, err = e2efixture.Setup("e2e-test-preemption", serialconfig.Config.NRTList)
+		Expect(err).ToNot(HaveOccurred(), "unable to setup test fixture")
+
+		Expect(fxt.Client.List(ctx, &nrtList)).ToNot(HaveOccurred())
+
+		By("creating custom priority classes")
+		Expect(fxt.Client.Create(ctx, &schedulingv1.PriorityClass{
+			ObjectMeta: metav1.ObjectMeta{Name: customLowPriorityClassName},
+			Value:      100,
+		})).To(Succeed(), "cannot create low priority class %q", customLowPriorityClassName)
+
+		Expect(fxt.Client.Create(ctx, &schedulingv1.PriorityClass{
+			ObjectMeta: metav1.ObjectMeta{Name: customMediumPriorityClassName},
+			Value:      200,
+		})).To(Succeed(), "cannot create medium priority class %q", customMediumPriorityClassName)
+
+		Expect(fxt.Client.Create(ctx, &schedulingv1.PriorityClass{
+			ObjectMeta: metav1.ObjectMeta{Name: customHighPriorityClassName},
+			Value:      300,
+		})).To(Succeed(), "cannot create high priority class %q", customHighPriorityClassName)
+	})
+
+	AfterEach(func(ctx context.Context) {
+		By("deleting custom priority classes")
+		for _, pcName := range []string{customLowPriorityClassName, customMediumPriorityClassName, customHighPriorityClassName} {
+			if pcName == "" {
+				continue
+			}
+			pc := &schedulingv1.PriorityClass{}
+			err := fxt.Client.Get(ctx, client.ObjectKey{Name: pcName}, pc)
+			if apierrors.IsNotFound(err) {
+				continue
+			}
+			Expect(err).ToNot(HaveOccurred(), "cannot get priority class %q", pcName)
+			Expect(fxt.Client.Delete(ctx, pc)).To(Succeed(), "cannot delete priority class %q", pcName)
+		}
+
+		Expect(e2efixture.Teardown(fxt)).ToNot(HaveOccurred())
+	})
+
+	When("one node is schedulable and the others are unschedulable", func() {
+		var (
+			targetNodeName string
+			nrtCandidates  []nrtv1alpha2.NodeResourceTopology
+		)
+		type testCase struct {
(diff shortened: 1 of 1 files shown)


ffromani:
what happens if we run a patched NROP with a scheduler image which does NOT support preFilter?

shajmakh:
> what happens if we run a patched NROP with a scheduler image which does NOT support preFilter?

spot on; the scheduler fails:
`Error: initializing profiles: creating profile for scheduler name topo-aware-scheduler: plugin "NodeResourceTopologyMatch" does not extend PreFilterPlugin plugin`

For GA releases that support preemption since day-1 there should not be a problem because the general recommendation is to use aligned versions of operator and scheduler. For zstreams however (in case of backporting the feature), this needs to be documented to use the scheduler image with the preFilter implementation.

shajmakh:
> > what happens if we run a patched NROP with a scheduler image which does NOT support preFilter?
> 
> spot on; the scheduler fails: `Error: initializing profiles: creating profile for scheduler name topo-aware-scheduler: plugin "NodeResourceTopologyMatch" does not extend PreFilterPlugin plugin`
> 
> For GA releases that support preemption since day-1 there should not be a problem because the general recommendation is to use aligned versions of operator and scheduler. For zstreams however (in case of backporting the feature), this needs to be documented to use the scheduler image with the preFilter implementation.

in fact, there is a way to manage that error silently by updating the `multipoint` instead, so if the scheduler image implements the plugin, it will be picked up, otherwise it will be skipped. will update this and run some tests.

vendor/github.com/k8stopologyawareschedwg/numaplacement/leb89/leb89.go
```diff
@@ -0,0 +1,102 @@
...
+			if nCont == 0 {
+				return int32(idx), pos
+			}
+			return int32(NumTerminal) + contValue*int32(NumTerminal) + int32(idx), pos
+		}
+		contValue = contValue*int32(NumContinuation) + int32(idx-NumTerminal)
+		nCont++
+	}
+	return Unmapped, pos
+}
```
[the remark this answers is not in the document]
shajmakh: will address as dependency PR

shajmakh:
ci requires: https://github.com/openshift-kni/numaresources-operator/pull/4784

shajmakh:
infra issue
/retest

test/e2e/serial/tests/priority-based-placement.go
```diff
@@ -0,0 +1,617 @@
...
+				// error should never happen by now
+				Expect(err).ToNot(HaveOccurred(), "no NRT data for target node %q", targetNodeName)
+
+				nrtWithTMContainerScope := e2enrt.FilterByTopologyManagerScope([]nrtv1alpha2.NodeResourceTopology{*targetNRT}, intnrt.Container)
+				if len(nrtWithTMContainerScope) < 1 {
+					e2efixture.Skipf(fxt, "not enough nodes with Container scope - found %d", len(nrtWithTMContainerScope))
+				}
+			})
+
+			It("should remain pending when the new pod is not priority enough despite the node-level sufficient space", func(ctx context.Context) {
```
rshemtov13: We can just use the ginkgo label to trigger this when it's via container scope right?

test/e2e/serial/tests/priority-based-placement.go
```diff
@@ -0,0 +1,617 @@
...
+)
+
+const (
+	systemNodeCriticalPriorityClassName = "system-node-critical"
+	customLowPriorityClassName          = "e2e-preemption-low"
+	customMediumPriorityClassName       = "e2e-preemption-medium"
+	customHighPriorityClassName         = "e2e-preemption-high"
+
+	pendingTimeout   = 5 * time.Minute
+	pendingPollIntvl = 10 * time.Second
```
rshemtov13: This is unused

test/e2e/serial/tests/priority-based-placement.go
```diff
@@ -0,0 +1,617 @@
...
+
+					By("verifying the rest of the filler pods continue to run")
+					Consistently(func(g Gomega) {
+						for _, fp := range fillerPods {
+							if evicted.Has(fp.Name) {
+								continue
+							}
+							pod := &corev1.Pod{}
+							g.Expect(fxt.Client.Get(ctx, client.ObjectKeyFromObject(fp), pod)).To(Succeed())
+							g.Expect(pod.Status.Phase).To(Equal(corev1.PodRunning), "pod %s/%s is not running", testPod.Namespace, testPod.Name)
```
rshemtov13: I believe here you expected to use the ```fp``` and not testPod which here is the preemptor pod right?

test/e2e/serial/tests/priority-based-placement.go
shajmakh: I don't recall how d/s pipeline works WRT filtering tests for different scopes, yet this is here to ensure that if the cluster is configured in a non-matching configuration then the test will be skipped

test/e2e/serial/tests/priority-based-placement.go
shajmakh: correct, the interval const isn't used, thanks for catching.

test/e2e/serial/tests/priority-based-placement.go
shajmakh: the preemptor pod was verified it's running a step ago, in this `Consistently` we are verifying filler pods (`fp`) are still running, so we need to take the object ky from the corresponding `fp`. does that answer your question?

rshemtov13: Yes I see thanks for answering.

test/e2e/serial/tests/priority-based-placement.go
rshemtov13: I see in some places in the suite it's happening in runtime exactly like u have had here. I was mistaking it with something else.

go.mod
Tal-or, line 1: lets add the vendor code on separate commit to make the PR a little bit more organized.

test/e2e/serial/tests/priority-based-placement.go
```diff
@@ -0,0 +1,530 @@
...
+
+			var ok bool
+			targetNodeName, ok = e2efixture.PopNodeName(schedulableNodeNames)
+			Expect(ok).To(BeTrue(), "cannot select target node from candidates")
+			klog.InfoS("selected target node for preemption test", "nodeName", targetNodeName)
+
+			By("cordoning all nodes except the target node")
+			var cordonedNodeNames []string
+			nodeList := &corev1.NodeList{}
+			Expect(fxt.Client.List(ctx, nodeList)).To(Succeed(), "cannot list cluster nodes")
```
Tal-or, line 150: You can consolidate the `Get` above with this call.

The `List` gives all the nodes so we can iterate and patch them at the same time.

test/e2e/serial/tests/priority-based-placement.go
```diff
@@ -0,0 +1,530 @@
...
+				// this will trigger the preemption logic
+				testPodRes := fillerPods[0].Spec.Containers[0].Resources.Limits
+				testPod.Spec.Containers[0].Resources = corev1.ResourceRequirements{
+					Limits:   testPodRes,
+					Requests: testPodRes,
+				}
+
+				Expect(fxt.Client.Create(ctx, testPod)).To(Succeed(), "cannot create test workload pod")
+
+				if tc.evictionExpected {
```
Tal-or, line 228: would it be possible to consolidate the logic within the `if` and `else` into functions so it would be more readable?

```
if tc.evictionExpected  {
     evictionExpectedFunc()
} else {
evictionNotExpectedFunc()
```

test/e2e/serial/tests/priority-based-placement.go
```diff
@@ -0,0 +1,530 @@
...
+				// part of testing the preemption functionality
+				baseload, err := intbaseload.ForNode(fxt.Client, ctx, targetNodeName)
+				Expect(err).ToNot(HaveOccurred(), "cannot get base load for %q", targetNodeName)
+				klog.InfoS("node base load", "nodeName", targetNodeName, "resources", baseload.Resources)
+
+				leastEnoughResourcesPerZone := corev1.ResourceList{}
+				for resName, resQty := range baseload.Resources {
+					// 1 baseload + 2 for default-priority fillers + 4 for test pod; this calculation guarantees that the
+					// eviction will happen to exactly one pod while the rest should be kept running without any evictions.
+					resQty.Mul(7)
```
Tal-or, line 327: That looks fragile, why do we need to multiply each resource by 7? I don't get why "this calculation guarantees that the eviction will happen to exactly one pod" ?

test/e2e/serial/tests/priority-based-placement.go
```diff
@@ -491,7 +486,99 @@ var _ = Describe("[serial][disruptive][preemption] priority-based workload place
...
+
+				klog.Info("ensure all pods are kept running")
+				for _, pod := range pods {
+					Expect(fxt.Client.Get(ctx, client.ObjectKeyFromObject(pod), pod)).To(Succeed(), "cannot get pod %q", pod.Name)
+					Expect(pod.Status.Phase).To(Equal(corev1.PodRunning), "pod %s/%s is not running", pod.Namespace, pod.Name)
+				}
+			})
+		})
 	})
+
```
Tal-or: nit: redundant new line

test/e2e/serial/tests/priority-based-placement.go
shajmakh: no problem

test/e2e/serial/tests/priority-based-placement.go
shajmakh: `Get` and `List` each works on a different set of nodes. `Get` works on nodes with NRT (nrtCandidates) and picks the target node that should be schedulable; while `List` is used to list all the cluster nodes and only keep the target as schedulable, no additional filters matter. 
I simplified the target node selection (random to first node found).

test/e2e/serial/tests/priority-based-placement.go
shajmakh: partially yes, I've added a helper function for unexpected eviction verification only because it is used elsewhere. Extracting a helper for the expected eviction is not practical here for readability because:
1. it will force a jump to the helper and back
2. different tests checks this differently, so better not to generalize for more controlled maintenance 
3. `By()` already says what the next block of code should does, so the helper name does not add a value in this context.

test/e2e/serial/tests/priority-based-placement.go
shajmakh: addressed by adding a desription to the test. I don't see how the test design can be further simplified or split to smaller tests because it depends on having intentional mixed-priority pods. I'd say that the setup of the test has a lot of details but all is captured in the description, thus I believe it should be easier to understand and maintain. please have another look and let me know if it make sense

test/e2e/serial/tests/priority-based-placement.go
```diff
@@ -0,0 +1,627 @@
...
+								// high enough number of pods to trigger a chance of a race and challenge the preemption logic
+								if count < 4 {
+									e2efixture.Skipf(fxt, "not enough available resources in zone %s to fit the test pods: %s", zone.Name, ri.Available.String())
+								}
+								break
+							}
+							count++
+							klog.InfoS("current count calculation", "zone", zone.Name, "resourceName", ri.Name, "currentCount", count)
+							total.Add(perPodCPURequest.DeepCopy())
+						}
```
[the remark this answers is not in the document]
shajmakh: it's never zero.

test/e2e/serial/tests/priority-based-placement.go
Tal-or: the sole reason for consolidating this logic into a function is because it's hard to find the `else` that correlates to the `if` but on modernize IDE that is suppose to be easier I guess.

Tal-or (approved):
/hold
/lgtm
/approve

Thank you.
Letting @rshemtov13 to give the final ack since he's working on the extensive TP for this feature.

rshemtov13:
@shajmakh I believe we are okay to merge, I'll also work on adding the additional test cases.
Let me know if you want to consolidate [this](https://github.com/openshift-kni/numaresources-operator/pull/4579#discussion_r3721778920) before we unhold.
/lgtm

shajmakh:
> @shajmakh I believe we are okay to merge, I'll also work on adding the additional test cases. Let me know if you want to consolidate [this](https://github.com/openshift-kni/numaresources-operator/pull/4579#discussion_r3721778920) before we unhold. /lgtm

it's done: https://github.com/openshift-kni/numaresources-operator/pull/4579/changes#diff-79dfe2f9bff8ea8a7818b43494d5337efff36e65cb9806c2771dd3187799b1cfR260-R262

/unhold

PR #4580, opened 2026-07-02
e2e: skip HyperShift install test until OCPBUGS-88738

SargunNarula:
After openshift/hypershift#8672 (OCPBUGS-86949), HCCO no longer deletes mirrored kubelet ConfigMaps when they are detached from a NodePool. As a result, the operator retains the derived RTE ConfigMap, causing the install test cleanup assertion to fail.

This PR therefore skips all `[Install] durability` specs on HyperShift until the platform fix lands in [hypershift#8890](https://github.com/openshift/hypershift/pull/8890) (OCPBUGS-88738).

### DIFF
--- a/test/e2e/install/install_test.go
+++ b/test/e2e/install/install_test.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"time"
 
+	ginkgotypes "github.com/onsi/ginkgo/v2/types"
+
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


Tal-or (commented):
I'm not sure the fix is correct, because we'll end with stale RTE configMaps on the production as well. 
Fixing that as part of the tests is not enough.
The main concern here is that the operator will refer to the stale RTE configMaps

test/e2e/install/install_test.go
```diff
@@ -64,6 +67,10 @@ import (
 
 const (
 	containerNameRTE = "resource-topology-exporter"
+
+	// skipHyperShiftKCDetachReason documents why HyperShift durability specs are skipped.
+	// Remove once hypershift#8890 (OCPBUGS-88738) is in CI payloads and teardown passes.
+	skipHyperShiftKCDetachReason = "Skipped until OCPBUGS-88738 (hypershift#8890): HCCO does not delete mirrored kubelet ConfigMap on guest cluster after KC detach"
```
ffromani, line 73: in this case the constant is unnecessary, but no big deal

ffromani (commented):
/approve
/lgtm

Tal-or (approved):

PR #4592, opened 2026-07-06
OCPBUGS-94181: fix: disallow custom exporter image by default

ffromani:
Stronger version of https://github.com/openshift-kni/numaresources-operator/pull/4227, which supercedes it.

ExporterImage is an escape hatch we had from day0 but which we never really leveraged. It made on the stable API, but still we can at narrow down the perimeter and make it obvious and loud it should not be used.
Force the status to Degraded if the ExporterImage is set by the user, and refuse to reconcile.

Cluster admins can temporarily restore the old behavior by setting the env var `EXPORTER_IMAGE_RESTRICTION` to `false`.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -399,12 +399,13 @@ func main() {
 			Core:    rteManifestsRendered,
 			Metrics: rteMetricsManifests,
 		},
-		Platform:        discoveredCluster.Platform,
-		Images:          imgs,
-		ImagePullPolicy: pullPolicy,
-		Namespace:       namespace,
-		ForwardMCPConds: params.enableMCPCondsForward,
-		RTEMetricsTLS:   tlsSettings,
+		Platform:            discoveredCluster.Platform,
+		Images:              imgs,
+		ImagePullPolicy:     pullPolicy,
+		Namespace:           namespace,
+		ForwardMCPConds:     params.enableMCPCondsForward,
+		RTEMetricsTLS:       tlsSettings,
+		OverridableRTEImage: controller.IsRTEImageOverridable(),
 	}).SetupWithManager(mgr); err != nil {
 		klog.ErrorS(err, "unable to create controller", "controller", "NUMAResourcesOperator")
 		exitWithCancel(cancel, 1)
--- a/internal/controller/numaresourcesoperator_controller.go
+++ b/internal/controller/numaresourcesoperator_controller.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"os"
 	"reflect"
 	"strings"
 	"time"
@@ -74,6 +75,13 @@ import (
 	"github.com/openshift-kni/numaresources-operator/rte/pkg/config"
 )
 
+const (
+	// ExporterImageRestrictionEnvVar is the environment variable that controls whether the exporter image restriction is enabled.
+	// If it is set to "false", the exporter image restriction is disabled; any other value is considered as enabled.
+	// Remove at after 5.4, preferably at 5.2.
+	ExporterImageRestrictionEnvVar = "EXPORTER_IMAGE_RESTRICTION"
+)
+
 const numaResourcesRetryPeriod = 1 * time.Minute
 
 // poolDaemonSet a struct to hold the target MCP of a configured node group and its created respective RTE daemonset
@@ -85,15 +93,16 @@ type poolDaemonSet struct {
 // NUMAResourcesOperatorReconciler reconciles a NUMAResourcesOperator object
 type NUMAResourcesOperatorReconciler struct {
 	client.Client
-	Scheme          *runtime.Scheme
-	Platform        platform.Platform
-	APIManifests    apimanifests.Manifests
-	RTEManifests    rtestate.Manifests
-	Namespace       string
-	Images          images.Data
-	ImagePullPolicy corev1.PullPolicy
-	Recorder        record.EventRecorder
-	ForwardMCPConds bool
+	Scheme              *runtime.Scheme
+	Platform            platform.Platform
+	APIManifests        apimanifests.Manifests
+	RTEManifests        rtestate.Manifests
+	Namespace           string
+	Images              images.Data
+	ImagePullPolicy     corev1.PullPolicy
+	Recorder            record.EventRecorder
+	ForwardMCPConds     bool
+	OverridableRTEImage bool
 	// RTEMetricsTLS is derived from apiserver.config.openshift.io/cluster TLS profile at operator
 	// startup (same source as the operator metrics server and scheduler).
 	RTEMetricsTLS objtls.Settings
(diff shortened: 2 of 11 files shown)


shajmakh:
/lgtm
Thanks!
When would we be removing the env-var option? maybe worth adding a comment to remember to cleanup the logic.
plus need to fix the CI

ffromani:
> /lgtm Thanks! When would we be removing the env-var option? maybe worth adding a comment to remember to cleanup the logic. plus need to fix the CI

yes, need to add a note about this and also address a couple good points raised by coderabbit

ffromani:
some e2e failures are expected. Working on them.

cmd/main.go
```diff
@@ -97,6 +97,13 @@ const (
 	defaultEnableLeaderElection = true
 )
 
+const (
+	// exporterImageRestrictionEnvVar is the environment variable that controls whether the exporter image restriction is enabled.
+	// If it is set to "false", the exporter image restriction is disabled; any other value is considered as enabled.
+	// Remove at after 5.4, preferably at 5.2.
+	exporterImageRestrictionEnvVar = "EXPORTER_IMAGE_RESTRICTION"
```
Tal-or: We also have annotations that we're planning to deprecate in future releases. maybe it would be better to consolidate them all together in a central fie.

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -164,6 +165,11 @@ func (r *NUMAResourcesOperatorReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, nil
 	}
 
+	if !r.UnrestrictRTEImage && instance.Spec.ExporterImage != "" {
+		err := fmt.Errorf("custom operand image is not supported: %s", instance.Spec.ExporterImage)
```
Tal-or, line 179: Although we don't want to encourage the a custom exporter image usage, from UX perspective it's more correct to expand the error message to include what should be done to enable the support, i.e. inject the `exporterImageRestrictionEnvVar`

cmd/main.go
ffromani: true, but mixing env vars and annotations doesn't seem great. A common abstraction/grab bag is not a great fit either. I'd rather keep the env var in `cmd/main.go` lacking better place for now

internal/controller/numaresourcesoperator_controller.go
ffromani: I was thinking about different personas here, but actually (and we need to better document that) only cluster-admin can both edit the CRs and set the env vars, so yes, worth adding a note

Tal-or (approved):
/lgtm
/approve

ffromani:
rebased to fix conflicts

Tal-or (approved):
/lgtm

Tal-or:
looks like an actual failure in ci/prow/ci-install-e2e
```
• [FAILED] [0.361 seconds]
[Install] durability with a running cluster with all the components and overall deployment [BeforeEach] should be able to delete NUMAResourceOperator CR and redeploy without polluting cluster state [Serial]
  [BeforeEach] /go/src/github.com/openshift-kni/numaresources-operator/test/e2e/install/install_test.go:182
  [It] /go/src/github.com/openshift-kni/numaresources-operator/test/e2e/install/install_test.go:381
  [FAILED] Expected success, but got an error:
      <*errors.StatusError | 0xc0004b8140>: 
      kubeletconfigs.machineconfiguration.openshift.io "kc-test" already exists
      {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "kubeletconfigs.machineconfiguration.openshift.io \"kc-test\" already exists",
              Reason: "AlreadyExists",
              Details: {
                  Name: "kc-test",
                  Group: "machineconfiguration.openshift.io",
                  Kind: "kubeletconfigs",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 409,
          },
      }
  In [BeforeEach] at: /go/src/github.com/openshift-kni/numaresources-operator/test/e2e/install/install_test.go:187 @ 07/09/26 10:03:20.978
```

ffromani:
> looks like an actual failure in ci/prow/ci-install-e2e
> 
> ```
> • [FAILED] [0.361 seconds]
> [Install] durability with a running cluster with all the components and overall deployment [BeforeEach] should be able to delete NUMAResourceOperator CR and redeploy without polluting cluster state [Serial]
>   [BeforeEach] /go/src/github.com/openshift-kni/numaresources-operator/test/e2e/install/install_test.go:182
>   [It] /go/src/github.com/openshift-kni/numaresources-operator/test/e2e/install/install_test.go:381
>   [FAILED] Expected success, but got an error:
>       <*errors.StatusError | 0xc0004b8140>: 
>       kubeletconfigs.machineconfiguration.openshift.io "kc-test" already exists
>       {
>           ErrStatus: {
>               TypeMeta: {Kind: "", APIVersion: ""},
>               ListMeta: {
>                   SelfLink: "",
>                   ResourceVersion: "",
>                   Continue: "",
>                   RemainingItemCount: nil,
>               },
>               Status: "Failure",
>               Message: "kubeletconfigs.machineconfiguration.openshift.io \"kc-test\" already exists",
>               Reason: "AlreadyExists",
>               Details: {
>                   Name: "kc-test",
>                   Group: "machineconfiguration.openshift.io",
>                   Kind: "kubeletconfigs",
>                   UID: "",
>                   Causes: nil,
>                   RetryAfterSeconds: 0,
>               },
>               Code: 409,
>           },
>       }
>   In [BeforeEach] at: /go/src/github.com/openshift-kni/numaresources-operator/test/e2e/install/install_test.go:187 @ 07/09/26 10:03:20.978
> ```

yeah, worth a deeper look.

test/e2e/install/install_test.go
```diff
@@ -185,7 +185,7 @@ var _ = Describe("[Install] durability", Serial, func() {
 		})
 
 		AfterEach(func() {
-			if CurrentSpecReport().State.Is(ginkgotypes.SpecStateSkipped) {
+			if deployer == nil {
```
ffromani, line 189: cc @SargunNarula

test/e2e/install/install_test.go
```diff
@@ -192,13 +192,97 @@ var _ = Describe("[Install] durability", Serial, func() {
...
+				if len(nroObj.Status.DaemonSets) != 1 {
+					klog.InfoS("unsupported daemonsets (/MCP)", "count", len(nroObj.Status.DaemonSets))
+					return false, nil
+				}
+				return true, nil
+			})
+			if err != nil {
+				logRTEPodsLogs(e2eclient.Client, e2eclient.K8sClient, ctx, nroObj, "NRO not available")
+			}
+			Expect(err).NotTo(HaveOccurred(), "inconsistent NRO instance:\n%s", objects.ToYAML(nroObj))
```
Tal-or, line 214: Why mixing assertions and error return? lets stay consistent with one approach (I don't mind which)

test/e2e/install/install_test.go
```diff
@@ -192,13 +192,97 @@ var _ = Describe("[Install] durability", Serial, func() {
...
+					if err != nil {
+						return err
+					}
+					nroObj.Spec.ExporterImage = ""
+					return e2eclient.Client.Update(ctx, nroObj)
+				}).WithTimeout(3 * time.Minute).WithPolling(10 * time.Second).ShouldNot(HaveOccurred())
+
+				Eventually(func(g Gomega) {
+					updatedNroObj := &nropv1.NUMAResourcesOperator{}
+					g.Expect(e2eclient.Client.Get(ctx, nroKey, updatedNroObj)).To(Succeed())
```
Tal-or: ditto
In the `Eventually` above (line 238) we're returning error and here we're using assertions.

Tal-or:
prerequisite to make CI vulnerability happy

https://github.com/openshift-kni/numaresources-operator/pull/4625

test/e2e/install/install_test.go
ffromani: because I haven't found a way to run code (`logRTEPodsLogs`) after the assertion fires. If the assertion fires, we can't log. So we need to check right before. Let's see if there's a smarter way though.

test/e2e/install/install_test.go
ffromani: yes, this is something I can and should make consistent

test/e2e/install/install_test.go
ffromani: there's an awkward way with defer/DeferCleanup which doesn't seem much an improvement to me, more like a sidegrade. Since this is an existing pattern and we are adding unclear gains, I'd defer to another cleanup PR (if possible at all)

Tal-or (approved):
/lgtm

Thanks for the changes

swatisehgal (commented):
/lgtm

PR #4593, opened 2026-07-06
test: numacell: remove old context dep

ffromani:
we can just use the stdlib. Also, fix the unnecessary parameters.


shajmakh:
/lgtm
Thanks.

PR #4594, opened 2026-07-06
e2e: drop workload partitioning resources in DropHostLevelResources

Tal-or:
On clusters with workload partitioning enabled, node allocatable includes `management.workload.openshift.io/cores` which gets picked up by the baseload computation.
However, NRT zone data does not report this resource, causing `SaturateZoneUntilLeft` to fail with "resource not found in zone" during padding pod creation.

Strip all `management.workload.openshift.io/*` resources alongside `ephemeral-storage` so padding works on workload-partitioned clusters.

### DIFF
--- a/test/internal/noderesourcetopologies/nrts.go
+++ b/test/internal/noderesourcetopologies/nrts.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"strconv"
 	"strings"
 	"time"
@@ -43,6 +44,8 @@ import (
 
 // ErrNotEnoughResources means a NUMA zone or a node has not enough resouces to reserve
 var ErrNotEnoughResources = errors.New("nrt: Not enough resources")
+
+const WorkloadPartitioningResourcePrefix = "management.workload.openshift.io/"
 
 func GetZoneIDFromName(zoneName string) (int, error) {
 	for _, prefix := range []string{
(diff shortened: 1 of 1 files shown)


test/internal/noderesourcetopologies/nrts.go
```diff
@@ -257,9 +257,21 @@ func DropHostLevelResources(res corev1.ResourceList) corev1.ResourceList {
 	newList := res.DeepCopy() // since res is a map, updating it directly would mutate its instance in all places
 	klog.Info("drop host-level resources")
 	delete(newList, corev1.ResourceEphemeralStorage)
+	deleteWorkloadPartitioningResources(newList)
```
ffromani: we assume golang >= 1.24 and we don't backport e2e test PRs, so we can use
```
  maps.DeleteFunc(newList, func(resName corev1.ResourceName, _ resource.Quantity) bool {
        return strings.HasPrefix(string(resName), "management.workload.openshift.io/")
  })
```

Tal-or: I'm all good for using the `maps.DeleteFunc` but I think we shouldn't use anonymous function since `deleteWorkloadPartitioningResources` explicitly emphasis the intent here.

Tal-or: on second thought, having this specific implementation as a standalone function doesn't make sense either. scratch my previous comment.

ffromani (commented):
/approve
/lgtm

test/internal/noderesourcetopologies/nrts.go
```diff
@@ -257,6 +258,11 @@ func DropHostLevelResources(res corev1.ResourceList) corev1.ResourceList {
 	newList := res.DeepCopy() // since res is a map, updating it directly would mutate its instance in all places
 	klog.Info("drop host-level resources")
 	delete(newList, corev1.ResourceEphemeralStorage)
+	// workload partitioning resources (e.g. management.workload.openshift.io/cores) are
+	// reported in node allocatable but not in NRT zone data, so we must drop them.
+	maps.DeleteFunc(newList, func(resName corev1.ResourceName, _ resource.Quantity) bool {
+		return strings.HasPrefix(string(resName), "management.workload.openshift.io/")
```
ffromani: if you resubmit, please turn this into a constant (`workloadPartitioningPrefix`?)

ffromani (commented):
/lgtm

thanks for the updates

PR #4595, opened 2026-07-06
OCPBUGS-94176: sched: validate scheduler image 

shajmakh:
run validation on the user-specified scheduler image. The PR provides an option to the cluster-admin to extend the valid set of digests or completely drop the image validation. See commits for more details.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -64,6 +64,7 @@ import (
 	intkloglevel "github.com/openshift-kni/numaresources-operator/internal/kloglevel"
 	"github.com/openshift-kni/numaresources-operator/internal/platforminfo"
 
+	schedulerapi "github.com/openshift-kni/numaresources-operator/internal/api/scheduler"
 	"github.com/openshift-kni/numaresources-operator/pkg/hash"
 	"github.com/openshift-kni/numaresources-operator/pkg/images"
 	rtemetricsmanifests "github.com/openshift-kni/numaresources-operator/pkg/metrics/manifests/monitor"
@@ -446,6 +447,11 @@ func main() {
 		}
 		klog.InfoS("manifests loaded", "component", "Scheduler")
 
+		iv := schedulerapi.GetImageValidationData()
+		if !iv.Enabled {
+			klog.InfoS("scheduler image validation is disabled. To enable it unset the environment variable", "controller", "NUMAResourcesScheduler", "envvar", schedulerapi.SchedulerImageValidationEnvVar)
+		}
+
 		if err = (&controller.NUMAResourcesSchedulerReconciler{
 			Client:             mgr.GetClient(),
 			Scheme:             mgr.GetScheme(),
@@ -453,6 +459,7 @@ func main() {
 			Namespace:          namespace,
 			PlatformInfo:       platforminfo.New(discoveredCluster.Platform, discoveredCluster.LongVersion),
 			TLSSettings:        tlsSettings,
+			ImageValidation:    iv,
 		}).SetupWithManager(mgr); err != nil {
 			klog.ErrorS(err, "unable to create controller", "controller", "NUMAResourcesScheduler")
 			exitWithCancel(cancel, 1)
--- a/internal/api/scheduler/digests.go
+++ b/internal/api/scheduler/digests.go
@@ -20,16 +20,22 @@ import (
 	"encoding/json"
 	"os"
 	"regexp"
+	"strings"
 
 	_ "embed"
 
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
 )
 
 const (
 	// SchedulerImageValidationEnvVar is the environment variable that controls whether the scheduler image validation is enabled.
 	// If it is set to "false", the scheduler image validation is disabled; any other value is considered as enabled.
 	SchedulerImageValidationEnvVar = "SCHEDULER_IMAGE_VALIDATION"
+
+	// CustomSchedulerDigestsEnvVar is the environment variable that contains a user-provided comma-separated list of SHA256
+	// digests of the scheduler images. It is used to extend the embedded digests.
+	CustomSchedulerDigestsEnvVar = "SCHEDULER_IMAGE_DIGESTS"
 )
 
 var (
@@ -65,9 +71,44 @@ func loadDigests() sets.Set[string] {
 
 	digests := sets.New(d.CurrentChannel...)
 	digests.Insert(d.PreviousChannelLast)
+
+	userDigests := getUserImageDigests()
+	digests.Insert(userDigests...)
+
+	klog.V(4).InfoS("trusted set of scheduler images",
+		"fromCurrentChannel", d.CurrentChannel,
+		"latestFromPreviousChannel", d.PreviousChannelLast,
+		"customDigests", userDigests)
+
 	return digests
 }
 
 func IsValidDigest(digest string) bool {
 	return sha256DigestRE.MatchString(digest)
+}
+
+func getUserImageDigests() []string {
+	val, ok := os.LookupEnv(CustomSchedulerDigestsEnvVar)
+	if !ok {
+		return []string{}
+	}
+	return parseCustomSchedulerDigests(val)
+}
+
+func parseCustomSchedulerDigests(customList string) []string {
+	ret := []string{}
+	digests := strings.Split(strings.TrimSpace(customList), ",")
+	for _, d := range digests {
+		d = strings.TrimSpace(d)
+		if d == "" {
+			continue
+		}
+		if !sha256DigestRE.MatchString(d) {
+			klog.Warningf("custom scheduler image digest %q is malformed: expected sha256:<64 hex chars>", d)
+			continue
+		}
+		ret = append(ret, d)
+	}
+
+	return ret
 }
--- a/internal/api/scheduler/digests.go
+++ b/internal/api/scheduler/digests.go
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2026 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package scheduler
+
+import (
+	"encoding/json"
+	"os"
+	"regexp"
+
+	_ "embed"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	// SchedulerImageValidationEnvVar is the environment variable that controls whether the scheduler image validation is enabled.
+	// If it is set to "false", the scheduler image validation is disabled; any other value is considered as enabled.
+	SchedulerImageValidationEnvVar = "SCHEDULER_IMAGE_VALIDATION"
+)
+
+var (
+	//go:embed _digests.json
+	digestsData string
+
+	// sha256DigestRE is a regular expression to match a sha256 digest at the end of a string as
+	// per the OCI and docker registry specifications, e.g. it matches strings of exactly this shape:
+	// sha256:a3f1b2c4d5e6...  (64 hex characters)
+	sha256DigestRE = regexp.MustCompile(`^sha256:[a-f0-9]{64}$`)
+)
+
+func GetImageValidationData() ImageValidation {
+	val, ok := os.LookupEnv(SchedulerImageValidationEnvVar)
+	if ok && val == "false" {
+		return ImageValidation{
+			Enabled: false,
+			Digests: sets.New[string](),
+		}
+	}
+
+	return ImageValidation{
+		Enabled: true,
+		Digests: loadDigests(),
+	}
+}
+
+func loadDigests() sets.Set[string] {
+	var d Digests
+	if err := json.Unmarshal([]byte(digestsData), &d); err != nil {
+		panic(err)
+	}
+
+	digests := sets.New(d.CurrentChannel...)
+	digests.Insert(d.PreviousChannelLast)
+	return digests
+}
+
+func IsValidDigest(digest string) bool {
+	return sha256DigestRE.MatchString(digest)
+}
--- a/internal/api/scheduler/types.go
+++ b/internal/api/scheduler/types.go
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2026 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package scheduler
+
+import "k8s.io/apimachinery/pkg/util/sets"
+
+type ImageValidation struct {
+	Enabled bool
+	Digests sets.Set[string]
+}
+
+type Digests struct {
+	CurrentChannel      []string `json:"current_channel"`
+	PreviousChannelLast string   `json:"prev_channel_last"`
+}
+
+func (ti *Digests) AddCurrentChannel(digest string) {
+	ti.CurrentChannel = append(ti.CurrentChannel, digest)
+}
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -48,9 +48,10 @@ const (
 
 // TODO: are we duping these?
 const (
-	ReasonAsExpected      = "AsExpected"
-	ReasonInternalError   = "InternalError"
-	ReasonCustomUserImage = "CustomUserImage"
+	ReasonAsExpected            = "AsExpected"
+	ReasonInternalError         = "InternalError"
+	ReasonCustomUserImage       = "CustomUserImage"
+	ReasonInvalidSchedulerImage = "InvalidSchedulerImage"
 )
 
 const (
--- a/pkg/validation/validation.go
+++ b/pkg/validation/validation.go
@@ -255,7 +255,7 @@ func SchedulerImage(image string, digests sets.Set[string]) error {
 	}
 
 	if !digests.Has(digest) {
-		return fmt.Errorf("image %q digest %q is not in the trusted list", image, digest)
+		return fmt.Errorf("image %q digest %q is not in the trusted list. To extend the list, update the environment variable %q with comma separated list of digests", image, digest, schedulerapi.CustomSchedulerDigestsEnvVar)
 	}
 	return nil
 }
--- a/pkg/validation/validation.go
+++ b/pkg/validation/validation.go
@@ -19,14 +19,17 @@ package validation
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
 
 	nropv1 "github.com/openshift-kni/numaresources-operator/api/v1"
 	nodegroupv1 "github.com/openshift-kni/numaresources-operator/api/v1/helper/nodegroup"
 	"github.com/openshift-kni/numaresources-operator/internal/api/annotations"
+	schedulerapi "github.com/openshift-kni/numaresources-operator/internal/api/scheduler"
 )
 
 const (
@@ -236,4 +239,23 @@ func nodeGroupsAnnotations(nodeGroups []nropv1.NodeGroup) error {
 		err = errors.Join(err, fmt.Errorf("pool #%d has too many annotations %d max %d", idx, len(nodeGroup.Annotations), nropv1.NodeGroupMaxAnnotations))
 	}
 	return err
+}
+
+// SchedulerImage checks if the scheduler image is pinned by a valid sha256 digest and is
+// found in the trusted digests list
+func SchedulerImage(image string, digests sets.Set[string]) error {
+	idx := strings.Index(image, "@")
+	if idx == -1 {
+		return fmt.Errorf("image %q must be referenced by digest, not by tag", image)
+	}
+
+	digest := image[idx+1:]
+	if !schedulerapi.IsValidDigest(digest) {
+		return fmt.Errorf("image %q has malformed digest %q: expected sha256:<64 hex chars>", image, digest)
+	}
+
+	if !digests.Has(digest) {
+		return fmt.Errorf("image %q digest %q is not in the trusted list", image, digest)
+	}
+	return nil
 }
--- a/test/e2e/serial/e2e_serial_test.go
+++ b/test/e2e/serial/e2e_serial_test.go
@@ -24,8 +24,11 @@ import (
 	_ "github.com/openshift-kni/numaresources-operator/test/e2e/serial/tests"
 	_ "github.com/openshift-kni/numaresources-operator/test/internal/configuration"
 
+	intsched "github.com/openshift-kni/numaresources-operator/internal/api/scheduler"
 	serialconfig "github.com/openshift-kni/numaresources-operator/test/e2e/serial/config"
 	e2eclient "github.com/openshift-kni/numaresources-operator/test/internal/clients"
+	e2eenvvar "github.com/openshift-kni/numaresources-operator/test/internal/envvar"
+	e2esched "github.com/openshift-kni/numaresources-operator/test/internal/nrosched"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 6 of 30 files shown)


shajmakh:
e2e is failing unexpectedly in prow, although the test is passing locally. The failure is due to a degraded condition due to image not pinned by digest, while the validation was disabled. Investigating the matter.

shajmakh:
@coderabbitai review commits 3e4945dd92356b51f88de1ecc9080b5f528baa50 and bb4e6bedc4076ac32883f74b7575f5c4f5a463b7

cmd/main.go
```diff
@@ -428,13 +429,19 @@ func main() {
 		}
 		klog.InfoS("manifests loaded", "component", "Scheduler")
 
+		iv := schedulerapi.GetImageValidation()
+		if !iv.Enabled {
+			klog.InfoS("scheduler image validation is disabled", "controller", "NUMAResourcesScheduler")
```
Tal-or: It would improve UX  to explain (one line sentence) what should be done to enable validation.

internal/api/scheduler/digests.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+var (
+	//go:embed _digests.json
+	digestsData string
+
+	// sha256DigestRE is a regular expression to match a sha256 digest at the end of a string as
+	// per the OCI and docker registry specifications, e.g. it matches strings of exactly this shape:
+	// sha256:a3f1b2c4d5e6...  (64 hex characters)
+	sha256DigestRE = regexp.MustCompile(`^sha256:[a-f0-9]{64}$`)
```
Tal-or, line 48: we can use the https://pkg.go.dev/github.com/opencontainers/go-digest to check that

pkg/status/status.go
```diff
@@ -58,6 +58,7 @@ const (
 
 const (
 	ConditionTypeIncorrectNUMAResourcesSchedulerResourceName = "IncorrectNUMAResourcesSchedulerResourceName"
+	ReasonUntrustedSchedulerImage                            = "UntrustedSchedulerImage"
```
Tal-or: untrusted sounds a bit offensive, may unvalidated/nonValidated

tools/getdigests/getdigests.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	digests.SetPreviousChannelLast(latestOfPrev)
+
+	return digests, nil
+}
+
+// previousVersion returns the version string for the immediately preceding minor release.
+// e.g. "4.20" -> "4.19"
+func previousVersion(versionString string) (string, error) {
```
Tal-or, line 113: don't we have a case when customers are going from EUS -> EUS versions?
like from 4.20 -> 4.22 will this logic still work in such case?

tools/getdigests/getdigests.go
```diff
@@ -0,0 +1,219 @@
...
+	pullSecretPath := ""
+	outputFile := ""
+	versionString := ""
+	prevVersionOverride := ""
+
+	flag.StringVar(&imageURL, "image", "", "image URL to query (e.g. registry.redhat.io/openshift4/noderesourcetopology-scheduler-rhel9)")
+	flag.StringVar(&pullSecretPath, "pull-secret", "", "path to the pull secret JSON file for registry authentication")
+	flag.StringVar(&outputFile, "output", "", "path to output file; if not provided, output is written to stdout")
+	flag.StringVar(&versionString, "version", "", "version string to filter tags (e.g. 4.20)")
+	flag.StringVar(&prevVersionOverride, "prev-version", "", "previous channel version to use instead of auto-calculating (required when --version has minor=0, e.g. --version 5.0 --prev-version 4.22)")
```
Tal-or, line 177: Maybe just to be on the safe side (because this tool will be used by users) it is better to always require the previous channel and not trying to calculate it automatically.

Tal-or (commented):
looking good overall
left couple of comments

pkg/status/status.go
shajmakh: I wouldn't go with unvalidated, because every image is validated against the rules (if validation is not disabled) and may pass or fail. we can go with InvalidSchedulerImage instead for a more generic expression, do you think that can work?

cmd/main.go
shajmakh: it's documented in later commit  https://github.com/openshift-kni/numaresources-operator/pull/4595/changes/bfcf400b4a882aa68df06f936a7a7419f2176830

internal/api/scheduler/digests.go
shajmakh: I'll check about that WRT dependencies vs local code trade-offs. Thanks for pointing out.

tools/getdigests/getdigests.go
shajmakh: we will be the ones in control of the built-in digests, and when generating the json we can choose what "previuos version to check. the worst case is that we add another entry in the json file for previous EUS (where relevant). I'll double check that.

tools/getdigests/getdigests.go
shajmakh: we are the primary users of the tool (actually via GH action or manual update), and I don't mind adding a mandatory condition to have the previous (and EUS-prev) provided. will update that.

cmd/main.go
Tal-or: yes, that's great, but a brief explanation in the log itself, could guide the user immediately

pkg/status/status.go
Tal-or: sounds good to me

tools/getdigests/getdigests.go
Tal-or: if that's for devs then we have more flexibility. 
I thought this tool is for users to configure their trusted images.

shajmakh: yeah we don't ship the tool, but it's public and optional. Even when executed, it's output can not update the embedded list of valid digests shipped by the operator.

cmd/main.go
shajmakh: the validation is enabled by default, so if it is disabled then the user must be aware of what changes he did to disable it because it is hard to miss (only if SCHEDULER_IMAGE_VALIDATION is set explicitly to `false`), a breif hint won't hurt though.

shajmakh:
/hold 
investigating ci-e2e failures

shajmakh:
> /hold investigating ci-e2e failures

The failures are due to OLM behavior failing to propagate the new values to the deployment in during the dedicated timeout period. A quick workaround (which may still have the effect of the OLm behavior but at least more visible to us to debug) is to force the deletion of the manager deployment - which is fine for test usage-, and by that, force it to get recreated by OLM handling with the new values from the subscription.

tools/getdigests/getdigests.go
shajmakh: for the EUS support, I'm addressing that separately in a followup PR, should also be backported to the older versions.

tools/getdigests/getdigests.go
shajmakh: https://github.com/openshift-kni/numaresources-operator/pull/4595#discussion_r3576467591

internal/api/scheduler/digests.go
ffromani: go-digest seems nice but let's try to avoid panics in the critical, non-startup path: https://github.com/opencontainers/go-digest/issues/110

test/e2e/serial/e2e_serial_test.go
```diff
@@ -46,6 +47,19 @@ var _ = BeforeSuite(func() {
...
+	// production registry is troublesome, and so is handling changing the digests on each branch. Testing e2e
+	// for the image validation will be handled in a separate test.
+	By("identify the source of operator installation and disable scheduler image validation")
+	var err error
+	restoreFunc, err := e2eenvvar.SetOperatorEnvVar(ctx, e2eclient.Client, e2eenvvar.SchedulerImageValidationEnvVar, "false")
+	Expect(err).NotTo(HaveOccurred())
+	DeferCleanup(func() {
+		By("restoring scheduler image validation settings")
+		Expect(restoreFunc()).To(Succeed())
+	})
```
ffromani: you can write `DeferCleanup(func(cleanupCtx context.Context) { ...` and ginkgo will give us a fresh context when invoking the callback. Then we can use the fresh valid context inside `restoreFunc`

ffromani: (here and in all the other similar places)

test/e2e/serial/e2e_serial_test.go
```diff
@@ -46,6 +47,19 @@ var _ = BeforeSuite(func() {
 	Expect(serialconfig.CheckNodesTopology(ctx)).Should(Succeed())
 	serialconfig.Setup()
 	setupExecuted = true
+
+	// for e2e verification we disable the image validation because handling pulling images in the CI from the
+	// production registry is troublesome, and so is handling changing the digests on each branch. Testing e2e
+	// for the image validation will be handled in a separate test.
+	By("identify the source of operator installation and disable scheduler image validation")
```
ffromani: conceptually fine, but we need to have a way to run with default configuration, so we can create lanes which exercise the user flows. IOW, the env var-based validation disable needs to be allowed to be disabled (!!).
In our default CI it should run as you set it indeed.

test/internal/envvar/envvar.go
```diff
@@ -0,0 +1,478 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openshift-kni/numaresources-operator/internal/podlist"
+	"github.com/openshift-kni/numaresources-operator/internal/wait"
+)
+
+const (
+	SchedulerImageValidationEnvVar = "SCHEDULER_IMAGE_VALIDATION"
+	CustomSchedulerDigestsEnvVar   = "SCHEDULER_IMAGE_DIGESTS"
```
ffromani: are these duped?

test/internal/envvar/envvar.go
```diff
@@ -0,0 +1,478 @@
...
+
+	operatorSubscriptionNameEnvVar = "OPERATOR_SUBSCRIPTION_NAME"
+
+	// This is the name and the package name of the subscription spec in the operator subscription CRD.
+	// in operator subscription it would be found under `spec.name` or `spec.package`
+	subscriptionSpecNamePackage = "numaresources-operator"
+
+	operatorDeploymentName   = "numaresources-controller-manager"
+	operatorManagerContainer = "manager"
```
ffromani: likewise, are these duped?

test/internal/envvar/envvar.go
```diff
@@ -0,0 +1,478 @@
...
+)
+
+var subscriptionGVK = schema.GroupVersionKind{
+	Group:   "operators.coreos.com",
+	Version: "v1alpha1",
+	Kind:    "Subscription",
+}
+
+type restoreFunc func() error
```
ffromani: should this func take a context as parameter?

test/internal/envvar/envvar.go
```diff
@@ -0,0 +1,478 @@
...
+		name, nameFound, nameErr := unstructured.NestedString(subList.Items[i].Object, "spec", "name")
+		if pkgErr != nil && nameErr != nil {
+			return nil, fmt.Errorf("failed to get package or name from subscription: pkgErr=%v nameErr=%v sub=%v", pkgErr, nameErr, subList.Items[i].Object)
+		}
+		if pkgFound {
+			if pkg == subscriptionSpecNamePackage {
+				return subList.Items[i].DeepCopy(), nil
+			}
+		}
+		if nameFound {
```
ffromani, line 109: in which cases matching by name is good enough?

test/internal/envvar/envvar.go
```diff
@@ -0,0 +1,478 @@
...
+			if name == subscriptionSpecNamePackage {
+				return subList.Items[i].DeepCopy(), nil
+			}
+		}
+	}
+	klog.InfoS("no subscription found with the correct package name", "subscriptions", subList.Items)
```
ffromani, line 115: dumping ALL the subscriptions may turn out in large and unhelpful output, especially if done unconditionally

test/internal/envvar/envvar.go
```diff
@@ -0,0 +1,478 @@
...
+	return nil, errors.NewNotFound(schema.GroupResource{Group: subscriptionGVK.Group, Resource: "subscriptions"}, subName)
+}
+
+func findOperatorDeployment(ctx context.Context, cli client.Client) (*appsv1.Deployment, error) {
+	dpList := &appsv1.DeploymentList{}
+	if err := cli.List(ctx, dpList, client.MatchingLabels{
+		"control-plane": "controller-manager",
+		"app":           "numaresources-operator",
```
ffromani: are we duping these constants? IIRC we should have an existing helper to find the operator deployment/pod, can we reuse it?

test/internal/envvar/envvar.go
```diff
@@ -0,0 +1,478 @@
...
+	}
+	if err := ensureSubscriptionIsUpdated(ctx, cli, subKey, updatedEnv); err != nil {
+		return nil, err
+	}
+	if err := waitForDeploymentRolloutAfterSubscriptionUpdate(ctx, cli, initialDp, initialPods, []corev1.EnvVar{{Name: envName, Value: envValue}}); err != nil {
+		return nil, err
+	}
+
+	return func() error {
+		ctx := context.TODO()
```
ffromani: yes, we can use a context as parameter there

test/internal/envvar/envvar.go
```diff
@@ -0,0 +1,478 @@
...
+	}
+
+	if !updatedSuccessfully {
+		return fmt.Errorf("failed to update deployment %q", key.String())
+	}
+
+	return waitForDeploymentRollout(ctx, cli, dp, initialPods, newContainer.Env)
+}
+
+func getSubscriptionEnv(sub *unstructured.Unstructured) ([]interface{}, error) {
```
ffromani: please use `any` in new code

test/internal/envvar/envvar.go
```diff
@@ -0,0 +1,478 @@
+/*
+ * Copyright 2026 Red Hat, Inc.
```
ffromani, line 2: envvar.go is clever but it adds a bunch of new code with no tests (and, in all fairness, quite few functions are hard to test). Can we reuse existing packages, or slightly extend them, to reduce the risk?

test/e2e/sched/sched_test.go
```diff
@@ -52,13 +54,10 @@ import (
...
 
 var _ = Describe("[Scheduler] scheduler object updates", func() {
-	var initialized bool
 	nroSchedObj := &nropv1.NUMAResourcesScheduler{}
 	BeforeEach(func() {
-		if !initialized {
-			Expect(e2eclient.ClientsEnabled).To(BeTrue(), "failed to create runtime-controller client")
-		}
-		initialized = true
+		Expect(e2eclient.ClientsEnabled).To(BeTrue(), "failed to create runtime-controller client")
```
ffromani: we do we do this unconditionally now?

test/e2e/sched/sched_test.go
```diff
@@ -136,6 +135,42 @@ var _ = Describe("[Scheduler] scheduler object updates", func() {
...
 		})
 
+		It("should degrade status when scheduler image is invalid and validation is enabled", func(ctx context.Context) {
+			restoreFunc, err := e2eenvvar.SetOperatorEnvVar(ctx, e2eclient.Client, e2eenvvar.SchedulerImageValidationEnvVar, "true")
+			Expect(err).NotTo(HaveOccurred())
+			// eventhough BeforeEach handles the restoration of the environment variable to the initial state, we need to restore it here
+			// because we don't want to depend on the order of the cleanup functions in BeforeEach.
+			DeferCleanup(func() {
+				Expect(restoreFunc()).To(Succeed())
+			})
```
ffromani: better:
```
			DeferCleanup(func(cleanupCtx context.Context) {
				Expect(restoreFunc(cleanupCtx)).To(Succeed())
			})
```

test/e2e/sched/sched_test.go
```diff
@@ -136,6 +135,42 @@ var _ = Describe("[Scheduler] scheduler object updates", func() {
...
+
+			testSchedObj := &nropv1.NUMAResourcesScheduler{}
+			By(fmt.Sprintf("switching the NROS image to %s which does not pass the image validation criteria", e2eimages.SchedTestImageCI))
```
ffromani: is it intentional we don't use `e2efixture.By`?

test/e2e/sched/sched_test.go
```diff
@@ -136,6 +135,42 @@ var _ = Describe("[Scheduler] scheduler object updates", func() {
...
+				g.Expect(e2eclient.Client.Get(ctx, client.ObjectKeyFromObject(nroSchedObj), testSchedObj)).To(Succeed())
+				klog.InfoS("scheduler conditions", "conditions", testSchedObj.Status.Conditions)
+
+				var degradedCondition *metav1.Condition
+				for _, condition := range testSchedObj.Status.Conditions {
+					if condition.Type == status.ConditionDegraded {
+						degradedCondition = &condition
+						break
+					}
+				}
```
ffromani: don't we have helpers to find conditions?

ffromani (commented):
thanks for this PR. `envvar.go` is clever and looks like  the right direction, but right now it's a bunch of new code. Let's try to de-risk it

test/e2e/sched/sched_test.go
shajmakh: I recall the helper was in the controller tests, I'll double check. Thanks.

test/e2e/sched/sched_test.go
shajmakh: not intentional, and I need to pay more attention while me repeating this

test/e2e/sched/sched_test.go
shajmakh: used the shared helper `metahelper.FindStatusCondition`

test/e2e/serial/e2e_serial_test.go
shajmakh: I was depending on the creating the context when I build the restore function, but this way is more reasonable. Addressed, thanks!

test/e2e/serial/e2e_serial_test.go
shajmakh: good point, addressed using env var. The var should be switched ON on d/s CI from production which is the most meaningful run to check the end product. And on other runs (from stage) it can be switched ON during the pipeline setup.

test/internal/envvar/envvar.go
shajmakh: yes, wrote them before exporting the ones used by the controller code. addressed that.

test/internal/envvar/envvar.go
shajmakh: you're right, can be done using the namespace under the deployment. will address this.

test/internal/envvar/envvar.go
shajmakh: only in case of the name of the subscription is exactly as expected in the tests , but this depends on the user/tester's decision. I think we can enhance this when we pick the subscription based on the namespace. Will address this.

test/internal/envvar/envvar.go
shajmakh: I would keep this code separate because I don't see where it can belong better, unless you see that, and then I don't mind extending it. Regarding the unit tests, I hear you, so I can add tests with a fake client to reduce the risk, but it likely won't be 100% coverage of the real update especially on the subscription path. I'm open for better suggestions.

test/internal/envvar/envvar.go
shajmakh: we don't, I'll create a consts file to group all of these

test/internal/envvar/envvar.go
ffromani: Are these helpers we can reuse? so we can get rid of some code here, hopefully

shajmakh: we might reuse to edit the env-vars in the subscription, so the only use case is this PR.
the reason why I thought to automate this here is because we don't know what is going to run the tests (u/s or d/s or local) and where the installation source is (local or OLM), so instead of handling each case with a special setup before we run the tests, we have that autodetected and automated

shajmakh: at least prow CI is going to fail with the default config, and in one way or another we need to disable the validation and IMO owing that through the tests is more controllable and easier to manage than managing that from different lanes (prow and d/s and local). We could have done that separately as a make target and call it before each prow job but I don't see how is that better than having the test suites control that

test/internal/envvar/envvar.go
shajmakh: extracted into consts.go file

test/e2e/sched/sched_test.go
shajmakh: I didn't see the use for it especially because the assertion is quick, so why the extra noisy condition? 
it actually belongs better to the beforeSuite not beforeEach. will update that.

shajmakh:
e2e test is split to https://github.com/openshift-kni/numaresources-operator/pull/4693 for isolated investigation of the subscription rollout. The problem isn't related to the NRO/S functionality but to OLM.

internal/api/scheduler/digests.go
shajmakh: I'v looked at the go-digest, and the panics are not part of our check here, however I'm in favor of keeping the minimal local regular expression check, because this is minimal, straightforward, doesn't need require dependency (although small), and we maintain it. OTOH, go-digest has been idle for a few years, and, most importantly, we do not want to expose the risk of misusing the verification route and trigger a panic. If you see a strong reason for pushing to use go-digest please share it, otherwise I think we can proceed with the reges version.

ffromani:
please avoid merge commits or commits which have "merge XXX" in their commit message

internal/api/scheduler/digests.go
ffromani: I'm fine starting with the regex path

internal/api/scheduler/types.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+type Digests struct {
+	CurrentChannel      []string `json:"current_channel"`
+	PreviousChannelLast string   `json:"prev_channel_last"`
+}
+
+func (ti *Digests) AddCurrentChannel(digest string) {
+	ti.CurrentChannel = append(ti.CurrentChannel, digest)
+}
```
ffromani, line 33: this is useful because the API contract of the `Digest` object is that you can add digests, but never remove them. The function enforces the contract (to be airtight the field should be private, but still acceptable)

internal/api/scheduler/types.go
```diff
@@ -0,0 +1,37 @@
...
+
+func (ti *Digests) SetPreviousChannelLast(digest string) {
+	ti.PreviousChannelLast = digest
+}
```
ffromani: OTOH this method adds no value, it's a trivial wrapper. The invariant here the field should be set only once at creation time and never reset.

This is an internal API, so we can improve later.

internal/api/scheduler/digests.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	digests := sets.New(d.CurrentChannel...)
+	digests.Insert(d.PreviousChannelLast)
+	return digests
+}
+
+func GetValidDigestRE() *regexp.Regexp {
+	return sha256DigestRE
+}
```
ffromani: the RE should never be exposed. We need a function which checks a sha256, not this getter.
IOW, the digest must never leak outside this package.

pkg/status/status.go
```diff
@@ -59,6 +59,7 @@ const (
 
 const (
 	ConditionTypeIncorrectNUMAResourcesSchedulerResourceName = "IncorrectNUMAResourcesSchedulerResourceName"
+	ReasonInvalidSchedulerImage                              = "InvalidSchedulerImage"
```
ffromani: please move the reason in the block above, lines 50-53

pkg/validation/validation.go
```diff
@@ -237,3 +240,22 @@ func nodeGroupsAnnotations(nodeGroups []nropv1.NodeGroup) error {
...
+// SchedulerImage checks if the scheduler image is pinned by a valid sha256 digest and is
+// found in the trusted digests list
+func SchedulerImage(image string, digests sets.Set[string]) error {
+	idx := strings.Index(image, "@")
+	if idx == -1 {
+		return fmt.Errorf("image %q must be referenced by digest, not by tag", image)
+	}
+
+	digest := image[idx+1:]
+	if !schedulerapi.GetValidDigestRE().MatchString(digest) {
```
ffromani: this if should be moved inside `schedulerapi`

internal/api/scheduler/digests.go
Tal-or: >I'm fine starting with the regex path

+1 
go-digest was merely a suggestion

ffromani (commented):
LGTM once the inline comments have been addressed. All minor, should be quick and painless

ffromani:
do we have any form of non-unit (integration, e2e preferred) test which demonstrates that the scheduler does not consume an untrusted image and goes degraded?

[EDIT] if not, test/e2e/install/install_test.go / "should reject custom ExporterImage and go Degraded without changing the DaemonSet" should be a good example

internal/api/scheduler/types.go
shajmakh: > to be airtight the field should be private

both fields need to be public because we unmarshal the json file

internal/api/scheduler/digests.go
shajmakh: created a validation function inside the file so I don't have to export it. Thanks for pointing this out.

internal/api/scheduler/types.go
shajmakh: Removed the method for now; the struct is being used only to unmarshal the embedded data which is fixed. The digests are loaded and shared then with the validation logic so the scheduler imae can be validated against them but there is no way to delete or update the embedded list.

shajmakh:
> do we have any form of non-unit (integration, e2e preferred) test which demonstrates that the scheduler does not consume an untrusted image and goes degraded?
> 
> [EDIT] if not, test/e2e/install/install_test.go / "should reject custom ExporterImage and go Degraded without changing the DaemonSet" should be a good example

there is controller tests:
https://github.com/openshift-kni/numaresources-operator/pull/4595/changes#diff-2eeebee0dc20cda093f35d6a21e81f9bc200716ecc3be36d60f1a4555d53b18fR143
and an e2e in a separate PR: https://github.com/openshift-kni/numaresources-operator/pull/4693 only because https://github.com/openshift-kni/numaresources-operator/pull/4595#issuecomment-5019925191

ffromani:
/approve
/lgtm

thanks for the updates

PR #4625, opened 2026-07-09
ci: use Go 1.25 (latest patch) to fix govulncheck failures

Tal-or:
## Summary
Set `GOLANG_VERSION` to 1.25 in CI workflows so the setup-go action
resolves to the latest 1.25.x patch, and drop the pinned toolchain
directive from go.mod. 

This fixes [GO-2026-5856 (crypto/tls)](https://pkg.go.dev/vuln/GO-2026-5856) and
[GO-2026-4970 (os)](https://pkg.go.dev/vuln/GO-2026-4970) without requiring a bump for every future patch.

## Test plan
- [x] CI govulncheck pass should now succeed
- [ ] Verify no regressions in unit/e2e tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,7 @@ on:
   workflow_dispatch:
 
 env:
-  GOLANG_VERSION: 1.25.11
+  GOLANG_VERSION: 1.25
 
 defaults:
   run:
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -35,7 +35,7 @@ jobs:
     - name: Set up golang
       uses: actions/setup-go@v5
       with:
-        go-version: 1.25.11
+        go-version: 1.25
 
     - name: Verify modules
       run: go mod verify
--- a/.github/workflows/govulncheck.yaml
+++ b/.github/workflows/govulncheck.yaml
@@ -14,7 +14,7 @@ on:
       - main
 
 env:
-  GOLANG_VERSION: 1.25.11
+  GOLANG_VERSION: 1.25
 
 defaults:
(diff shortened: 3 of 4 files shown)


ffromani:
can we just set `toolchain` to 1.25 and let it (hopefully) pick up the latest Z?

Tal-or:
> can we just set `toolchain` to 1.25 and let it (hopefully) pick up the latest Z?

using 1.25 without the z-stream makes go mod tidy to drop the `toolchain` directive completely, and it doesn't seems like it's taking the latest because CI fail again

ffromani:
> > can we just set `toolchain` to 1.25 and let it (hopefully) pick up the latest Z?
> 
> using 1.25 without the z-stream makes go mod tidy to drop the `toolchain` directive completely, and it doesn't seems like it's taking the latest because CI fail again

I see, thanks for checking

Tal-or:
> > > can we just set `toolchain` to 1.25 and let it (hopefully) pick up the latest Z?
> > 
> > 
> > using 1.25 without the z-stream makes go mod tidy to drop the `toolchain` directive completely, and it doesn't seems like it's taking the latest because CI fail again
> 
> I see, thanks for checking

I stand corrected. you were right it does take the latest, but we had the z version pinned elsewhere. now its suppose to be fine.

ffromani (commented):
I reviewed how the `toolchain` directive works. It sets the minimum version, so we really need to set 1.25.12: setting 1.25 would be equivalent to set 1.25.0

ffromani:
/approve
/lgtm

ok, let's see if this helps. If not, we will pin to 1.25.12

PR #4230, opened 2026-05-27
OCPBUGS-12986: narrow down RBAC permissions

ffromani:
Observe the usage pattern of the operator and narrow down as much as possible the RBAC permissions

AI-attribution: AIA PAI Ce Hin R claude-4.6-opus-1M v1.0

PR #4639, opened 2026-07-13
OCPBUGS-94182: narrow down RBAC permissions aggressively

ffromani:
Stronger version of https://github.com/openshift-kni/numaresources-operator/pull/4230.

Narrow down more aggressively the RBAC footprint, narrowing down cluster to namespace permissions, listing verbs explicitly and pinning as much as possible to resource names.

Two key decisions made with the intent to narrow down RBAC further
- disable events, which were always half-baked and which require permissions in the default namespace, or a rewrite. We do neither and just remove
- disable the pod readiness: RTE would need to patch its own pod status, but we can't predict the name and the operator must have the permissions it grants, which would broaden the scope too much.


cmd/main.go
```diff
@@ -343,8 +345,7 @@ func main() {
 	mgr, err := ctrl.NewManager(restConfig, ctrl.Options{
 		Cache: cache.Options{
 			DefaultNamespaces: map[string]cache.Config{
-				namespace:            {},
-				metav1.NamespaceNone: {},
+				namespace: {},
 			},
```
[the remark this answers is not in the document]
Tal-or: Bot has a point, the NS can be added conditionally when running on HyperShift.
Add this change only if you resubmit since this is out of scope for this work (although related in some extent)

internal/controller/kubeletconfig_controller.go
```diff
@@ -88,10 +88,10 @@ type kubeletConfigHandler struct {
 }
 
 // Namespace Scoped
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=create;delete;get;list;update;watch,namespace="numaresources"
```
Tal-or: That would break NROP on HyperShift, we should add the `openshift-config-managed` namespace.

internal/controller/numaresourcesoperator_controller.go
```diff
@@ -777,6 +779,11 @@ func validateMachineConfigLabels(mc client.Object, trees []nodegroupv1.Tree) err
 func daemonsetUpdater(poolName string, gdm *rtestate.GeneratedDesiredManifest, rteMetricsTLS objtls.Settings) error {
 	rteupdate.AllContainersTerminationMessagePolicy(gdm.DaemonSet)
 	rteupdate.DaemonSetTolerations(gdm.DaemonSet, gdm.NodeGroup.Config.Tolerations)
+	// The vendored DaemonSet manifest defines readiness gates for PodresourcesFetched and
+	// NodeTopologyUpdated conditions. DaemonSetArgs disables podreadiness (--podreadiness=false)
+	// to avoid needing pods/status RBAC, so these conditions are never set and the gates must
+	// be cleared or pods will never reach full Ready status.
+	gdm.DaemonSet.Spec.Template.Spec.ReadinessGates = []corev1.PodReadinessGate{}
```
Tal-or: So far we had PodReadiness enabled by default, and the its status reported as part of the API (`PodresourcesFetched` and  `NodeTopologyUpdated` conditions).

I'm not sure we can simply remove a stable API for a GA grade product, unless switching something from opt-out to opt-in doesn't count.

Tal-or (commented):
Some changes required to avoid breaking operation on HyperShift.

The u/s CI passing doesn't provide a strong evidence here because it's not using the updated RBAC rules.

from the must-gather extra I pulled:
```
           "apiVersion": "rbac.authorization.k8s.io/v1",
            "kind": "Role",
            "metadata": {
                "creationTimestamp": "2026-07-13T17:18:28Z",
                "labels": {
                    "olm.managed": "true",
                    "olm.owner": "numaresources-operator.v5.0.999-snapshot",
                    "olm.owner.kind": "ClusterServiceVersion",
                    "olm.owner.namespace": "numaresources-operator",
...
                    "operators.coreos.com/numaresources-operator.numaresources-operator": ""
                },
   ...           
            "rules": [
...
                {
                    "apiGroups": [
                        ""
                    ],
                    "resources": [
                        "events"
                    ],
                    "verbs": [
                        "create",
                        "patch"
                    ]
                },
```

Which IIRC this PR removes events completely

internal/controller/numaresourcesoperator_controller.go
ffromani: Right, and most notably this change was not sufficient to drop the `pod/status` permission it seems, so I removed this commit

internal/controller/kubeletconfig_controller.go
ffromani: thanks for pointing this out. I think I addressed in the last commit.

shajmakh:
Thanks for addressing.
/lgtm
/hold
please unhold if you see fit or no if you have other updates

Tal-or (approved):
/hold cancel 
Looking good, thanks for the updates

PR #4693, opened 2026-07-20
e2e: add test to check the image validation

shajmakh:
Add an e2e test to verify the image validation causing the scheduler CR
to be degraded after not complying to the validation criteria.

### DIFF
--- a/hack/run-test-e2e.sh
+++ b/hack/run-test-e2e.sh
@@ -54,6 +54,7 @@ if [ "$ENABLE_SCHED_TESTS" = true ]; then
   # --flake-attempts: rerun the test if it fails
   # -requireSuite: fail if tests are not executed because of missing suite
   ${BIN_DIR}/e2e-nrop-sched.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.flake-attempts=2 --ginkgo.junit-report=${REPORT_DIR}/e2e-sched.xml
+  ${BIN_DIR}/e2e-nrop-sched-image-validation.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.flake-attempts=2 --ginkgo.junit-report=${REPORT_DIR}/e2e-sched-image-validation.xml
 
   if [ "$ENABLE_CLEANUP" = true ]; then
     echo "Running NROScheduler uninstall test suite";
--- a/test/e2e/sched/image_validation/test_suite_validation_test.go
+++ b/test/e2e/sched/image_validation/test_suite_validation_test.go
@@ -1,0 +1,43 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package sched
+
+import (
+	"context"
+	"testing"
+
+	e2eclient "github.com/openshift-kni/numaresources-operator/test/internal/clients"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// TestSchedulerImageValidation tests the image validation logic for the scheduler with default validation behavior, which is
+// enabled at the time of writing this test. We could have added this test under sched/sched_test.go, but because that suite
+// is disabling the validation at suite level, we would need to enable that again via updating the subcription multiple
+// consecutive times, which proved in prow and local CI runs to be flaky in terms of the time it takes for the updates of
+// the subscription environment variables to be reflected on the operator deployment, after consecutive updates of the same
+// variable.
+
+func TestScheduler(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Scheduler")
+}
+
+var _ = BeforeSuite(func(ctx context.Context) {
+	Expect(e2eclient.ClientsEnabled).To(BeTrue(), "failed to create runtime-controller client")
+})
--- a/test/e2e/sched/image_validation/validation_test.go
+++ b/test/e2e/sched/image_validation/validation_test.go
@@ -1,0 +1,131 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package sched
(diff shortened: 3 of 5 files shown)


test/e2e/must-gather/must_gather_suite_test.go
```diff
@@ -81,7 +84,23 @@ var _ = BeforeSuite(func() {
...
+		// for the image validation will be handled in a separate test.
+		By("identify the source of operator installation and disable scheduler image validation")
+		var err error
+		restoreFunc, err := e2eenvvar.SetOperatorEnvVar(ctx, e2eclient.Client, intsched.SchedulerImageValidationEnvVar, "false")
+		Expect(err).NotTo(HaveOccurred())
+		DeferCleanup(func(cleanupCtx context.Context) {
+			By("restoring scheduler image validation settings")
+			Expect(restoreFunc(cleanupCtx)).To(Succeed())
+		})
+	}
```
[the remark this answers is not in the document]
shajmakh: handling in a different PR

shajmakh:
e2e passed, rerunning to ensure its success is stable:
/retest ci-e2e

test/e2e/sched/image_validation/test_suite_validation_test.go
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright 2021 Red Hat, Inc.
```
Tal-or: IIRC we agreed to remove the year, which is wrong date anyway.

hack/run-test-e2e.sh
```diff
@@ -54,6 +54,7 @@ if [ "$ENABLE_SCHED_TESTS" = true ]; then
   # --flake-attempts: rerun the test if it fails
   # -requireSuite: fail if tests are not executed because of missing suite
   ${BIN_DIR}/e2e-nrop-sched.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.flake-attempts=2 --ginkgo.junit-report=${REPORT_DIR}/e2e-sched.xml
+  ${BIN_DIR}/e2e-nrop-sched-image-validation.test ${NO_COLOR} --ginkgo.v --ginkgo.timeout=5h --ginkgo.flake-attempts=2 --ginkgo.junit-report=${REPORT_DIR}/e2e-sched-image-validation.xml
```
Tal-or, line 57: Is there a reason to have the sched-image-validation test as a separate binary from the rest of sched tests?

test/e2e/sched/image_validation/validation_test.go
```diff
@@ -0,0 +1,131 @@
+/*
+ * Copyright 2021 Red Hat, Inc.
```
Tal-or: ditto

test/e2e/sched/image_validation/validation_test.go
```diff
@@ -0,0 +1,131 @@
...
+			klog.InfoS("scheduler conditions", "conditions", testSchedObj.Status.Conditions)
+
+			degradedCondition := metahelper.FindStatusCondition(testSchedObj.Status.Conditions, status.ConditionDegraded)
+			g.Expect(degradedCondition).ToNot(BeNil(), "degraded condition not found")
+			g.Expect(degradedCondition.Status).To(Equal(metav1.ConditionTrue), "degraded condition status is not true")
+			g.Expect(degradedCondition.Reason).To(Equal(string(status.ReasonInvalidSchedulerImage)), "degraded condition reason is not untrusted scheduler image")
+		}).WithTimeout(5 * time.Minute).WithPolling(time.Second * 10).Should(Succeed())
+
+		digest := "sha256:c1b0456fa8b7b96325a9cd2509c19c1e407e9d2cd7523ca62dd5db30cadb73b9"
+		invalidImage = "quay.io/openshift-kni/scheduler-plugins@" + digest
```
rshemtov13, line 86: Maybe here just keep the digest and remove the sha256 and pass it into invalidImage.
Another quick question can we get the digest dynamically instead of hardcoding or is it too much overhead?

test/e2e/sched/image_validation/validation_test.go
```diff
@@ -0,0 +1,131 @@
...
+			g.Expect(degradedCondition.Reason).To(Equal(string(status.ReasonInvalidSchedulerImage)), "degraded condition reason is not untrusted scheduler image")
+		}).WithTimeout(5 * time.Minute).WithPolling(time.Second * 10).Should(Succeed())
+
+		By("identify the source of operator installation and disable scheduler image validation")
+		var err error
+		restoreFunc, err := e2eenvvar.SetOperatorEnvVar(ctx, e2eclient.Client, intsched.CustomSchedulerDigestsEnvVar, digest)
+		Expect(err).NotTo(HaveOccurred())
+		DeferCleanup(func(cleanupCtx context.Context) {
+			By("restoring scheduler image validation settings")
+			Expect(restoreFunc(cleanupCtx)).To(Succeed())
```
Tal-or, line 115: this is the only place where we call `restoreFunc` why not simply inline the function?

Tal-or (commented):
overall looking good
minor comments

test/e2e/sched/image_validation/validation_test.go
shajmakh: > Maybe here just keep the digest and remove the sha256 and pass it into invalidImage.

we need the digest with the sha because we will use that in the next step to verify that intsched.CustomSchedulerDigestsEnvVar works as expected. 

> Another quick question can we get the digest dynamically instead of hardcoding or is it too much overhead?

not with the available infra, because pulling/inspecting the image dynamically needs a special infra. in this case all what we care about is the conditions of the scheduler so it doesn't matter the custom image we are using if it is old or not. I added a comment to clarify this. Thanks for raising this topic.

test/e2e/sched/image_validation/test_suite_validation_test.go
shajmakh: I'll remove it entirely, thanks!

test/e2e/sched/image_validation/validation_test.go
shajmakh: it's a returned function from SetOperatorEnvVar we connot inline it because we don't know what is it: https://github.com/shajmakh/numaresources-operator/blob/main/test/internal/envvar/envvar.go#L52

test/e2e/sched/image_validation/validation_test.go
rshemtov13: Ack, yes adding a comment explaining that for future purposes is beneficial.
Thanks!

hack/run-test-e2e.sh
shajmakh: good question, yes, because there seems an OLM issue updating the subscription with the different values of env var consecutively, as raised n https://github.com/openshift-kni/numaresources-operator/pull/4595#issuecomment-5019925191. This test requires re-enabling the image validation right after having it disabled so we can check  the default behavior of the validation. after many retries, I've seen that the multiple updates of the subcription and hence the operator deployment gets stuck for a while (sometimes over 15 mins) and is not stable in the prow ci. The last resort I had (before moving this test to d/s pipeline handling for more control over the operator installation - skipping subscription updates) was to have a new suite that works with the default configuration (meaning enabled image validation), so we don't hit the consecutive subscription updates. 
I haven't spent enough time investigating the OLM timing issue because after a while, the updates resolve to the expected state.

shajmakh: I added a note about this in the suite in case we wondered about this in the future.

Tal-or: sure. lets add a comment and I'll approve (I don't see the comment atm)

shajmakh: https://github.com/openshift-kni/numaresources-operator/pull/4693/changes#diff-43b6b54cf9ce892d32144d3cd7ed0800a9a7221b091a5d0dd13fd59349b01523R29

Tal-or (approved):
/lgtm

Thank you for the changes

shajmakh:
rebased to get this update: https://github.com/openshift-kni/numaresources-operator/pull/4784
no change in the reviewed and approved code.

Tal-or (approved):
/lgtm

PR #4702, opened 2026-07-21
go: toolchain: pin to 1.25.12

ffromani:
consume last fixes from the build chain 1.25.z

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,7 @@ on:
   workflow_dispatch:
 
 env:
-  GOLANG_VERSION: 1.25.11
+  GOLANG_VERSION: 1.25.12
 
 defaults:
   run:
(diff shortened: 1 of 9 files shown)


Tal-or (commented):
checkout what we did in main:
https://github.com/openshift-kni/numaresources-operator/pull/4625
no need to pin the version. it takes the latest automatically

ffromani:
> checkout what we did in main: #4625 no need to pin the version. it takes the latest automatically

There are 2 key differences though:
- main branch can follow a floating tag within reason (Y stream namely), release branches should be pinned as precisely as possible, with pins updated
- elaborating what I wrote in https://github.com/openshift-kni/numaresources-operator/pull/4625#pullrequestreview-4663807302 , yes, it works, but likely by side effect. The intended behavior of removing the `toolchain` directive is to make the toolchain match the declared go version, in our case 1.25.0 (not "last 1.25.z"). Then it's likey that github automation accidentally covers by always fetching the latest version of the 1.25.z series, which likely happened in parallel. xref: https://go.dev/doc/toolchain

Tal-or (approved):
/approve
/lgtm

PR #4784, opened 2026-07-29
govulcheck: bump deps to fix vulnerabilities

shajmakh:
bump grpc to address govulcheck failure.


Tal-or (approved):
/lgtm

this should be backport to 4.22 as well, but we're past codefreeze today IIRC

PR #4843, opened 2026-08-04
sched: enable default preemption

shajmakh:
Update the scheduler config to activate the prefilter plugin needed for preemption support, and bump the `deployer` dependency to consume the latest configuration option. See commits for additional details.

### DIFF
--- a/api/v1/numaresourcesscheduler_defaults.go
+++ b/api/v1/numaresourcesscheduler_defaults.go
@@ -53,6 +53,8 @@ func SetDefaults_NUMAResourcesSchedulerSpec(spec *NUMAResourcesSchedulerSpec) {
 			Type: defaultScoringStrategy,
 		}
 	}
+	// do not set preemptionMode default value to differentiate between unset and set values
+
 	// do not set replicas default value
 	// it will be computed later
 }
--- a/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/numaresources-operator.clusterserviceversion.yaml
@@ -66,7 +66,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2026-07-14T08:24:59Z"
+    createdAt: "2026-08-18T10:28:38Z"
     features.operators.openshift.io/cnf: "true"
     features.operators.openshift.io/cni: "false"
     features.operators.openshift.io/csi: "false"
@@ -295,6 +295,13 @@ spec:
           Defaults to "Normal".
         displayName: Scheduler log verbosity
         path: logLevel
+      - description: |-
+          PreemptionMode controls NUMA-aware preemption in the NodeResourceTopologyMatch plugin.
+          When unset, preemption is disabled by default. Set explicitly to Enabled to enable preemption.
+        displayName: Scheduler preemption mode setting
+        path: preemptionMode
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:text
       - description: Replicas control how many scheduler pods must be configured for
           High Availability (HA)
         displayName: Scheduler replicas
--- a/config/crd/bases/nodetopology.openshift.io_numaresourcesschedulers.yaml
+++ b/config/crd/bases/nodetopology.openshift.io_numaresourcesschedulers.yaml
@@ -75,6 +75,14 @@ spec:
                 - Trace
                 - TraceAll
                 type: string
+              preemptionMode:
+                description: |-
+                  PreemptionMode controls NUMA-aware preemption in the NodeResourceTopologyMatch plugin.
+                  When unset, preemption is disabled by default. Set explicitly to Enabled to enable preemption.
+                enum:
+                - Enabled
+                - Disabled
+                type: string
               replicas:
                 description: Replicas control how many scheduler pods must be configured
                   for High Availability (HA)
--- a/config/manifests/bases/numaresources-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/numaresources-operator.clusterserviceversion.yaml
@@ -230,6 +230,13 @@ spec:
           Defaults to "Normal".
         displayName: Scheduler log verbosity
         path: logLevel
+      - description: |-
+          PreemptionMode controls NUMA-aware preemption in the NodeResourceTopologyMatch plugin.
+          When unset, preemption is disabled by default. Set explicitly to Enabled to enable preemption.
+        displayName: Scheduler preemption mode setting
+        path: preemptionMode
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:text
       - description: Replicas control how many scheduler pods must be configured for
           High Availability (HA)
         displayName: Scheduler replicas
--- a/internal/controller/numaresourcesscheduler_controller.go
+++ b/internal/controller/numaresourcesscheduler_controller.go
@@ -514,6 +514,13 @@ func configParamsFromSchedSpec(schedSpec nropv1.NUMAResourcesSchedulerSpec, cach
 		})
 	}
 	params.ScoringStrategy.Resources = resources
+
+	if schedSpec.PreemptionMode != nil && *schedSpec.PreemptionMode == nropv1.PreemptionEnabled {
+		params.PreemptionMode = ptr.To(k8swgmanifests.PreemptionEnabled)
+	} else {
+		params.PreemptionMode = nil
+	}
+
 	params.Cache.ResyncMethod = &resyncMethod
 	params.Cache.ForeignPodsDetectMode = &foreignPodsDetect
 	params.Cache.InformerMode = &informerMode
--- a/internal/controller/numaresourcesscheduler_controller_test.go
+++ b/internal/controller/numaresourcesscheduler_controller_test.go
@@ -708,6 +708,81 @@ var _ = Describe("Test NUMAResourcesScheduler Reconcile", func() {
 
 		})
 
+		It("should not configure preemptionMode when unset", func(ctx context.Context) {
+			key := client.ObjectKeyFromObject(nrs)
+			_, err := reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: key})
+			Expect(err).ToNot(HaveOccurred())
+
+			cm := schedulerConfigMap(ctx, reconciler.Client)
+			Expect(cm).To(HaveSchedulerPreemptionMode(""))
+		})
+
+		It("should allow to change preemptionMode from Enabled to Disabled explicitly", func(ctx context.Context) {
+			key := client.ObjectKeyFromObject(nrs)
+			nrs := nrs.DeepCopy()
+			nrs.Spec.PreemptionMode = ptr.To(nropv1.PreemptionEnabled)
+			Eventually(func(g Gomega) {
+				g.Expect(reconciler.Client.Update(ctx, nrs)).To(Succeed())
+			}).WithTimeout(30 * time.Second).WithPolling(5 * time.Second).Should(Succeed())
+			_, err := reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: key})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(schedulerConfigMap(ctx, reconciler.Client)).To(HaveSchedulerPreemptionMode(string(nropv1.PreemptionEnabled)))
+
+			Expect(reconciler.Client.Get(ctx, key, nrs)).To(Succeed())
+			nrs.Spec.PreemptionMode = ptr.To(nropv1.PreemptionDisabled)
+			Eventually(func(g Gomega) {
+				g.Expect(reconciler.Client.Update(ctx, nrs)).To(Succeed())
+			}).WithTimeout(30 * time.Second).WithPolling(5 * time.Second).Should(Succeed())
+			_, err = reconciler.Reconcile(ctx, reconcile.Request{NamespacedName: key})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(schedulerConfigMap(ctx, reconciler.Client)).To(HaveSchedulerPreemptionMode(""))
+		})
+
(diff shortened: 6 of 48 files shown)


shajmakh:
split from https://github.com/openshift-kni/numaresources-operator/pull/4579

rte/main.go
```diff
@@ -120,14 +120,16 @@ func main() {
...
+
 	hnd := resourcetopologyexporter.Handle{
 		ResMon: resourcemonitor.Handle{
 			PodResCli: cli,
 			K8SCli:    k8scli,
 		},
 		NRTCli: nrtcli,
 	}
-	err = resourcetopologyexporter.Execute(hnd, parsedArgs.NRTupdater, parsedArgs.Resourcemonitor, parsedArgs.RTE)
+	err = resourcetopologyexporter.Execute(ctx, hnd, parsedArgs.NRTupdater, parsedArgs.Resourcemonitor, parsedArgs.RTE)
```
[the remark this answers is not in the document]
shajmakh: it was intentional to use `context.background` because u/s RTE is using it like this; but using a cancellable context is additional enhancement and also aligns with how the operator main.go is designed: 
https://github.com/openshift-kni/numaresources-operator/blob/cb30d32ddba4eb710add135f221b39488aef67b6/cmd/main.go#L268

rte/main.go
```diff
@@ -99,9 +98,11 @@ func main() {
 		cli = podexclude.NewFromLister(cli, parsedArgs.Global.Debug, parsedArgs.Resourcemonitor.PodExclude)
 	}
 
+	ctx := ctrl.SetupSignalHandler()
```
Tal-or, line 101: That's new? I read about it but first time I'm seeing it.

Do we have any trusted sources that recommends about this approach?

shajmakh: yes, the operator main:
https://github.com/openshift-kni/numaresources-operator/pull/4843#discussion_r3720172712

pkg/numaresourcesscheduler/manifests/yaml/configmap.nrt.yaml
ffromani: since we set multipoint we can delete these, can't we?

ffromani (commented):
LGTM once the inline comment is addressed

vendor/github.com/k8stopologyawareschedwg/numaplacement/leb89/leb89.go
```diff
@@ -0,0 +1,102 @@
...
+			if nCont == 0 {
+				return int32(idx), pos
+			}
+			return int32(NumTerminal) + contValue*int32(NumTerminal) + int32(idx), pos
+		}
+		contValue = contValue*int32(NumContinuation) + int32(idx-NumTerminal)
+		nCont++
+	}
+	return Unmapped, pos
+}
```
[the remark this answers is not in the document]
shajmakh: should be addressed on the numaplacement repo:
https://github.com/k8stopologyawareschedwg/numaplacement/issues/7

vendor/github.com/k8stopologyawareschedwg/resource-topology-exporter/test/e2e/utils/pods/rtepod/rtepod.go
```diff
@@ -88,7 +93,39 @@ func FindNotificationFilePath(rtePod *corev1.Pod) (string, error) {
...
+		if !ok {
+			continue
+		}
+		notifyPath, ok := te["notifyFilePath"].(string)
+		if ok && notifyPath != "" {
+			return notifyPath, nil
+		}
+	}
+	return "", fmt.Errorf("cannot find notification file path from CLI args or daemon configmap")
 }
```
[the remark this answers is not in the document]
shajmakh: vendor issue created: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/issues/405

vendor/github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/resourcemonitor/resourcemonitor.go
```diff
@@ -252,30 +275,49 @@ func WithNodeName(name string) func(*resourceMonitor) {
...
+	numaEligiblePodRes := []*podresourcesapi.PodResources{}
+	for _, pr := range respRawPodRes {
+		nl := numalocfilter.Verify(pr)
+		if !nl.Allow {
+			logger.V(8).Info("podresources item: not eligible for NUMA", "pod", pr.Namespace+"/"+pr.Name, "ident", nl.Ident, "reason", nl.Reason)
+			continue
+		}
+		numaEligiblePodRes = append(numaEligiblePodRes, pr)
+	}
+	logger.V(6).Info("NUMA eligible podresources list", "podresources", stringifyPodResources(numaEligiblePodRes))
```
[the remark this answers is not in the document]
shajmakh: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/issues/405

vendor/github.com/k8stopologyawareschedwg/resource-topology-exporter/pkg/resourcetopologyexporter/resourcetopologyexporter.go
```diff
@@ -113,9 +119,13 @@ func Execute(hnd Handle, nrtupdaterArgs nrtupdater.Args, resourcemonitorArgs res
...
-	eventSource.Wait()  // will never return
-	eventSource.Close() // still we try to clean after ourselves :)
-	return nil          // unreachable
+	select {
+	case <-ctx.Done():
+	}
+	eventSource.Stop()
+	eventSource.Wait()
+	eventSource.Close()
+	return nil
```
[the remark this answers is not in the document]
shajmakh: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/issues/405

internal/controller/numaresourcesscheduler_controller.go
```diff
@@ -514,6 +514,16 @@ func configParamsFromSchedSpec(schedSpec nropv1.NUMAResourcesSchedulerSpec, cach
...
 	params.ScoringStrategy.Resources = resources
+
+	var preemptionMode string
+	switch *schedSpec.PreemptionMode {
+	case nropv1.PreemptionDisabled:
+		preemptionMode = k8swgmanifests.PreemptionDisabled
+	default:
+		preemptionMode = k8swgmanifests.PreemptionEnabled
+	}
+	params.PreemptionMode = &preemptionMode
```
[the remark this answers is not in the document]
shajmakh: good catch, fixing it

pkg/numaresourcesscheduler/manifests/yaml/configmap.nrt.yaml
```diff
@@ -15,6 +15,8 @@ data:
           multiPoint:
             disabled:
               - name: DynamicResources
+            enabled:
+              - name: NodeResourceTopologyMatch
```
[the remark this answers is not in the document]
shajmakh: >  If any allowed digests include z-stream scheduler images without preFilter, enabling multiPoint can register an unsupported extension point.
 
 @coderabbitai but even when multipoint is added and an old scheduler image is used, the framework would pick up the plugin only if it is registred/implemented, right? so I expect the prefilter point to be ignored in such case and not cause a failure

pkg/numaresourcesscheduler/manifests/yaml/configmap.nrt.yaml
shajmakh: I believe yes because these are already registered by multipoint, `filter` however, must be kept explicit to define the allow list and the plugin in the right positions

shajmakh:
/hold 
Needs a new deployer patch

api/v1/numaresourcesscheduler_defaults.go
```diff
@@ -53,6 +53,9 @@ func SetDefaults_NUMAResourcesSchedulerSpec(spec *NUMAResourcesSchedulerSpec) {
 			Type: defaultScoringStrategy,
 		}
 	}
+	// do not set preemptionMode default value
+	// it must be set explicitly to preserve compatibility with legacy scheduler images
+
```
ffromani: can we set the default to "disabled" and switch to enabled later on when we are confident (and backward compatible, or high chance to)?

bundle/manifests/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -295,6 +295,14 @@ spec:
           Defaults to "Normal".
         displayName: Scheduler log verbosity
         path: logLevel
+      - description: |-
+          PreemptionMode controls NUMA-aware preemption in the NodeResourceTopologyMatch plugin.
+          When unset, preemption is not configured in the scheduler to preserve compatibility with
+          scheduler images that do not support this feature. Set explicitly to Enabled to enable preemption.
```
ffromani: the compatibility thing is an accident and should not be mentioned in the API docs, but rather in the release notes

config/crd/bases/nodetopology.openshift.io_numaresourcesschedulers.yaml
```diff
@@ -75,6 +75,15 @@ spec:
                 - Trace
                 - TraceAll
                 type: string
+              preemptionMode:
+                description: |-
+                  PreemptionMode controls NUMA-aware preemption in the NodeResourceTopologyMatch plugin.
+                  When unset, preemption is not configured in the scheduler to preserve compatibility with
+                  scheduler images that do not support this feature. Set explicitly to Enabled to enable preemption.
```
ffromani: ditto

config/manifests/bases/numaresources-operator.clusterserviceversion.yaml
```diff
@@ -230,6 +230,14 @@ spec:
           Defaults to "Normal".
         displayName: Scheduler log verbosity
         path: logLevel
+      - description: |-
+          PreemptionMode controls NUMA-aware preemption in the NodeResourceTopologyMatch plugin.
+          When unset, preemption is not configured in the scheduler to preserve compatibility with
+          scheduler images that do not support this feature. Set explicitly to Enabled to enable preemption.
```
ffromani: ditto

internal/controller/numaresourcesscheduler_controller_test.go
```diff
@@ -1421,6 +1464,32 @@ func expectScoringStrategyParams(cli client.Client, scoringStrategyType string,
...
+	cfg := cfgs[0]
+
+	if preemptionMode == "" {
+		Expect(cfg.PreemptionMode).To(BeNil())
+		return
+	}
+
+	Expect(cfg.PreemptionMode).ToNot(BeNil())
+	Expect(*cfg.PreemptionMode).To(Equal(preemptionMode))
+}
```
ffromani: good. Turning into a ginkgo helper would be even better. (stretch goal, non-blocking for this PR)

ffromani (commented):
LGTM once the inline comments are addressed

I haven't reviewed the e2e tests, I'll do separately. Arguably, there's a super-thin path to merge this code with disabled by default and defer e2e tests verification to a later stage.

shajmakh:
requires https://github.com/k8stopologyawareschedwg/deployer/pull/384

api/v1/numaresourcesscheduler_defaults.go
shajmakh: with the current deployer base, making that disabled would configure the scheduler config map to have the field. that is fine for scheduler images that support digesting this field; however, it is not for legacy scheduler images. 
what can be done is a change in the deployer to unset the field in CM in two cases: 
- nil, or 
- Disabled
And only when set to Enabled, it will be shown in the CM as enabled. 
shall I update the deployer to support that? just note that the different with the current version is that when mode is explicitly set to `Disabled` it will act as nil in CM instead of showing it is explcitly disabled; in turn the scheduler will know how to support that:
- if the preemption mode is supported it will autofill it with Disabled
- if we are using legacy scheduler, it will make no change

bundle/manifests/numaresources-operator.clusterserviceversion.yaml
shajmakh: ack, addressed. Thanks for highlighting.

api/v1/numaresourcesscheduler_defaults.go
shajmakh: or instead of updating deployer, in the controller code we can pass nil param when Disabled is set explicitly:

```
		if *schedSpec.PreemptionMode == nropv1.PreemptionEnabled {
			preemptionMode = k8swgmanifests.PreemptionEnabled
		}else{
		params.PreemptionMode = nil
               }
```

internal/controller/numaresourcesscheduler_controller_test.go
shajmakh: it is a ginkgo helper, did I miss anything?

api/v1/numaresourcesscheduler_defaults.go
ffromani: yes, this works. Furthermore, in the near future we should stop depending on the deployer: import the packages we need in the operator, copying the code (same license and often same authors anyway)
this will allow us to shred a dep and simpify our work.

internal/controller/numaresourcesscheduler_controller_test.go
ffromani: ginkgo helper matcher, but not blocking.

api/v1/numaresourcesscheduler_defaults.go
shajmakh: addressed as such: 
Only when the mode is explicitly enabled the CM param will be set to enabled. This assumes the user is aware if what he is doing and once enabled he is sure he is using the non-legacy scheduler.

internal/controller/numaresourcesscheduler_controller_test.go
shajmakh: thanks for clarifying, addressed.

ffromani:
/lgtm

thanks!
e2e review intentionally deferred

PR #4863, opened 2026-08-05
AI SDLC: ci: integrate OpenCode agent for automated govulncheck remediation

Tal-or:
Introduce `govulncheck-autofix.yaml` workflow that uses an OpenCode AI agent to automatically fix vulnerabilities detected by govulncheck. The workflow runs daily on schedule, can be triggered manually via workflow_dispatch, or by commenting `/fix-vulns` on a PR.

AIA Human-AI blend, Content edits, New content, Human-initiated, Reviewed, Claude Opus 4.6 v1.0

### DIFF
--- a/.github/workflows/govulncheck-autofix.yaml
+++ b/.github/workflows/govulncheck-autofix.yaml
@@ -1,0 +1,114 @@
+name: Autofix govulncheck vulnerabilities
+
+on:
+  schedule:
+    - cron: "0 6 * * *" # 06:00 UTC daily
+  workflow_dispatch: {}
+  issue_comment:
+    types: [created]
+
+env:
+  GOLANG_VERSION: 1.25.12
+  MODEL: google/gemini-3.6-flash
+
+jobs:
+  autofix:
+    runs-on: ubuntu-latest
+    concurrency:
+      group: govulncheck-autofix
+      cancel-in-progress: true
+    if: |
+      github.event_name == 'schedule' ||
+      github.event_name == 'workflow_dispatch' ||
+      (
+        github.event.issue.pull_request &&
+        contains(github.event.comment.body, '/fix-vulns') &&
+        contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)
+      )
+    permissions:
+      id-token: write
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GOLANG_VERSION }}
+
+      - name: Resolve base branch
+        id: base
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          if [ "${{ github.event_name }}" = "issue_comment" ]; then
+            REF=$(gh pr view "${{ github.event.issue.number }}" --json baseRefName -q .baseRefName)
+          else
+            REF="${{ github.event.repository.default_branch }}"
+          fi
+          echo "ref=$REF" >> "$GITHUB_OUTPUT"
+
+      - name: Acknowledge trigger
+        if: github.event_name == 'issue_comment'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh pr comment "${{ github.event.issue.number }}" \
+            --body "ai-bot: govulncheck autofix triggered — running vulnerability scan."
+
+      - name: Run govulncheck
+        id: govulncheck
+        uses: golang/govulncheck-action@v1
+        with:
+          go-package: "./..."
+          go-version-input: ${{ env.GOLANG_VERSION }}
+          output-file: govulncheck-output.txt
+          repo-checkout: false
+        continue-on-error: true
+
+      - name: Run OpenCode to fix vulnerabilities
+        if: steps.govulncheck.outcome == 'failure'
+        uses: anomalyco/opencode/github@latest
+        env:
+          GOOGLE_GENERATIVE_AI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          model: ${{ env.MODEL }}
+          use_github_token: true
+          prompt: |
+            The file `govulncheck-output.txt` contains a govulncheck vulnerability report.
+            The target branch is `${{ steps.base.outputs.ref }}`.
+
+            Fix ALL reported vulnerabilities by bumping the affected Go packages in go.mod
+            to versions that resolve them. Run `go mod tidy` and `go mod vendor`, then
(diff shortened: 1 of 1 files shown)


Tal-or:
/hold

Requires a secret to access GEMINI API. 
Secret will be added only after reaching consensus with reviewers.

shajmakh:
Thanks for looking into this. 

> can be triggered manually via workflow_dispatch, or by commenting /fix-vulns on a PR.

will this add a commit on the PR? or create a different PR, merge it and rebase the current PR?

.github/workflows/govulncheck-autofix.yaml
```diff
@@ -0,0 +1,74 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          model: ${{ env.MODEL }}
+          use_github_token: true
+          prompt: |
+            The file `govulncheck-output.txt` contains a govulncheck vulnerability report.
+
+            Fix the reported vulnerabilities by bumping the affected Go packages in go.mod
+            to versions that resolve them. Run `go mod tidy` and `go mod vendor`, then
+            verify the build passes with `go build ./...`.
```
shajmakh, line 83: do we want it to ensure lint and unit tests pass too, and if not fix that as well?

.github/workflows/govulncheck-autofix.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+            Requirements:
+            - Branch from `origin/main` using a branch name like `fix/govulncheck-<vuln-id>`.
```
shajmakh: what if there are multiple vulnerabilities?

.github/workflows/govulncheck-autofix.yaml
```diff
@@ -0,0 +1,74 @@
...
+            - If that branch already exists remotely, stop — a fix is already in progress.
```
shajmakh: what if while the a first PR is open a new volunerability is discovered for the same package and needs a new bump? 
I think the open PR should be continuously rebased and updated, and then it will be always in a race between lgtm being added and the next day the content is re-pushed, leading to removing the label.

.github/workflows/govulncheck-autofix.yaml
shajmakh: I'd go with `fix/govulncheck-<target-branch-name>` and delete it once the PR is merge. 
I assume that way new updates should be pushed to the open PR, and if no PR is there then it would open one.

.github/workflows/govulncheck-autofix.yaml
```diff
@@ -0,0 +1,74 @@
...
+            - Commit message format: "deps: bump <pkg> to <version> to fix <vuln-id>".
+            - Push the branch and open a PR against main with title prefix `[ai-bot]`.
```
shajmakh: may need to adjut the verify-commits script to allow the bot

shajmakh:
@Tal-or do you have a trial run of the action? perhaps on your forked repo just to visualise how this would behave

.github/workflows/govulncheck-autofix.yaml
Tal-or: unit tests and lint are not running on `vendor` directories which is the only change that is needed for fixing go vulnerabilities.

Tal-or:
> Thanks for looking into this.
> 
> > can be triggered manually via workflow_dispatch, or by commenting /fix-vulns on a PR.
> 
> will this add a commit on the PR? or create a different PR, merge it and rebase the current PR?

by commenting  `/fix-vulns` on any PR, you'll be triggering the bot to run the workflow and trying to fix the vulnerability. 

See this example on my fork:
https://github.com/Tal-or/numaresources-operator/pull/5#issuecomment-5193097645

.github/workflows/govulncheck-autofix.yaml
Tal-or: The agent will figure it our, but I can make the instructions more accurate.

shajmakh:
> 
> See this example on my fork: [Tal-or#5 (comment)](https://github.com/Tal-or/numaresources-operator/pull/5#issuecomment-5193097645)

nice! can we get the agent to add the govulcheck report to the PR body just to justify the update?

.github/workflows/govulncheck-autofix.yaml
Tal-or: > what if while the a first PR is open a new volunerability is discovered for the same package and needs a new bump?
I think the open PR should be continuously rebased and updated, and then it will be always in a race between lgtm being added and the next day the content is re-pushed, leading to removing the label.

This is extremely rare, but in such case the scheduled workflow that will be trigger in the day after, or by a comment, will detect the new vulnerability and issue a PR that fixes both.
And I'll update the prompt for the agent to handle multiple vulnerabilities.

.github/workflows/govulncheck-autofix.yaml
Tal-or: probably. this is an iterative effort, we'll have to adapt as we go.

Tal-or:
> > See this example on my fork: [Tal-or#5 (comment)](https://github.com/Tal-or/numaresources-operator/pull/5#issuecomment-5193097645)
> 
> nice! can we get the agent to add the govulcheck report to the PR body just to justify the update?

This is how the PR look: https://github.com/Tal-or/numaresources-operator/pull/6 
We can tell the agent to add whatever we want.

.github/workflows/govulncheck-autofix.yaml
ffromani: agreed that lint is unnecessary, but a unit test pass won't hurt

ffromani (commented):
The idea is neat, but I wonder if we need actual agents here or if we can just script the action (perhaps with AI assistance to build the script in the first place)

Tal-or:
> The idea is neat, but I wonder if we need actual agents here or if we can just script the action (perhaps with AI assistance to build the script in the first place)

That's probably possible, but you'll have to parse the CI vulnerability output, to detect the right GO-* vuln, which is not a standardized API.

Also this is just the first step in my opinion, there're other more complex tasks we can leverage this tool for.

.github/workflows/govulncheck-autofix.yaml
```diff
@@ -0,0 +1,114 @@
...
+    runs-on: ubuntu-latest
+    concurrency:
+      group: govulncheck-autofix
+      cancel-in-progress: true
+    if: |
+      github.event_name == 'schedule' ||
+      github.event_name == 'workflow_dispatch' ||
+      (
+        github.event.issue.pull_request &&
+        contains(github.event.comment.body, '/fix-vulns') &&
```
Tal-or, line 25: Using this approach we can invoke the agent to perform different tasks.

ffromani: these tasks need to be however well defined and well contained though.

Tal-or:
There could be a way to automate this work and not involving AI at all. 
Besides, there are other initiatives going in the same direction so lets consider existing approaches first.

PR #4887, opened 2026-08-06
E2E: TLS: move tier2 label from context to It scope

mrniranjan:
Move tier2 label from Context scope to It scope

### DIFF
--- a/test/e2e/serial/tests/configuration.go
+++ b/test/e2e/serial/tests/configuration.go
@@ -1818,14 +1818,14 @@ var _ = Describe("[serial][disruptive] numaresources configuration management", 
 		})
 	})
 
-	Context("[tlscompliance][rte] rte complies with TLS Profile modifications", Label(label.Tier2, "feature:tlscompliance"), func() {
+	Context("[tlscompliance][rte] rte complies with TLS Profile modifications", Label("feature:tlscompliance"), func() {
 		const (
 			rteDaemonSetCheckTimeout  = 30 * time.Second
 			rteDaemonSetCheckInterval = 5 * time.Second
 			apiServerUpdateTimeout    = 10 * time.Minute
 			apiServerUpdateInterval   = 10 * time.Second
 		)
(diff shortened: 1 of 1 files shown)


shajmakh:
I see there is only one `It` in this context, so it is a change for future hygiene in case other nested specs are added. can you please update the commit message to illustrate that instead of empty message?

test/e2e/serial/tests/configuration.go
```diff
@@ -1818,14 +1818,14 @@ var _ = Describe("[serial][disruptive] numaresources configuration management",
...
-	Context("[tlscompliance][rte] rte complies with TLS Profile modifications", Label(label.Tier2, "feature:tlscompliance"), func() {
+	Context("[tlscompliance][rte] rte complies with TLS Profile modifications", Label("feature:tlscompliance"), func() {
 		const (
 			rteDaemonSetCheckTimeout  = 30 * time.Second
 			rteDaemonSetCheckInterval = 5 * time.Second
 			apiServerUpdateTimeout    = 10 * time.Minute
 			apiServerUpdateInterval   = 10 * time.Second
 		)
-		It("[test_id:88383] should update RTE daemonset TLS args to adhere to modified TLS profile", Label(label.Reboot, label.Slow), func(ctx context.Context) {
+		It("[test_id:88383] should update RTE daemonset TLS args to adhere to modified TLS profile", Label(label.Reboot, label.Slow, label.Tier2, label.OpenShift), func(ctx context.Context) {
```
shajmakh: Why add the openshift label? can't this test run on hypershift?

mrniranjan: I have no idea, we don't have hypershift environment

Tal-or: Let's not narrow-down the scope without a good reason, since we'll have to support that on HCP in the future anyway.

I would suggest to keep it for HCP as well, and when you'll have the HCP environment we'll reevaluate.

mrniranjan: ack, updated

shajmakh: +1, if we just disable it now, it will be very hard to know about it when the suite runs on hypershift, because it will be skipped; having it fail in this case and address it later is better.

PR #4888, opened 2026-08-06
api: ng-config: allow disabling NUMA placement reporting

shajmakh:
RTE provides an flag to control NUMA placement reporting in NRT. This is particularly handy to disable NRT scheduler preemption because we know that the scheduler needs the conatiners localities to operate on postFilter preemption, and when NUMA placement data is not provided it will do the normal flow of filtering without any eviction. The goal of this is to have a backup path in case any false evictions are happening, especially because the preemption is enabled by default.

AI-attribution: AIA PAI Hin R Cursor 3.1.7 v1.0

### DIFF
--- a/api/v1/numaresourcesoperator_defaults.go
+++ b/api/v1/numaresourcesoperator_defaults.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 )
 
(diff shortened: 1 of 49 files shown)


shajmakh:
requires RTE bump to the new tag that would include: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/402

shajmakh:
/hold
requires: https://github.com/openshift-kni/numaresources-operator/pull/4843

ffromani:
superceded by https://github.com/openshift-kni/numaresources-operator/pull/5082

PR #4963, opened 2026-08-11
makefile: add target for scheduler digests update

shajmakh:
Add a make target to update the scheduler digests json file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,9 +240,18 @@ binary-numacell: build-tools ## Build the numacell test device plugin binary.
 	CGO_ENABLED=0 go build -mod=vendor -o bin/numacell -ldflags "$$LDFLAGS" test/deviceplugin/cmd/numacell/main.go
 
 .PHONY: binary-getdigests
-binary-getdigests: 
+binary-getdigests: ## Build the getdigests helper binary.
 	LDFLAGS="-s -w"; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -240,10 +240,19 @@ binary-numacell: build-tools ## Build the numacell test device plugin binary.
...
+binary-getdigests: ## Build the getdigests helper binary.
 	LDFLAGS="-s -w"; \
 	go build -mod=vendor -o bin/getdigests -ldflags "$$LDFLAGS" tools/getdigests/getdigests.go
 
+.PHONY: update-scheduler-digests
+update-scheduler-digests: binary-getdigests ## Refresh embedded scheduler image digests from the production registry.
+	bin/getdigests \
+		--current-url registry.redhat.io/openshift4/noderesourcetopology-scheduler-rhel9:v4.22 \
+		--prev-url registry.redhat.io/openshift4/noderesourcetopology-scheduler-rhel9:v4.21 \
+		--eus-url registry.redhat.io/openshift4/noderesourcetopology-scheduler-rhel9:v4.20 \
```
Tal-or, line 252: These paths will have to be updated with each release right? my concern is that they'll get rotten quickly. 

Who should call this make target? is it suppose to be done manually or by an automated process?

PR #4968, opened 2026-08-12
ci: ship /bin/numazone alias in numacell test image

swatisehgal:
The numacell productization work(https://github.com/openshift-kni/numaresources-operator/pull/4747) renames the device plugin to numazone and updates DaemonSet manifests to run Command /bin/numazone. That change cannot land cleanly while quay.io/openshift-kni/numacell-device-plugin:test-ci only provides /bin/numacell, and the new release workflow cannot be used from main until productization merges.

Install the existing numacell binary at /bin/numazone as well so a refresh of the current test image unblocks productization e2e without renaming the quay repository yet.

ENTRYPOINT is kept intact so that the current main stays compatible with existing manifests and default image behavior. Productization PR will override the entrypoint and can use the new path immediately after the image is republished.


ffromani:
/approve
/lgtm

an alias is cheap, especially in the test-only image, even if the productization doesn't materialize for whatever reason.

PR #4969, opened 2026-08-12
e2e: use operator/RTE image for numazone device plugin

swatisehgal:
Point serial e2e infra at the same multi-entrypoint image the operator
already uses for RTE, where numazone is shipped alongside the exporter.
Keep explicit E2E_*_DEVICE_PLUGIN env overrides and the standalone quay
image only as a last-resort fallback.

Depends on https://github.com/openshift-kni/numaresources-operator/pull/4747. This will be rebased once https://github.com/openshift-kni/numaresources-operator/pull/4747 is merged.

Comment for reviewers: For this PR only the last commit needs to be reviewed.


Tal-or (commented):
Hey @swatisehgal could you please rebase this one?

PR #4979, opened 2026-08-13
e2e: serial: fix discover RuntimeClass handler from node CRI runtimes

rshemtov13:
On OpenShift clusters whose nodes only expose `crun` (and optionally `high-performance`), pods that must start fail with `FailedCreatePodSandBox: failed to find runtime handler runc`.

+ fixing a small typo in APIVersion

### DIFF
--- a/test/e2e/serial/tests/workload_overhead.go
+++ b/test/e2e/serial/tests/workload_overhead.go
@@ -49,6 +49,13 @@ import (
 	. "github.com/onsi/gomega"
 )
 
+type runtimeHandler = string
+
+const (
+	crunHandler runtimeHandler = "crun"
+	runcHandler runtimeHandler = "runc"
+)
+
 var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhead", Serial, Label("disruptive", "scheduler"), Label("feature:overhead"), func() {
 	var fxt *e2efixture.Fixture
 	var padder *e2epadder.Padder
@@ -107,16 +114,29 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
 
 		When("a RuntimeClass exist in the cluster", func() {
 			var rtClass *nodev1.RuntimeClass
-			BeforeEach(func() {
+			BeforeEach(func(ctx context.Context) {
+				node := &corev1.Node{}
+				Expect(fxt.Client.Get(ctx, client.ObjectKey{Name: nrtTwoZoneCandidates[0].Name}, node)).To(Succeed())
+
+				var handler string
+				for _, h := range node.Status.RuntimeHandlers {
+					if h.Name == crunHandler || h.Name == runcHandler {
+						handler = h.Name
+						break
+					}
+				}
+				Expect(handler).NotTo(BeEmpty(), "no %s/%s runtime handler on node %q", crunHandler, runcHandler, node.Name)
+
+				klog.InfoS("creating RuntimeClass", "name", "test-rtclass", "handler", handler, "node", node.Name)
 				rtClass = &nodev1.RuntimeClass{
 					TypeMeta: metav1.TypeMeta{
 						Kind:       "RuntimeClass",
-						APIVersion: "node.k8s.io/vi",
+						APIVersion: "node.k8s.io/v1",
 					},
 					ObjectMeta: metav1.ObjectMeta{
 						Name: "test-rtclass",
 					},
-					Handler: "runc",
+					Handler: handler,
 					Overhead: &nodev1.Overhead{
 						PodFixed: corev1.ResourceList{
 							corev1.ResourceCPU:    resource.MustParse("500m"),
@@ -125,7 +145,7 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
 					},
 				}
 
-				err := fxt.Client.Create(context.TODO(), rtClass)
+				err := fxt.Client.Create(ctx, rtClass)
 				Expect(err).NotTo(HaveOccurred())
 			})
 			AfterEach(func() {


test/e2e/serial/tests/workload_overhead.go
```diff
@@ -108,15 +108,17 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
...
 					TypeMeta: metav1.TypeMeta{
 						Kind:       "RuntimeClass",
-						APIVersion: "node.k8s.io/vi",
+						APIVersion: "node.k8s.io/v1",
 					},
 					ObjectMeta: metav1.ObjectMeta{
 						Name: "test-rtclass",
 					},
-					Handler: "runc",
+					Handler: handler,
```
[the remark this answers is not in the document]
rshemtov13: Our d/s CI clusters never have mix runtimes: workers expose a single general-purpose handler (`crun` or `runc`). The later target-node pick can’t land on a different one.

test/e2e/serial/tests/workload_overhead.go
```diff
@@ -450,3 +452,22 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
...
+	for _, h := range nodeList.Items[0].Status.RuntimeHandlers {
+		if h.Name == "crun" || h.Name == "runc" {
+			handler = h.Name
+			break
+		}
+	}
+
+	Expect(handler).NotTo(BeEmpty(), "no crun/runc runtime handler on node %q", nodeList.Items[0].Name)
+	return handler
+}
```
shajmakh: The content can be inlined for better readability, no need for a function

test/e2e/serial/tests/workload_overhead.go
```diff
@@ -108,6 +108,8 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
 		When("a RuntimeClass exist in the cluster", func() {
 			var rtClass *nodev1.RuntimeClass
 			BeforeEach(func() {
+				handler := runtimeClassHandler(context.TODO(), fxt.Client)
```
shajmakh: let's please use the BeforeEach context level:
```	
		BeforeEach(func(ctx context.Context) {
                      ...ctx, fxt.Client)
```

test/e2e/serial/tests/workload_overhead.go
```diff
@@ -450,3 +452,22 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
 		})
 	})
 })
+
+func runtimeClassHandler(ctx context.Context, cli client.Client) string {
+	GinkgoHelper()
+
+	nodeList := &corev1.NodeList{}
+	Expect(cli.List(ctx, nodeList)).To(Succeed(), "listing nodes to discover runtime handlers")
```
shajmakh: why do we need to list all the nodes and on the first runtimeclass break the loop? can't we useany node from the nrt candidates?

test/e2e/serial/tests/workload_overhead.go
mrniranjan: why are we hardcoding the container runtimes ?

test/e2e/serial/tests/workload_overhead.go
rshemtov13: Right we can use, switching to an NRT candidate nrtTwoZoneCandidates[0] instead of NodeList[0].

test/e2e/serial/tests/workload_overhead.go
```diff
@@ -107,16 +107,29 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
...
 		When("a RuntimeClass exist in the cluster", func() {
 			var rtClass *nodev1.RuntimeClass
-			BeforeEach(func() {
+			BeforeEach(func(ctx context.Context) {
+				node := &corev1.Node{}
+				Expect(fxt.Client.Get(ctx, client.ObjectKey{Name: nrtTwoZoneCandidates[0].Name}, node)).To(Succeed())
+
+				var handler string
+				for _, h := range node.Status.RuntimeHandlers {
+					if h.Name == "crun" || h.Name == "runc" {
```
ffromani: do we have or can we use constants here?

test/e2e/serial/tests/workload_overhead.go
```diff
@@ -107,16 +107,29 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
...
+					}
+				}
+				Expect(handler).NotTo(BeEmpty(), "no crun/runc runtime handler on node %q", node.Name)
+
+				klog.InfoS("creating RuntimeClass", "name", "test-rtclass", "handler", handler, "node", node.Name)
 				rtClass = &nodev1.RuntimeClass{
 					TypeMeta: metav1.TypeMeta{
 						Kind:       "RuntimeClass",
-						APIVersion: "node.k8s.io/vi",
+						APIVersion: "node.k8s.io/v1",
```
ffromani, line 134: nice catch!

test/e2e/serial/tests/workload_overhead.go
rshemtov13: We don't have but we definitely can use constants here

rshemtov13: Added the constants.

test/e2e/serial/tests/workload_overhead.go
```diff
@@ -107,16 +114,29 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
 
 		When("a RuntimeClass exist in the cluster", func() {
 			var rtClass *nodev1.RuntimeClass
-			BeforeEach(func() {
+			BeforeEach(func(ctx context.Context) {
+				node := &corev1.Node{}
+				Expect(fxt.Client.Get(ctx, client.ObjectKey{Name: nrtTwoZoneCandidates[0].Name}, node)).To(Succeed())
```
Tal-or, line 119: Are all nodes using same handler that we can use the first one as an indication for the rest?

test/e2e/serial/tests/workload_overhead.go
```diff
@@ -107,16 +114,29 @@ var _ = Describe("[serial][disruptive][scheduler] numaresources workload overhea
...
+
+				var handler string
+				for _, h := range node.Status.RuntimeHandlers {
```
Tal-or, line 122: this is a `for` loop because there could be other runtime handler but it is guaranteed that both crunHandler and runcHandler cannot co-exist?

test/e2e/serial/tests/workload_overhead.go
rshemtov13: Yes all will have the same handler.

test/e2e/serial/tests/workload_overhead.go
rshemtov13: They can co-exist, but we will never have both in the d/s CI. The loop just picks the first `crun`/`runc` and skips the rest (`""`, `high-performance`, …). Either is fine, we only need a handler that exists on the node.
Also the default changed in OCP 4.18, which is why both are options.

Tal-or (approved):
/lgtm
/approve

Thanks!

PR #4987, opened 2026-08-14
e2e: serial: adding the preemption for active topics

rshemtov13:
D/s skip list is (tests topics.json active) − (operator
inspect-features active). Without preemption in active, it
never gets skipped and runs on all releases. Disable on 5.0
d/s until ready.

### DIFF
--- a/internal/api/features/_topics.json
+++ b/internal/api/features/_topics.json
@@ -24,6 +24,7 @@
     "netpols",
     "introspection",
     "mgselinuxcollect",
-    "tlscompliance"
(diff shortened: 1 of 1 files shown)


shajmakh:
I'm not sure the fix should be u/s because this is related to how the filtering is done in d/s pipeline. What I'd expect is that if a test belongs to a feature that is not among the topic's list then it should not run.
However, besides that this is needed anyway when the feature is merged, I understand it removes noise on other releases runs so no harm to merge. 
/lgtm

PR #5041, opened 2026-08-18
ci: fix mcp crd path

Tal-or:
Path has been moved again, we need to find a better way to make it more stable. for now lets set the new path to unblock ci.


ffromani:
/approve
/lgtm

thanks

PR #5082, opened 2026-08-24
api: ng-config: allow disabling NUMA placement reporting [20260824]

ffromani:
RTE provides an flag to control NUMA placement reporting in NRT. This is particularly handy to disable NRT scheduler preemption because we know that the scheduler needs the conatiners localities to operate on postFilter preemption, and when NUMA placement data is not provided it will do the normal flow of filtering without any eviction. The goal of this is to have a backup path in case any false evictions are happening, especially because the preemption is enabled by default.

AI-attribution: AIA PAI Hin R Cursor 3.1.7 v1.0

### DIFF
--- a/api/v1/numaresourcesoperator_types.go
+++ b/api/v1/numaresourcesoperator_types.go
@@ -77,6 +77,17 @@ const (
 	InfoRefreshPauseEnabled InfoRefreshPauseMode = "Enabled"
 )
 
+// +kubebuilder:validation:Enum=container;none
+type NUMAPlacementMode string
+
+const (
+	// NUMAPlacementContainer enables container NUMA placement reporting in NRT. It is the default.
+	NUMAPlacementContainer NUMAPlacementMode = "container"
+
+	// NUMAPlacementNone disables container NUMA placement reporting in NRT. Used to disable NUMA-aware eviction during default preemption.
+	NUMAPlacementNone NUMAPlacementMode = "none"
+)
+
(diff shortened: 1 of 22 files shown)


ffromani:
replaces https://github.com/openshift-kni/numaresources-operator/pull/4888

ffromani:
requires: https://github.com/openshift-kni/numaresources-operator/pull/5083

api/v1/numaresourcesoperator_types.go
```diff
@@ -77,6 +77,17 @@ const (
 	InfoRefreshPauseEnabled InfoRefreshPauseMode = "Enabled"
 )
 
+// +kubebuilder:validation:Enum=container;none
+type NUMAPlacementMode string
```
Tal-or, line 81: Is there a concrete reason to go with string value and not a boolean?

I presume it's more convenient in case of future expansion, just wondering if there's such already

ffromani: The kubernetes API generally recommends to avoid booleans. It makes future expansion _possible_ and it is less ambiguous with default values (something which can be mitigated with pointers to bools though). In our case we have no obvious expansion in mind, but no reason to diverge from the recommendation either.

Tal-or: Good rational indeed.

Tal-or (approved):
/lgtm
/approve

PR #5135, opened 2026-08-31
e2e: serial: add compact PFP regression for OCPBUGS-90597

SargunNarula:
Default EnabledExclusiveResources PFP on compact clusters could leave Guaranteed topo-aware pods Pending with invalid node topology data when the scheduler counted OpenShift infra extended resources as exclusive.

Add tier0 serial test [test_id:90597] that runs only on compact clusters, verifies default exclusive-resources PFP, and asserts a Guaranteed pod schedules successfully. Run it from the compact e2e lane (ci-e2e-compact).


shajmakh:
Thanks Sargun for the efforts here, writing the key points here following our offline conversation:
1. generate the buggy environment: the current change assumes that the environment the bug is going to run on is enough to reproduce the bug if the scheduler is not fixed. This is incorrect, because the compact lane does not have the kind of host-level resources that look like devices but not part of the NRT. 
2. The steps that need to be done are: 
- try to simulate the buggy environment (check sample-device-plugin for host-level); run the bug scenario on the environment and ensure it fails for the same reason.
- automate the simulation d/s to avoid comlicating the implementation u/s
- automate the test u/s gating it with an env var that indicates if the environment to run the test is fulfiling the needs of the tests. If it is, continue to run the test, if not then skip. 

Simulating the buggy env can be done by creating host-level resources. The question is whether sample-device-plugin supports that. If it does, the solution is straightforward since we already use device-plugin setup in d/s.

PR #5179, opened 2026-09-03
pkg: introduce and use the iter package

ffromani:
the iter package borrows the code from podutil.ContainerIter,
to make us avoid to depend (again) on `k8s.io/kubernetes`.

Besides trivial rename, the package is intentional backward compatible
with the borrowed kube code; we may consider to extend in the future.

The benefit of this approach (iterator) is the following:
the pre-iterator standard pattern of using `append(pod.Spec.InitContainers, pod.Spec.Containers...)`
to iterate over containers was causeing allocation of backing slices and
copies, creating GC pressure.

We now have a kube utility to iterate in place, let's use it.
No intended change of behavior.
Note: this means the previous code MUST not mutate containers - that was
never the intent anyway, but this change may now expose latent hidden bugs.

### DIFF
--- a/api/v1/numaresourcesoperator_types.go
+++ b/api/v1/numaresourcesoperator_types.go
@@ -18,6 +18,7 @@ package v1
 
 import (
 	"fmt"
+	"maps"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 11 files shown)


shajmakh:
/lgtm
Thanks for handling.

PR #5191, opened 2026-09-04
chore: fix iterator copyright boilerplate

ffromani:
it was already easily inferrable, but let's make it really obvious and explicit.

followup of #5179


shajmakh:
I missed that, thanks.
/lgtm

PR #5233, opened 2026-09-09
NO-JIRA: ci: remove ginkgo focus from test-install-e2e

Tal-or:
Since the very first commit of `run-test-install-e2e.sh` it was limited to a subset of the tests by a ginkgo focus.

As we move along towards the apporach of pre-merge verification state of mine, there's no good reason to not runnnig all the tests we have in this suite allowing catching errors sooner rather than later.


Tal-or:
depends on #5239

Tal-or:
rebase was needed

Tal-or:
looks like there are actual failures that need to be examined

Tal-or:
@ffromani PTAL 
I had to refactor the tests a little bit to make CI happy

PR #5301, opened 2026-09-15
e2e: fix: padding pod scheduling failures

rshemtov13:
Tests were failing because padding pods for multiple NUMA zones could request more CPU or memory than the node had available.

This change creates padding pods one at a time, waits for each pod to run, caps the next pod to the node’s remaining resources, and sets explicit resource requests for correct scheduler accounting.

All tests that were affected by these changes were validated and passed.

85779, 49068, 47616, 47614, 84015, 84016, 84017, 74249, 74250, 74251, 74252


ffromani (commented):
/approve
/lgtm

there are minor improvement we can add later

